PR #5, opened 2017-10-04
Allow user to hide explorer

EricJizbaMSFT:


EricJizbaMSFT:
It's related to localization. We should add it to app services eventually

PR #6, opened 2017-10-10
Add zip deploy

### LINKED ISSUE #3: Deploy function app as zip package to Azure

EricJizbaMSFT:
This leverages the shared package I've created here: https://github.com/EricJizbaMSFT/vscode-azure-sdk

Obviously we'll have to discuss if that shared package will work and find an actual home for it instead of my fork.

Fixes #3

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -7,5 +7,6 @@
     "azFunc.startFunctionApp": "Start",
     "azFunc.stopFunctionApp": "Stop",
     "azFunc.restartFunctionApp": "Restart",
-    "azFunc.showExplorerDescription": "Show or hide the Azure Functions Explorer"
+    "azFunc.showExplorerDescription": "Show or hide the Azure Functions Explorer",
+    "azFunc.deployZip": "Deploy as Zip Package"
 }
--- a/src/commands/deployZip.ts
+++ b/src/commands/deployZip.ts
@@ -1,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line:no-require-imports
+import WebSiteManagementClient = require('azure-arm-website');
+import * as vscode from 'vscode';
+import { AzureFunctionsExplorer } from '../AzureFunctionsExplorer';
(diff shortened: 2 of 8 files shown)


src/commands/deployZip.ts
```diff
@@ -0,0 +1,31 @@
...
+    const newFolderId: string = 'newFolder';
+    let folderPicks: util.PickWithData<string>[] = [new util.PickWithData(newFolderId, localize('azFunc.selectOtherFolder', '$(plus) Select other folder'))];
+    const folders: vscode.WorkspaceFolder[] | undefined = vscode.workspace.workspaceFolders;
+    if (folders) {
+        folderPicks = folderPicks.concat(folders.map((f: vscode.WorkspaceFolder) => new util.PickWithData<string>('', f.uri.fsPath)));
+    }
+    const folder: util.PickWithData<string> = await util.showQuickPick<string>(folderPicks, localize('azFunc.deployZipSelectFolder', 'Select a workspace folder to deploy to your Function App'));
+    const folderPath: string = folder.data === newFolderId ? await util.showFolderDialog() : folder.label;
+
+    await node.siteWrapper.deployZip(folderPath, client, outputChannel);
```
StephenWeatherford: outputChannel.show();

Should also add it to create function app, etc.

EricJizbaMSFT: Added this inside of siteWrapper.deployZip

package.nls.json
```diff
@@ -7,5 +7,6 @@
     "azFunc.startFunctionApp": "Start",
     "azFunc.stopFunctionApp": "Stop",
     "azFunc.restartFunctionApp": "Restart",
-    "azFunc.showExplorerDescription": "Show or hide the Azure Functions Explorer"
+    "azFunc.showExplorerDescription": "Show or hide the Azure Functions Explorer",
+    "azFunc.deployZip": "Deploy as Zip Package"
```
StephenWeatherford: How about "Deploy here as zip package"?

StephenWeatherford: (but capitalized properly)

EricJizbaMSFT: But 'here' doesn't make sense when the command is run from the command palette.

I created a bug to address the entry points separately: https://github.com/Microsoft/vscode-azurefunctions/issues/15

EricJizbaMSFT:
Okay the PR for the shared packages is out: https://github.com/Microsoft/vscode-azuretools/pull/1

This PR is updated with latest from master and I removed link to package in my custom fork. CI build will fail until shared package is published.

PR #13, opened 2017-10-12
Refactor error handling and add unit tests for the first time

EricJizbaMSFT:
If we let errors bubble out of our extension, VSCode will sometimes show the error and sometimes not. To be safe, we should show the message ourselves. I also like showing messages in one place (extension.ts) so that our experience is consistent across the extension and so that telemetry always catches it.

However, when I initially implemented this I didn't do the correct typing all over the place and thus an error of type 'string' wouldn't be displayed to the user. Now it should work for whatever type of error is thrown at us. I also took this as an excuse to experiment with unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,20 @@ language: node_js
 node_js:
   - 'stable'
 
+before_install:
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CXX="g++-4.9" CC="gcc-4.9" DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
+      sleep 3;
+    fi
+
 install:
   - npm install
 
 script:
-  - tsc -p ./
-  - tslint --project tsconfig.json -e src/*.d.ts --type-check -t verbose
+  - npm run vscode:prepublish
+  - npm run lint
+  - npm run test
 
 notifications:
   email:
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,4 @@ package-lock.json
 # Output
 out
 *.vsix
+.vscode-test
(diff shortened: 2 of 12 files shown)


.travis.yml
```diff
@@ -3,12 +3,20 @@ language: node_js
 node_js:
   - 'stable'
 
+before_install:
```
EricJizbaMSFT: FYI - I got this code from here: https://code.visualstudio.com/docs/extensions/testing-extensions#_running-tests-automatically-on-travis-ci-build-machines

src/util/errorUtil.ts
```diff
@@ -0,0 +1,33 @@
...
+    public errorType: string;
+}
+
+// tslint:disable-next-line:no-any
+export function getErrorData(error: any): ErrorData {
+    const errorData: ErrorData = new ErrorData();
+    if (error instanceof Error) {
+        errorData.errorType = error.constructor.name;
+        errorData.message = error.message;
+    } else if (typeof (error) === 'object' && error !== null) {
```
StephenWeatherford: nit: if typeof error is object, it can't be null.

src/util/errorUtil.ts
```diff
@@ -0,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { localize } from '../util';
+
+export class ErrorData {
```
StephenWeatherford: How about exposing ErrorData via ErrorData.ts and having a constructor that takes an error/any?

EricJizbaMSFT: Makes complete sense. Resolved

PR #17, opened 2017-10-17
Improve entry points for zip deploy

### LINKED ISSUE #15: Fix entry points for zip deploy
My first iteration of zip deploy has two entry points (command palette and functions explorer) with the same "Deploy as Zip Package" label.

Ideally we should have three entry points with a context-specific label for each. Initial proposal:
1. Command palette "Deploy as zip package"
1. Functions explorer: "Deploy here as zip package"
1. File explorer: "Deploy to Function App as zip package"

EricJizbaMSFT:
1. Add 'right click on folder' entry point
1. Add context-specific wording for each entry point
1. Make folder selection consistent across several commands
1. Refactor util.ts into multiple files

Fixes #15

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,16 @@
                 "command": "azureFunctions.deployZip",
                 "title": "%azFunc.deployZip%",
                 "category": "Azure Functions"
+            },
+            {
+                "command": "azureFunctions.deployZipFromExplorer",
+                "title": "%azFunc.deployZipFromExplorer%",
+                "category": "Azure Functions"
+            },
+            {
+                "command": "azureFunctions.deployZipFromFolder",
+                "title": "%azFunc.deployZipFromFolder%",
+                "category": "Azure Functions"
             }
         ],
         "views": {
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -144,10 +154,27 @@
...
+                {
+                    "command": "azureFunctions.deployZipFromFolder",
+                    "when": "explorerResourceIsFolder == true",
+                    "group": "zzz_@1"
+                }
+            ],
+            "commandPalette": [
+                {
+                    "command": "azureFunctions.deployZipFromExplorer",
+                    "when": "never"
```
StephenWeatherford: Why never?

src/utils/workspace.ts
```diff
@@ -0,0 +1,38 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { localize } from '../localize';
+import * as uiUtil from './ui';
+
+export async function selectWorkspaceFolder(placeholder: string): Promise<string> {
+    const selectFolderId: string = 'selectFolder';
+    let folder: uiUtil.PickWithData<string> | undefined;
+    if (vscode.workspace.workspaceFolders) {
+        let folderPicks: uiUtil.PickWithData<string>[] = [new uiUtil.PickWithData(selectFolderId, localize('azFunc.selectFolder', '$(file-directory) Select other folder'))];
```
StephenWeatherford: "Browse..."?

src/utils/workspace.ts
```diff
@@ -0,0 +1,38 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { localize } from '../localize';
+import * as uiUtil from './ui';
+
+export async function selectWorkspaceFolder(placeholder: string): Promise<string> {
+    const selectFolderId: string = 'selectFolder';
```
StephenWeatherford: Use something invalid like ":selectedFolder"?

package.json
EricJizbaMSFT: We have three separate 'deployZip' commands with a context-specific label for each. However, we only want _one_ of them to show in the command palette

PR #26, opened 2017-10-19
Fix restart to be in correct order

### LINKED ISSUE #25: Restart stops the app without starting again
State of app stays at (Stopped).  Confirmed in Portal that the func app is indeed stopped still.

EricJizbaMSFT:
Fixes #25


EricJizbaMSFT:
Nice catch @nturinski!!! 🎉

PR #36, opened 2017-10-26
Use color icons from the portal

EricJizbaMSFT:
screen shot 2017-10-26 at 10 24 22 am screen shot 2017-10-26 at 10 24 06 am


EricJizbaMSFT:
The portal has yellow on white sometimes, too:

screen shot 2017-10-26 at 12 08 26 pm

Jay thinks the added color is worth it, and I tend to agree (see: https://github.com/Microsoft/vscode-azureappservice/pull/151)

We can always change back if people complain :)

PR #42, opened 2017-11-01
Leverage fs-extra instead of defining our own fsUtils

EricJizbaMSFT:
Also:
1. Use secure random string
1. Slightly modify newline rules that are being annoying
1. Remove '--type-check' from lint command since it's deprecated and no longer necessary

### DIFF
--- a/package.json
+++ b/package.json
@@ -222,15 +222,15 @@
     "scripts": {
         "vscode:prepublish": "tsc -p ./",
         "compile": "tsc -watch -p ./",
-        "lint": "tslint --project tsconfig.json -e src/*.d.ts --type-check -t verbose",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -207,15 +207,15 @@
     "scripts": {
         "vscode:prepublish": "tsc -p ./",
         "compile": "tsc -watch -p ./",
-        "lint": "tslint --project tsconfig.json -e src/*.d.ts --type-check -t verbose",
+        "lint": "tslint --project tsconfig.json -e src/*.d.ts -t verbose",
```
StephenWeatherford: why this change?

EricJizbaMSFT: I was getting this message when I ran the lint task:
```
--type-check is deprecated. You only need --project to enable rule which need type information.
```

PR #60, opened 2017-11-09
Add 'Create Function App' to node picker

### LINKED ISSUE #58: Allow user to create new function app during deploy
Today we only show them existing function apps

EricJizbaMSFT:
And remove deploy from function app node (like I meant to do in #56)

Fixes #58

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
         "onCommand:azureFunctions.startFunctionApp",
         "onCommand:azureFunctions.stopFunctionApp",
         "onCommand:azureFunctions.restartFunctionApp",
-        "onCommand:azureFunctions.deployZip",
+        "onCommand:azureFunctions.deploy",
         "onView:azureFunctionsExplorer"
     ],
     "main": "./out/src/extension",
@@ -96,7 +96,7 @@
                 "category": "Azure Functions"
             },
             {
-                "command": "azureFunctions.deployZip",
+                "command": "azureFunctions.deploy",
                 "title": "%azFunc.deployZip%",
                 "category": "Azure Functions",
                 "icon": {
@@ -127,7 +127,7 @@
                     "group": "navigation@2"
                 },
                 {
-                    "command": "azureFunctions.deployZip",
+                    "command": "azureFunctions.deploy",
                     "when": "view == azureFunctionsExplorer",
                     "group": "navigation@3"
                 },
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -167,16 +167,11 @@
...
-                    "when": "view == azureFunctionsExplorer && viewItem == azureFunctionsFunctionApp",
-                    "group": "3@1"
                 }
             ],
             "explorer/context": [
                 {
-                    "command": "azureFunctions.deployZip",
+                    "command": "azureFunctions.deploy",
                     "when": "explorerResourceIsFolder == true",
                     "group": "zzz_@1"
```
nturinski: Just curious, but what's with the ground name 'zzz'?

EricJizbaMSFT: It makes sure its the last command in the context menu alphabetically (since this is the 'folder' context menu and we don't control that menu)

src/nodes/NodeBase.ts
```diff
@@ -3,25 +3,25 @@
...
-    public parent: NodeBase;
+    public readonly parent: NodeBase | undefined;
     public command?: vscode.Command;
     public childType?: string;
+    public readonly createChild?: CreateChildFunction;
 
     private children: NodeBase[] | undefined;
 
-    constructor(id: string, label: string, contextValue?: string, commandId?: string) {
+    constructor(parent: NodeBase | undefined, id: string, label: string, contextValue?: string, commandId?: string) {
```
nturinski: Can't we just make this an optional parameter at the end?

package.json
```diff
@@ -127,7 +127,7 @@
                     "group": "navigation@2"
                 },
                 {
-                    "command": "azureFunctions.deployZip",
+                    "command": "azureFunctions.deploy",
                     "when": "view == azureFunctionsExplorer",
                     "group": "navigation@3"
                 },
```
nturinski: nit: I feel like "Create Function App in Azure" should go above "Refresh" in the context menu

src/nodes/NodeBase.ts
EricJizbaMSFT: I intentionally put it at the front because it's an easy thing to forget and theoretically the only nodes we're adding in the future are child nodes (aka 'parent' is essentially required going forward)

EricJizbaMSFT:
@nturinski nice catches!

PR #63, opened 2017-11-09
Remove reliance on func cli for 'Create new project'

### LINKED ISSUE #18: Output channel should show when creating a function or function app
Per feedback from @StephenWeatherford on a previous PR

### LINKED ISSUE #21: Show better error message when func cli is not installed
Today we show: "'func' is not recognized as an internal or external command"

We should instruct them to install the tools instead

### LINKED ISSUE #24: Prompt for Creating Functions suck
Make the prompt more explicit so the users know that the missing files are required.

EricJizbaMSFT:
Benefits:
1. User can now create project and function without func cli (they only need the cli for debugging)
1. We can now prompt the user to overwrite existing files
1. We can automatically detect if we should git init the folder

This results in a _little bit_ of copied logic between us and the func cli, but the benefits outweigh that drawback

Fixes #24 
Fixes #21 (since func cli is not required until debugging anymore)
Fixes #18 (since func cli output isn't logged in the output window anymore)


README.md
```diff
@@ -1,6 +1,6 @@
 # Azure Functions for Visual Studio Code (Preview)
 
-## Prerequisites
+## Prerequisites for local debugging
 * [Node v8.0+](https://nodejs.org/)
   * Older versions of node will be supported soon. See issue [#1](https://github.com/Microsoft/vscode-azurefunctions/issues/1)
 * [.NET Core 2.0](https://www.microsoft.com/net/download/core)
```
StephenWeatherford: Are you sure this is true?  I had thought apps were supposed to install .net core libs with their app.

src/commands/createFunction.ts
```diff
@@ -40,16 +37,13 @@ function validateTemplateName(rootPath: string, name: string | undefined): strin
...
-        const yes: string = localize('azFunc.yes', 'Yes');
-        const no: string = localize('azFunc.no', 'No');
-        const message: string = localize('azFunc.missingFuncAppFiles', 'The current folder is missing the following function app files: \'{0}\'. Add the missing files?', missingFiles.join(','));
-        const result: string | undefined = await vscode.window.showWarningMessage(message, yes, no);
-        if (result === yes) {
-            await FunctionsCli.createNewProject(outputChannel, functionAppPath);
-        } else {
+async function validateIsFunctionApp(outputChannel: vscode.OutputChannel, functionAppPath: string, ui: IUserInterface): Promise<void> {
+    if (requiredFunctionAppFiles.find((file: string) => !fse.existsSync(path.join(functionAppPath, file))) !== undefined) {
+        const message: string = localize('azFunc.notFunctionApp', 'The current folder is not a function app project. Initialize Project?');
```
StephenWeatherford: selected folder?

src/utils/fs.ts
```diff
@@ -30,3 +47,13 @@ export function getRandomHexString(length: number = 10): string {
...
 }
+
+export function isPathEqual(fsPath1: string, fsPath2: string): boolean {
+    const relativePath: string = path.relative(fsPath1, fsPath2);
+    return relativePath === '';
+}
+
+export function isSubPath(fsPath1: string, fsPath2: string): boolean {
+    const relativePath: string = path.relative(fsPath1, fsPath2);
+    return relativePath !== '' && !relativePath.startsWith('..');
```
StephenWeatherford: nit: seems like !== '' not necessary

src/utils/fs.ts
```diff
@@ -30,3 +47,13 @@ export function getRandomHexString(length: number = 10): string {
     const buffer: Buffer = crypto.randomBytes(Math.ceil(length / 2));
     return buffer.toString('hex').slice(0, length);
 }
+
+export function isPathEqual(fsPath1: string, fsPath2: string): boolean {
+    const relativePath: string = path.relative(fsPath1, fsPath2);
+    return relativePath === '';
+}
+
+export function isSubPath(fsPath1: string, fsPath2: string): boolean {
```
StephenWeatherford: Please comment which is being checked as parent and which as child.

test/fsUtils.test.ts
```diff
@@ -0,0 +1,64 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import * as fsUtils from '../src/utils/fs';
+
+suite('fsUtils Tests', () => {
```
StephenWeatherford: Might want to test with "c:\", etc., and paths without a common prefix

README.md
EricJizbaMSFT: Yep: https://docs.microsoft.com/azure/azure-functions/functions-run-local

They're _supposed_ to bundle .net core libs with their app, but they haven't done that quite yet (After all, the cli is still technically in preview)

src/utils/fs.ts
EricJizbaMSFT: If I remove that check, then you get 'true' if the paths are the same.

It doesn't seem like `/example/path/` is a subfolder of `/example/path/`

Issue #49, opened 2017-11-03
Support add Java Function into Java Function Project


jdneo:
May include following steps:
- select project folder
- if contains ```pom.xml```, treat as Java project, otherwise treat as Javascript
- Using Maven command (leveraging maven function plugin) to add function (in the future may change to use function cli)

Issue #48, opened 2017-11-03
Support Create Java Function Project in extension


jdneo:
May include following steps:
- let user select language(```Node.js```, ```Java```)
- If is Java, using Maven archetype to create a Java project
- generate task.json and launch.json for Java Function Project

PR #66, opened 2017-11-10
Support create java function project & add java function

jdneo:
This PR add the following feature:

- Create java function project ( #48 )
- Add java function ( #49 )

For creating java function project, we leverage [maven archetype](https://github.com/Microsoft/azure-maven-archetypes/tree/master/azure-functions-archetype).

For adding java function, we leverage [azure function maven plugin](https://github.com/Microsoft/azure-maven-plugins/tree/master/azure-functions-maven-plugin). The reason we call it in terminal is that the maven plugin does not support batch mode for this maven goal.

As to the tasks.json file for Java function, we do not use ```dependOn``` for tasks because this may cause ```preLaunchTask cannot be tracked``` error in VS Code 1.18 ([see issue here](https://github.com/Microsoft/vscode/issues/37997)). 

------------- Update on Wed Nov 15 2017 09:12:33 GMT+0800 -------------
Remove the add Java function part, since the Maven plugin not support adding java function non-interactively.


src/commands/createNewProject.ts
```diff
@@ -100,45 +147,137 @@ const localSettingsJson: {} = {
...
+                `-DappName="${appName}"`,
+                '-B' // in Batch Mode
+            );
+
+            functionAppPath = path.join(functionAppPath, artifactId);
+            javaTargetPath = `target/azure-functions/${appName}/`;
+            break;
+        default:
+            // the maven archetype contains these files, so not check them when language is Java
+            const hostJsonPath: string = path.join(functionAppPath, 'host.json');
```
EricJizbaMSFT: Why does the maven archetype have these files? Shouldn't it let VS Code (or the func cli) handle it? Same with the `.gitignore`

src/utils/executor.ts
```diff
@@ -0,0 +1,36 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import * as cp from 'child_process';
+import * as vscode from 'vscode';
+import { localize } from '../localize';
+
+const terminal: vscode.Terminal = vscode.window.createTerminal('Azure Functions');
+
+export function runCommandInTerminal(command: string, addNewLine: boolean = true): void {
+    terminal.show();
```
EricJizbaMSFT: We don't like using terminal because then you have no idea when the command finishes

src/utils/workspace.ts
```diff
@@ -32,3 +35,15 @@ export function isFolderOpenInWorkspace(fsPath: string): boolean {
         return false;
     }
 }
+
+export function getProjectType(projectPath: string): string {
+    let language: string = TemplateLanguage.JavaScript;
+    fse.readdirSync(projectPath).forEach((file: string) => {
```
EricJizbaMSFT: You should use async methods wherever possible (in this case, you would use `readdir` instead of `readdirsync`

src/utils/workspace.ts
```diff
@@ -32,3 +35,15 @@ export function isFolderOpenInWorkspace(fsPath: string): boolean {
...
+        const stat: fse.Stats = fse.statSync(path.join(projectPath, file));
+        // Currently checking the existing pom.xml to determine whether the function project is Java language based.
```
EricJizbaMSFT: Isn't pom.xml at the root of the project? Couldn't you just do `fse.pathExists(path.join(projectPath, 'pom.xml'))`?

If it's not always at the root, then we need to find a different way to detect Java projects - iterating through an entire folder has bad performance for the user

test/createNewProject.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
+        inputs.unshift(undefined); // If the test environment has an open workspace, select the 'Browse...' option
+    }
+
+    const ui: TestUI = new TestUI(inputs);
+    await createNewProject(outputChannel, undefined, false, ui);
+    assert.equal(inputs.length, 0, 'Not all inputs were used.');
+}
+
+async function testInitFileExist(projectPath: string): Promise<void> {
```
EricJizbaMSFT: nit: `testProjectFilesExist`

(I don't know what `InitFile` means)

src/commands/createFunction.ts
```diff
@@ -102,31 +104,40 @@ export async function createFunction(
...
 
-    const localAppSettings: LocalAppSettings = new LocalAppSettings(ui, azureAccount, functionAppPath);
-
-    const templatePicks: PickWithData<Template>[] = (await templateData.getTemplates()).map((t: Template) => new PickWithData<Template>(t, t.name));
-    const templatePlaceHolder: string = localize('azFunc.selectFuncTemplate', 'Select a function template');
-    const template: Template = (await ui.showQuickPick<Template>(templatePicks, templatePlaceHolder)).data;
-
-    if (template.bindingType !== 'httpTrigger') {
-        await localAppSettings.validateAzureWebJobsStorage();
+    const languageType: string = workspaceUtil.getProjectType(functionAppPath);
```
EricJizbaMSFT: Why do you infer the project type? I can easily create C# and JavaScript functions in the same project from the command line today.

Shouldn't I be able to create 'Java' and 'JavaScript' functions in the same project from VS Code?

src/utils/workspace.ts
```diff
@@ -32,3 +35,15 @@ export function isFolderOpenInWorkspace(fsPath: string): boolean {
         return false;
     }
 }
+
+export function getProjectType(projectPath: string): string {
```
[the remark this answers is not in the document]
jdneo: I think I'll use this method for deploying Java functions. How about make a new file called 'project.ts' and move to it?

src/utils/workspace.ts
jdneo: Yes it's at the root path. I like the idea of using ```fse.pathExists()```

src/utils/workspace.ts
jdneo: Using ```fse.pathExists()```.

test/createNewProject.test.ts
jdneo: Make sense.

src/utils/executor.ts
```diff
@@ -0,0 +1,36 @@
...
+    terminal.sendText(command, addNewLine);
+}
+
+export async function executeCommand(outputChannel: vscode.OutputChannel, workingDirectory: string, command: string, ...args: string[]): Promise<void> {
```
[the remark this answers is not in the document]
jdneo: 👍

src/commands/createFunction.ts
jdneo: How about let the user choose the language he/she wants to add by showing Picks?

src/commands/createNewProject.ts
```diff
@@ -105,6 +152,50 @@ export async function createNewProject(outputChannel: OutputChannel, functionApp
...
+                '-DarchetypeArtifactId="azure-functions-archetype"',
+                `-DgroupId="${groupId}"`,
+                `-DartifactId="${artifactId}"`,
+                `-Dversion="${version}"`,
+                `-Dpackage="${packageName}"`,
+                `-DappName="${appName}"`,
+                '-B' // in Batch Mode
+            );
+
+            functionAppPath = path.join(functionAppPath, artifactId);
```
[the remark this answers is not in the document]
jdneo: Do you mean the extra folder of ```artifactId```?

src/utils/executor.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+const terminal: vscode.Terminal = vscode.window.createTerminal('Azure Functions');
```
EricJizbaMSFT: If the user closes/deletes the 'Azure Functions' terminal, then create new project will fail every time until they restart VS Code

src/commands/createNewProject.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yes

src/commands/createNewProject.ts
EricJizbaMSFT: Do you own the maven archetype or does someone else? I thought you guys created it

src/commands/createNewProject.ts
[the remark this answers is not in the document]
EricJizbaMSFT: https://github.com/Microsoft/vscode-azurefunctions/issues/69

Issue #50, opened 2017-11-03
Support deploy Java Function


jdneo:
May contain following steps:
- select folder
- if not contains ```pom.xml```, zip and upload
- if contains ```pom.xml```, treat as Java Function Project, list folders under ./target/azure-functions(which are output of the ```mvn package``` command)
- select a folder, zip and deploy

PR #70, opened 2017-11-16
Add support for java function deploy

jdneo:
Work item: #50 

- Infer the language type by checking if the folder has 'pom.xml'
- If is Java type, run `mvn clean package`, point to the target folder (the folder contains packaged *.jar file)
- The deployment API uses the same API as JS function

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -3,28 +3,87 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// tslint:disable-next-line:no-require-imports
-import WebSiteManagementClient = require('azure-arm-website');
+import * as fse from 'fs-extra';
+import * as path from 'path';
 import * as vscode from 'vscode';
+import { SiteWrapper } from 'vscode-azureappservice';
 import { AzureFunctionsExplorer } from '../AzureFunctionsExplorer';
-import { IUserInterface } from '../IUserInterface';
+import { NoPackagedJavaFunctionError } from '../errors';
+import { IUserInterface, Pick } from '../IUserInterface';
 import { localize } from '../localize';
 import { FunctionAppNode } from '../nodes/FunctionAppNode';
 import { getWebSiteClient } from '../nodes/SubscriptionNode';
+import { TemplateLanguage } from '../templates/Template';
+import { cpUtils } from '../utils/cpUtils';
+import { getProjectType } from '../utils/project';
 import * as workspaceUtil from '../utils/workspace';
 import { VSCodeUI } from '../VSCodeUI';
+
+/* tslint:disable:no-require-imports */
+import WebSiteManagementClient = require('azure-arm-website');
+import StringDictionary = require('azure-arm-website/lib/models/StringDictionary');
 
 export async function deploy(explorer: AzureFunctionsExplorer, outputChannel: vscode.OutputChannel, context?: FunctionAppNode | vscode.Uri, ui: IUserInterface = new VSCodeUI()): Promise<void> {
     const uri: vscode.Uri | undefined = context && context instanceof vscode.Uri ? context : undefined;
     let node: FunctionAppNode | undefined = context && context instanceof FunctionAppNode ? context : undefined;
 
-    const folderPath: string = uri ? uri.fsPath : await workspaceUtil.selectWorkspaceFolder(ui, localize('azFunc.selectZipDeployFolder', 'Select the folder to zip and deploy'));
+    let folderPath: string = uri ? uri.fsPath : await workspaceUtil.selectWorkspaceFolder(ui, localize('azFunc.selectZipDeployFolder', 'Select the folder to zip and deploy'));
 
     if (!node) {
         node = <FunctionAppNode>(await explorer.showNodePicker(FunctionAppNode.contextValue));
     }
 
     const client: WebSiteManagementClient = getWebSiteClient(node);
+    const siteWrapper: SiteWrapper = node.siteWrapper;
+    const languageType: string = await getProjectType(folderPath);
+    if (languageType === TemplateLanguage.Java) {
+        folderPath = await getJavaFolderPath(outputChannel, folderPath, ui);
+        await updateJavaFunctionAppSettings(outputChannel, client, siteWrapper);
+    }
 
-    await node.siteWrapper.deployZip(folderPath, client, outputChannel);
+    await siteWrapper.deployZip(folderPath, client, outputChannel);
+}
+
+async function getJavaFolderPath(outputChannel: vscode.OutputChannel, basePath: string, ui: IUserInterface): Promise<string> {
+    outputChannel.show();
+    await cpUtils.executeCommand(outputChannel, basePath, 'mvn', 'clean', 'package', '-B');
+    const targetFolder: string = path.join(basePath, 'target', 'azure-functions');
+    if (!fse.pathExistsSync(targetFolder)) {
+        throw new NoPackagedJavaFunctionError();
+    }
+    const packagedFolders: string[] = fse.readdirSync(targetFolder);
+    if (packagedFolders.length === 0) {
+        throw new NoPackagedJavaFunctionError();
+    } else if (packagedFolders.length === 1) {
+        return path.join(targetFolder, packagedFolders[0]);
+    } else {
+        return path.join(targetFolder, await promptForPackagedFolder(ui, packagedFolders));
+    }
+}
+
+async function promptForPackagedFolder(ui: IUserInterface, folders: string[]): Promise<string> {
+    const picks: Pick[] = folders.reduce(
+        (ret: Pick[], item: string): Pick[] => {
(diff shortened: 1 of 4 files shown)


src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
-    await node.siteWrapper.deployZip(folderPath, client, outputChannel);
+    const placeHolder: string = localize('azFunc.PackagedFolderPlaceholder', 'Select packaged folder you want to deploy');
+    return (await ui.showQuickPick(picks, placeHolder, false)).label;
+}
+
+async function updateJavaFunctionAppSettings(outputChannel: vscode.OutputChannel, client: WebSiteManagementClient, siteWrapper: SiteWrapper): Promise<void> {
+    const appSettings: StringDictionary = await client.webApps.listApplicationSettings(siteWrapper.resourceGroup, siteWrapper.appName);
+    // tslint:disable-next-line:no-string-literal
+    if (appSettings['properties']['FUNCTIONS_EXTENSION_VERSION'] !== 'beta') {
+        outputChannel.appendLine(localize('azFunc.updateJavaFunctionRuntime', 'Updating FUNCTIONS_EXTENSION_VERSION to beta to support Java runtime...'));
```
EricJizbaMSFT: You should prompt the user before changing their function app to a beta runtime version

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
     if (!node) {
         node = <FunctionAppNode>(await explorer.showNodePicker(FunctionAppNode.contextValue));
     }
 
     const client: WebSiteManagementClient = getWebSiteClient(node);
+    const siteWrapper: SiteWrapper = node.siteWrapper;
+    const languageType: string = await getProjectType(folderPath);
+    if (languageType === TemplateLanguage.Java) {
+        folderPath = await getJavaFolderPath(outputChannel, folderPath, ui);
+        await updateJavaFunctionAppSettings(outputChannel, client, siteWrapper);
```
EricJizbaMSFT: Change this function name to something like `verifyBetaRuntime`

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
+        throw new NoPackagedJavaFunctionError();
+    } else if (packagedFolders.length === 1) {
+        return path.join(targetFolder, packagedFolders[0]);
+    } else {
+        return path.join(targetFolder, await promptForPackagedFolder(ui, packagedFolders));
+    }
+}
+
+async function promptForPackagedFolder(ui: IUserInterface, folders: string[]): Promise<string> {
+    const picks: Pick[] = folders.reduce(
```
EricJizbaMSFT:
```
folders.map((f: string) => new Pick(f))
```

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
+    const languageType: string = await getProjectType(folderPath);
+    if (languageType === TemplateLanguage.Java) {
+        folderPath = await getJavaFolderPath(outputChannel, folderPath, ui);
+        await updateJavaFunctionAppSettings(outputChannel, client, siteWrapper);
+    }
+
+    await siteWrapper.deployZip(folderPath, client, outputChannel);
+}
+
+async function getJavaFolderPath(outputChannel: vscode.OutputChannel, basePath: string, ui: IUserInterface): Promise<string> {
```
EricJizbaMSFT: Theres should only ever be one package per function app project. You could use `project.properties.functionAppName` in `pom.xml` to determine the name of the package.

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
 
     if (!node) {
         node = <FunctionAppNode>(await explorer.showNodePicker(FunctionAppNode.contextValue));
     }
 
     const client: WebSiteManagementClient = getWebSiteClient(node);
+    const siteWrapper: SiteWrapper = node.siteWrapper;
+    const languageType: string = await getProjectType(folderPath);
+    if (languageType === TemplateLanguage.Java) {
+        folderPath = await getJavaFolderPath(outputChannel, folderPath, ui);
```
EricJizbaMSFT: So if I understand this correctly, I can _only_ have Java functions in my project? Seems like other languages would not get deployed

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
 import { localize } from '../localize';
 import { FunctionAppNode } from '../nodes/FunctionAppNode';
 import { getWebSiteClient } from '../nodes/SubscriptionNode';
+import { TemplateLanguage } from '../templates/Template';
+import { cpUtils } from '../utils/cpUtils';
+import { getProjectType } from '../utils/project';
 import * as workspaceUtil from '../utils/workspace';
 import { VSCodeUI } from '../VSCodeUI';
 
+/* tslint:disable:no-require-imports */
```
EricJizbaMSFT: Don't disable this for the entire file. We still prefer non-require imports where possible

And keep all imports in the same grouping (they're automatically alphabetized when you save the file, but only if they're together)

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
+import WebSiteManagementClient = require('azure-arm-website');
+import StringDictionary = require('azure-arm-website/lib/models/StringDictionary');
```
EricJizbaMSFT: Don't use a require import for this.
```
import { StringDictionary } from 'azure-arm-website/lib/models';
```

src/utils/project.ts
```diff
@@ -0,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
+import * as path from 'path';
+import { TemplateLanguage } from '../templates/Template';
+
+// tslint:disable-next-line:export-name
```
EricJizbaMSFT: If you wrap this in 
```
export namespace projectUtils {

}
```

Then you don't have to disable this rule

src/commands/deploy.ts
```diff
@@ -3,28 +3,87 @@
...
+    }
+
+    await siteWrapper.deployZip(folderPath, client, outputChannel);
+}
+
+async function getJavaFolderPath(outputChannel: vscode.OutputChannel, basePath: string, ui: IUserInterface): Promise<string> {
+    outputChannel.show();
+    await cpUtils.executeCommand(outputChannel, basePath, 'mvn', 'clean', 'package', '-B');
+    const targetFolder: string = path.join(basePath, 'target', 'azure-functions');
+    if (!fse.pathExistsSync(targetFolder)) {
```
EricJizbaMSFT: Don't use the 'Sync' versions of methods

src/errors.ts
```diff
@@ -17,7 +17,11 @@ export class ArgumentError extends Error {
     }
 }
 
-// tslint:disable-next-line:max-classes-per-file
+/* tslint:disable:max-classes-per-file */
```
EricJizbaMSFT: nit: Why do you use block comments like this? Just use a single line comment

src/commands/deploy.ts
jdneo: Emmm, actually, I think yes you can (at least partially). For example, you can have a js function named blobTriggerJs. That means you will have a folder named blobTriggerJs where contains the js source and function property files. If you select this blobTriggerJs folder, this folder will be zipped and deployed (because the blobTriggerJs folder has no pom file in it).

But if we want to deploy all the functions in the project folder in one deployment (choose the project root folder), no we cannot achieve that.

src/commands/deploy.ts
jdneo: Yes, that's a good idea. I Will confirm with the Maven plugin team to check if it is safe to do that. Let's file a issue for this, shall we?

src/commands/deploy.ts
EricJizbaMSFT: You can't deploy blobTriggerJs by itself - you have to deploy a project (that contains functions).

That's fine - let's just make sure to only display Java function templates if the current project is Java

src/commands/deploy.ts
EricJizbaMSFT: Here's what the outputDirectory is in pom.xml:
```
<outputDirectory>${project.build.directory}/azure-functions/${functionAppName}</outputDirectory>
```

You already hard-coded `azure-functions` into the path from your code, so I don't see why it would be a problem to also use functionAppName. Technically it is possible the user could change this, but I think the benefits for most users outweighs the negatives in this specific edge case.

EricJizbaMSFT: I'm okay with occasionally filing issues and addressing feedback after a PR is merged, but I don't like approving a PR before the issue is filed (otherwise it could be forgotten).

I want to merge code as soon as it's ready so I filed the issue for you this time (https://github.com/Microsoft/vscode-azurefunctions/issues/74). In the future, please just file the issue rather than asking "Let's file an issue for this, shall we?"

Issue #69, opened 2017-11-15
Create Java Project adds extra folder


EricJizbaMSFT:
<img width="609" alt="screen shot 2017-11-06 at 6 13 59 pm" src="https://user-images.githubusercontent.com/11282622/32473862-8385ec76-c31e-11e7-94d7-37e59f24c436.png">

I will never put anything else in 'myFunctionProject' folder. There should be no intermediate folder

EricJizbaMSFT:
We should prompt before overwriting. Just use the exisiting 'confirmOverwriteFile': https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/utils/fs.ts#L18

jdneo:
That's good.

PR #79, opened 2017-11-22
fix issue #69

jdneo:
Related discussion can be referenced in #69

### DIFF
--- a/src/commands/createNewProject.ts
+++ b/src/commands/createNewProject.ts
@@ -181,7 +181,7 @@ async function promotForMavenParameters(ui: IUserInterface, functionAppPath: str
 async function createJavaFunctionProject(outputChannel: OutputChannel, functionAppPath: string, ui: IUserInterface): Promise<string> {
     // Get parameters for Maven command
     const { groupId, artifactId, version, packageName, appName } = await promotForMavenParameters(ui, functionAppPath);
-    const tempFolder: string = path.join(os.tmpdir(), Date.now().toString());
+    const tempFolder: string = path.join(os.tmpdir(), fsUtil.getRandomHexString());
     await fse.ensureDir(tempFolder);
     // Use maven command to init Java function project.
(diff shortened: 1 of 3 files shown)


src/commands/createNewProject.ts
```diff
@@ -177,6 +178,31 @@ async function promotForMavenParameters(ui: IUserInterface, functionAppPath: str
     };
 }
 
+async function createJavaFunctionProject(outputChannel: OutputChannel, functionAppPath: string, ui: IUserInterface): Promise<string> {
+    // Get parameters for Maven command
+    const { groupId, artifactId, version, packageName, appName } = await promotForMavenParameters(ui, functionAppPath);
+    const tempFolder: string = path.join(os.tmpdir(), Date.now().toString());
```
EricJizbaMSFT: Use `fsUtil.getRandomHexString()` instead of `Date.now()`

src/commands/createNewProject.ts
```diff
@@ -177,6 +178,31 @@ async function promotForMavenParameters(ui: IUserInterface, functionAppPath: str
     };
 }
 
+async function createJavaFunctionProject(outputChannel: OutputChannel, functionAppPath: string, ui: IUserInterface): Promise<string> {
+    // Get parameters for Maven command
+    const { groupId, artifactId, version, packageName, appName } = await promotForMavenParameters(ui, functionAppPath);
```
EricJizbaMSFT: Do we still need to prompt for the artifactId if we're inferring it from the original path selected?

src/commands/createNewProject.ts
jdneo: 👍

src/commands/createNewProject.ts
jdneo: I think, yes we need. Because it is a property in the pom file, and can be different from the name of the project folder.

PR #87, opened 2017-11-28
Add function and app setting tree items

EricJizbaMSFT:
The app setting icons are the exact same as the azure app service extension. The function icons I got from the portal (the 'f' is not great in dark mode, but we're consistent with the portal):
screen shot 2017-11-28 at 11 01 36 amscreen shot 2017-11-28 at 11 02 55 am

Here's what it looks like in the portal 
screen shot 2017-11-28 at 11 02 13 amscreen shot 2017-11-28 at 11 02 37 am

Builds will fail until PR from shared package is merged: https://github.com/Microsoft/vscode-azuretools/pull/31


src/extension.ts
```diff
@@ -54,12 +58,17 @@ export function activate(context: vscode.ExtensionContext): void {
         initCommand<IAzureNode<FunctionAppTreeItem>>(context, outputChannel, 'azureFunctions.openInPortal', async (node?: IAzureNode<FunctionAppTreeItem>) => await openInPortal(tree, node));
         initAsyncCommand<IAzureNode>(context, outputChannel, 'azureFunctions.createFunction', async () => await createFunction(outputChannel, azureAccount, templateData));
         initAsyncCommand<IAzureNode>(context, outputChannel, 'azureFunctions.createNewProject', async () => await createNewProject(outputChannel));
-        initAsyncCommand<IAzureParentNode>(context, outputChannel, 'azureFunctions.createFunctionApp', async (node?: IAzureParentNode) => await createFunctionApp(tree, node));
+        initAsyncCommand<IAzureParentNode>(context, outputChannel, 'azureFunctions.createFunctionApp', async (node?: IAzureParentNode) => await createChildNode(tree, 'azureSubscription', node));
```
nturinski: How come this is the only context value that is hardcoded?

src/tree/FunctionAppTreeItem.ts
```diff
@@ -6,38 +6,65 @@
...
+
+    public async loadMoreChildren(_node: IAzureNode<IAzureTreeItem>, _clearCache: boolean | undefined): Promise<IAzureTreeItem[]> {
+        return [this._functionsTreeItem, this._appSettingsTreeItem];
+    }
+
+    public pickTreeItem(expectedContextValue: string): IAzureParentTreeItem | undefined {
+        switch (expectedContextValue) {
+            case FunctionTreeItem.contextValue:
+                return this._functionsTreeItem;
+            case AppSettingsTreeItem.contextValue:
```
nturinski: Aren't these different TreeItems?  Should it fall through like that?

EricJizbaMSFT: The node picker is traveling down the tree. In this example, I'm basically saying "If you're looking for a AppSettingsTreeItem _or_ AppSettingTreeItem, go to this. _appSettingsTreeItem next". After that, the node picker will continue down the tree as applicable

PR #89, opened 2017-11-29
Send telemetry event with new project/function details

EricJizbaMSFT:


src/commands/createFunction.ts
```diff
@@ -182,4 +184,8 @@ export async function createFunction(
 
     const newFileUri: vscode.Uri = vscode.Uri.file(newFilePath);
     vscode.window.showTextDocument(await vscode.workspace.openTextDocument(newFileUri));
+
+    if (telemetryReporter) {
```
StephenWeatherford: Would it be better to hang this onto the main azureFunctions.createFunction/createProject events?  We'd need a way to do that, though, so may not be worth it.

src/commands/createNewProject.ts
```diff
@@ -204,7 +204,7 @@ async function createJavaFunctionProject(outputChannel: OutputChannel, functionA
     return appName;
 }
 
-export async function createNewProject(outputChannel: OutputChannel, functionAppPath?: string, openFolder: boolean = true, ui: IUserInterface = new VSCodeUI()): Promise<void> {
+export async function createNewProject(outputChannel: OutputChannel, telemetryReporter: TelemetryReporter | undefined, functionAppPath?: string, openFolder: boolean = true, ui: IUserInterface = new VSCodeUI()): Promise<void> {
```
nturinski: Why not just make this optional?

src/commands/createFunction.ts
EricJizbaMSFT: Refactored to do just that - let me know what you think

src/commands/createNewProject.ts
EricJizbaMSFT: I refactored this anyways, but there's a difference between 'optional' parameters and 'required parameters that could be undefined'. The reporter should _always_ be passed (it just happens to be undefined if there was an error creating it).

src/commands/createFunction.ts
StephenWeatherford: Cool.  Are we still guaranteed to receive the event even if you open a new workspace?

EricJizbaMSFT: Not sure. I'll look into that separately: https://github.com/Microsoft/vscode-azurefunctions/issues/93

I don't want to create a separate telemetry event if I don't have to

Issue #94, opened 2017-11-29
Error message for not having Maven installed not user friendly


nturinski:
If Maven is not installed when attempting to create a Java Project, the error message looks like this

![image](https://user-images.githubusercontent.com/5290572/33404757-6046748e-d51a-11e7-8fad-f65cb4b143c8.png)

We should make the message more user-legible.

EricJizbaMSFT:
@jdneo I put this in 0.3.0 in case you have time to fix it - but not a big deal if we have to punt it to the next release

EricJizbaMSFT:
In the interest of locking down for the release, I'm moving this to 0.4.0

PR #99, opened 2017-11-30
Check if command exist in cpUtils

jdneo:
work item: #94

image

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -19,6 +19,7 @@ import { localize } from '../localize';
 import { TemplateLanguage } from '../templates/Template';
 import { FunctionAppTreeItem } from '../tree/FunctionAppTreeItem';
 import { cpUtils } from '../utils/cpUtils';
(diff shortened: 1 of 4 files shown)


jdneo:
yes this seems better than a generic solution

src/commands/deploy.ts
```diff
@@ -46,6 +47,9 @@ export async function deploy(tree: AzureTreeDataProvider, outputChannel: vscode.
 }
 
 async function getJavaFolderPath(outputChannel: vscode.OutputChannel, basePath: string, ui: IUserInterface): Promise<string> {
+    if (!(await mavenUtils.isMavenInstalled(basePath))) {
+        throw new Error(localize('azFunc.mvnNotFound', 'Failed to find "maven" on path.'));
```
EricJizbaMSFT: Don't copy code like this - you could just refactor `isMavenInstalled` to `validateMavenInstalled` and have it throw the error

Issue #102, opened 2017-12-01
Error occurs if folder path contains whitespace


Eskibear:
[platform: osx]

I selected a folder named `test folder`  in `Create New Project` step. Two errors block me along the way.
* In inputbox of artifactId, the default value is same as folder name, i.e. `test folder`, which should be illegal for artifactId. Should add validation, otherwise it fails the maven command.

* In `tasks.json`, the whitespace also breaks the maven command, should be quoted or escaped.
    ```
    "osx" : {
        "command": "sh -c \"mvn clean package -B && func host start --script-root target/azure-
    functions/test folder-1512092704461/\""
    }
    ```
    It's also supposed to fail in other platform.

jdneo:
The first point has already been fixed in #97

PR #104, opened 2017-12-01
Add function name check and fix some bugs

jdneo:
In this PR, I did the following changes here:

**In ```src/commands/createFunction.ts```:**
change the function name validation according to the portal hit:

**In ```src/commands/createNewProject.ts```:** 
add ```"``` around %path% in case that the path contains spaces #102 
Show the outpput channel when creating java function project

**In ```test/createFunction.test.ts```:**
Update the unit test code according to the code change

**In ```src/utils/javaNameUtils.ts```:**
Since the function name will also used for the Java class name, so the name should not be a reserved keyword, and both obey java name identifier rule and function name rule (no '-' and '$').

### DIFF
--- a/src/commands/createFunction.ts
+++ b/src/commands/createFunction.ts
@@ -25,6 +25,8 @@ import * as workspaceUtil from '../utils/workspace';
 import { VSCodeUI } from '../VSCodeUI';
 import { createNewProject } from './createNewProject';
 
+const functionNameRegex: RegExp = /^[a-zA-Z][a-zA-Z\d_\-]*$/;
+
 const requiredFunctionAppFiles: string[] = [
     'host.json',
     'local.settings.json',
@@ -41,6 +43,9 @@ function validateTemplateName(rootPath: string, name: string | undefined, langua
     } else {
         if (fse.existsSync(path.join(rootPath, name))) {
             return localize('azFunc.existingFolderError', 'A folder with the name \'{0}\' already exists.', name);
+        }
+        if (!functionNameRegex.test(name)) {
+            return localize('azFunc.functionNameInvalidError', 'Function name must start with a letter and can contain letters, digits, \'_\' and \'-\'');
         }
         return undefined;
     }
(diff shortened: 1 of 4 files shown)


src/utils/javaNameUtils.ts
```diff
@@ -50,7 +51,13 @@ export function getJavaClassName(name: string): string {
 }
 
 export function validateJavaFunctionName(name: string): string | undefined {
-    return validateJavaName(name);
+    if (isKeyword(name)) {
+        return localize('azFunc.JavaNameIsKeywordError', '\'{0}\' is a reserved keyword.', name);
+    }
+    if (!javaFunctionRegex.test(name)) {
```
EricJizbaMSFT: Don't copy this code. It's the same as the other languages and should only be in createFunction.ts

src/commands/createFunction.ts
```diff
@@ -25,6 +25,8 @@ import * as workspaceUtil from '../utils/workspace';
 import { VSCodeUI } from '../VSCodeUI';
 import { createNewProject } from './createNewProject';
 
+const functionNameRegex: RegExp = /^[a-zA-Z][a-zA-Z\d_\-]*$/;
```
EricJizbaMSFT: Don't use global variables when you don't have to. Put this in the context of `validateTemplateName`

(`requiredFunctionAppFiles` is a bad example because it was used in multiple places when the code was originally written)

src/utils/javaNameUtils.ts
```diff
@@ -50,7 +51,13 @@ export function getJavaClassName(name: string): string {
 }
 
 export function validateJavaFunctionName(name: string): string | undefined {
-    return validateJavaName(name);
+    if (isKeyword(name)) {
+        return localize('azFunc.JavaNameIsKeywordError', '\'{0}\' is a reserved keyword.', name);
```
EricJizbaMSFT: Refactor to get rid of this copied code - the error is the exact same as above

src/utils/javaNameUtils.ts
jdneo: Emm, I leave this code here is because currently the maven plugin use the function name to be the Java class name as well. Which means the java function name should both follow the function naming rule and Java class nameing rule (they are different since Java class not allow '-' in it). 

I can refactor this code when the maven plugin change the logic to name the java class.

EricJizbaMSFT: Ah - I thought the regex was the exact same. Didn't realize the hyphen was different. This is fine

PR #114, opened 2017-12-04
Log 'delete function' in output channel

### LINKED ISSUE #110: Delete Function does not show user any output
When you delete a Function in a Function App, after the confirming the prompt, there is no output in the output channel confirming that the Function was deleted.

EricJizbaMSFT:
Fixes #110

### DIFF
--- a/src/tree/FunctionTreeItem.ts
+++ b/src/tree/FunctionTreeItem.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as vscode from 'vscode';
+import { OutputChannel } from 'vscode';
 import { SiteWrapper } from 'vscode-azureappservice';
 import { IAzureNode, IAzureTreeItem, UserCancelledError } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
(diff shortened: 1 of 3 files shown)


src/tree/FunctionTreeItem.ts
```diff
@@ -49,8 +52,11 @@ export class FunctionTreeItem implements IAzureTreeItem {
     public async deleteTreeItem(node: IAzureNode<FunctionTreeItem>): Promise<void> {
         const message: string = localize('azFunc.ConfirmDelete', 'Are you sure you want to delete "{0}"?', this._name);
         if (await vscode.window.showWarningMessage(message, DialogResponses.yes, DialogResponses.cancel) === DialogResponses.yes) {
+            this._outputChannel.show(true);
+            this._outputChannel.appendLine(localize('DeletingFunction', 'Deleting function "{0}"...', this._name));
```
nturinski: Is this worth moving to the shared npm package?  The functionality is the same, we would just need to check when to use client.functionModel and client.webApps, right?

EricJizbaMSFT: IMO - not worth it. The code isn't that complex and it is slightly different for functions vs. sites.

We could move it if other extensions ever need to delete a function.

PR #122, opened 2017-12-07
Show in-between states for stop/start

EricJizbaMSFT:


src/tree/FunctionAppTreeItem.ts
```diff
@@ -40,7 +40,7 @@ export class FunctionAppTreeItem implements IAzureParentTreeItem {
     }
 
     public get label(): string {
-        return this._state === 'Running' ? this.siteWrapper.name : `${this.siteWrapper.name} (${this._state})`;
+        return this.state === 'Running' ? this.siteWrapper.name : `${this.siteWrapper.name} (${this.state})`;
```
StephenWeatherford: I think Jay may only be showing when not running?  Can you make sure we're consistent across extensions?

StephenWeatherford: What if it's empty or undefined?

src/tree/FunctionAppTreeItem.ts
```diff
@@ -18,18 +18,18 @@ export class FunctionAppTreeItem implements IAzureParentTreeItem {
...
-    private readonly _state: string | undefined;
     private readonly _outputChannel: OutputChannel;
 
     public constructor(site: Site, outputChannel: OutputChannel) {
         this.siteWrapper = new SiteWrapper(site);
         if (!site.state) {
             throw new ArgumentError(site);
         }
-        this._state = site.state;
+        this.state = site.state;
```
StephenWeatherford: Don't quite get the usage here.  Is site.state obtained from Azure when the wrapper is originally created?  Perhaps it would be better to require using getState always?

EricJizbaMSFT: Yes, site.state is the state from when the instance of 'site' is originally created. I think it makes sense to use that in the constructor of the node? Using 'getState' feels redundant in this case to me...

Also we can't _always_ use getState because it doesn't give use the in-between values of 'starting' or 'stopping'

src/tree/FunctionAppTreeItem.ts
EricJizbaMSFT: This also only shows state when it's not running...I don't see how it's different than Jay's...

I don't know why it would ever be empty or undefined?

PR #124, opened 2017-12-08
Add 'Copy function url' command

### LINKED ISSUE #92: Add 'Copy Url' command to http functions in explorer

EricJizbaMSFT:
And make the parsing of 'function.json' more robust

Fixes #92

### DIFF
--- a/src/FunctionConfig.ts
+++ b/src/FunctionConfig.ts
@@ -1,0 +1,95 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ErrorData } from "./ErrorData";
+import { localize } from "./localize";
+
+interface IFunctionJson {
+    disabled?: boolean;
+    bindings: IFunctionBinding[];
+}
+
+interface IFunctionBinding {
+    // tslint:disable-next-line:no-reserved-keywords
+    type?: string;
+    name?: string;
+    direction?: string;
+    authLevel?: string;
+    [propertyName: string]: string | undefined;
+}
+
+export enum HttpAuthLevel {
+    admin = 'admin',
+    function = 'function',
+    anonymous = 'anonymous'
+}
+
+enum BindingDirection {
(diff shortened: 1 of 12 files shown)


src/commands/createFunction.ts
```diff
@@ -142,7 +142,7 @@ export async function createFunction(
     const template: Template = (await ui.showQuickPick<Template>(templatePicks, templatePlaceHolder)).data;
     telemetryProperties.templateId = template.id;
 
-    if (template.bindingType !== 'httpTrigger') {
+    if (template.functionConfig.isHttpTrigger) {
```
nturinski: The logic has flipped here.  Is that intentional?

EricJizbaMSFT: Noooooopeeeee not intentional :)

src/FunctionConfig.ts
```diff
@@ -0,0 +1,83 @@
...
+    public readonly inBindingType: string;
+    public readonly isHttpTrigger: boolean = false;
+    public readonly authLevel: HttpAuthLevel = HttpAuthLevel.function;
+
+    // tslint:disable-next-line:no-any
+    public constructor(data: any) {
+        let parseError: string | undefined;
+
+        this.functionJson = <IFunctionJson>data;
+        this.disabled = this.functionJson.disabled === true;
```
StephenWeatherford: Could data be null or undefined?

src/FunctionConfig.ts
```diff
@@ -0,0 +1,83 @@
...
+
+        if (!this.functionJson.bindings || !(this.functionJson.bindings instanceof Array)) {
```
StephenWeatherford: I think it would be better if you verified data before assigning it to this.functionJson, so that after it's assigned we know it's valid.

In which case you should change IFunctionJson.bindings to be non-optional.

src/FunctionConfig.ts
```diff
@@ -0,0 +1,83 @@
...
+            parseError = localize('expectedBindings', 'Expected "bindings" element of type "Array".');
```
StephenWeatherford: There are other things that can go wrong here, such as unexpected types.  How about putting the whole thing in a try/catch?

EricJizbaMSFT:
ping @StephenWeatherford

PR #130, opened 2017-12-13
Add language and runtime user settings

EricJizbaMSFT:
The only languages that show up when creating a new project are 'JavaScript' and 'Java', so they're still the only "officially" supported languages. However, this enables users to modify their settings and create functions from other languages/runtimes. They still can't debug it in VS Code, but they can create/deploy.

I initially did this work to support C#, but it turns out the C# templates from the functions portal are not the ones we want to use. We want to mirror the behavior of VS and that will take more time to iron out.


package.json
```diff
@@ -333,6 +333,29 @@
                         "type": "string",
                         "default": "",
                         "description": "Defines which files in the workspace to ignore for Zip deploy. This applies to Zip deploy only, has no effect on other deployment methods."
+                    },
+                    "azureFunctions.projectRuntime": {
+                        "type": "string",
+                        "enum": [
+                            "~1",
```
StephenWeatherford: Do the runtime enum values depend on the language?

StephenWeatherford: I thought there was a ~2 available?

package.nls.json
```diff
@@ -14,6 +14,8 @@
     "azFunc.deleteFunction": "Delete Function",
     "azFunc.showExplorerDescription": "Show or hide the Azure Functions Explorer",
     "azFunc.templateFilterDescription": "Specify the templates to display when creating a new function. The supported values are 'Verified', 'Core', and 'All'. The 'Verified' category is a subset of 'Core' that has been verified to work with the latest VS Code extension.",
+    "azFunc.projectRuntimeDescription": "The default runtime to use when performing operations in the Azure Functions extension (e.g. \"Create New Function\"). \"~1\" is currently the latest version approved for production use.",
+    "azFunc.projectLanguageDescription": "The default language to use when performing operations in the Azure Functions extension (e.g. \"Create New Function\"). \"F#\", and \"JavaScript\" are currently the only languages approved for production use with runtime \"~1\".",
```
StephenWeatherford: Need to say something about not being able to debug?

package.json
```diff
@@ -333,6 +333,29 @@
                         "type": "string",
                         "default": "",
                         "description": "Defines which files in the workspace to ignore for Zip deploy. This applies to Zip deploy only, has no effect on other deployment methods."
+                    },
+                    "azureFunctions.projectRuntime": {
+                        "type": "string",
+                        "enum": [
```
StephenWeatherford: Why not allow "latest"?

package.json
EricJizbaMSFT: I thought so, too, but it has weird behavior.

screen shot 2017-12-14 at 11 14 11 am

'~2' is _supposed_ to mean v2, but that message says it's older than v1. It's possible '~2' won't be supported as a version until 'beta' is GA'ed

package.json
EricJizbaMSFT: I chose to be consistent with the portal, which only allows '~1' or 'beta':
screen shot 2017-12-14 at 11 09 03 am

Let me know if you still think I should allow latest and I will address it separately

package.nls.json
EricJizbaMSFT: Sure

PR #131, opened 2017-12-15
Show http trigger urls in deploy log

### LINKED ISSUE #72: Include function URL in logs
Love this extension! Thank you for creating it. 

Right now when publishing a new function to Azure the logs will give you the link to the function app but not for the individual functions. So this means that in order to test that my function has been deployed successfully I still need to go open the portal and copy manually the URL. Would be great to have the function url included in the logs.

EricJizbaMSFT:
Fixes #72


src/commands/deploy.ts
```diff
@@ -49,7 +52,29 @@ export async function deploy(telemetryProperties: { [key: string]: string; }, tr
 
     await verifyRuntimeIsCompatible(runtime, outputChannel, client, siteWrapper);
 
-    await siteWrapper.deploy(folderPath, client, outputChannel, extensionPrefix);
+    node.treeItem.state = localize('deploying', 'Deploying...');
+    try {
+        node.refresh();
+        await siteWrapper.deploy(folderPath, client, outputChannel, extensionPrefix);
+    } finally {
+        node.treeItem.state = await node.treeItem.siteWrapper.getState(client);
```
StephenWeatherford: Do you care that setting the state here could get an error, leaving it at deploying...?

src/commands/deploy.ts
```diff
@@ -49,7 +52,29 @@ export async function deploy(telemetryProperties: { [key: string]: string; }, tr
...
+        node.refresh();
+    }
+
+    const children: IAzureNode[] = await node.getCachedChildren();
+    const functionsNode: IAzureParentNode<FunctionsTreeItem> = <IAzureParentNode<FunctionsTreeItem>>children.find((n: IAzureNode) => n.treeItem instanceof FunctionsTreeItem);
+    const functions: IAzureNode<FunctionTreeItem>[] = <IAzureNode<FunctionTreeItem>[]>await functionsNode.getCachedChildren();
```
StephenWeatherford: functionsNode is guaranteed to be non-null?

src/commands/deploy.ts
EricJizbaMSFT: Hmm good point. I don't think I care for now. I created a new issue:
https://github.com/Microsoft/vscode-azurefunctions/issues/133

There's a larger problem where refresh should update the state anyways (aka I wouldn't have to update it here) and I'll fix it as a part of that

src/commands/deploy.ts
EricJizbaMSFT: Yep

Issue #144, opened 2017-12-25
Make the Java function name validation align with other languages


jdneo:
Previously, the Java function name validation is different from the validation for JS functions. 

The reason is that the Maven azure function plugin will use the input both for the Java class name and function name, so it must follow both Java class naming rule and function naming rule.

Start from Maven Azure Function Plugin v0.1.9, the Java class name will be converted from the input of Function name. So there is no need to make the Java function name validation seperate from other languages.

jdneo:
Here is the way to set azure function maven plugin version.

in the pom.xml: 

```xml
<build>
    <pluginManagement>
        <plugins>
            ...
            <plugin>
                <groupId>com.microsoft.azure</groupId>
                <artifactId>azure-functions-maven-plugin</artifactId>
                <version>0.1.9</version>
            </plugin>
        </plugins>
    </pluginManagement>
    ...
</build>
```

PR #145, opened 2017-12-25
fix issue: #144: uniform the function name validation for different languages

jdneo:
fix issue: #144

### DIFF
--- a/src/commands/createFunction/AbstractFunctionCreator.ts
+++ b/src/commands/createFunction/AbstractFunctionCreator.ts
@@ -6,11 +6,20 @@
 import { IUserInterface } from "../../IUserInterface";
 import { Template } from "../../templates/Template";
 
-export interface IFunctionCreator {
+export abstract class AbstractFunctionCreator {
+    protected readonly _functionNameRegex: RegExp = /^[a-zA-Z][a-zA-Z\d_\-]*$/;
+    protected _functionAppPath: string;
+    protected _template: Template;
+
+    constructor(functionAppPath: string, template: Template) {
+        this._functionAppPath = functionAppPath;
+        this._template = template;
+    }
+
     /**
      * Prompt for any settings that are specific to this creator
      * This includes the function name (Since the name could have different restrictions for different languages)
      */
-    promptForSettings(functionAppPath: string, template: Template, ui: IUserInterface): Promise<void>;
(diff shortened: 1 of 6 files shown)


src/commands/createFunction.ts
```diff
@@ -33,22 +33,17 @@ const requiredFunctionAppFiles: string[] = [
...
-            return localize('azFunc.existingFolderError', 'A folder with the name \'{0}\' already exists.', name);
-        }
-        if (!functionNameRegex.test(name)) {
-            return localize('azFunc.functionNameInvalidError', 'Function name must start with a letter and can contain letters, digits, \'_\' and \'-\'');
-        }
-        return undefined;
+    if (fse.existsSync(path.join(rootPath, name))) {
+        return localize('azFunc.existingFolderError', 'A folder with the name \'{0}\' already exists.', name);
+    }
+    if (!functionNameRegex.test(name)) {
```
EricJizbaMSFT: nit: I've noticed you doing `if/if` several times when you should be doing `if/else if` like this:
```
if (...) {
    ...
} else if (...) {
    ...
}
```
Please adopt this pattern going forward. Here's a few links I found explaining why it's better:
https://stackoverflow.com/questions/9169249/why-we-use-if-else-if-instead-of-multiple-if-block-if-the-body-is-a-return-stat
https://www.codecademy.com/en/forum_questions/5595581e76b8fea07700004c

src/commands/createFunction.ts
```diff
@@ -33,22 +33,17 @@ const requiredFunctionAppFiles: string[] = [
...
-        return validateJavaFunctionName(name);
-    } else {
-        if (fse.existsSync(path.join(rootPath, name))) {
-            return localize('azFunc.existingFolderError', 'A folder with the name \'{0}\' already exists.', name);
-        }
-        if (!functionNameRegex.test(name)) {
-            return localize('azFunc.functionNameInvalidError', 'Function name must start with a letter and can contain letters, digits, \'_\' and \'-\'');
-        }
-        return undefined;
+    if (fse.existsSync(path.join(rootPath, name))) {
```
EricJizbaMSFT: This case still does not apply for Java, correct?

EricJizbaMSFT: (I'm guessing you should be using `getNewJavaFunctionFilePath` to test the path's existence rather than `rootPath`)

jdneo:
Update: I extract some common fields into a base class and update the Java name validation.

PR #150, opened 2018-01-04
Leverage shared AzureActionHandler

EricJizbaMSFT:
See this PR for more details: https://github.com/Microsoft/vscode-azuretools/pull/53
Builds will fail until that is merged.


src/FunctionConfig.ts
```diff
@@ -47,45 +46,41 @@ export class FunctionConfig {
     public constructor(data: any) {
         let parseError: string | undefined;
 
-        try {
-            if (data === null || data === undefined) {
-                parseError = localize('noDataError', 'No data was supplied.');
+        if (data === null || data === undefined) {
```
StephenWeatherford: why remove the try/catch?  I thought the structure of data was not guaranteed to be a valid IFunctionJSon?

StephenWeatherford: Looks like you improved the error checking.  Personally I would still leave in the try/catch because it would be easy to miss a catch, especially as people add new fields/subfields to the data structure.

EricJizbaMSFT: Nope didn't improve the error checking. All I did was remove the try/catch (which decreased the indent on the code and made the github diff-ing look terrible).

The try/catch doesn't do a whole lot. It basically just wraps error messages with "Failed to parse function.json: {0}". I wanted to remove the dependency on ErrorData.ts since I moved it to the shared package and this was the path of least resistence. However, I can just export ErrorData from the shared package (I think it might be useful going forward) and add the try/catch back.

Issue #154, opened 2018-01-05
Set ignoreFocusOut to true


chrisdias:
There are a lot of steps to create a trigger (blob for example). The quickPick/inputBox prompts don't set the option `ignoreFocusOut` to `true`. As a result, if I get a long way through the set of questions and I click away, I lose all the information I entered.

A good example of this is when creating a blob trigger. At some point it asks me for the path within the blob (it defaults to `samples-workitems/{name}` and I have to switch over to the Storage extension to remember the container name.  Once I do that, I've lost everything I've entered in the wizard.

To repro:

`F1` to open command palette
`Azure Functions: Create Function`
Select a folder
Choose `Blob Trigger`
Provide a Function Name
Pick a Storage account (assuming you have `AzureWebJobsStorage` already defined in `local.settings.json`)
On the `Provide a 'Path'` quick pick, click away to the storage extension

Result: profound sadness as you need to enter the roughly ~7 steps again.

PR #161, opened 2018-01-11
Add support for C# Class library projects

### LINKED ISSUE #43: Add support for C# scripts
Was just testing the extension and it looks like when adding a new function it add for js only. Would be good to see it also support C#.

### LINKED ISSUE #162: Add local debug support for C# class libraries

EricJizbaMSFT:
This PR supports project/function creation and local debugging. I have not implemented logic for deploy.

A brief summary:
1. We default C# projects to beta runtime and class library (instead of C# script)
1. We use dotnet templates for project/function creation. We will automatically install the templates for the user if they are not on their machine (I don't prompt at all - let me know if you think we should prompt).
1. We use the parameter information from the functions portal (Aka C# Script templates) since it's easier to parse than the dotnet cli and it gives us more information (like validation). This requires us to assume that parameters for C# Scripts are the same as the parameters for the C# Class libraries. Since that might not always be the case, I mitigated this with unit tests and hard-coding the version of the dotnet templates.
1. Unlike JavaScript debugging, we have to attach to a specific process instead of attaching with a port. I implemented a 'pickProcess' command to search for the functions host process.
1. This only works on Windows. There's a few issues on a Mac I still need to iron out.

Fixes #162


src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -0,0 +1,91 @@
...
+                    type: 'coreclr',
+                    request: 'attach',
+                    processId: '\${command:azureFunctions.pickProcess}',
+                    preLaunchTask: 'build'
+                }
+            ]
+        };
+    }
+
+    public getRedommendedExtensions(): string[] {
```
nturinski: Redommended? 😆

src/commands/createNewProject/IProjectCreator.ts
```diff
@@ -3,14 +3,35 @@
...
 export interface IProjectCreator {
     /**
      * Add all project files not included in the '.vscode' folder
      */
     addNonVSCodeFiles(functionAppPath: string): Promise<void>;
-    getTasksJson(launchTaskId: string, funcProblemMatcher: {}): {};
-    getLaunchJson(launchTaskId: string): {};
+    getTasksJson(): {};
+    getLaunchJson(): {};
+    getRedommendedExtensions?(): string[];
```
nturinski: 😄

EricJizbaMSFT:
Per offline discussion, I'll fix these two bugs separately: #167, #154. I'm not too worried about the language suffix thing you mentioned - almost all users will stick to one language and have no problem remembering.

PR #172, opened 2018-01-18
Add createFunctionApp to api

EricJizbaMSFT:
It takes the subscription id as an optional parameter and returns the function app id

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,4 +1,4 @@
-# Azure Functions API
+# Azure Functions API (Preview)
 
 The following extension commands are supported for programatic use. If a parameter is not specified, the user will be prompted for the value. You must list 'ms-azuretools.vscode-azurefunctions' under the 'extensionDependencies' section of your package.json to ensure these apis are available to your extension.
 > NOTE: The functions extension is still in preview and the apis are subject to change.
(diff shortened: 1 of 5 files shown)


src/commands/createFunctionApp.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureTreeDataProvider, IAzureNode, IAzureParentNode } from 'vscode-azureextensionui';
+import { nodeUtils } from '../utils/nodeUtils';
+
+export async function createFunctionApp(tree: AzureTreeDataProvider, arg?: IAzureParentNode | string): Promise<string> {
+    let node: IAzureParentNode;
+    if (typeof (arg) === 'string') {
```
StephenWeatherford: nit: most people don't use () with typeof, they're not needed.

docs/api.md
```diff
@@ -5,7 +5,8 @@ The following extension commands are supported for programatic use. If a paramet
```
StephenWeatherford: Increment version?

EricJizbaMSFT: I prefer to do that closer to the release date (in the same PR where we update the changelog)

PR #173, opened 2018-01-18
Add preview support for several script languages

### LINKED ISSUE #43: Add support for C# scripts

EricJizbaMSFT:
We've had requests to add support for C# Scripts. ~~Unfortunately, the func cli doesn't support local debugging for C# Scripts (AFAIK)~~. However, we can still provide the following:
1. 'Create New Project': It provides a 'tasks.json' file with the 'Run Functions Host' task, but no 'launch.json' file. (Users must change their `azureFunctions.ProjectLangauge` setting for this support.)
1. 'Create New Function': Most of this support already existed - the only addition is C# Scripts
1. Deploy: Works for all script languages

I added unit tests for each language, but only felt the need to do one function per language (since those could change and I don't think we need to keep our unit tests that up-to-date for preview languages).

Fixes #43 

Update: The func cli _does_ support debugging for C# Scripts. This PR includes now includes a 'launch.json' file for C# Scripts.


src/ProjectSettings.ts
```diff
@@ -90,15 +91,18 @@ export async function selectTemplateFilter(ui: IUserInterface): Promise<Template
     return <TemplateFilter>result;
 }
 
+export function tryGetLanguageSetting(): ProjectLanguage | undefined {
```
nturinski: You have three functions that getConfiguration based off extensionPrefix and the setting name.  Is it time to create a getWorkspaceSetting function?

src/commands/createNewProject/CSharpScriptProjectCreator.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public getLaunchJson(): {} {
+        return {
+            version: '0.2.0',
+            configurations: [
+                {
+                    name: localize('azFunc.attachToNetCoreFunc', "Attach to .NET Core Functions"),
+                    type: 'coreclr',
+                    request: 'attach',
```
nturinski: I'm probably doing something incorrectly, butt seems like C# Scripts attach, but I can't set breakpoints.

src/ProjectSettings.ts
nturinski: https://github.com/Microsoft/vscode-azurefunctions/pull/177/files#diff-67685aa049a8b0e5dfd64a65893c8717R94 😆

EricJizbaMSFT: Great minds think alike

src/commands/createNewProject/CSharpScriptProjectCreator.ts
EricJizbaMSFT: It was working for me. I'll look into separately
https://github.com/Microsoft/vscode-azurefunctions/issues/180

PR #174, opened 2018-01-19
Add deploy support for C# class libraries

### LINKED ISSUE #160: Add deploy support for C# class libraries
With the current implementation, it deploys the whole project, but it should deploy just the output folder instead

EricJizbaMSFT:
When creating a new project, we set the 'azureFunctions.deploySubPath' setting to 'bin/debug/netstandard2.0'. The user can obviously change this setting if they change the path, or just use the 'Browse...' option.

The deploy subpath shows up in the workspace folder picker. This is the current version for JavaScript:
screen shot 2018-01-18 at 3 38 37 pm

This is the new version (works on mac and windows) for c# class libraries:
screen shot 2018-01-18 at 3 39 05 pm screen shot 2018-01-18 at 3 55 13 pm

Fixes #160

### DIFF
--- a/package.json
+++ b/package.json
@@ -372,6 +372,10 @@
                             "TypeScript"
                         ],
                         "description": "%azFunc.projectLanguageDescription%"
+                    },
+                    "azureFunctions.deploySubPath": {
+                        "type": "string",
+                        "description": "%azFunc.deploySubPathDescription%"
                     }
                 }
             }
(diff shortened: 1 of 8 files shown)


nturinski:
This might not be an issue since people probably won't have different projects with different languages often, but if you have C#, Java, and JS projects opened, this is what happens:

image

EricJizbaMSFT:
@nturinski personally I'm not worried about it. Like you said, I don't think many people work with multiple languages. In any case, they can just change the setting or use 'Browse...' if they don't like the behavior

PR #176, opened 2018-01-20
C# functions - prompt for namespace

### LINKED ISSUE #165: New C# class library function should prompt for namespace
Today it always uses 'Company.Function'. We should still default to that, but at least prompt the user if they want to change it.

EricJizbaMSFT:
Fixes #165

### DIFF
--- a/src/commands/createFunction/CSharpFunctionCreator.ts
+++ b/src/commands/createFunction/CSharpFunctionCreator.ts
@@ -6,6 +6,8 @@
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import { OutputChannel } from "vscode";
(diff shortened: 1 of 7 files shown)


src/commands/createFunction/CSharpFunctionCreator.ts
```diff
@@ -6,6 +6,8 @@
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import { OutputChannel } from "vscode";
+// tslint:disable-next-line:no-require-imports
+import XRegExp = require('xregexp');
```
StephenWeatherford: Don't see where this is used...

PR #177, opened 2018-01-23
Only use global settings when creating a new project

EricJizbaMSFT:
Since a project is basically a new workspace, it doesn't make sense to use settings from the "previous" workspace. Instead, only respect global settings.

This is based on top of my currently opened PR here: https://github.com/Microsoft/vscode-azurefunctions/pull/173

### DIFF
--- a/src/ProjectSettings.ts
+++ b/src/ProjectSettings.ts
@@ -91,17 +91,22 @@ export async function selectTemplateFilter(ui: IUserInterface): Promise<Template
     return <TemplateFilter>result;
 }
 
-export function tryGetLanguageSetting(): ProjectLanguage | undefined {
+export function getFuncExtensionSetting<T>(key: string, ignoreWorkspaceSettings: boolean = false): T | undefined {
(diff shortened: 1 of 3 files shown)


src/ProjectSettings.ts
```diff
@@ -91,17 +91,22 @@ export async function selectTemplateFilter(ui: IUserInterface): Promise<Template
...
 }
 
-export function tryGetLanguageSetting(): ProjectLanguage | undefined {
+export function getFuncExtensionSetting<T>(key: string, ignoreWorkspaceSettings: boolean = false): T | undefined {
     const projectConfiguration: WorkspaceConfiguration = vscode.workspace.getConfiguration(extensionPrefix);
-    // tslint:disable-next-line:no-backbone-get-set-outside-model
-    return projectConfiguration.get(projectLanguageSetting);
+    if (ignoreWorkspaceSettings) {
+        const result: { globalValue?: T } | undefined = projectConfiguration.inspect<T>(key);
+        return result ? result.globalValue : undefined;
```
StephenWeatherford: Normal JavaScript would be: return result && result.globalValue

PR #178, opened 2018-01-24
Fix C# blob trigger path

### LINKED ISSUE #166: Default path for C# blob trigger is incorrect
If you create a blob trigger with all default values, the path in the generated BlobTrigger.cs file will look like this: 'sample-workitems/{name}/{name}'

EricJizbaMSFT:
Fixes #166

### DIFF
--- a/src/commands/createFunction/CSharpFunctionCreator.ts
+++ b/src/commands/createFunction/CSharpFunctionCreator.ts
@@ -60,8 +60,9 @@ export class CSharpFunctionCreator extends FunctionCreatorBase {
             }
 
             // https://github.com/Microsoft/vscode-azurefunctions/issues/166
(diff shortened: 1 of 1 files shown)


src/commands/createFunction/CSharpFunctionCreator.ts
```diff
@@ -59,6 +59,11 @@ export class CSharpFunctionCreator extends FunctionCreatorBase {
                 parameter = 'Path';
             }
 
+            // https://github.com/Microsoft/vscode-azurefunctions/issues/166
+            if (userSettings[key].endsWith('/{name}')) {
```
StephenWeatherford: Would be better to create a const for '/{name}' and use .length

PR #181, opened 2018-01-24
Improve C# template installation

EricJizbaMSFT:
1. Allow windows users to install either the .NET Core _or_ .NET Framework templates
1. Provide commands to install/uninstall the templates
1. Limit function templates to just the main 4
1. Detect if the project created was .NET Core or .NET Framework and set runtime accordingly

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -23,7 +23,7 @@
     "azFunc.appSettings.rename": "Rename setting...",
     "azFunc.appSettings.delete": "Delete setting",
     "azFunc.pickProcess": "Pick Process",
-    "azFunc.deploySubPathDescription": "The default subPath of a workspace folder to user when deploying.",
+    "azFunc.deploySubpathDescription": "The default subpath of a workspace folder to use when deploying.",
     "azFunc.installDotnetTemplates": "Install templates for the dotnet cli",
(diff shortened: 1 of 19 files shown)


package.nls.json
```diff
@@ -23,5 +23,7 @@
     "azFunc.appSettings.rename": "Rename setting...",
     "azFunc.appSettings.delete": "Delete setting",
     "azFunc.pickProcess": "Pick Process",
-    "azFunc.deploySubPathDescription": "The default subPath of a workspace folder to user when deploying."
+    "azFunc.deploySubPathDescription": "The default subPath of a workspace folder to user when deploying.",
```
StephenWeatherford: "use"

StephenWeatherford: BTW, "subpath" or "sub-path" would be better.

src/commands/pickFuncProcess.ts
```diff
@@ -16,7 +16,7 @@ export async function pickFuncProcess(): Promise<string | undefined> {
         // If the functions host isn't running, start the task for the user
         await vscode.commands.executeCommand('workbench.action.tasks.runTask', funcHostTaskId);
 
-        const timeout: number = 15;
+        const timeout: number = 60;
```
StephenWeatherford: suggest: timeoutSeconds

PR #189, opened 2018-01-26
Add more logging

### LINKED ISSUE #186: We should have better visibility about being able to uninstall the CLI templates for C#
Perhaps a comment in the output window?

EricJizbaMSFT:
1. How to uninstall/reinstall templates
1. Show that we're searching for the func process on F5 (since it can take a while)

Fixes #186

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -9,7 +9,9 @@ import * as vscode from 'vscode';
 import { localize } from '../localize';
 import { funcHostTaskId } from './createNewProject/IProjectCreator';
 
-export async function pickFuncProcess(): Promise<string | undefined> {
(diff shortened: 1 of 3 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -9,7 +9,9 @@ import * as vscode from 'vscode';
 import { localize } from '../localize';
 import { funcHostTaskId } from './createNewProject/IProjectCreator';
 
-export async function pickFuncProcess(): Promise<string | undefined> {
+export async function pickFuncProcess(outputChannel: vscode.OutputChannel): Promise<string | undefined> {
+    outputChannel.show();
+    outputChannel.appendLine(localize('searchingFuncHost', 'Searching for Azure Functions host process...'));
```
StephenWeatherford: Cool.  Perhaps consider logging an issue against vscode to indicate sooner in the UI that something's happening, too.

PR #196, opened 2018-01-31
Improve C# F5 experience

### LINKED ISSUE #185: Can't build/run C# function app after changes
If you have a C# function app, F5, stop, make some changes (e.g. new function) and F5 again, you get this multiple times until it gives up:

The process cannot access the file 'C:\Temp\csfunc4\bin\Debug\net461\bin\csfunc4.dll' because it is being used by another process.

### LINKED ISSUE #194: "Failed to detect running Functions host within "60" seconds" after cancelling build and attempting rebuild
I run azure functions locally (.net core 2.0), after doing a build and running it (F5) I terminate the app by going into terminal and hitting ctrl-c

When I try to run the functions again I get an error after a while saying 'error failed to detect running functions host within "60" seconds', if I hit close and F5 again the functions run fine.

Ideally I would like the functions to hot compile without needing to be  terminated and run again but failing that would be nice to get consistent  rebuilds.

EricJizbaMSFT:
1. Switch away from ps-node on windows because it is very slow
1. Stop and restart the functions host if it is running when the user starts debugging
1. Only build once when the user F5's instead of twice

Fixes #194 
Fixes #185

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -7,61 +7,112 @@
 import ps = require('ps-node');
 import * as vscode from 'vscode';
 import { localize } from '../localize';
+import { cpUtils } from '../utils/cpUtils';
 import { funcHostTaskId } from './createNewProject/IProjectCreator';
 
-export async function pickFuncProcess(outputChannel: vscode.OutputChannel): Promise<string | undefined> {
-    outputChannel.show();
-    outputChannel.appendLine(localize('searchingFuncHost', 'Searching for Azure Functions host process...'));
-    let funcProcess: string | undefined = await getFuncPid();
-
-    if (funcProcess === undefined) {
-        // If the functions host isn't running, start the task for the user
-        await vscode.commands.executeCommand('workbench.action.tasks.runTask', funcHostTaskId);
-
-        const timeoutSeconds: number = 60;
-        const maxTime: number = Date.now() + timeoutSeconds * 1000;
-        while (Date.now() < maxTime) {
-            // Wait one second between each attempt
-            await new Promise((resolve: () => void): void => { setTimeout(resolve, 1000); });
-
-            funcProcess = await getFuncPid();
-            if (funcProcess !== undefined) {
-                return funcProcess;
-            }
-        }
-
-        throw new Error(localize('failedToFindFuncHost', 'Failed to detect running Functions host within "{0}" seconds.', timeoutSeconds));
-    } else {
-        return funcProcess;
+export async function pickFuncProcess(): Promise<string | undefined> {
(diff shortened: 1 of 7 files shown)


src/utils/cpUtils.ts
```diff
@@ -4,12 +4,14 @@
...
 import * as cp from 'child_process';
+import * as os from 'os';
 import * as vscode from 'vscode';
 import { localize } from '../localize';
 
 export namespace cpUtils {
-    export async function executeCommand(outputChannel: vscode.OutputChannel | undefined, workingDirectory: string, command: string, ...args: string[]): Promise<string> {
+    export async function executeCommand(outputChannel: vscode.OutputChannel | undefined, workingDirectory: string | undefined, command: string, ...args: string[]): Promise<string> {
         let result: string = '';
+        workingDirectory = workingDirectory !== undefined ? workingDirectory : os.tmpdir();
```
StephenWeatherford: suggest: workingDirectory = workingDirectory || os.tmpdir();

src/commands/pickFuncProcess.ts
```diff
@@ -7,61 +7,112 @@
...
-                if (error) {
-                    reject(error);
-                } else {
-                    resolve(result);
-                }
-            });
-    });
+function isWindows(): boolean {
+    return /^win/.test(process.platform);
+}
```
StephenWeatherford: How about adding some CTI test cases?  Esp. cause of the win/Linux split?

src/utils/cpUtils.ts
EricJizbaMSFT: I've been avoiding that pattern because of a tslint rule. It sometimes complains with this:
```
ERROR: (strict-boolean-expressions): This type is not allowed in the operand for the '||' operator because it is always truthy. Allowed types are boolean, undefined-union, or string.
```

I definitely like the rule because it is handy in situations like this:
```typescript
enum ExampleEnum {
    a = 0,
    b = 1,
    c = 2
}

const example: ExampleEnum | undefined = tryGetEnumValue();
if (example) { // <-- tslint will complain about this. It should be "example === undefined"
    console.log('Enum is defined'); // <-- this won't display if tryGetEnumValue returns 'ExampleEnum.a'
}
```

However, I'm less sure what to do for cases like this file. For example, what should happen if workingDirectory gets passed in as an empty string? On one hand - it was explicitly passed in so why would we ignore that value? On the other hand...I'm not sure it makes sense to ever use a value of an empty string.

Like I said - I want to keep the rule on. However, I don't have a strong opinion for how we handle this specific case. Thoughts?

src/commands/pickFuncProcess.ts
EricJizbaMSFT: Done

PR #198, opened 2018-01-31
Update prereqs section of README

### LINKED ISSUE #188: We should add more info in the README about FX vs Core C# templates
E.g. about the uninstall option.  And that the CLI v2 only works with .net CORE.

EricJizbaMSFT:
* Include more detail about v1.0 vs v2.0
* Explain how to uninstall/reinstall C# templates
* Move language specific sections down the page

Fixes #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
     ```
 * Install the prerequisites for your desired language:
   * [JavaScript](#javascript)
-  * [C#](#csharp)
+  * [C#](#c)
   * [Java](#java)
 > NOTE: You may change your `azureFunctions.projectLanguage` user setting to multiple other 'preview' languages not listed above. This allows you to create a project/function in that language, but we do not yet support local debugging for these languages.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -46,6 +44,33 @@
...
+## Language-Specific Prerequisites
+
+### JavaScript
+
+* [Node 8.0+](https://nodejs.org/)
+
+### C#
+
+* [VS Code Debugger for C#](https://marketplace.visualstudio.com/items?itemName=ms-vscode.csharp)
+* [Dotnet cli](https://docs.microsoft.com/dotnet/core/tools/?tabs=netcore2x)
```
nturinski: nit: Typically CLI is capitalized and all documentation refers to it as .NET instead of spelling it phonetically.  Though I see that the command for it is "dotnet"

whiskeyjay: Is ".NET Core CLI" the official name?

README.md
```diff
@@ -46,6 +44,33 @@
...
+* Dotnet Templates for Azure Functions
```
whiskeyjay: .NET?

README.md
EricJizbaMSFT: I changed it to '.NET CLI'. I'm not sure if 'Core' is supposed to be in there because you can use 'dotnet new' for '.NET Framework' templates

PR #207, opened 2018-02-06
Add log streaming

### LINKED ISSUE #11: View streaming traces for a function
The portal displays traces per function (as opposed to per app like in app service)

EricJizbaMSFT:
Builds will fail until the appservice package is published with logstream support: https://github.com/Microsoft/vscode-azuretools/pull/67

Fixes #11

Edit: More fixes for log stream in the shared package here: https://github.com/Microsoft/vscode-azuretools/pull/68

### DIFF
--- a/src/commands/logstream/startStreamingLogs.ts
+++ b/src/commands/logstream/startStreamingLogs.ts
@@ -3,9 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+// tslint:disable-next-line:no-require-imports
+import WebSiteManagementClient = require('azure-arm-website');
+import { SiteLogsConfig } from 'azure-arm-website/lib/models';
 import * as vscode from 'vscode';
-import { AzureActionHandler, AzureTreeDataProvider, IAzureNode } from 'vscode-azureextensionui';
+import { AzureActionHandler, AzureTreeDataProvider, IAzureNode, UserCancelledError } from 'vscode-azureextensionui';
 import KuduClient from 'vscode-azurekudu';
+import { DialogResponses } from '../../DialogResponses';
 import { localize } from '../../localize';
 import { FunctionAppTreeItem } from '../../tree/FunctionAppTreeItem';
 import { nodeUtils } from '../../utils/nodeUtils';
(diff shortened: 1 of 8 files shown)


src/commands/logstream/stopStreamingLogs.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureTreeDataProvider, IAzureNode } from 'vscode-azureextensionui';
+import { FunctionAppTreeItem } from '../../tree/FunctionAppTreeItem';
+import { ILogStreamTreeItem } from './ILogStreamTreeItem';
+
+export async function stopStreamingLogs(tree: AzureTreeDataProvider, node?: IAzureNode<ILogStreamTreeItem>): Promise<void> {
```
PrashanthCorp: Looks like this function (and corresponding context menu item) would be visible for every Function App. Is there a way to filter the node picker and context menu presence down to Apps that have a logStream present? (I see that being done with the `node.treeItem.logStream` conditional - I think the UX might be cleaner though)

src/commands/logstream/startStreamingLogs.ts
```diff
@@ -0,0 +1,23 @@
...
+import { localize } from '../../localize';
+import { FunctionAppTreeItem } from '../../tree/FunctionAppTreeItem';
+import { nodeUtils } from '../../utils/nodeUtils';
+import { ILogStreamTreeItem } from './ILogStreamTreeItem';
+
+export async function startStreamingLogs(context: vscode.ExtensionContext, actionHandler: AzureActionHandler, tree: AzureTreeDataProvider, node?: IAzureNode<ILogStreamTreeItem>): Promise<void> {
+    if (!node) {
+        node = <IAzureNode<ILogStreamTreeItem>>await tree.showNodePicker(FunctionAppTreeItem.contextValue);
+    }
+
```
PrashanthCorp: I don't see a check to verify that the Streaming Log for the chosen node hasn't already been started. I don't see anything in the docs suggesting that `createOutputChannel` would perform an 'already-present' check.

PrashanthCorp: I don't see a check to verify that the Streaming Log for the chosen node hasn't already been started. I don't see anything in the docs suggesting that `createOutputChannel` would perform an 'already-present' check.

src/commands/logstream/stopStreamingLogs.ts
EricJizbaMSFT: It's not recommended to hide things like context-menu commands based on state. Users will get confused about why it displays sometimes and not other times. Best practice is actually to gray-out the option, but I don't think that's possible in VS Code

In any case, I'll just display a warning to the user in this case.

src/commands/logstream/startStreamingLogs.ts
EricJizbaMSFT: Fixed

src/commands/logstream/stopStreamingLogs.ts
PrashanthCorp: Yeah, greying out an option makes waay more sense. I stand corrected. :)

src/commands/logstream/startStreamingLogs.ts
PrashanthCorp: Is this fixed in [this](https://github.com/Microsoft/vscode-azuretools/pull/68) AzureTools PR?

EricJizbaMSFT: A combination of that PR and the commit I pushed to this PR [here](https://github.com/Microsoft/vscode-azurefunctions/pull/207/commits/eef0de9f45d8ad6285f19c7df6eafc9d303641bd)

Issue #184, opened 2018-01-25
Create new Project (C# version ~1) fails on existing project


PrashanthCorp:
1.  Click on 'Create new Project'
2. Choose the directory
3. Choose C# and version ~1 when prompted.

I see this error : 
![functionsc 1](https://user-images.githubusercontent.com/30420973/35416794-5ef97314-01df-11e8-903e-122779a2e83e.PNG)

Text: 
___start___
Downloading dotnet templates for Azure Functions...
Installing "microsoft.azurefunctions.projecttemplates.1.0.3.10152"
Installing "azure.functions.templates.1.0.3.10152"
Finished installing Azure Functions templates.
Creating this template will make changes to existing files:
  Overwrite   .gitignore
  Overwrite   host.json
  Overwrite   local.settings.json

Rerun the command and pass --force to accept and create.
Error: Command "dotnet new,azureFunctionsProjectTemplates" failed with exit code "2147500036".
___end___

EricJizbaMSFT:
The error is pretty clear in the logs. You tried to create a new project in a folder that already had a project. However, we could automatically check this case and ask the user if they want to overwrite those files

PR #211, opened 2018-02-07
 Improve error handling for child processes

EricJizbaMSFT:
## Scenario 1

User tries to create a new C# project in a folder that already has a project (related to #184).

BEFORE (NOTE: I explicitly opened up the output channel for the screenshot. It was not opened for me):
screen shot 2018-02-07 at 1 59 42 pm

AFTER:
screen shot 2018-02-07 at 1 34 32 pm

## Scenario 2

User tries to install .NET templates with no internet connection
> NOTE: This scenario covers the case where we don't pass 'outputChannel' to 'cpUtils.executeCommand'. (Aka situations where we  don't want the user to see output from the command - maybe because it's ugly and we print better output for them manually)

BEFORE:
screen shot 2018-02-07 at 1 58 45 pm

AFTER:
screen shot 2018-02-07 at 1 57 49 pm

### DIFF
--- a/src/utils/cpUtils.ts
+++ b/src/utils/cpUtils.ts
@@ -39,12 +39,13 @@ export namespace cpUtils {
             childProc.on('error', reject);
             childProc.on('close', (code: number) => {
                 if (code !== 0) {
+                    // We want to make sure the full error message is displayed to the user, not just the error code.
+                    // If outputChannel is defined, then we simply call 'outputChannel.show()' and throw a generic error telling the user to check the output window
+                    // If outputChannel is _not_ defined, then we include the command's output in the error itself and rely on AzureActionHandler to display it properly
                     if (outputChannel) {
                         outputChannel.show();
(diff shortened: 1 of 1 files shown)


src/utils/cpUtils.ts
```diff
@@ -34,8 +39,18 @@ export namespace cpUtils {
             childProc.on('error', reject);
             childProc.on('close', (code: number) => {
                 if (code !== 0) {
-                    reject(new Error(localize('azFunc.commandError', 'Command "{0} {1}" failed with exit code "{2}".', command, args.toString(), code)));
+                    if (outputChannel) {
+                        outputChannel.show();
+                        reject(new Error(localize('azFunc.commandErrorWithOutput', 'Failed to run "{0}" command. Check output window for more details.', command)));
```
StephenWeatherford: Not sure I'm understanding why having an output channel or not should affect the amount of detail you display.

EricJizbaMSFT: In both cases we're showing the same amount of details. See the 'after' screenshots for both scenarios. They have similar amounts of details. The only difference is where we display those details - from this file or in AzureActionHandler

Issue #179, opened 2018-01-24
Dynamically retrieve C# template information


EricJizbaMSFT:
We've hard-coded info for C# templates because the dotnet cli doesn't provide the following dynamically:
1. List of templates
1. Version of templates
1. Parameter information (parameter name, default value, regexp for validating, etc.)

The dotnet cli team has said they will add support for that (hopefully) soon. We should update once that's ready.

PR #220, opened 2018-02-12
Update functions sdk reference so that C# works on all OS's

### LINKED ISSUE #164: Fail to run C# Class library function on mac/linux
For some reason the generated 'function.json' file has the dll path specified as `bin/example.dll` when it should be `../bin/example.dll`. On windows, it has the correct path.

EricJizbaMSFT:
And bump the functions templates versions to the latest (We won't have to hard-code the versions anymore once [this](https://github.com/Microsoft/vscode-azurefunctions/issues/179) is fixed)

Fixes #164

### DIFF
--- a/src/commands/createNewProject/CSharpProjectCreator.ts
+++ b/src/commands/createNewProject/CSharpProjectCreator.ts
@@ -9,6 +9,7 @@ import * as path from 'path';
 import { SemVer } from 'semver';
 import * as vscode from 'vscode';
 import { OutputChannel } from 'vscode';
+import { parseError, TelemetryProperties } from 'vscode-azureextensionui';
 import { DialogResponses } from '../../DialogResponses';
 import { IUserInterface } from '../../IUserInterface';
 import { localize } from "../../localize";
@@ -24,10 +25,12 @@ export class CSharpProjectCreator implements IProjectCreator {
 
     private _outputChannel: OutputChannel;
     private _ui: IUserInterface;
+    private _telemetryProperties: TelemetryProperties;
 
-    constructor(outputChannel: OutputChannel, ui: IUserInterface) {
+    constructor(outputChannel: OutputChannel, ui: IUserInterface, telemetryProperties: TelemetryProperties) {
         this._outputChannel = outputChannel;
         this._ui = ui;
+        this._telemetryProperties = telemetryProperties;
     }
 
     public async addNonVSCodeFiles(functionAppPath: string): Promise<void> {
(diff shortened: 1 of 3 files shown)


src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -120,4 +124,28 @@ export class CSharpProjectCreator implements IProjectCreator {
...
+     * Validates the project has the minimum Functions SDK version that works on all OS's
+     * See this bug for more info: https://github.com/Microsoft/vscode-azurefunctions/issues/164
+     */
+    private async validateFuncSdkVersion(csprojPath: string, csprojContents: string): Promise<void> {
+        try {
+            const minVersion: string = '1.0.8';
+            const lineMatches: RegExpMatchArray | null = /^.*Microsoft\.NET\.Sdk\.Functions.*$/gm.exec(csprojContents);
+            if (lineMatches !== null && lineMatches.length > 0) {
+                const line: string = lineMatches[0];
+                const versionMatches: RegExpMatchArray | null = /Version="(.*)"/g.exec(line);
```
StephenWeatherford: Should you be more specific on the version regex?  e.g. Version=([0-9.]+) or something?

src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -120,4 +124,28 @@ export class CSharpProjectCreator implements IProjectCreator {
...
+                if (versionMatches !== null && versionMatches.length > 1) {
+                    const version: SemVer = new SemVer(versionMatches[1]);
+                    if (version.compare(minVersion) < 0) {
+                        const newContents: string = csprojContents.replace(line, line.replace(version.raw, minVersion));
+                        await fse.writeFile(csprojPath, newContents);
+                    }
+                }
+            }
+        } catch (err) {
+            // ignore errors and assume the version of the templates installed on the user's machine works for them
```
StephenWeatherford: telemetry?

src/commands/createNewProject/CSharpProjectCreator.ts
EricJizbaMSFT: I made it more specific, but not quite that specific (Sometimes they append things like "-beta" to the end)

src/commands/createNewProject/CSharpProjectCreator.ts
EricJizbaMSFT: done

PR #235, opened 2018-02-16
Ship backup versions of templates with vsix

EricJizbaMSFT:

PR #237, opened 2018-02-21
Ship backup versions of templates with vsix (version 2)

EricJizbaMSFT:
These backup templates will be used if the user has no internet or the func portal api changes.

This PR also has two main differences than my [original PR](https://github.com/Microsoft/vscode-azurefunctions/pull/235) from last week:
1. Much more testing - all function unit tests are run against the production portal, staging portal, and backup files
1. Telemetry for each case (getting templates from portal, from cache, and from backup files). NOTE: I added a custom 'isActivationEvent' property since these events always run on extension activation and we don't necessarily want to track that for usage


package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurefunctions",
     "displayName": "Azure Functions",
     "description": "%extension.description%",
-    "version": "0.6.0",
+    "version": "0.6.1",
```
nturinski: Do we need to update vscode engine to 1.20.0?

src/extension.ts
```diff
@@ -63,14 +63,19 @@ export function activate(context: vscode.ExtensionContext): void {
         context.subscriptions.push(tree);
         context.subscriptions.push(vscode.window.registerTreeDataProvider('azureFunctionsExplorer', tree));
 
-        const templateData: TemplateData = new TemplateData(context.globalState);
-
         const actionHandler: AzureActionHandler = new AzureActionHandler(context, outputChannel, reporter);
+
+        const funcPortalTemplateDataTask: Promise<TemplateData | undefined> = tryGetTemplateDataFromFuncPortal(actionHandler, context.globalState);
+        const cachedTemplateDataTask: Promise<TemplateData | undefined> = tryGetTemplateDataFromCache(actionHandler, context.globalState);
```
nturinski: Calling all 3 functions regardless of failure trivializes the telemetry on the tryGetTemplateDataFrom**** events.  I think it could be useful to see how often the portal/cache breaks and when we have to fall back on the stored templates.

I guess it could also be a somewhat expensive operation since you're digging through an array of all the different templates.

package.json
EricJizbaMSFT: Nope - this is still using the old versions of the ui/appservice packages

src/extension.ts
EricJizbaMSFT: Each event reports if it succeeds/fails - so we still get the information you want.

However, I agree that it's not the most efficient. I'm going to keep it as-is for 0.6.1 since I've already done quite a bit of testing and don't want to make changes and re-test before releasing, but I'll fix when I merge this to master

EricJizbaMSFT:
The backup templates should rarely be used and we should keep them up-to-date (the unit tests help ensure this). I don't want users to even know they're using backup templates since there should be little difference.

I filed a separate issue to fix the unit tests on windows: https://github.com/Microsoft/vscode-azurefunctions/issues/239

Issue #203, opened 2018-02-02
Better support for exsting Functions apps


fiveisprime:
There is some degradation in the experience for users that have existing Functions Apps when it comes to locally debugging the app. Here are a few ideas for improving that.

Note that `$ func init` creates a launch.json that doesn't include starting up the functions host. I can work with them to get the tasks.json added that matches ours, that may help a bit.

### Contribute Functions launch config to the debugger

There is a way to contribute generic launch configs to the debugger that show up if a user hits F5 and VS Code doesn't know how to launch. 

![image](https://user-images.githubusercontent.com/1186948/35749989-460c8bbe-0808-11e8-9265-4ef939c7d456.png)

### Allow user to setup launch config

I know you can reinitialize the app to write out the launch config, but it's not super discoverable for users. Maybe add a button to "setup Functions debugging" that will add/overwrite the launch config and tasks.
 
### Better handling of language and runtime selection

The prompts for understanding the workspace ask for the runtime twice. Repro:

```sh
$ func init
$ code .
```

Now add a new function to the app, you'll be promoted to select the runtime version, set the runtime. Next you get prompted for the language, then again for the runtime version.

It seems like the first check could be skipped completely.

ahmelsayed:
let me know what the `func` cli can do to help here. I just got sometime to use the vscode extension and it's doing an awesome job so far 👍. I'd love to have a better experience for users starting from the cli.

EricJizbaMSFT:
@ahmelsayed there's two things for sure the cli can do:

1. Add this extension as a recommended extension for the workspace. This is done by adding an 'extensions.json' file in the '.vscode' folder with the following content:
```
{
    "recommendations": [
        "ms-azuretools.vscode-azurefunctions"
    ]
}
```
1. Remove '.vscode' from the default gitignore. We definitely want users to check in code in their '.vscode' folder since it contains debug configurations, etc.

I planned on submitting a PR to the cli repo eventually, but feel free to take care of it yourself if its easy for you.

EricJizbaMSFT:
Opened a separate issue to track the work in the functions cli: https://github.com/Azure/azure-functions-core-tools/issues/394

PR #241, opened 2018-02-23
Remove '.vscode' folder from default gitignore

EricJizbaMSFT:
We add a gitignore to each new project and we originally copied this from what 'func init' put down. However, we definitely want users to commit their vscode workspace settings, launch config, etc.

### DIFF
--- a/src/commands/createNewProject/ScriptProjectCreatorBase.ts
+++ b/src/commands/createNewProject/ScriptProjectCreatorBase.ts
@@ -17,7 +17,6 @@ csx
 .vs
 edge
 Publish
-.vscode
 
 *.user
 *.suo


src/commands/createNewProject/ScriptProjectCreatorBase.ts
```diff
@@ -17,7 +17,6 @@ csx
 .vs
 edge
 Publish
-.vscode
```
StephenWeatherford: Consider entering a bug against func init as well?  This definitely seems wrong.

EricJizbaMSFT: We have an existing bug to improve the relationship between the func cli and our extension. I added a comment to that issue to make sure we don't forget

EricJizbaMSFT: https://github.com/Microsoft/vscode-azurefunctions/issues/203

PR #242, opened 2018-02-23
Change 'Create function api' to accept object for function settings

EricJizbaMSFT:
This is better in a few ways:
1. It eliminates the need for the settings to always be in the same order
1. It allows us to change the function signature in a backwards compatible format in the future

This is a breaking change and even though the api is still in preview, I wanted to change it before any partner team (that we know of) actually starts using the api

### DIFF
--- a/src/commands/createFunction/CSharpFunctionCreator.ts
+++ b/src/commands/createFunction/CSharpFunctionCreator.ts
@@ -29,7 +29,7 @@ export class CSharpFunctionCreator extends FunctionCreatorBase {
         this._ui = ui;
     }
 
-    public async promptForSettings(ui: IUserInterface, functionName: string | undefined, functionSettings: {}): Promise<void> {
+    public async promptForSettings(ui: IUserInterface, functionName: string | undefined, functionSettings: { [key: string]: string | undefined; }): Promise<void> {
         if (!functionName) {
             const defaultFunctionName: string | undefined = await fsUtil.getUniqueFsPath(this._functionAppPath, removeLanguageFromId(this._template.id), '.cs');
             const placeHolder: string = localize('azFunc.funcNamePlaceholder', 'Function name');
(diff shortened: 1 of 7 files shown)


src/commands/createFunction/CSharpFunctionCreator.ts
```diff
@@ -29,7 +29,7 @@ export class CSharpFunctionCreator extends FunctionCreatorBase {
         this._ui = ui;
     }
 
-    public async promptForSettings(ui: IUserInterface, functionName: string | undefined, functionSettings: string[]): Promise<void> {
+    public async promptForSettings(ui: IUserInterface, functionName: string | undefined, functionSettings: {}): Promise<void> {
```
StephenWeatherford: Are functionSettings values just strings or any type?  Consider being more specific.

Also consider specifying known keys.

src/commands/createFunction/CSharpFunctionCreator.ts
```diff
@@ -39,8 +39,9 @@ export class CSharpFunctionCreator extends FunctionCreatorBase {
             this._functionName = functionName;
         }
 
-        if (functionSettings.length > 0) {
-            this._namespace = <string>functionSettings.shift();
+        const namespaceKey: string = 'namespace';
```
StephenWeatherford: Is empty string a valid setting value for namespace?

src/commands/createFunction/createFunction.ts
```diff
@@ -146,8 +151,8 @@ export async function createFunction(
         const setting: ConfigSetting | undefined = await templateData.getSetting(runtime, template.functionConfig.inBindingType, settingName);
         if (setting) {
             let settingValue: string | undefined;
-            if (functionSettings.length > 0) {
-                settingValue = functionSettings.shift();
+            if (functionSettings[settingName.toLowerCase()]) {
```
StephenWeatherford: Are empty string or zero valid setting values?

EricJizbaMSFT: Depends on the language/setting. I'll change this - ultimately as an api I think it would be better for us to use what was passed in (even if it's an empty string that's not valid)

src/commands/createFunction/CSharpFunctionCreator.ts
EricJizbaMSFT: It is not valid. However, now that you mention it I'm torn on what we should do here. I'm leaning towards we should pass along the empty string (which will eventually result in an exception). I'm not sure we want to prompt for the namespace when we're being called as an api like this

src/commands/createFunction/CSharpFunctionCreator.ts
EricJizbaMSFT: Yeah it should be strings. I'll be more specific.

However, I don't plan on specifying known keys. There's over 20 templates and they each have different keys. It's up to user's of the api to know what keys to specify for the template they're using (which is good enough for now IMO)

PR #244, opened 2018-02-26
View and delete proxies

### LINKED ISSUE #91: Show Proxies in Explorer
Open in portal isn't possible at the moment for proxies. We have to figure out a basic command that we can do to warrant showing these in the explorer

EricJizbaMSFT:
Builds will fail until this PR is merged/published: https://github.com/Microsoft/vscode-azuretools/pull/82

Fixes #91

### DIFF
--- a/src/tree/ProxiesTreeItem.ts
+++ b/src/tree/ProxiesTreeItem.ts
@@ -1,0 +1,101 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import { IFileResult, SiteWrapper } from 'vscode-azureappservice';
+import { IAzureNode, IAzureParentTreeItem, IAzureTreeItem, parseError, UserCancelledError } from 'vscode-azureextensionui';
+import KuduClient from 'vscode-azurekudu';
+import { DialogResponses } from '../DialogResponses';
+import { localize } from '../localize';
+import { nodeUtils } from '../utils/nodeUtils';
+import { ProxyTreeItem } from './ProxyTreeItem';
+
+export class ProxiesTreeItem implements IAzureParentTreeItem {
+    public static contextValue: string = 'azFuncProxies';
+    public readonly contextValue: string = ProxiesTreeItem.contextValue;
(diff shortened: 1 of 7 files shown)


src/tree/ProxiesTreeItem.ts
```diff
@@ -0,0 +1,101 @@
...
+                rawProxyConfig.proxies = {};
+            }
+            this._proxyConfig = <IProxyConfig>rawProxyConfig;
+            return Object.keys(this._proxyConfig.proxies).map((name: string) => new ProxyTreeItem(name));
+        } catch (err) {
+            throw new Error(localize('failedToParseProxyConfig', 'Failed to parse "proxies.json" file: {0}', parseError(err).message));
+        }
+    }
+
+    public async deleteProxy(node: IAzureNode, name: string): Promise<void> {
```
nturinski: Is it a limitation of proxies to only be able to delete one at a time because we're using the kuduClient api?

src/tree/ProxiesTreeItem.ts
```diff
@@ -0,0 +1,101 @@
...
+                this._deletingProxy = true;
+                try {
+                    this._outputChannel.show(true);
+                    this._outputChannel.appendLine(localize('DeletingProxy', 'Deleting proxy "{0}"...', name));
+                    delete this._proxyConfig.proxies[name];
+                    const client: KuduClient = await nodeUtils.getKuduClient(node, this._siteWrapper);
+                    const data: string = JSON.stringify(this._proxyConfig);
+                    this._etag = await this._siteWrapper.putFile(client, data, this._proxiesJsonPath, this._etag);
+                    this._outputChannel.appendLine(localize('DeleteProxySucceeded', 'Successfully deleted proxy "{0}".', name));
+                } finally {
```
nturinski: With no catch statement, the error will get swallowed if an exception occurs during the delete, right?

src/tree/ProxiesTreeItem.ts
EricJizbaMSFT: It's because all proxies are controlled by one 'proxies.json' file on the backend. We can't make multiple edits to the file at the same time

src/tree/ProxiesTreeItem.ts
EricJizbaMSFT: No - the exception will not get swallowed. This try/finally has nothing to do with error handling and everything to do with the 'deletingProxy' flag. You need to make sure the flag is _always_ set back to the original value, otherwise a failure will block the user from ever performing the operation again

PR #246, opened 2018-02-26
Improve support for projects created outside of VS Code

### LINKED ISSUE #203: Better support for exsting Functions apps

EricJizbaMSFT:
On extension activation, we will check for un-initialized functions projects based on the following criteria:
1. host.json and local.settings.json exist at the root of the folder
1. projectLanguage and projectRuntime are not set

Then we will display this message:
screen shot 2018-02-26 at 3 06 41 pm

If they select yes, we will initialize the project in the following manner:
1. If we can uniquely detect the language - use that. If we can't, prompt the user for the langauge
1. If we can detect the installed runtime - use that. Otherwise, prompt the user for the runtime
1. Remove '.vscode' from the gitignore if applicable

Related to that work, this PR also includes the following:
1. A command to "Initialize for use with VS Code" that can be run from the command palette
1. A new docs page that will be used with the 'See more info' aka.ms link
1. More tests 😊

Fixes #203


src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -57,11 +54,11 @@ export class CSharpProjectCreator implements IProjectCreator {
             throw new Error(localize('unrecognizedTargetFramework', 'Unrecognized target framework in project file "{0}".', csProjName));
         } else {
             const targetFramework: string = matches[1];
```
nturinski: Why matches[1] and not [0] out of curiosity?

src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -129,32 +128,34 @@ export class CSharpProjectCreator implements IProjectCreator {
...
-            if (lineMatches !== null && lineMatches.length > 0) {
-                const line: string = lineMatches[0];
-                const versionMatches: RegExpMatchArray | null = /Version=(?:"([^"]+)"|'([^']+)')/g.exec(line);
-                if (versionMatches !== null && versionMatches.length > 2) {
-                    const version: SemVer = new SemVer(versionMatches[1] || versionMatches[2]);
-                    this._telemetryProperties.cSharpFuncSdkVersion = version.raw;
-                    if (version.compare(minVersion) < 0) {
-                        const newContents: string = csprojContents.replace(line, line.replace(version.raw, minVersion));
-                        await fse.writeFile(csprojPath, newContents);
+        if (!isWindows) {
```
nturinski: Does this function fail silently if it's a Windows OS or is this validation not necessary?

src/commands/createNewProject/IProjectCreator.ts
```diff
@@ -3,21 +3,46 @@
...
 import { localize } from "../../localize";
-import { extensionPrefix, ProjectRuntime, TemplateFilter } from "../../ProjectSettings";
+import { extensionPrefix, ProjectRuntime, promptForProjectRuntime, TemplateFilter } from "../../ProjectSettings";
+import { functionRuntimeUtils } from "../../utils/functionRuntimeUtils";
 
-export interface IProjectCreator {
-    deploySubpath?: string;
-    runtime: ProjectRuntime;
-    templateFilter: TemplateFilter;
+export abstract class ProjectCreatorBase {
```
nturinski: Should change the name of the file as this is no longer an interface

src/commands/createNewProject/IProjectCreator.ts
```diff
@@ -3,21 +3,46 @@
...
+        this.telemetryProperties = telemetryProperties;
+    }
+
+    public async getRuntime(): Promise<ProjectRuntime> {
+        // tslint:disable-next-line:strict-boolean-expressions
+        return await functionRuntimeUtils.tryGetLocalRuntimeVersion() || await promptForProjectRuntime(this.ui);
+    }
+
+    public getLaunchJson(): {} | undefined {
+        return undefined;
```
nturinski: Is this always supposed to default to undefined?

src/commands/createNewProject/CSharpProjectCreator.ts
EricJizbaMSFT: matches[0] is always the _full_ match and the rest of the matches in a RegExpMatchArray are from the capturing groups. So this case:
```
console.log(matches[0]); // Prints "<TargetFramework>net461<\/TargetFramework>"
console.log(matches[1]); // Prints "net461"
```

src/commands/createNewProject/CSharpProjectCreator.ts
EricJizbaMSFT: Validation isn't necessary on Windows - 1.0.8 is only required for Mac and Linux

src/commands/createNewProject/IProjectCreator.ts
EricJizbaMSFT: Sounds good. Just filed an issue to address separately: https://github.com/Microsoft/vscode-azurefunctions/issues/251

This isn't the only file I want to rename and I'd prefer to do it separately

src/commands/createNewProject/IProjectCreator.ts
EricJizbaMSFT: Yep - only like 3 out of 10-ish languages actually support a launch.json (granted those 3 languages are the 3 we care about most - C#, Java, and JavaScript. The rest are just preview langagues)

PR #248, opened 2018-02-28
Leverage refactored telemetry

EricJizbaMSFT:
This lets us use `callWithTelemetryAndErrorHandling` for validating projects. I also added the validateProjects command to run with the onDidChangeWorkspaceFolders event

Builds will fail until the shared package PRs are merged:
https://github.com/Microsoft/vscode-azuretools/pull/88
https://github.com/Microsoft/vscode-azuretools/pull/89


src/commands/createNewProject/validateFunctionProjects.ts
```diff
@@ -3,33 +3,31 @@
...
-                for (const folder of vscode.workspace.workspaceFolders) {
-                    const folderPath: string = folder.uri.fsPath;
-                    if (await isFunctionProject(folderPath) && !isInitializedProject(folderPath) && await promptToInitializeProject(folderPath)) {
-                        await initProjectForVSCode(properties, outputChannel, folderPath);
-                    }
+export async function validateFunctionProjects(actionContext: IActionContext, outputChannel: vscode.OutputChannel, folders: vscode.WorkspaceFolder[]): Promise<void> {
+    for (const folder of folders) {
+        const folderPath: string = folder.uri.fsPath;
+        if (await isFunctionProject(folderPath)) {
+            actionContext.suppressTelemetry = false;
```
StephenWeatherford: Would be better not to assume that supressTelemetry was set to true before calling this function, and set it here explicitly.

PR #250, opened 2018-03-01
Add pickFuncProcessTimeout setting

EricJizbaMSFT:
We continue to see this timeout error in our telemetry. We'll never get the error percentage completely down to zero since the host won't start if the user had build errors, but hopefully this will mitigate the problem for some of our users. For example:
1. If they have a long build, they can increase the timeout
1. If they have a short build, they can decrease the timeout and we'll fail faster in the case of build failures

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -8,6 +8,7 @@ import ps = require('ps-node');
 import * as vscode from 'vscode';
 import { isWindows } from '../constants';
 import { localize } from '../localize';
+import { extensionPrefix, getFuncExtensionSetting } from '../ProjectSettings';
 import { cpUtils } from '../utils/cpUtils';
 import { funcHostTaskId } from './createNewProject/IProjectCreator';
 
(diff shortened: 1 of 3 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -21,10 +22,17 @@ export async function pickFuncProcess(): Promise<string | undefined> {
     // Start (or restart) functions host (which will also trigger a build)
     await vscode.commands.executeCommand('workbench.action.tasks.runTask', funcHostTaskId);
 
-    const timeoutSeconds: number = 60;
-    const maxTime: number = Date.now() + timeoutSeconds * 1000;
+    const settingKey: string = 'pickFuncProcessTimeout';
```
StephenWeatherford: how about functionStartTimeout?  Or at least "find" instead of "pick".

EricJizbaMSFT: So this is what a user's launch.json looks like for C#:
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Attach to C# Functions",
      "type": "coreclr",
      "request": "attach",
      "processId": "${command:azureFunctions.pickProcess}"
    }
  ]
}
```
You'll notice that it explicitly references the "azureFunctions.pickProcess" command. I would prefer to keep the setting consistent with the command name (which now that I look closer, I realize I should remove 'Func' and change it to "pickProcessTimeout"). In other words, I don't disagree with your comment, but it's too late to change the command name (without breaking existing projects) and I think being consistent is more important. Sound fine to you?

StephenWeatherford: Sure

PR #252, opened 2018-03-02
Allow runtime in createNewProject api

### LINKED ISSUE #249: Allow createNewProject api to pass in runtime

EricJizbaMSFT:
Currently the IoT trigger is only available for C#Script and runtime "~1", so our partner team needs to pass in the runtime. Ideally it was available for C# and runtime "beta" that works on all OS's, but in any case we should let consumers of the api pass in the runtime they want.

Fixes #249

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -16,13 +16,14 @@ Commands:
 |Name|Type|Description|
 |---|---|---|
 |projectPath|string|Absolute file path that will contain your new project. If the path doesn't exist, it will be created.|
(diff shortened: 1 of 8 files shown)


EricJizbaMSFT:
ping @StephenWeatherford @nturinski I want to get this in ASAP so the partner team can test the API tonight

EricJizbaMSFT:
Yeah it's for a specific partner team that wants to call this api

PR #259, opened 2018-03-07
Fix pickProcess on latest func cli

EricJizbaMSFT:
The functions cli released a self-contained cross-plat version yesterday that no longer requires .NET Core 2.0 to be installed on the user's machine.  That's great, but it broke "pickProcess" on mac/linux when debugging C# projects. Luckily we have a release planned and we can get this fix out quickly :)

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -5,6 +5,7 @@
 
 // tslint:disable-next-line:no-require-imports
 import ps = require('ps-node');
+import { isNullOrUndefined } from 'util';
 import * as vscode from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { isWindows } from '../constants';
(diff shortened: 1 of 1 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -47,27 +48,46 @@ export async function pickFuncProcess(actionContext: IActionContext): Promise<st
 }
 
 async function getFuncHostPid(): Promise<string | undefined> {
-    const multipleProcError: Error = new Error(localize('multipleFuncHost', 'Detected multiple processes running the Functions host. Stop all but one process in order to debug.'));
+    let pids: string[] = [];
+    // If the cli is self-contained, the command will look like this: func host start
+    pids = pids.concat(await getMatchingPids(/.*func.*/, /.*host.*start.*/));
```
StephenWeatherford: How fast are these awaits?  Should you run them in parallel?

EricJizbaMSFT: They're very fast

EricJizbaMSFT: Can't hurt though

PR #275, opened 2018-03-12
Leverage refactored siteWrapper

EricJizbaMSFT:
Based on this PR in the shared package: https://github.com/Microsoft/vscode-azuretools/pull/118

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -3,14 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// tslint:disable-next-line:no-require-imports
-import WebSiteManagementClient = require('azure-arm-website');
 import { StringDictionary } from 'azure-arm-website/lib/models';
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { MessageItem } from 'vscode';
-import { SiteWrapper } from 'vscode-azureappservice';
+import { SiteClient } from 'vscode-azureappservice';
+import * as appservice from 'vscode-azureappservice';
 import { AzureTreeDataProvider, IAzureNode, IAzureParentNode, TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
 import * as xml2js from 'xml2js';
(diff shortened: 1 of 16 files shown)


src/commands/deploy.ts
```diff
@@ -3,14 +3,13 @@
...
-// tslint:disable-next-line:no-require-imports
-import WebSiteManagementClient = require('azure-arm-website');
 import { StringDictionary } from 'azure-arm-website/lib/models';
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import * as vscode from 'vscode';
 import { MessageItem } from 'vscode';
-import { SiteWrapper } from 'vscode-azureappservice';
+import { SiteClient } from 'vscode-azureappservice';
+import * as appservice from 'vscode-azureappservice';
```
nturinski: Import from the same package twice?

src/tree/FunctionAppTreeItem.ts
```diff
@@ -3,15 +3,12 @@
...
  *--------------------------------------------------------------------------------------------*/
 
-// tslint:disable-next-line:no-require-imports
-import WebSiteManagementClient = require('azure-arm-website');
-import { Site } from 'azure-arm-website/lib/models';
-import { OutputChannel } from 'vscode';
 import * as vscode from 'vscode';
-import { AppSettingsTreeItem, AppSettingTreeItem, ILogStream, SiteWrapper } from 'vscode-azureappservice';
+import { OutputChannel } from 'vscode';
+import { AppSettingsTreeItem, AppSettingTreeItem, ILogStream, SiteClient } from 'vscode-azureappservice';
```
nturinski: Here too.

src/commands/deploy.ts
EricJizbaMSFT: I fixed a few cases, but I actually do this on purpose sometimes. In most cases I would prefer to do this:
`import { deploy } from 'vscode-azureappservice';`

But in this case I couldn't do that because I already have a `deploy` function defined in this file. So I have to do `import * as appservice from 'vscode-azureappservice';` and call `appservice.deploy`. But I only use that syntax if absolutely necessary and leave the rest as I would prefer it above.

So yeah that's my reasoning, but at the end of the day I don't have a strong preference on this

PR #303, opened 2018-03-29
Allow function to be created with project from api

### LINKED ISSUE #296: createNewProject and createFunction in the same API call
Per request from partner team, they want to be able to add a function in the same call that they create a project. This is primarily because VS Code settings only apply after the workspace is open, but opening the workspace restarts the extension host.

EricJizbaMSFT:
Fixes #296 

I'm also experimenting with an 'extensionVariables' file so that we don't have to pass around common variables as much. Something similar has been discussed in several places throughout our repos. I chose as short a name as possible for the namespace since we'll probably be using this a lot ('ext', short for 'extension'). If people like this pattern we can shift the whole repo to do it eventually

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -19,6 +19,9 @@ Commands:
 |language|string|The currently supported languages are 'JavaScript', 'C#', and 'Java'.|
 |runtime|string|The currently supported runtimes are "~1" and "beta".|
 |openFolder|boolean|(Defaulted to true) Represents whether or not to open the project folder after it has been created. If true, the extension host may be restarted when the folder is opened.|
+|templateId|string|The id of an optional template you want to create with the new project.|
+|functionName|string|The name of the optional function to be created with the new project.|
+|functionSettings|{ [key: string]: string; }|Any settings unique to the optional function template. For example, the HttpTrigger template requires an AuthorizationLevel parameter.|
 
 ### Example Usage
 
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -7,15 +7,15 @@ import * as fse from 'fs-extra';
 import { isString } from 'util';
 import { QuickPickItem } from 'vscode';
 import * as vscode from 'vscode';
-import { AzureTreeDataProvider, DialogResponses, IAzureQuickPickItem, IAzureUserInput, TelemetryProperties } from 'vscode-azureextensionui';
+import { DialogResponses, IAzureQuickPickItem, TelemetryProperties } from 'vscode-azureextensionui';
 import { ProjectLanguage, projectLanguageSetting, ProjectRuntime, projectRuntimeSetting, TemplateFilter } from '../../constants';
+import { ext } from '../../extensionVariables';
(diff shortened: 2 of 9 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -93,22 +91,29 @@ export async function createFunction(
...
-    await validateIsFunctionApp(telemetryProperties, outputChannel, functionAppPath, ui);
+    await validateIsFunctionApp(telemetryProperties, functionAppPath);
 
-    const localAppSettings: LocalAppSettings = new LocalAppSettings(ui, tree, functionAppPath);
+    const localAppSettings: LocalAppSettings = new LocalAppSettings(ext.ui, ext.tree, functionAppPath);
 
-    let language: ProjectLanguage = await getProjectLanguage(functionAppPath, ui);
-    let runtime: ProjectRuntime = await getProjectRuntime(language, functionAppPath, ui);
-    let templateFilter: TemplateFilter = await getTemplateFilter(functionAppPath);
+    if (language === undefined) {
```
StephenWeatherford: !language
?

Don't you want the same functionality for null and ""?

StephenWeatherford: I rarely compare directly against undefined

EricJizbaMSFT: tslint doesn't like `!language` for enums because of the `!0` case (even though these are string enums, not number enums)

I'm fine with checking directly against undefined as long as strictNullChecks is turned on. You can't really make mistakes if the type checker complains about null vs undefined

docs/api.md
```diff
@@ -19,6 +19,9 @@ Commands:
 |language|string|The currently supported languages are 'JavaScript', 'C#', and 'Java'.|
```
StephenWeatherford: Mixed feelings about the globalness here. Will you have troubles injecting ui and trees into test cases?  This is less of an issue with JavaScript.  Certainly be a lot less of a pain, though. I think it's worth experimenting with.

src/commands/createFunction/createFunction.ts
StephenWeatherford: Sorry, missed that this was an enum, thought it was a string.

docs/api.md
EricJizbaMSFT: No problems injecting for testing. In some ways its actually easier this way

src/extension.ts
```diff
@@ -49,16 +50,19 @@ export function activate(context: vscode.ExtensionContext): void {
     }
 
     const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('Azure Functions');
+    ext.outputChannel = outputChannel;
     context.subscriptions.push(outputChannel);
 
     callWithTelemetryAndErrorHandling('azureFunctions.activate', reporter, outputChannel, async function (this: IActionContext): Promise<void> {
         this.properties.isActivationEvent = 'true';
```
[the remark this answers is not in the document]
EricJizbaMSFT: All of the 'registerCommand' calls are within this callback. There's no chance of registering commands if creating the ui or tree fails. rethrowOnError only matters for code _outside_ the callback

PR #306, opened 2018-04-02
Upload/Download app settings

### LINKED ISSUE #137: Add ability to synchronize local app settings and remote app settings
Example scenario:
1. Create new project
1. Create local blob trigger
1. Select a storage account
1. Deploy

After these steps, I want some way to synchronize the storage account setting to my remote function app that I just deployed to (otherwise the blob trigger won't work on the remote).

Still need to iron out what the experience would be

EricJizbaMSFT:
Fixes #137 

I had to handle the case where settings were encrypted, so I went ahead and just added encrypt/decrypt commands as well.

Here's some example output:
```
Uploading settings to "emj328"...
Added the following settings:
- emj328_STORAGE
Ignored the following settings that were already the same:
- AzureWebJobsStorage
WARNING: This operation will not delete any settings in "emj328". You must manually delete settings if desired.
```


src/commands/deploy.ts
```diff
@@ -31,14 +31,15 @@ export async function deploy(ui: IAzureUserInput, telemetryProperties: Telemetry
     let confirmDeployment: boolean = true;
     let node: IAzureParentNode<FunctionAppTreeItem> | undefined;
 
+    const workspaceMessage: string = localize('azFunc.selectZipDeployFolder', 'Select the folder to zip and deploy');
```
nturinski: Thanks, boss :)

src/utils/workspace.ts
```diff
@@ -7,16 +7,52 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import { IAzureQuickPickItem, IAzureUserInput } from 'vscode-azureextensionui';
 import { localize } from '../localize';
-import { getFuncExtensionSetting } from '../ProjectSettings';
 import * as fsUtils from './fs';
 
-export async function selectWorkspaceFolder(ui: IAzureUserInput, placeHolder: string, subpathSettingKey?: string): Promise<string> {
+export async function selectWorkspaceFolder(ui: IAzureUserInput, placeHolder: string, getSubPath?: (f: vscode.WorkspaceFolder) => string | undefined): Promise<string> {
```
nturinski: What's all this new code for??

src/commands/appSettings/confirmOverwriteSettings.ts
```diff
@@ -0,0 +1,81 @@
...
+                destinationSettings[key] = sourceSettings[key];
+            } else {
+                userIgnoredKeys.push(key);
+            }
+        } else {
+            matchingKeys.push(key);
+        }
+    }
+
+    if (addedKeys.length > 0) {
```
nturinski: If you're going to show all the added keys, might be worth to show the values as well.

PR #309, opened 2018-04-03
Update install instructions for azure functions core tools

EricJizbaMSFT:
Copied the content from [here](https://github.com/Azure/azure-functions-core-tools), but changed the formatting to better fit our readme. For example, I don't like having a giant prereqs section before the feature list, so I added a bullet-point list that references to sections further down

Also fixed some markdown lint stuff


README.md
```diff
@@ -7,26 +8,19 @@ to get started with the Azure Functions extension.
...
-  * For version 1.0 (This version is only supported on Windows):
-    ```bash
-    npm install -g azure-functions-core-tools
-    ```
-* Install the prerequisites for your desired language:
-  * [JavaScript](#javascript)
-  * [C#](#c)
-  * [Java](#java)
-> NOTE: You may change your `azureFunctions.projectLanguage` user setting to multiple other 'preview' languages not listed above. This allows you to create a project/function in that language, but we do not yet support local debugging for these languages.
+Follow the OS-specific instructions to install the [Azure Functions Core Tools](https://docs.microsoft.com/azure/azure-functions/functions-run-local):
```
nturinski: Do we need to link the doc if all the information is in our README?  Also because the source material seems to be more directly copied from here https://github.com/Azure/azure-functions-core-tools#installing

PR #310, opened 2018-04-03
Prep for 0.8.0 release

EricJizbaMSFT:
And fix some markdown lint rules in the changelog. I ignored [this rule](https://github.com/DavidAnson/markdownlint/blob/v0.8.1/doc/Rules.md#md024)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,67 +1,91 @@
+<!-- markdownlint-disable MD024 -->
+
 # Change Log
+
 All notable changes to the "azurefunctions" extension will be documented in this file.
 
 ## 0.7.0 - 2018-03-05
+
 ### Added
+
 - Automatically detect projects created outside of VS Code and prompt to initialize
 - View/Delete Proxies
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,67 +1,109 @@
...
 # Change Log
+
 All notable changes to the "azurefunctions" extension will be documented in this file.
 
+## 0.8.0 - 2018-04-05
+
+### Added
+
+- Upload and download application settings
+- Prompt to automatically install the Azure Functions Core Tools when possible
```
nturinski: Should we be more specific here?  I'm worried users won't be able to anticipate when this will happen since it's not specified that it occurs after a new project is created.

arroyc:
nice

CHANGELOG.md
EricJizbaMSFT: I'll let you handle that as a part of your PR if you want. I honestly wasn't sure what the final feature would look like haha

PR #330, opened 2018-04-19
Download zip from Func Cli Feed

nturinski:
I need some help with the use of unsafe anys.  I thought I declared the type correctly, but it's saying that tags and releases are anys since I defined it as an object with more properties.

### DIFF
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -46,8 +46,7 @@ export async function createNewProject(
             const languagePicks: QuickPickItem[] = [
                 { label: ProjectLanguage.JavaScript, description: '' },
                 { label: ProjectLanguage.CSharp, description: '' },
-                { label: ProjectLanguage.Java, description: '' },
-                { label: ProjectLanguage.Python, description: '(Preview)' }
+                { label: ProjectLanguage.Java, description: '' }
             ];
             const options: QuickPickOptions = { placeHolder: localize('azFunc.selectFuncTemplate', 'Select a language for your function project') };
             language = (await ext.ui.showQuickPick(languagePicks, options)).label;
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -46,7 +46,8 @@ export async function createNewProject(
             const languagePicks: QuickPickItem[] = [
                 { label: ProjectLanguage.JavaScript, description: '' },
                 { label: ProjectLanguage.CSharp, description: '' },
-                { label: ProjectLanguage.Java, description: '' }
+                { label: ProjectLanguage.Java, description: '' },
+                { label: ProjectLanguage.Python, description: '' }
             ];
             const options: QuickPickOptions = { placeHolder: localize('azFunc.selectFuncTemplate', 'Select a language for your function project') };
             language = (await ext.ui.showQuickPick(languagePicks, options)).label;
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -47,7 +47,7 @@ export async function createNewProject(
                 { label: ProjectLanguage.JavaScript, description: '' },
                 { label: ProjectLanguage.CSharp, description: '' },
                 { label: ProjectLanguage.Java, description: '' },
-                { label: ProjectLanguage.Python, description: '' }
+                { label: ProjectLanguage.Python, description: '(Preview)' }
             ];
             const options: QuickPickOptions = { placeHolder: localize('azFunc.selectFuncTemplate', 'Select a language for your function project') };
             language = (await ext.ui.showQuickPick(languagePicks, options)).label;
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -26,7 +26,7 @@ export enum ProjectLanguage {
 }
 
 export enum ProjectRuntime {
-    one = '~v1',
+    one = '~1',
     beta = 'beta'
 }
 
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -26,8 +26,8 @@ export enum ProjectLanguage {
 }
 
 export enum ProjectRuntime {
-    one = 'v1',
-    two = 'v2'
+    one = '~v1',
+    beta = 'beta'
 }
 
 export enum TemplateFilter {
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -26,8 +26,8 @@ export enum ProjectLanguage {
 }
 
 export enum ProjectRuntime {
-    one = '~1',
-    beta = 'beta'
+    one = 'v1',
+    two = 'v2'
 }
 
 export enum TemplateFilter {
--- a/src/templates/TemplateData.ts
+++ b/src/templates/TemplateData.ts
@@ -3,13 +3,13 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as compareVersions from 'compare-versions';
 import * as extract from 'extract-zip';
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 // tslint:disable-next-line:no-require-imports
 import request = require('request-promise');
+import { gt } from 'semver';
 import * as vscode from 'vscode';
 import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
 import TelemetryReporter from 'vscode-extension-telemetry';
@@ -29,7 +29,7 @@ const resourcesKey: string = 'FunctionTemplateResources';
 const funcCliFeedUrl: string = 'https://functionscdn.azureedge.net/public/cli-feed-v3.json';
 const tempPath: string = path.join(os.tmpdir(), 'vscode-azurefunctions-templates');
 
-type cliFeedJsonResponse = {
+export type cliFeedJsonResponse = {
     tags: {
         [tag: string]: {
             release: string,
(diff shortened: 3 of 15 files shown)


StephenWeatherford:
You want this:

type cliFeedJsonResponse = {
    tags: {
        [tag: string]: {
            release: string,
            displayName: string,
            hidden: boolean
        }
    },
    releases: {
        [release: string]: {
            templateApiZip: string
        }
    }
};

Then you're missing "releases" in your use of it.

StephenWeatherford:
https://www.typescriptlang.org/docs/handbook/advanced-types.html:
**Index types and string index signatures**
keyof and T[K] interact with string index signatures. If you have a type with a string index signature, keyof T will just be string. And T[string] is just the type of the index signature:
interface Map<T> {
    [key: string]: T;
}
let keys: keyof Map<number>; // string
let value: Map<number>['foo']; // number
Mapped types

src/templates/TemplateData.ts
```diff
@@ -147,29 +168,37 @@ export async function tryGetTemplateDataFromCache(reporter: TelemetryReporter |
     }
 }
 
-export async function tryGetTemplateDataFromFuncPortal(reporter: TelemetryReporter | undefined, globalState?: vscode.Memento, hostname: string = 'functions.azure.com'): Promise<TemplateData | undefined> {
+export async function tryGetTemplateDataFromFuncPortal(reporter: TelemetryReporter | undefined, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
     try {
         return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetTemplateDataFromFuncPortal', reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
+            if (await cachedTemplatesAreCurrent(globalState)) {
+                throw new Error('Cached templates are current-- no need');
```
EricJizbaMSFT: This shouldn't be an error. It will mess up our telemetry numbers and we don't want to risk accidentally showing this to the user. You'll have to refactor a bit, but you should probably just return undefined

src/templates/TemplateData.ts
```diff
@@ -202,16 +231,56 @@ function getRuntimeKey(baseKey: string, runtime: ProjectRuntime): string {
...
+                uri: funcCliFeedUrl,
+                headers: {
+                    'User-Agent': 'Mozilla/5.0' // Required otherwise we get Unauthorized
+                }
+            };
+
+            const cliFeedJson: cliFeedJsonResponse = <cliFeedJsonResponse>JSON.parse(await <Thenable<string>>request(funcJsonOptions).promise());
+            const currentRelease: string = <string>cliFeedJson.tags[runtime].release;
+            const cachedRelease: string | undefined = globalState.get(`${runtime}-release`);
+            if (!cachedRelease || parseInt(cachedRelease, 10) < parseInt(currentRelease, 10)) {
```
EricJizbaMSFT: Use the semver module for comparing versions. There's a few examples in this repo already

src/templates/TemplateData.ts
```diff
@@ -202,16 +231,56 @@ function getRuntimeKey(baseKey: string, runtime: ProjectRuntime): string {
...
         method: 'GET',
-        uri: `https://${hostname}/api/${subpath}?runtime=${runtime}&${param}`,
+        uri: funcCliFeedUrl,
         headers: {
             'User-Agent': 'Mozilla/5.0' // Required otherwise we get Unauthorized
         }
     };
 
-    return <T>(JSON.parse(await <Thenable<string>>request(options).promise()));
+    const cliFeedJson: cliFeedJsonResponse = <cliFeedJsonResponse>JSON.parse(await <Thenable<string>>request(funcJsonOptions).promise());
```
EricJizbaMSFT: Can you create a function to get 'cliFeedJson'? I feel like this code is duplicated

src/templates/TemplateData.ts
```diff
@@ -202,16 +231,56 @@ function getRuntimeKey(baseKey: string, runtime: ProjectRuntime): string {
...
+    const currentRelease: string = <string>cliFeedJson.tags[runtime].release;
+    const urlWithPythonTemplates: string = 'https://functionscdn.azureedge.net/public/TemplatesApi/2.0.0-beta-10180.zip';
+    // temp ternary for dev purposes
+    const templateUrl: string = runtime === ProjectRuntime.two ? urlWithPythonTemplates : <string>cliFeedJson[currentRelease].templateApiZip;
```
EricJizbaMSFT: Not sure what "temp ternary for dev purposes" means. I suggest getting rid of this or filing an issue on github to address before the release

src/templates/TemplateData.ts
```diff
@@ -202,16 +231,56 @@ function getRuntimeKey(baseKey: string, runtime: ProjectRuntime): string {
...
+                return false;
+            }
+        }
+        // only return true if ALL runtimes are current
+        return true;
+    }
+    return false;
+}
+
+async function getCurrentReleaseTemplates(runtime: string): Promise<string> {
```
EricJizbaMSFT: nit: I feel like getCurrentReleaseTemplates implies it would return the templates, not the version. How about "downloadLatestTemplates". I think that makes it more clear that the templates are a side-effect rather than the return value

src/templates/TemplateData.ts
```diff
@@ -230,3 +299,32 @@ function parseTemplates(rawResources: object, rawTemplates: object[], rawConfig:
...
+            }
+        };
+        request(templateOptions, (err: Error) => {
+            if (err !== undefined) {
+                reject(err);
+            }
+        }).pipe(fse.createWriteStream(path.join(tempPath, 'templates.zip')).on('finish', () => {
+            ext.outputChannel.appendLine('Downloading templates zip file. . .');
+            // tslint:disable-next-line:no-unsafe-any
+            extract(path.join(tempPath, 'templates.zip'), { dir: tempPath }, (err: Error) => {
```
EricJizbaMSFT: nit: make a constant for 'templates.zip'

test/createFunction/FunctionTesterBase.ts
```diff
@@ -27,7 +27,7 @@ suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
     // Ensure template data is initialized before any 'Create Function' test is run
     backupTemplateData = await getTemplateDataFromBackup(undefined, path.join(__dirname, '..', '..', '..'));
     funcPortalTemplateData = await tryGetTemplateDataFromFuncPortal(undefined);
-    funcStagingPortalTemplateData = await tryGetTemplateDataFromFuncPortal(undefined, undefined, 'functions-staging.azure.com');
+    funcStagingPortalTemplateData = await tryGetTemplateDataFromFuncPortal(undefined, undefined);
```
EricJizbaMSFT: At least file an issue on this because right now it's just going to run the tests twice on the same templateData which is pointless. Ideally 'funcStagingPortalTemplateData' tests against the pre-release of the templates

src/templates/TemplateData.ts
```diff
@@ -202,16 +231,56 @@ function getRuntimeKey(baseKey: string, runtime: ProjectRuntime): string {
...
+    return false;
+}
+
+async function getCurrentReleaseTemplates(runtime: string): Promise<string> {
+    const funcJsonOptions: request.OptionsWithUri = {
         method: 'GET',
-        uri: `https://${hostname}/api/${subpath}?runtime=${runtime}&${param}`,
+        uri: funcCliFeedUrl,
         headers: {
             'User-Agent': 'Mozilla/5.0' // Required otherwise we get Unauthorized
```
EricJizbaMSFT: Can you confirm if this user-agent is required for the new api? It might've just been the old api

src/templates/TemplateData.ts
```diff
@@ -147,29 +168,37 @@ export async function tryGetTemplateDataFromCache(reporter: TelemetryReporter |
     }
 }
 
-export async function tryGetTemplateDataFromFuncPortal(reporter: TelemetryReporter | undefined, globalState?: vscode.Memento, hostname: string = 'functions.azure.com'): Promise<TemplateData | undefined> {
+export async function tryGetTemplateDataFromFuncPortal(reporter: TelemetryReporter | undefined, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
```
EricJizbaMSFT: How about we rename this to 'tryGetLatestTemplateData'

I'm not sure 'FuncPortal' still has the same meaning

src/constants.ts
```diff
@@ -26,8 +26,8 @@ export enum ProjectLanguage {
 }
 
 export enum ProjectRuntime {
-    one = '~1',
-    beta = 'beta'
+    one = 'v1',
```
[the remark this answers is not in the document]
nturinski: These are the tags that the Azure Functions team is using for the cli-json.  Should I just create a new enum for "tags"?

src/templates/TemplateData.ts
nturinski: Should I make them two separate functions?

getCurrentReleaseVersion which calls downloadLatestTemplates?

EricJizbaMSFT: I think we should revisit this after you create a separate function to get cliFeedJson. You're getting it multiple times (once for each runtime in getCurrentReleaseTemplates and once in cachedTemplatesAreCurrent) and you only need to get it once and pass it along. That's actually most of the logic in this function, which means this function might not be necessary after your refactor. (In other words, you could move the remaining few lines directly into tryGetTemplateDataFromFuncPortal

src/templates/TemplateData.ts
nturinski: Yeah, I think it's just the old api

nturinski:
So I noticed that a lot of the tests are broken... It seems that the templates may have changed?

src/templates/TemplateData.ts
```diff
@@ -147,33 +175,44 @@ export async function tryGetTemplateDataFromCache(reporter: TelemetryReporter |
     }
 }
 
-export async function tryGetTemplateDataFromFuncPortal(reporter: TelemetryReporter | undefined, globalState?: vscode.Memento, hostname: string = 'functions.azure.com'): Promise<TemplateData | undefined> {
+export async function tryGetLatestTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
     try {
-        return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetTemplateDataFromFuncPortal', reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
+        return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetLatestTemplateData', reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
             this.suppressErrorDisplay = true;
+            this.rethrowError = true;
```
EricJizbaMSFT: Remove this before merging

src/commands/createNewProject/createNewProject.ts
```diff
@@ -46,7 +46,8 @@ export async function createNewProject(
             const languagePicks: QuickPickItem[] = [
                 { label: ProjectLanguage.JavaScript, description: '' },
                 { label: ProjectLanguage.CSharp, description: '' },
-                { label: ProjectLanguage.Java, description: '' }
+                { label: ProjectLanguage.Java, description: '' },
+                { label: ProjectLanguage.Python, description: '(Preview)' }
```
EricJizbaMSFT: Remove python until we know we'll support it

PR #339, opened 2018-04-26
Check local version with func and update if possible

nturinski:
Factored out the passing of outputChannel and UI in favor of extensionVariables
Just making the PR now, but no plans to merge until I find a way to test out brew outdated-- azure-functions-core-tools should be pushing an update out soon


src/commands/createNewProject/validateFuncCoreToolsInstalled.ts
```diff
@@ -31,33 +31,41 @@ export async function validateFuncCoreToolsInstalled(ui: IAzureUserInput, output
...
-            const winput: MessageItem = await ui.showWarningMessage(localize('windowsVersion', 'Which version of the runtime do you want to install?'), v1, v2);
+            let winput: string;
+            const v1: string = 'v1';
+            const v2: string = 'v2';
+            if (!runtimeVersion) {
+                const v1MsgItm: MessageItem = { title: v1 };
+                const v2MsgItm: MessageItem = { title: v2 };
+                winput = (await ext.ui.showWarningMessage(localize('windowsVersion', 'Which version of the runtime do you want to install?'), v1MsgItm, v2MsgItm)).title;
+            } else {
+                winput = runtimeVersion;
```
EricJizbaMSFT: What will happen here? What's the point of setting it to 'undefined'?

src/utils/functionRuntimeUtils.ts
```diff
@@ -82,26 +103,22 @@ export namespace functionRuntimeUtils {
     }
 
     async function getLocalFunctionRuntimeVersion(): Promise<string | null> {
-        const versionInfo: string = await cpUtils.executeCommand(undefined, undefined, 'npm', 'ls', runtimePackage, '-g');
-        const matchResult: RegExpMatchArray | null = versionInfo.match(/(?:.*)azure-functions-core-tools@(.*)/);
+        const versionInfo: string = await cpUtils.executeCommand(undefined, undefined, 'func');
+        const matchResult: RegExpMatchArray | null = versionInfo.match(/(?:.*)Azure Functions Core Tools (.*)/);
```
EricJizbaMSFT: We should use `func --version` and fall back to this logic only if it doesn't work. Since `func --version` isn't released yet, I say just file an issue on it for now

tatuq
```diff
@@ -0,0 +1,1769 @@
+[33mcommit e75426680e84dda1c13fdbc4f3cca7992862e99d[m[33m ([m[1;36mHEAD -> [m[1;32mnturinski/runtimeCheck[m[33m, [m[1;31morigin/nturinski/runtimeCheck[m[33m)[m
```
EricJizbaMSFT: gif

src/utils/functionRuntimeUtils.ts
```diff
@@ -29,28 +34,49 @@ export namespace functionRuntimeUtils {
...
-                        );
 
-                        const result: vscode.MessageItem = await ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
-                        if (result === DialogResponses.learnMore) {
-                            // tslint:disable-next-line:no-unsafe-any
-                            opn('https://aka.ms/azFuncOutdated');
-                        } else if (result === DialogResponses.dontWarnAgain) {
-                            await updateGlobalSetting(settingKey, false);
+                    // Code disabled untiled we can verify a consistent way to determine when a new release is available
+                    const enabled: boolean = false;
```
EricJizbaMSFT: File an issue for this. Otherwise looks good

PR #353, opened 2018-05-04
Update message updated

nturinski:
Fixes #250 

I really don't like how the message looks.  It's very dense--

image

Any suggestions?

### DIFF
--- a/src/utils/functionRuntimeUtils.ts
+++ b/src/utils/functionRuntimeUtils.ts
@@ -42,7 +42,7 @@ export namespace functionRuntimeUtils {
 
                     if (semver.gt(newestVersion, localVersion)) {
                         const canUpdate: boolean = await brewOrNpmInstalled();
-                        const message: string = canUpdate ? localize(
+                        let message: string = canUpdate ? localize(
                             'azFunc.outdatedFunctionRuntimeUpdate',
                             'Your version of the Azure Functions Core Tools ({0}) does not match the latest ({1}). Would you like to update now?',
                             localVersion,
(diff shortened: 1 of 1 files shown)


src/utils/functionRuntimeUtils.ts
```diff
@@ -42,7 +42,7 @@ export namespace functionRuntimeUtils {
 
                     if (semver.gt(newestVersion, localVersion)) {
                         const canUpdate: boolean = await brewOrNpmInstalled();
-                        const message: string = canUpdate ? localize(
+                        let message: string = canUpdate ? localize(
                             'azFunc.outdatedFunctionRuntimeUpdate',
                             'Your version of the Azure Functions Core Tools ({0}) does not match the latest ({1}). Would you like to update now?',
```
EricJizbaMSFT: Since the button says "Update" you could remove "Would you like to update now?". That might help

src/utils/functionRuntimeUtils.ts
```diff
@@ -52,23 +52,33 @@ export namespace functionRuntimeUtils {
                             'Your version of the Azure Functions Core Tools ({0}) does not match the latest ({1}). Please update for the best experience.',
                             localVersion,
                             newestVersion);
+                        const v2: string = localize('v2BreakingChanges', 'Caution: As v2 is still in preview, updating could result in breaking changes.');
```
EricJizbaMSFT: I'm struggling with this message - the hard part is that we want it to make them _more_ likely to update rather than less likely (since their Azure function app always gets updated with the latest and may be in a broken state).

Again, here's the shortest message I've come up with:
> v2 is in preview and may have breaking changes (which are applied automatically in Azure).

src/utils/functionRuntimeUtils.ts
EricJizbaMSFT: Here's the shortest message I can think of 
> Please update your Azure Functions Core Tools ({0}) to the latest ({1}).

src/utils/functionRuntimeUtils.ts
nturinski: This means that their Azure Function app will always be updated to the most current release?

src/utils/functionRuntimeUtils.ts
nturinski: We don't typically use please in our dialogs, do we?  What about
> Update your Azure Functions Core Tools ({0}) to the latest ({1}) for the best experience.

src/utils/functionRuntimeUtils.ts
EricJizbaMSFT: Yep it will

src/utils/functionRuntimeUtils.ts
EricJizbaMSFT: Sounds good to me

nturinski:
Sorry, don't know how that little guy slithered his way in there,

src/utils/functionRuntimeUtils.ts
```diff
@@ -42,33 +42,39 @@ export namespace functionRuntimeUtils {
...
+                            } else if (result === DialogResponses.dontWarnAgain) {
+                                await updateGlobalSetting(settingKey, false);
                             }
-                        } else if (result === DialogResponses.dontWarnAgain) {
-                            await updateGlobalSetting(settingKey, false);
                         }
+                        while (result === DialogResponses.learnMore);
                     }
                 } catch (error) {
                     ext.outputChannel.appendLine(`Error occurred when checking the version of 'Azure Functions Core Tools': ${parseError(error).message}`);
```
EricJizbaMSFT: Just noticed this behavior:
screen shot 2018-05-07 at 11 04 31 am

Feels like an easy thing you could tack on to this PR, but I'm fine if you fix separately, too.

nturinski: Do we not want to output an error message in general or just for when the user cancels?

EricJizbaMSFT: I was just talking about the canceled scenario

PR #354, opened 2018-05-05
Nturinski/backout backups

### LINKED ISSUE #349: Remove backup templates from extension
Since the cli-feed supports 'versioned' templates, we don't have to worry about the 'latest' templates breaking us. Thus we can remove the backup templates we currently ship with the extension

### LINKED ISSUE #359: TemplateData is not retrieved from the cache because of invalid SemVer
@nturinski and I have already discussed this offline, but I wanted to file an issue so we don't forget it for this milestone. Basically the cli-feed isn't using valid semver which means we're failing to get the cached templates (and thus it's downloading new templates every time). For now, we can probably just do a 'string equals' comparison since we're always using the latest templates.

nturinski:
* Deleted resources/templates folder
* getTemplateDataFromBackup get the hard-coded version release and downloads/extracts the zips
* Refactored downloadAndExtractZip to accept a releaseVersion and cliJsonFeed to get the templateUrl so that the release number can be passed and appropriate release can be downloaded programmatically

Fixes #349 & Fixes #359


src/templates/TemplateData.ts
```diff
@@ -200,25 +200,37 @@ export async function tryGetLatestTemplateData(reporter: TelemetryReporter | und
...
+                const rawTemplates: object[] = <object[]>await fse.readJSON(path.join(tempPath, 'templates', 'templates.json'));
+                const rawConfig: object = <object>await fse.readJSON(path.join(tempPath, 'bindings', 'bindings.json'));
+
+                [templatesMap[runtime], configMap[runtime]] = parseTemplates(rawResources, rawTemplates, rawConfig);
+            }
+
+            return new TemplateData(templatesMap, configMap);
+        });
+    } catch (error) {
+        throw error;
```
EricJizbaMSFT: Remove the catch block - it's redundant

src/templates/TemplateData.ts
```diff
@@ -17,6 +17,7 @@ import { ScriptProjectCreatorBase } from '../commands/createNewProject/ScriptPro
 import { ProjectLanguage, ProjectRuntime, TemplateFilter } from '../constants';
 import { ext } from '../extensionVariables';
```
EricJizbaMSFT: I think you should still delete templates.zip. However, I _do_ think you should cache the backup templates so that you don't have to download/extract the next time

src/templates/TemplateData.ts
```diff
@@ -243,8 +255,9 @@ export function removeLanguageFromId(id: string): string {
 }
 
 // tslint:disable-next-line:no-unsafe-any
-async function downloadAndExtractZip(templateUrl: string, release: string): Promise<{}> {
+async function downloadAndExtractZip(cliFeedJson: cliFeedJsonResponse, release: string): Promise<{}> {
```
EricJizbaMSFT: nit: Can you rename this function? It bothers me that it's a generic name that should work with literally any zip file I want to download, but the implementation only works for the templates zip file. How about `downloadAndExtractTemplates` or just `downloadTemplates`

package.nls.json
```diff
@@ -40,5 +40,7 @@
     "azFunc.enableRemoteDebugging": "Enable remote debugging, an experimental feature that only supports Java-based Functions Apps.",
     "azFunc.deleteProxy": "Delete Proxy",
     "azFunc.pickProcessTimeoutDescription": "The timeout (in seconds) to be used when searching for the Azure Functions host process. Since a build is required every time you F5, you may need to adjust this based on how long your build takes.",
-    "azFunc.showFuncInstallationDescription": "Show a warning to install Azure Functions Core Tools CLI when you create a new project if the CLI is not installed."
+    "azFunc.showFuncInstallationDescription": "Show a warning to install Azure Functions Core Tools CLI when you create a new project if the CLI is not installed.",
+    "azFunc.backupTemplateRuntime-~1": "The v1 runtime template release to use if retrieving templates from cache and latest both result in failure.",
```
EricJizbaMSFT: I actually don't see the 'backup' behavior as the main driver for this setting. My main motivation is for users that don't want the latest templates. In other words, this setting should override the version in `tryGetLatestTemplateData`.

That behavior would actually serve as a more robust 'backup' anyways. Consider this scenario: The cliFeed continues to list all templates, but the 'index.js' file is missing for all JavaScript templates. With your current implementation, there's nothing the user can do since the `backupTemplateRuntime` only applies to the backup templates and those won't be used in this scenario. With my suggestion, they can actually change the setting to an older version that still has 'index.js' and it will be used.

package.json
```diff
@@ -575,10 +575,15 @@
...
-                    "azureFunctions.showFuncInstallation": {
-                        "type": "boolean",
-                        "description": "%azFunc.showFuncInstallationDescription%",
-                        "default": true
+                    "azureFunctions.backupTemplateRuntime-~1": {
+                        "type": "string",
+                        "description": "%azFunc.backupTemplateRuntime-~1%",
+                        "default": "1.0.12.1"
+                    },
+                    "azureFunctions.backupTemplateRuntime-beta": {
```
EricJizbaMSFT: I don't like making a specific setting for each runtime because then we're forced to add a new setting anytime the runtime version changes. (For example they will stop using 'beta' at some point and switch to '~2').

I think you can just do one generic setting that applies to all runtimes. People can't use multiple runtimes at the same time anyways. I suggest simply 'templateVersion'

nturinski: The current behavior of the function makes us download both runtimes though.  Which release should I download for the runtime that is not specified?  Just default to the latest?

EricJizbaMSFT: "The current behavior of the function" lol why does that matter? You have the ability to change EVERYTHING

I see a few options:
1. As you mentioned: default to the latest for the runtime that's not specified (which I'd be fine with).
1. Use the version specified for BOTH runtimes (I could see an advanced user wanting to use the v1 JavaScript templates even on 'beta'. There's not that much difference in JavaScript and more exist for v1)

nturinski:
An issue with this new implementation is that there is no guaranteed method of ensuring that we have templates to use.  The only solutions I can think of are either:
1.) Let ext.templateData: TemplateData | undefined
2.) Have another true backup function to fall back on.
I think the chances of this happening are pretty slim though.

src/templates/TemplateData.ts
```diff
@@ -133,11 +141,13 @@ export async function tryGetTemplateDataFromCache(reporter: TelemetryReporter |
             const templatesMap: { [runtime: string]: Template[] } = {};
             const configMap: { [runtime: string]: Config } = {};
             for (const key of Object.keys(ProjectRuntime)) {
+                // called within loop in case setting has changed between runtimes
+                const userTemplateVersion: string | undefined = getFuncExtensionSetting(templateVersionSetting);
                 const runtime: ProjectRuntime = <ProjectRuntime>ProjectRuntime[key];
                 const feedRuntime: string = getFeedRuntime(runtime);
-                const currentRelease: string = cliFeedJson.tags[feedRuntime].release;
+                const currentRelease: string = userTemplateVersion ? userTemplateVersion : cliFeedJson.tags[feedRuntime].release;
```
EricJizbaMSFT: nit: Can you rename this to 'releaseVersion' like you have in tryGetTemplateDataFromCache? It might not be the 'current' version if the user specified something else

src/templates/TemplateData.ts
```diff
@@ -243,8 +279,27 @@ export function removeLanguageFromId(id: string): string {
...
+        const releaseQuickPicks: vscode.QuickPickItem[] = [];
+        for (const rel of Object.keys(cliFeedJson.releases)) {
+            releaseQuickPicks.push({
+                label: rel,
+                description: ''
+            });
+        }
+        const input: vscode.QuickPickItem | undefined = await ext.ui.showQuickPick(releaseQuickPicks, { placeHolder: invalidVersion });
+        // tslint:disable-next-line:strict-boolean-expressions
+        if (!input) {
```
EricJizbaMSFT: You don't have to do this. ext.ui takes care of UserCancelled for you

package.nls.json
```diff
@@ -40,5 +40,6 @@
     "azFunc.enableRemoteDebugging": "Enable remote debugging, an experimental feature that only supports Java-based Functions Apps.",
     "azFunc.deleteProxy": "Delete Proxy",
     "azFunc.pickProcessTimeoutDescription": "The timeout (in seconds) to be used when searching for the Azure Functions host process. Since a build is required every time you F5, you may need to adjust this based on how long your build takes.",
-    "azFunc.showFuncInstallationDescription": "Show a warning to install Azure Functions Core Tools CLI when you create a new project if the CLI is not installed."
+    "azFunc.showFuncInstallationDescription": "Show a warning to install Azure Functions Core Tools CLI when you create a new project if the CLI is not installed.",
+    "azFunc.templateVersion": "A runtime release version (any runtime) that species which templates will be used rather than the latest templates.  This version will be used for ALL runtimes."
```
EricJizbaMSFT: This should mention that it requires a restart of VS Code to take effect

PR #357, opened 2018-05-07
Validate func installed when debugging C# projects

### LINKED ISSUE #340: Function app debugging not working.
Repro steps:

1. Create "New Function" app in Visual Studio code
2.  Debug
3. Following error popup.

If you access, Fucntion apps. Its works fine.

Action: azureFunctions.pickProcess
Error type: Error
Error Message: Failed to detect running Functions host within "60" seconds. You may want to adjust the "azureFunctions.pickProcessTimeout" setting.

Version: 0.8.1
OS: darwin

EricJizbaMSFT:
Fixes #340 (hopefully - this is my best guess unless we get more detailed feedback from the user)

I had to refactor `validateFuncCoreToolsInstalled` to accept a `forcePrompt` boolean that basically says "The user MUST be prompted and MUST interact with the prompt" since they can't debug until this prompt is done. I messed with the prompt a little bit because I thought it was too dense with the command in there. Let me know what you think:

||forcePrompt=true|forcePrompt=false|
|---|---|---|
|canInstall=true|screen shot 2018-05-07 at 2 38 29 pm|screen shot 2018-05-07 at 2 38 59 pm|
|canInstall=false|screen shot 2018-05-07 at 2 54 59 pm|screen shot 2018-05-07 at 2 53 30 pm|

Finally - I think this will be a good test to see if we should do the same for other languages. It's definitely easier to do for C# since we have pickProcess and I don't know how we'd do it otherwise, but at least this will tell us if it's worth investigating.

### DIFF
--- a/src/commands/createNewProject/validateFuncCoreToolsInstalled.ts
+++ b/src/commands/createNewProject/validateFuncCoreToolsInstalled.ts
@@ -5,7 +5,7 @@
 
 import * as opn from 'opn';
 import { MessageItem } from 'vscode';
-import { callWithTelemetryAndErrorHandling, DialogResponses, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { callWithTelemetryAndErrorHandling, DialogResponses, IActionContext } from 'vscode-azureextensionui';
 import { Platform } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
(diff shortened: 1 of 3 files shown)


nturinski:
https://github.com/Microsoft/vscode-azurefunctions/issues/358
I made an issue on this where we can discuss potential solutions

PR #368, opened 2018-05-09
Fix null ref exception when there are no workspace folders

### LINKED ISSUE #364: Pop up an error after removing all folders from workspace
Repro steps:
1. Remove all folders from Workspace.
2. Click 'Create New Project'.

Action: azureFunctions.createNewProject
Error type: TypeError
Error Message: Cannot read property 'uri' of undefined

Version: 0.9.0
OS: win32

More Info:
This issue also reproduces on Mac.

EricJizbaMSFT:
Fixes #364 

I think this is a low pri bug, but it's also a low risk fix.


nturinski:
Agreed.  We should have been checking the length though I'm surprised `vscode.workspace.workspaceFolders` doesn't return undefined if there are no workspaceFolders opened.

PR #373, opened 2018-05-09
Refactor tryGetTemplate actions and error for version number

### LINKED ISSUE #31: Create Function should allow the user to display all templates
Today it only displays the 'Core' templates. However, we should provide an option to display all templates (maybe a user setting? Or just a quickpick?)

### LINKED ISSUE #361: Refactor TemplateGetting functions
There is a lot of reused code between getting the templates from latest, cache, and backup.  We can refactor this to be a lot cleaner.

### LINKED ISSUE #372: Prompt when templateVersion is invalid is confusing
This pops up whenever the extension is activated, but the message doesn't ever say that these are Function templates. It's also super hard to read quick pick placeholders in general:
screen shot 2018-05-09 at 3 19 44 pm

I would prefer if we popped up an error message like `Failed to retrieve Azure Functions templates for version "{0}".` that has a button to "Select Version" and "Use latest". _Then_ we can show the quick pick with versions.

nturinski:
Fixes #372 and Fixes #361

### DIFF
--- a/src/templates/TemplateData.ts
+++ b/src/templates/TemplateData.ts
@@ -10,8 +10,7 @@ import * as path from 'path';
 // tslint:disable-next-line:no-require-imports
 import request = require('request-promise');
 import * as vscode from 'vscode';
-import { callWithTelemetryAndErrorHandling, IActionContext } from 'vscode-azureextensionui';
-import TelemetryReporter from 'vscode-extension-telemetry';
+import { callWithTelemetryAndErrorHandling, IActionContext, parseError } from 'vscode-azureextensionui';
 import { betaReleaseVersion, ProjectLanguage, ProjectRuntime, TemplateFilter, templateVersionSetting, v1ReleaseVersion } from '../constants';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
@@ -29,7 +28,6 @@ const configKey: string = 'FunctionTemplateConfig';
 const resourcesKey: string = 'FunctionTemplateResources';
 const templateVersionKey: string = 'templateVersion';
 const tempPath: string = path.join(os.tmpdir(), 'vscode-azurefunctions-templates');
-const noInternet: string = localize('retryInternet', 'There was an error in retrieving the templates.  Recheck your internet connection and try again.');
 
 const verifiedTemplates: string[] = [
     'BlobTrigger-JavaScript',
@@ -60,23 +58,30 @@ const verifiedJavaTemplates: string[] = [
  * We cache the template data retrieved from the portal so that the user can create functions offline.
  */
 export class TemplateData {
-    private readonly _templatesMap: { [runtime: string]: Template[] } = {};
-    private readonly _configMap: { [runtime: string]: Config } = {};
-    constructor(templatesMap: { [runtime: string]: Template[] }, configMap: { [runtime: string]: Config }) {
+    private readonly _templatesMap: { [runtime: string]: Template[] | undefined } = {};
+    private readonly _configMap: { [runtime: string]: Config | undefined } = {};
+    // if there are no templates, then there is likely no internet or a problem with the clifeed url
+    private readonly _noInternetErrMsg: string = localize('retryInternet', 'There was an error in retrieving the templates.  Recheck your internet connection and try again.');
+    constructor(templatesMap: { [runtime: string]: Template[] | undefined }, configMap: { [runtime: string]: Config | undefined }) {
         this._templatesMap = templatesMap;
         this._configMap = configMap;
     }
 
     public async getTemplates(language: string, runtime: string = ProjectRuntime.one, templateFilter?: string): Promise<Template[]> {
+        const templates: Template[] | undefined = this._templatesMap[runtime];
+        if (!templates) {
+            throw new Error(this._noInternetErrMsg);
+        }
+
         if (language === ProjectLanguage.Java) {
             // Currently we leverage JS templates to get the function metadata of Java Functions.
             // Will refactor the code here when templates HTTP API is ready.
             // See issue here: https://github.com/Microsoft/vscode-azurefunctions/issues/84
-            const javaTemplates: Template[] = this._templatesMap[runtime].filter((t: Template) => t.language === ProjectLanguage.JavaScript);
+            const javaTemplates: Template[] = templates.filter((t: Template) => t.language === ProjectLanguage.JavaScript);
             return javaTemplates.filter((t: Template) => verifiedJavaTemplates.find((vt: string) => vt === removeLanguageFromId(t.id)));
         } else if (language === ProjectLanguage.CSharp) {
             // https://github.com/Microsoft/vscode-azurefunctions/issues/179
-            return this._templatesMap[runtime].filter((t: Template) => verifiedCSharpTemplates.some((id: string) => id === t.id));
+            return templates.filter((t: Template) => verifiedCSharpTemplates.some((id: string) => id === t.id));
         } else {
             switch (language) {
                 case ProjectLanguage.CSharpScript:
@@ -87,24 +92,28 @@ export class TemplateData {
                 default:
             }
 
-            let templates: Template[] = this._templatesMap[runtime].filter((t: Template) => t.language.toLowerCase() === language.toLowerCase());
+            let filterTemplates: Template[] = templates.filter((t: Template) => t.language.toLowerCase() === language.toLowerCase());
             switch (templateFilter) {
                 case TemplateFilter.All:
                     break;
                 case TemplateFilter.Core:
-                    templates = templates.filter((t: Template) => t.isCategory(TemplateCategory.Core));
+                    filterTemplates = filterTemplates.filter((t: Template) => t.isCategory(TemplateCategory.Core));
                     break;
                 case TemplateFilter.Verified:
                 default:
-                    templates = templates.filter((t: Template) => verifiedTemplates.find((vt: string) => vt === t.id));
+                    filterTemplates = filterTemplates.filter((t: Template) => verifiedTemplates.find((vt: string) => vt === t.id));
             }
 
-            return templates;
+            return filterTemplates;
         }
     }
 
     public async getSetting(runtime: ProjectRuntime, bindingType: string, settingName: string): Promise<ConfigSetting | undefined> {
-        const binding: ConfigBinding | undefined = this._configMap[runtime].bindings.find((b: ConfigBinding) => b.bindingType === bindingType);
+        const config: Config | undefined = this._configMap[runtime];
+        if (!config) {
+            throw new Error(this._noInternetErrMsg);
+        }
+        const binding: ConfigBinding | undefined = config.bindings.find((b: ConfigBinding) => b.bindingType === bindingType);
         if (binding) {
             return binding.settings.find((bs: ConfigSetting) => bs.name === settingName);
         } else {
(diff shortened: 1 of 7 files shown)


src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
-    }
-}
-
-export async function tryGetLatestTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
+export async function tryGetTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento, isBackup: boolean = false): Promise<TemplateData | undefined> {
+    const templateReleaseVersion: string = 'templateReleaseVersion';
+    const v1ReleaseVersion: string = '1.0.12'; // known stable version
+    const betaReleaseVersion: string = '2.0.1-beta.25'; // known stable version
     try {
         return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetLatestTemplateData', reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
```
EricJizbaMSFT: Change the name of this to match the function ("tryGetTemplateData") and add telemetry so that we know if this is latest/backup/cache/user-setting and what version

src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
-            return new TemplateData(templatesMap, configMap);
-        });
-    } catch (error) {
-        return undefined;
-    }
-}
-
-export async function tryGetLatestTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
+export async function tryGetTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento, isBackup: boolean = false): Promise<TemplateData | undefined> {
+    const templateReleaseVersion: string = 'templateReleaseVersion';
```
EricJizbaMSFT: Rename the variable to `templateVersionKey` and the value to 'templateVersion'

src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
-                const rawConfig: object = <object>await fse.readJSON(path.join(tempPath, 'bindings', 'bindings.json'));
-
-                [templatesMap[runtime], configMap[runtime]] = parseTemplates(rawResources, rawTemplates, rawConfig);
-                verifyTemplatesByRuntime(templatesMap, runtime);
-                if (globalState) {
-                    globalState.update(`${runtime}-release`, releaseVersion);
-                    globalState.update(getRuntimeKey(templatesKey, runtime), rawTemplates);
-                    globalState.update(getRuntimeKey(configKey, runtime), rawConfig);
-                    globalState.update(getRuntimeKey(resourcesKey, runtime), rawResources);
+                let releaseVersion: string = userTemplateVersion ? userTemplateVersion : cliFeedJson.tags[feedRuntime].release;
```
EricJizbaMSFT: Rename releaseVersion to templateVersion

src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
     try {
         return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetLatestTemplateData', reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
             this.suppressErrorDisplay = true;
             this.properties.isActivationEvent = 'true';
             const templatesMap: { [runtime: string]: Template[] } = {};
             const configMap: { [runtime: string]: Config } = {};
             for (const key of Object.keys(ProjectRuntime)) {
+                const runtime: ProjectRuntime = <ProjectRuntime>ProjectRuntime[key];
                 // called within loop in case setting has changed between runtimes
                 const userTemplateVersion: string | undefined = getFuncExtensionSetting(templateVersionSetting);
```
EricJizbaMSFT: I'm still not happy with `invalidTemplateVersionPrompt` and the fact that you're getting the user setting multiple times. How about you rename `invalidTemplateVersionPrompt` to `getTemplateVersionSetting` and call it once. It gets the setting and validates it at the same time, _before_ you enter the loop or try to get any templates

src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
-        });
-    } catch (error) {
-        return undefined;
-    }
-}
-
-export async function tryGetLatestTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento): Promise<TemplateData | undefined> {
+export async function tryGetTemplateData(reporter: TelemetryReporter | undefined, cliFeedJson: cliFeedJsonResponse, globalState?: vscode.Memento, isBackup: boolean = false): Promise<TemplateData | undefined> {
+    const templateReleaseVersion: string = 'templateReleaseVersion';
+    const v1ReleaseVersion: string = '1.0.12'; // known stable version
```
EricJizbaMSFT: I think these should go in 'constants.ts'. They're pretty important and I think will be easier to find that way.

src/templates/TemplateData.ts
```diff
@@ -133,107 +133,35 @@ function verifyTemplatesByRuntime(templatesMap: { [runtime: string]: Template[]
...
-                [templatesMap[runtime], configMap[runtime]] = parseTemplates(rawResources, rawTemplates, rawConfig);
-                verifyTemplatesByRuntime(templatesMap, runtime);
-                if (globalState) {
-                    globalState.update(`${runtime}-release`, releaseVersion);
-                    globalState.update(getRuntimeKey(templatesKey, runtime), rawTemplates);
-                    globalState.update(getRuntimeKey(configKey, runtime), rawConfig);
-                    globalState.update(getRuntimeKey(resourcesKey, runtime), rawResources);
+                let releaseVersion: string = userTemplateVersion ? userTemplateVersion : cliFeedJson.tags[feedRuntime].release;
+                if (isBackup) {
+                    releaseVersion = runtime === ProjectRuntime.one ? v1ReleaseVersion : betaReleaseVersion;
```
EricJizbaMSFT: I think you originally said you would pass in 'templateVersion' to tryGetTemplateData, which I like better than 'isBackup'

nturinski: I did say that, but when I tried to implement it, I'd have to pass both templateVersions, one for each runtime as a string[] or call the function for each runtime.

EricJizbaMSFT: "call the function for each runtime" sounds the best to me

src/templates/TemplateData.ts
```diff
@@ -335,3 +270,49 @@ function getFeedRuntime(runtime: ProjectRuntime): string {
...
+        }
+    } else {
+        // if there was no cliJsonFeed, then there is likely no internet or a problem with the feed url
+        // fallback to the cache if there is one, if not, use the backup
+        if (!globalState) {
+            throw new Error(noInternet);
+        }
+        const cacheVersion: string | undefined = globalState.get(`${templateVersionKey}-${runtime}`);
+        if (!cacheVersion) {
+            throw new Error(noInternet);
```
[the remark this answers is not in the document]
nturinski: It would happen only on the first activation if they did not have backup templates.

src/templates/TemplateData.ts
```diff
@@ -111,146 +122,114 @@ export class TemplateData {
...
 }
 
-export async function tryGetTemplateDataFromCache(reporter: TelemetryReporter | undefined, globalState: vscode.Memento, cliFeedJson: cliFeedJsonResponse): Promise<TemplateData | undefined> {
-    try {
-        return <TemplateData | undefined>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetTemplateDataFromCache', reporter, undefined, async function (this: IActionContext): Promise<TemplateData | undefined> {
+export async function getTemplateData(globalState?: vscode.Memento): Promise<TemplateData> {
+    const templatesMap: { [runtime: string]: Template[] | undefined } = {};
+    const configMap: { [runtime: string]: Config | undefined } = {};
+    const cliFeedJson: cliFeedJsonResponse | undefined = await tryGetCliFeedJson();
+    return <TemplateData>await callWithTelemetryAndErrorHandling('azureFunctions.tryGetParsedTemplateDataFromCliFeed', ext.reporter, undefined, async function (this: IActionContext): Promise<TemplateData> {
```
EricJizbaMSFT: Let's move callWithTelemetryAndErrorHandling inside the for-loop so that we get  templateSource/errors per runtime

src/templates/TemplateData.ts
```diff
@@ -111,146 +122,114 @@ export class TemplateData {
...
+            if (!parsedTemplatesByRuntime && cliFeedJson) {
+                const backupVersion: string = runtime === ProjectRuntime.one ? v1ReleaseVersion : betaReleaseVersion;
+                parsedTemplatesByRuntime = await tryGetParsedTemplateDataFromCliFeed(this, cliFeedJson, backupVersion, runtime, globalState);
+                this.properties.templateSource = 'backupCliFeed';
+            }
+
+            if (parsedTemplatesByRuntime) {
+                [templatesMap[runtime], configMap[runtime]] = parsedTemplatesByRuntime;
+            } else {
+                // Failed to get templates for this runtime
```
EricJizbaMSFT: set `this.properties.templateSource` to something like 'None' here

src/templates/TemplateData.ts
```diff
@@ -111,146 +122,114 @@ export class TemplateData {
...
-                        globalState.update(getRuntimeKey(`${templatesKey}-backup`, runtime), rawTemplates);
-                        globalState.update(getRuntimeKey(`${configKey}-backup`, runtime), rawConfig);
-                        globalState.update(getRuntimeKey(`${resourcesKey}-backup`, runtime), rawResources);
-                    }
-                }
-            }
-            return new TemplateData(templatesMap, configMap);
-        });
+        context.properties.templateVersion = templateVersion;
+        context.properties.runtime = runtime;
```
EricJizbaMSFT: Move the 'runtime' telemetry out of this function and into getTemplateData (so that we get it even if we use cached templates)

src/templates/TemplateData.ts
```diff
@@ -335,3 +299,40 @@ function getFeedRuntime(runtime: ProjectRuntime): string {
             throw new RangeError();
     }
 }
+
+async function tryGetTemplateVersionSetting(cliFeedJson: cliFeedJsonResponse | undefined, runtime: ProjectRuntime): Promise<string | undefined> {
+    const feedRuntime: string = getFeedRuntime(runtime);
+    const userTemplateVersion: string | undefined = getFuncExtensionSetting(templateVersionSetting);
+    let templateVersion: string;
```
EricJizbaMSFT: Add telemetry for this:
```
this.properties.userTemplateVersion = userTemplateVersion
```

src/utils/getCliFeedJson.ts
```diff
@@ -23,10 +25,20 @@ export type cliFeedJsonResponse = {
...
+            this.properties.isActivationEvent = 'true';
+            this.suppressErrorDisplay = true;
+            const funcJsonOptions: request.OptionsWithUri = {
+                method: 'GET',
+                uri: funcCliFeedUrl
+            };
+            return <cliFeedJsonResponse>JSON.parse(await <Thenable<string>>request(funcJsonOptions).promise());
+        });
+
+    } catch (error) {
```
EricJizbaMSFT: nit: This try/catch is redundant since callWithTelemetryAndErrorHandling swallows errors and returns undefined by default

PR #389, opened 2018-05-21
Add try/catch to tryGetTemmplateVersionSetting

nturinski:
Try and catch added for the breaking case of a latest release version retrieved from npm not existing in the jsonCliFeed


src/templates/TemplateData.ts
```diff
@@ -302,39 +302,45 @@ function getFeedRuntime(runtime: ProjectRuntime): string {
...
+                }
             }
+        } else {
+            return undefined;
         }
-    } else {
+
+        return templateVersion;
+    } catch (error) {
+        // if cliJson does not have the template version being searched for, it will throw an error
```
EricJizbaMSFT: Add telemetry just like we have for cacheError/cliFeedError

Issue #387, opened 2018-05-19
Cannot connect to runtime process, timeout after 10000 ms - (reason: Cannot connect to the target: connect ECONNREFUSED 127.0.0.1:5858)


ahmadawais:
I'm facing a weird issue.

Error:

```
Cannot connect to runtime process, timeout after 10000 ms - (reason: Cannot connect to the target: connect ECONNREFUSED 127.0.0.1:5858).
```

```
VSCode 1.23.1
MacOS 
Node: 10.1.0
OS: macOS Sierra
npm: 6.0.1
Chrome: 66.0.3359.181
```

Launch.json

```json
{
	"version": "0.2.0",
	"configurations": [
		{
			"name": "Attach to JavaScript Functions",
			"type": "node",
			"request": "attach",
			"port": 5858,
			"protocol": "inspector",
			"preLaunchTask": "runFunctionsHost"
		}
	]
}
```

![image](https://on.ahmda.ws/riqo/c)

Trying to create an Azure function.

Several errors in my terminal output:

```sh

> Executing task: func host start <


                  %%%%%%
                 %%%%%%
            @   %%%%%%    @
          @@   %%%%%%      @@
       @@@    %%%%%%%%%%%    @@@
     @@      %%%%%%%%%%        @@
       @@         %%%%       @@
         @@      %%%       @@
           @@    %%      @@
                %%
                %

Hosting environment: Production
Content root path: /Users/$USER/Documents/Web/Sandbox/afn
Now listening on: http://localhost:7071
Application started. Press Ctrl+C to shut down.
[5/19/18 8:24:57 PM] Reading host configuration file '/Users/$USER/Documents/Web/Sandbox/afn/host.json'
[5/19/18 8:24:57 PM] Host configuration file read:
[5/19/18 8:24:57 PM] {}
[5/19/18 8:24:57 PM] Starting Host (HostId=ahmadsmacbookpro-763423413, InstanceId=50434726-525b-4afa-9566-a137447a193f, Version=2.0.11651.0, ProcessId=98058, AppDomainId=1, Debug=False, ConsecutiveErrors=0, StartupCount=1, FunctionsExtensionVersion=)
[5/19/18 8:24:57 PM] Unable to configure java worker. Could not find JAVA_HOME app setting.
[5/19/18 8:24:57 PM]
[5/19/18 8:24:57 PM] Could not configure language worker Java.
[5/19/18 8:24:57 PM]
[...]

PR #392, opened 2018-05-23
Prep for 0.9.1 release

EricJizbaMSFT:
The release is due to [this bug](https://github.com/Microsoft/vscode-azurefunctions/issues/387), but doesn't actually fix the bug itself (the real fix needs to be in the func cli). However, this release mitigates the bug by giving us the ability to 'turn off' the prompt to install the latest func cli (since Nathan [just changed](https://github.com/Microsoft/vscode-azurefunctions/pull/391) it to an aka.ms link).

I did a few additional things:
1. Stop displaying errors in the output channel. I think they're more confusing than helpful (NOTE: `this.suppressErrorDisplay` is still on even though I removed the try/catch, so users will NOT see these errors)
1. Temporarily hard-code the kudu package to the version used with the last release. I want to keep changes as small as possible for this bug-fix release


src/utils/functionRuntimeUtils.ts
```diff
@@ -127,17 +120,22 @@ export namespace functionRuntimeUtils {
...
+            type distTags = { core: string, docker: string, latest: string };
+            const distTags: distTags = <distTags>JSON.parse((await <Thenable<string>>request(npmRegistryUri).promise()));
+            switch (major) {
+                case FunctionRuntimeTag.latest:
+                    return distTags.latest;
+                case FunctionRuntimeTag.core:
+                    return distTags.core;
+                default:
+            }
+        } catch (error) {
```
StephenWeatherford: Not really liking the idea of putting error messages into random telemetry properties. I'd rather we keep them standardized so that the current dashboard picks them up.  That suggests two possibilities (I'm looking at a similar scenario in ARM tools):

1) run this under a new telemetry event
2) run this under a new telemetry event, but only log the event on errors (via a new flag)

Thoughts?

EricJizbaMSFT: I generally agree with you, but I also don't like creating new events every time because it makes it harder to correlate properties together. I generally do it this way when I don't really care if the code fails (In this case, errors are never displayed to the user and worst case they don't get the prompt to update to the latest runtime).

I do this very infrequently, but this isn't the first time we do this in functions. I guess I don't have a super strong preference.

StephenWeatherford: Please comment on https://github.com/Microsoft/vscode-azuretools/issues/183.  For now we'll let this go.

PR #409, opened 2018-06-04
Make function.json parsing less strict

### LINKED ISSUE #386: BUG: Failed to parse function.json: Expected "bindings" element of type "Array".
If you have a functions.json with a bindings:[] (empty), you receive this message and the plugin doesn't work properly.

Repro steps:
<Enter steps to reproduce issue>

Action: azureFunctions.deploy
Error type: Error
Error Message: Failed to parse function.json: Expected "bindings" element of type "Array".

Version: 0.9.0
OS: win32

EricJizbaMSFT:
Fixes #386

### DIFF
--- a/src/FunctionConfig.ts
+++ b/src/FunctionConfig.ts
@@ -38,10 +38,12 @@ enum BindingDirection {
 export class FunctionConfig {
     public readonly functionJson: IFunctionJson;
     public readonly disabled: boolean;
-    public readonly inBinding: IFunctionBinding;
-    public readonly inBindingType: string;
     public readonly isHttpTrigger: boolean = false;
     public readonly authLevel: HttpAuthLevel = HttpAuthLevel.function;
+
+    private readonly _inBinding: IFunctionBinding | undefined;
+    private readonly _inBindingType: string | undefined;
+    private readonly _noInBindingError: Error = new Error(localize('noInBinding', 'Failed to find binding with direction "in" for this function.'));
 
     // tslint:disable-next-line:no-any
     public constructor(data: any) {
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
ping

src/FunctionConfig.ts
```diff
@@ -94,4 +98,20 @@ export class FunctionConfig {
...
         }
     }
+
+    public get inBinding(): IFunctionBinding {
+        if (!this._inBinding) {
+            throw this._noInBindingError;
+        } else {
+            return this._inBinding;
+        }
+    }
```
PrashanthCorp: Is this primarily for the tests below? I see no mention of `this.inBinding` in this file. 
I'm mainly confused between the `const inBinding`  in the diff, and the above getter function.

src/FunctionConfig.ts
```diff
@@ -55,31 +57,33 @@ export class FunctionConfig {
...
+                            this._inBindingType = this._inBinding.type;
+                            if (this._inBinding.type.toLowerCase() === 'httptrigger') {
+                                this.isHttpTrigger = true;
+                                if (this._inBinding.authLevel) {
+                                    const authLevel: HttpAuthLevel | undefined = <HttpAuthLevel>HttpAuthLevel[this._inBinding.authLevel.toLowerCase()];
+                                    if (authLevel === undefined) {
+                                        errMessage = localize('unrecognizedAuthLevel', 'Unrecognized auth level "{0}".', this._inBinding.authLevel);
+                                    } else {
+                                        this.authLevel = authLevel;
+                                    }
```
PrashanthCorp: It feels like `authLevel` has a union type just for the error handling, which can be handled by checking if the key `this._inBinding.authLevel.toLowerCase()` is in HttpAuthLevel. 
This is just personal preference, feel free to ignore. :)

src/FunctionConfig.ts
```diff
@@ -55,31 +57,33 @@ export class FunctionConfig {
...
+                    if (this._inBinding) {
+                        if (!this._inBinding.type) {
+                            errMessage = localize('inBindingTypeError', 'The binding with direction "in" must have a type.');
+                        } else {
+                            this._inBindingType = this._inBinding.type;
+                            if (this._inBinding.type.toLowerCase() === 'httptrigger') {
+                                this.isHttpTrigger = true;
+                                if (this._inBinding.authLevel) {
+                                    const authLevel: HttpAuthLevel | undefined = <HttpAuthLevel>HttpAuthLevel[this._inBinding.authLevel.toLowerCase()];
+                                    if (authLevel === undefined) {
```
PrashanthCorp: Is `this.authLevel = null` a valid use case that shouldn't throw an error?

EricJizbaMSFT: In this repo we almost always use 'undefined' and ignore 'null'. We can do that without too much worry since strictNullChecks is turned on

src/FunctionConfig.ts
EricJizbaMSFT: inBinding is a public member and is used outside this file. Before this change, we required it to be defined. Now, we don't require it to be defined _unless you try to access it_. This is because sometimes the code uses other parts of TemplateData and doesn't really need inBinding

src/FunctionConfig.ts
```diff
@@ -55,31 +57,33 @@ export class FunctionConfig {
...
-                                    this.authLevel = authLevel;
+                    if (this._inBinding) {
+                        if (!this._inBinding.type) {
+                            errMessage = localize('inBindingTypeError', 'The binding with direction "in" must have a type.');
+                        } else {
+                            this._inBindingType = this._inBinding.type;
+                            if (this._inBinding.type.toLowerCase() === 'httptrigger') {
+                                this.isHttpTrigger = true;
+                                if (this._inBinding.authLevel) {
+                                    const authLevel: HttpAuthLevel | undefined = <HttpAuthLevel>HttpAuthLevel[this._inBinding.authLevel.toLowerCase()];
```
nturinski: Any reason not to just set this.authLevel directly?  It'll be undefined if HttpAuthLevel is undefined anyway, right?

EricJizbaMSFT: I'm pretty sure I did it this way for tslint reasons. Since 'this.authLevel' is of type 'HttpAuthLevel', tslint doesn't think I should be checking `if (this.authLevel === undefined)`

PR #421, opened 2018-06-19
Automatically retrieve dotnet templates from cli feed

EricJizbaMSFT:
Rather than hard-coding the version, always update to the version of dotnet templates listed in the cli feed. NOTE: @fiveisprime and I decided to no longer prompt when installing dotnet templates (unless we don't know what version to install).

Required as a part of #179

### DIFF
--- a/src/utils/dotnetUtils.ts
+++ b/src/utils/dotnetUtils.ts
@@ -6,112 +6,114 @@
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { MessageItem, OutputChannel, QuickPickItem, QuickPickOptions } from "vscode";
-import { DialogResponses, IAzureUserInput } from 'vscode-azureextensionui';
-import { isWindows, ProjectRuntime } from '../constants';
+// tslint:disable-next-line:no-require-imports
+import request = require('request-promise');
+import { QuickPickOptions } from "vscode";
+import { IActionContext, IAzureQuickPickItem } from 'vscode-azureextensionui';
+import { ProjectRuntime } from '../constants';
+import { ext } from '../extensionVariables';
 import { localize } from "../localize";
+import { tryGetTemplateVersionSetting } from '../templates/TemplateData';
+import { functionRuntimeUtils } from '../utils/functionRuntimeUtils';
 import { cpUtils } from "./cpUtils";
 import * as fsUtil from './fs';
+import { cliFeedJsonResponse, getFeedRuntime, tryGetCliFeedJson } from './getCliFeedJson';
+
+const projectPackageId: string = 'microsoft.azurefunctions.projecttemplates';
+const functionsPackageId: string = 'azure.functions.templates';
+const templatesKey: string = 'cSharpFunctionsTemplates';
+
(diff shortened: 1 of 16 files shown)


src/utils/dotnetUtils.ts
```diff
@@ -6,112 +6,114 @@
...
+            method: 'GET',
+            uri: url
+        };
+
+        request(templateOptions, (err: Error) => {
+            // tslint:disable-next-line:strict-boolean-expressions
+            if (err) {
+                reject(err);
+            }
+        }).pipe(fse.createWriteStream(fullPath).on('finish', () => {
```
StephenWeatherford: Have you tested if there's an error?  I think with node you get an unhandled exception which won't be caught if you don't have an error listener....?

src/utils/dotnetUtils.ts
```diff
@@ -6,112 +6,114 @@
...
+        }
 
-        outputChannel.show();
-        outputChannel.appendLine(localize('uninstallFuncProject', 'Uninstalling "{0}"', projectPackageId));
-        await cpUtils.executeCommand(undefined, tempFolder, 'dotnet', 'new', '--uninstall', projectPackageId);
-        outputChannel.appendLine(localize('uninstallFuncs', 'Uninstalling "{0}"', functionsPackageId));
-        await cpUtils.executeCommand(undefined, tempFolder, 'dotnet', 'new', '--uninstall', functionsPackageId);
-        outputChannel.appendLine(localize('finishedUninstallingTemplates', 'Finished uninstalling Azure Functions templates.'));
-    }
+        const latestRelease: string = await tryGetTemplateVersionSetting(actionContext, cliFeedJson, runtime) || cliFeedJson.tags[getFeedRuntime(runtime)].release;
```
StephenWeatherford: Anything need to be guarded in this expression?

src/utils/dotnetUtils.ts
```diff
@@ -6,112 +6,114 @@
...
+            if (templatesInstalled) {
+                await uninstallTemplates();
+            }
 
-    export async function installDotnetTemplates(ui: IAzureUserInput, outputChannel: OutputChannel): Promise<void> {
-        await validateDotnetInstalled();
+            const tempFolder: string = path.join(os.tmpdir(), fsUtil.getRandomHexString());
+            await fse.ensureDir(tempFolder);
+            try {
+                ext.outputChannel.appendLine(localize('installFuncProject', 'Installing "{0}"', projectPackageId));
```
StephenWeatherford: This stuff all works without permission elevation?

src/utils/dotnetUtils.ts
EricJizbaMSFT: Fixed

src/utils/dotnetUtils.ts
EricJizbaMSFT: They've promised that the cli feed will be backwards compatible, so I'm not too worried. I added some additional telemetry/error handling just in case, though.

src/utils/dotnetUtils.ts
EricJizbaMSFT: Yep. Tested on mac, windows, and linux

PR #444, opened 2018-07-09
Retrieve the function version and node version from the cli feed

### LINKED ISSUE #419: Set node version when setting runtime version
The Functions runtime takes dependency on node at specific versions. Right now, when updating the runtime on Azure, I believe we are only updating the runtime, but we need to also update the node version using `WEBSITE_NODE_DEFAULT_VERSION`.

When setting functions runtime to **v1**, use node version **6.5.0**
When setting functions runtime to **v2** use node version **8.11.1**

I'm talking with the Functions team to find a flexible way to make sure we always choose the correct node version. I'll keep this issue updated as that conversation progresses.

Edit: Soon, v2 will expect node@10 and the Functions team may incorporate this version requirement into the typical CLI feed.

EricJizbaMSFT:
The old behavior always creates function apps for v1 and prompted users if they tried to deploy and their versions didn't match.

The new behavior detects their runtime and creates a function app that matches (although the warning on deploy still exists just in case).

As a part of this, we want to make sure the node version matches the runtime version as well. This information is contained in the cli-feed, although I've hard-coded backup versions just in case. This fixes #419 

Based on https://github.com/Microsoft/vscode-azuretools/pull/218

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -22,6 +22,7 @@ import { FunctionsTreeItem } from '../tree/FunctionsTreeItem';
 import { FunctionTreeItem } from '../tree/FunctionTreeItem';
 import { cpUtils } from '../utils/cpUtils';
 import { isSubpath } from '../utils/fs';
+import { getCliFeedAppSettings } from '../utils/getCliFeedJson';
 import { mavenUtils } from '../utils/mavenUtils';
 import * as workspaceUtil from '../utils/workspace';
 
@@ -165,8 +166,12 @@ async function verifyRuntimeIsCompatible(localRuntime: ProjectRuntime, ui: IAzur
                 telemetryProperties.cancelStep = 'learnMoreRuntime';
                 throw new UserCancelledError();
             } else {
-                outputChannel.appendLine(localize('azFunc.updateFunctionRuntime', 'Updating FUNCTIONS_EXTENSION_VERSION to "{0}"...', localRuntime));
-                appSettings.properties.FUNCTIONS_EXTENSION_VERSION = localRuntime;
+                const newAppSettings: { [key: string]: string } = await getCliFeedAppSettings(localRuntime);
+                for (const key of Object.keys(newAppSettings)) {
+                    const value: string = newAppSettings[key];
+                    outputChannel.appendLine(localize('updateFunctionRuntime', 'Updating "{0}" to "{1}"...', key, value));
+                    appSettings.properties[key] = value;
+                }
                 await client.updateApplicationSettings(appSettings);
             }
         }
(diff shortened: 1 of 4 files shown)


src/commands/deploy.ts
```diff
@@ -165,8 +166,12 @@ async function verifyRuntimeIsCompatible(localRuntime: ProjectRuntime, ui: IAzur
                 telemetryProperties.cancelStep = 'learnMoreRuntime';
                 throw new UserCancelledError();
             } else {
-                outputChannel.appendLine(localize('azFunc.updateFunctionRuntime', 'Updating FUNCTIONS_EXTENSION_VERSION to "{0}"...', localRuntime));
-                appSettings.properties.FUNCTIONS_EXTENSION_VERSION = localRuntime;
+                const newAppSettings: { [key: string]: string } = await getCliFeedAppSettings(localRuntime);
+                for (const key of Object.keys(newAppSettings)) {
+                    const value: string = newAppSettings[key];
+                    outputChannel.appendLine(localize('updateFunctionRuntime', 'Updating "{0}" to "{1}"...', key, value));
```
StephenWeatherford: Will this message make sense without saying what it's updating?

src/tree/FunctionAppProvider.ts
```diff
@@ -49,7 +53,26 @@ export class FunctionAppProvider implements IChildProvider {
...
+        const runtime: ProjectRuntime = await getDefaultRuntime();
+        const appSettings: { [key: string]: string } = await getCliFeedAppSettings(runtime);
+        const site: Site = await createFunctionApp(actionContext, parent.credentials, parent.subscriptionId, parent.subscriptionDisplayName, showCreatingNode, appSettings);
         return new FunctionAppTreeItem(new SiteClient(site, parent), this._outputChannel);
     }
 }
+
+async function getDefaultRuntime(): Promise<ProjectRuntime> {
+    // Try to get VS Code setting for runtime (aka if they have a project open)
+    let runtime: ProjectRuntime | undefined = getFuncExtensionSetting(projectRuntimeSetting);
```
StephenWeatherford: telemetry?

src/utils/getCliFeedJson.ts
```diff
@@ -50,3 +52,24 @@ export function getFeedRuntime(runtime: ProjectRuntime): string {
...
     }
 }
+
+/**
+ * Returns the app settings that should be used when creating or deploying to a Function App, based on runtime
+ */
+export async function getCliFeedAppSettings(projectRuntime: ProjectRuntime): Promise<{ [key: string]: string }> {
+    // Use these defaults in case we can't get the cli-feed
+    let funcVersion: string = projectRuntime;
+    let nodeVersion: string = projectRuntime === ProjectRuntime.one ? '6.5.0' : '8.11.1';
```
StephenWeatherford: Would be good to have these as constants at top of file

src/commands/deploy.ts
EricJizbaMSFT: The 'key' should explain what it's updating

PR #461, opened 2018-07-18
Fix error when downloading app settings to empty file

### LINKED ISSUE #439: Unexpected end of JSON input error on "Download Remote Settings" feature
Repro steps:
1. Open VS Code
2. Create a new empty file like, for example, `remote-settings.json`
3. Select the Azure Functions extension
4. Select a remote Azure Function > Application Settings
5. Right-click and choose *Download Remote Settings*
6. Select target file from step 2 (e.g. `remote-settings.json`)

The error as mentioned below occurs. Expected behaviour would be that the appsettings from Azure would be copied in the file.

The error:

> Action: azureFunctions.appSettings.download
> Error type: SyntaxError
> Error Message: /Users/.../remote-settings.json: Unexpected end of JSON input

Version: 0.9.1
OS: darwin

EricJizbaMSFT:
Fixes #439

### DIFF
--- a/src/LocalAppSettings.ts
+++ b/src/LocalAppSettings.ts
@@ -154,10 +154,15 @@ async function setAppSetting(settings: ILocalAppSettings, localSettingsPath: str
 }
 
 export async function getLocalSettings(localSettingsPath: string): Promise<ILocalAppSettings> {
-    return await fse.pathExists(localSettingsPath) ?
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
The RegEx is testing to see if the file has anything other than whitespace (`\s` is shorthand for any whitespace character). If it only has whitespace, I think it's safe to assume we can overwrite whatever they have

PR #470, opened 2018-07-18
Fix AccessRights not being passed to C# HttpTriggers

### LINKED ISSUE #466: AccessRights for HttpTriggers is ignored
Repro Steps:
1. Create a new C# project
1. Create a HttpTrigger
1. Specify 'Anonymous' as the AccessRights

Expected:
'Anonymous' is used

Actual:
'Function' is used for AccessRights

Not sure if this is a bug in our code our in the dotnet templates

EricJizbaMSFT:
It was a bug in Microsoft.TemplateEngine.JsonCli.dll that wasn't passing any parameter of type 'choice'

Fixes #466


nturinski:
How do you make .dll changes??

EricJizbaMSFT:
magic

Jk it's a project given to us by the .NET CLI team. I had to make minor changes to get it to work for us, but it's largely their code. We should put the source code somewhere, but still haven't decided on the best place

PR #472, opened 2018-07-19
Temporarily remove proxies.json

EricJizbaMSFT:
I ran into this issue: https://github.com/Azure/azure-functions-core-tools/issues/562

It only reproduces on a Mac, but it breaks debugging. It sounds like the issue has already been fixed in the functions cli - just not released yet. @nturinski can you see if this reproduces for you as well? If so, we just won't include 'proxies.json' in projects by default for this release and add it back in the next.

Repro steps:
1. Create a JavaScript project on a Mac
1. Add an HttpTrigger
1. F5

### DIFF
--- a/src/commands/createNewProject/ScriptProjectCreatorBase.ts
+++ b/src/commands/createNewProject/ScriptProjectCreatorBase.ts
@@ -5,7 +5,7 @@
 
 import * as fse from 'fs-extra';
 import * as path from 'path';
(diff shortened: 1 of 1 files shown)


nturinski:
This repros on the latest azure-functions-core-tools on my Mac machine as well.  It not only breaks debugging, but the func host start commands completely fails due to proxies.

EricJizbaMSFT:
Existing users shouldn't really be impacted. Proxies weren't supported on Mac until just recently and we just added the file for this release.

Yes an existing user will run into this if they manually add a proxies.json file, but I think that's out of our scope.

PR #478, opened 2018-07-23
Update debug config for node projects due to breaking changes

### LINKED ISSUE #475: JavaScript debugging requires "language-worker" argument in order to debug
As of v2.0.1-beta.30 of the func cli, we are required to pass `--language-worker -- "--inspect=5858"` when running `func host start` in order to debug (See [here](https://github.com/Azure/azure-functions-core-tools/issues/521) for more info). This was an intentional breaking change in the func cli, and is _not_ a temporary workaround.

That being said, there is another bug (see https://github.com/Azure/azure-functions-core-tools/issues/557) where even that is not working on a mac. The _workaround_ is to add `"NODE_OPTIONS": "--inspect=5858"` to local.settings.json, but long term (and starting with v2.0.1-beta.32) we need to add the language-worker argument as described above.

This affects new projects and existing projects. We should probably warn existing users and change their project for them.

EricJizbaMSFT:
Fixes #475 

See the docs page that I added for more information. New projects should work going forward. Old projects will see this dialog:
screen shot 2018-07-23 at 2 47 40 pm

And the following if they decide to update:
screen shot 2018-07-23 at 3 02 47 pm

I implemented Option 1 from my docs page. It's technically just a workaround, but it's better because it only changes 1 file that is tracked by git. I thought about implementing _both_ option 1 _and_ option 2, but ultimately decided that wasn't worth it since option 2 is still not finalized (according to https://github.com/Azure/azure-functions-host/issues/3120).

### DIFF
--- a/docs/debugConfig.md
+++ b/docs/debugConfig.md
@@ -1,0 +1,22 @@
+# Breaking changes to JavaScript Debug Configuration
+
+The debug configuration for v2 of the Azure Functions runtime [has changed](https://github.com/Azure/azure-functions-core-tools/issues/521) and old configurations will likely fail with the error "Cannot connect to runtime process, timeout after 10000 ms". You will automatically be prompted to update your configuration when you open a project. However, you can manually update your project with one of the following options:
+
+The first option is to add the following to your `runFunctionsHost` task in `.vscode\tasks.json`:
+
+```json
+{
+    "options": {
+        "env": {
+            "languageWorkers:node:arguments": "--inspect=5858"
+        }
+    }
+}
+```
+
+The second option requires two steps:
+
+1. Edit your `runFunctionsHost` task in `.vscode\tasks.json` so that `command` is `func host start --language-worker -- \"--inspect=5858\"` instead of just `func host start`
+1. Add `"FUNCTIONS_WORKER_RUNTIME": "node"` to your `local.settings.json` file
+
+The recommended debug configuration going forward has not been finalized. It will likely be similar to Option 2, but without a change to `local.settings.json` (since that file is not tracked by git). See [here](https://github.com/Azure/azure-functions-host/issues/3120) for more information.
--- a/docs/debugConfig.md
+++ b/docs/debugConfig.md
@@ -2,21 +2,98 @@
 
 The debug configuration for v2 of the Azure Functions runtime [has changed](https://github.com/Azure/azure-functions-core-tools/issues/521) and old configurations will likely fail with the error "Cannot connect to runtime process, timeout after 10000 ms". You will automatically be prompted to update your configuration when you open a project. However, you can manually update your project with one of the following options:
 
-The first option is to add the following to your `runFunctionsHost` task in `.vscode\tasks.json`:
+The first option is to add `languageWorkers:node:arguments` to your `runFunctionsHost` task in `.vscode\tasks.json` as seen below:
 
 ```json
 {
-    "options": {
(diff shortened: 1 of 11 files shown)


docs/debugConfig.md
```diff
@@ -0,0 +1,22 @@
+# Breaking changes to JavaScript Debug Configuration
```
StephenWeatherford: should this mention JavaScript?

docs/debugConfig.md
```diff
@@ -0,0 +1,22 @@
...
+
+The debug configuration for v2 of the Azure Functions runtime [has changed](https://github.com/Azure/azure-functions-core-tools/issues/521) and old configurations will likely fail with the error "Cannot connect to runtime process, timeout after 10000 ms". You will automatically be prompted to update your configuration when you open a project. However, you can manually update your project with one of the following options:
+
+The first option is to add the following to your `runFunctionsHost` task in `.vscode\tasks.json`:
+
+```json
```
StephenWeatherford: It's really helpful to see the entire task (even if just an example).

src/commands/createNewProject/JavaScriptProjectCreator.ts
```diff
@@ -28,4 +32,33 @@ export class JavaScriptProjectCreator extends ScriptProjectCreatorBase {
...
+
+        return {
+            version: '2.0.0',
+            tasks: [
+                {
+                    label: funcHostTaskLabel,
+                    identifier: funcHostTaskId,
+                    type: 'shell',
+                    command: 'func host start',
+                    options: options,
```
StephenWeatherford: So this can return 

    options: undefined

is that okay?

src/commands/createNewProject/validateFunctionProjects.ts
```diff
@@ -68,3 +75,67 @@ function isInitializedProject(folderPath: string): boolean {
     const runtime: string | undefined = getFuncExtensionSetting(projectRuntimeSetting, folderPath);
     return !!language && !!runtime;
 }
+
+/**
+ * JavaScript debugging in the func cli had breaking changes in v2.0.1-beta.30. This verifies users are up-to-date with the latest working debug config.
+ * See https://aka.ms/AA1vrxa for more info
+ */
+async function verifyDebugConfigIsValid(folderPath: string, actionContext: IActionContext): Promise<void> {
```
StephenWeatherford: Need to guard against this function throwing?

docs/debugConfig.md
EricJizbaMSFT: Yeah I think so. C# wasn't affected. Python isn't supported by this extension yet. And these instructions don't work for Java (The partner team is handling Java support and I think they fixed everything directly in Maven. In any case, we can always change this after the fact to include Java since the aka.ms link will point directly to master)

src/commands/createNewProject/JavaScriptProjectCreator.ts
EricJizbaMSFT: Yep. 'undefined' values are excluded when stringifying JSON, which is what we want in this case

docs/debugConfig.md
EricJizbaMSFT: done

src/commands/createNewProject/validateFunctionProjects.ts
EricJizbaMSFT: good point

PR #481, opened 2018-07-24
Fix range error when creating new function app

### LINKED ISSUE #480: Fail to create function apps with error
Repro steps:
1. Right click the applied subscription and select 'Create Function App in Azure...'.
2. Check whether there is an error or not.

Expect:
There is no error.

Actual:
Pop up an error.
error

More Info:
This issue also reproduces on Mac and Linux.

Action: azureFunctions.createFunctionApp
Error type: RangeError
Error Message: {}

Version: 0.10.0
OS: win32

EricJizbaMSFT:
Fixes #480

I figured since it was described as an enum [in package.json](https://github.com/Microsoft/vscode-azurefunctions/blob/master/package.json#L518) that it would return undefined, but turns out it returns an empty string.

@StephenWeatherford feel free to say "I told you so" since you're always advocating for `!foo` instead of `foo === undefined` 😂😂😂 I've been doing this because of the weird edge case where a valid enum value can be `0`, but that argument feels less and less worth it

### DIFF
--- a/src/templates/DotnetTemplateRetriever.ts
+++ b/src/templates/DotnetTemplateRetriever.ts
@@ -6,6 +6,7 @@
 import * as fse from 'fs-extra';
 import { ProjectRuntime } from '../constants';
 import { ext } from '../extensionVariables';
+import { localize } from '../localize';
 import { dotnetUtils } from '../utils/dotnetUtils';
 import { downloadFile } from '../utils/fs';
 import { cliFeedJsonResponse } from '../utils/getCliFeedJson';
@@ -80,7 +81,7 @@ export function getDotnetVerifiedTemplateIds(runtime: string): string[] {
             case ProjectRuntime.beta:
                 return `${id}.2.x`;
             default:
-                throw new RangeError();
+                throw new RangeError(localize('invalidRuntime', 'Invalid runtime "{0}".', runtime));
         }
     });
 }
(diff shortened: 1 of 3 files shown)


StephenWeatherford:
:-)

src/templates/DotnetTemplateRetriever.ts
```diff
@@ -80,7 +81,7 @@ export function getDotnetVerifiedTemplateIds(runtime: string): string[] {
             case ProjectRuntime.beta:
                 return `${id}.2.x`;
             default:
-                throw new RangeError();
+                throw new RangeError(localize('invalidRuntime', 'Invalid runtime "{0}".', runtime));
```
StephenWeatherford: Does localize handle it if runtime is undefined?  Maybe consider String(runtime)

src/tree/FunctionAppProvider.ts
```diff
@@ -68,22 +68,22 @@ export class FunctionAppProvider implements IChildProvider {
 
 async function getDefaultRuntime(actionContext: IActionContext): Promise<ProjectRuntime> {
     // Try to get VS Code setting for runtime (aka if they have a project open)
-    let runtime: ProjectRuntime | undefined = getFuncExtensionSetting(projectRuntimeSetting);
+    let runtime: string | undefined = getFuncExtensionSetting(projectRuntimeSetting);
     actionContext.properties.runtimeSource = 'VSCodeSetting';
 
-    if (runtime === undefined) {
+    if (!runtime) {
```
PrashanthCorp: Is this the same behavior you intend to have for 0, null and other false-y values?

src/templates/DotnetTemplateRetriever.ts
EricJizbaMSFT: Yeah it's handled

src/tree/FunctionAppProvider.ts
EricJizbaMSFT: Yep

PR #487, opened 2018-07-30
Add new java templates

### LINKED ISSUE #486: Support more Java trigger templates
Start from Maven Azure Functions plugin: `1.0.0-beta-3`, users can add more Java trigger templates, including:
- EventGrid
- EventHub
- Cosmos DB
- Service Bus Queue
- Service Bus Topic

For short-term plan, check the pom and get the version of the plugin to determine which templates are available.
For long-term plan, we will check the possibility to add Java templates into the tooling feed and consume the templates directly from the feed.

@EricJizbaMSFT What do you think?

jdneo:
resolve #486 

We will resolve the templates information with the help of the information given by `mvn azure-functions:list`. If the command runs failed, we fallback to the basic templates set.

function


src/templates/parseJavaTemplates.ts
```diff
@@ -0,0 +1,192 @@
...
+        name: getResourceValue(resources, rawTemplate.metadata.name),
+        defaultFunctionName: rawTemplate.metadata.defaultFunctionName,
+        language: ProjectLanguage.Java,
+        userPromptedSettings: userPromptedSettings,
+        categories: [TemplateCategory.Core]
+    };
+}
+
+function parseSetting(data: object, resources: IResources, variables: IVariables): IFunctionSetting {
+    const rawSetting: IRawSetting = <IRawSetting>data;
```
EricJizbaMSFT: This file has a lot of code copied from `parseScriptTemplates` which we should leverage directly. For now, I suggest you just put `parseJavaTemplates` inside the same file as `parseScriptTemplates`. We should refactor it to a different file eventually (perhaps a separate PR right after this), but that will make it easier to review

src/templates/parseJavaTemplates.ts
```diff
@@ -0,0 +1,192 @@
...
+    return getResourceValue(resources, <string>data);
+}
+
+function getResourceValue(resources: IResources, data: string): string {
+    const matches: RegExpMatchArray | null = data.match(/\$(.*)/);
+    return matches !== null ? resources.en[matches[1]] : data;
+}
+
+export async function parseJavaTemplates(allTemplates: IFunctionTemplate[], functionAppPath: string, telemetryProperties?: TelemetryProperties): Promise<IFunctionTemplate[]> {
+    const javaScriptTemplates: IFunctionTemplate[] = allTemplates.filter((t: IFunctionTemplate) => t.language === ProjectLanguage.JavaScript);
```
EricJizbaMSFT: Please rename `basicJavaTemplates` to `backupJavaTemplates`. Also we shouldn't need to get  these most of the time.

I suggest something like this at the end of this function:
```typescript
if (templates.length > 0) {
    return templates;
} else {
    // move logic to get backup templates here
}
```

src/utils/mavenUtils.ts
```diff
@@ -38,7 +38,7 @@ export namespace mavenUtils {
         });
     }
 
-    export async function executeMvnCommand(telemetryProperties: TelemetryProperties, outputChannel: vscode.OutputChannel | undefined, workingDirectory: string | undefined, ...args: string[]): Promise<string> {
+    export async function executeMvnCommand(telemetryProperties: TelemetryProperties | undefined, outputChannel: vscode.OutputChannel | undefined, workingDirectory: string | undefined, ...args: string[]): Promise<string> {
```
EricJizbaMSFT: Why did you change this function? When is telemetryProperties undefined?

src/templates/parseJavaTemplates.ts
```diff
@@ -0,0 +1,192 @@
...
+        }
+    } catch (error) {
+        // Swallow the exception if the plugin do not support list templates information.
+        if (telemetryProperties) {
+            telemetryProperties.parseJavaTemplateErrors = parseError(error).message;
+        }
+    }
+    const templates: IFunctionTemplate[] = [];
+    const cachedResources: object | undefined = ext.context.globalState.get<object>(`FunctionTemplateResources.${ProjectRuntime.beta}`);
+    const cachedConfig: object | undefined = ext.context.globalState.get<object>(`FunctionTemplateConfig.${ProjectRuntime.beta}`);
```
EricJizbaMSFT: So this information isn't included in the maven list function? That means there could be a mismatch if maven has older templates but the config is newer/different. I'm fine with that risk if you are, but I would add a comment explaining that

jdneo: Emm, that's a good question. Cache the bindings.json & Resources.json into Maven plugin seems much safer. Let's me update the Maven side first.

src/utils/mavenUtils.ts
jdneo: This is due to this method signature:
```javascript
getTemplates(language: string, runtime: string = ProjectRuntime.one, functionAppPath: string, templateFilter?: string, telemetryProperties?: TelemetryProperties)
```
If the `telemetryProperties` is not optional, we need to pass it in the unit test

jdneo:
@EricJizbaMSFT .

Now the maven plugin will print three file content when calling `mvn azure-functions:list`:
- [templates](https://github.com/Microsoft/azure-maven-plugins/blob/develop/azure-functions-maven-plugin/src/main/resources/templates.json)
- [bindings](https://github.com/Microsoft/azure-maven-plugins/blob/develop/azure-functions-maven-plugin/src/main/resources/bindings.json)
- [resources](https://github.com/Microsoft/azure-maven-plugins/blob/develop/azure-functions-maven-plugin/src/main/resources/resources.json)

src/templates/parseScriptTemplates.ts
```diff
@@ -193,3 +214,46 @@ export function parseScriptTemplates(rawResources: object, rawTemplates: object[
     }
     return templates;
 }
+
```
EricJizbaMSFT: Filed this issue to refactor as discussed https://github.com/Microsoft/vscode-azurefunctions/issues/505

PR #494, opened 2018-08-06
Integration tests for Azure commands

nturinski:
Leverages TestAzureAccount to use a test Azure subscription to run Azure commands against.  Currently only createFunctionApp and deleteFunctionApp are tested.

Seems like I broke a test in the process.  I'd like to figure out how to create an output channel just once at the beginning of all suites and dispose of it after all tests are completed rather than creating/disposing of it per testing suite.

The build is also broken due to npm 'any' issues since TestAccountAzure isn't published in the azureextensionui package yet.


test/createAzureResource.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            ext.outputChannel = outputChannel;
+            const testAccount: TestAzureAccount = new TestAzureAccount();
+            await testAccount.getTestSubscription();
+            ext.tree = new AzureTreeDataProvider(new FunctionAppProvider(ext.outputChannel), 'azureFunctions.startTesting', undefined, testAccount);
+            await fse.ensureDir(testFolderPath);
+        });
+
+        suiteTeardown(async () => {
+            ext.ui = new TestUserInput([appName]);
+            const subscriptionNode: IAzureNode<IAzureTreeItem> = (await ext.tree.getChildren())[0];
```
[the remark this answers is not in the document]
nturinski: Yeah, makes sense for the clean-up method so we don't rely on something that can break.  Though I think I would want to keep it that way for the actual tests since it's a good way to validate the end-to-end process of creating an app is functional.

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            await fse.remove(testFolderPath);
+
+        });
+
+        const createNewFunctionApp: string = 'Create New Function App';
+        test(createNewFunctionApp, async () => {
+            const testInputs: string[] = [appName, '$(plus) Create new resource group', appName, '$(plus) Create new storage account', appName, 'West US'];
+            ext.ui = new TestUserInput(testInputs);
+            const subscriptionNode: IAzureNode<IAzureTreeItem> = (await ext.tree.getChildren())[0];
+            await vscode.commands.executeCommand('azureFunctions.createFunctionApp', subscriptionNode);
```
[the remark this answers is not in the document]
nturinski: You know, I saw the tree.showNodePicker and assumed it wouldn't automatically use the subscription node, but I think you're right as those changes are in the ui package.  I'll try and see.

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,69 @@
...
+            ext.ui = new TestUserInput([appName]);
+            const newNode: IAzureNode<FunctionAppTreeItem> = <IAzureNode<FunctionAppTreeItem>>await ext.tree.showNodePicker(FunctionAppTreeItem.contextValue);
+            assert.equal(newNode.treeItem.label, appName);
```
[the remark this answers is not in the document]
nturinski: Okay, makes sense.... Man, I thought I was so clever too :(

Would it make sense to make that another test to verify that our code works though?

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,69 @@
...
+        test(deleteFunctionApp, async () => {
+            ext.ui = new TestUserInput([appName]);
+            const newNode: IAzureNode<FunctionAppTreeItem> = <IAzureNode<FunctionAppTreeItem>>await ext.tree.showNodePicker(FunctionAppTreeItem.contextValue);
+            ext.ui = new TestUserInput([DialogResponses.deleteResponse.title, DialogResponses.yes.title]);
+            await vscode.commands.executeCommand('azureFunctions.deleteFunctionApp', newNode);
+            const deletedNode: IAzureNode<IAzureTreeItem> | undefined = await ext.tree.findNode(newNode.id);
+            assert.equal(deletedNode, undefined);
+        });
+    });
+}
```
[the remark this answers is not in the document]
nturinski: Please add?  Am I rubbing off on you? 😉

test/createAzureResource.test.ts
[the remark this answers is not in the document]
EricJizbaMSFT: I guess I just don't want it to be the _only_ thing we verify, but I'm totally fine if you leave the assert in this test (not sure it's worth its own separate test)

test/createNewProject.test.ts
```diff
@@ -20,19 +20,17 @@ import { validateVSCodeProjectFiles } from './initProjectForVSCode.test';
...
     // tslint:disable-next-line:no-function-expression
     suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
         this.timeout(120 * 1000);
+        const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('Azure Functions Test');
+        ext.outputChannel = outputChannel;
         await fse.ensureDir(testFolderPath);
     });
 
     suiteTeardown(async () => {
-        outputChannel.dispose();
```
EricJizbaMSFT: I feel like the call to dispose should still be here

test/createNewProject.test.ts
```diff
@@ -20,19 +20,17 @@ import { validateVSCodeProjectFiles } from './initProjectForVSCode.test';
...
     this.timeout(60 * 1000);
-
     const testFolderPath: string = path.join(os.tmpdir(), `azFunc.createNewProjectTests${fsUtil.getRandomHexString()}`);
-    const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('Azure Functions Test');
-    ext.outputChannel = outputChannel;
 
     // tslint:disable-next-line:no-function-expression
     suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
         this.timeout(120 * 1000);
+        const outputChannel: vscode.OutputChannel = vscode.window.createOutputChannel('Azure Functions Test');
```
EricJizbaMSFT: Honestly you probably don't need to create and dispose of outputChannels in these tests anymore. You can just pass in `ext.OutputChannel` if necessary (it should already be initialized since you made sure the extension is activated as a part of this PR). That might fix the failing tests in the CI, too

test/_init.test.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
```
nturinski: This is necessary because you can't await commands in index.ts as it's not an async function.  This basically just makes sure the extension is activated before we start trying to do any tests.

src/commands/createFunctionApp.ts
```diff
@@ -4,14 +4,15 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureTreeDataProvider, IActionContext, IAzureNode, IAzureParentNode } from 'vscode-azureextensionui';
```
StephenWeatherford: Minor grievance with your title...  These are by definition not unit tests because you're enabling running against Azure.

test/_init.test.ts
```diff
@@ -0,0 +1,12 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ISuiteCallbackContext } from 'mocha';
+import { commands } from 'vscode';
+
+suite('Initialize extension', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(120 * 1000); // two minutes should be enough time
+    await commands.executeCommand('azureFunctions.refresh'); //activate the extension before testing begins to make sure activation doesn't change any extensionVariables
```
StephenWeatherford: FYI, my solution was to simply use "if (!ext.tree) { ext.tree = new … }" kind of code in Extension.ts.

test/_init.test.ts
StephenWeatherford: So are you depending on this to get run first by the fact that "_init.test.ts" is alphabetized first?  Better would be to use a suiteSetup(), which can appear in any test file.

nturinski: I considered using suiteSetup, but I didn't want to have to do it per test file.  In that case, I had to call suiteSetup in the first test that would be run (which seemed to be in alphabetical order) and came to this solution.

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Site } from 'azure-arm-website/lib/models';
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as vscode from 'vscode';
+import { AzureTreeDataProvider, DialogResponses, TestAzureAccount, TestUserInput } from 'vscode-azureextensionui';
+import { ext } from '../src/extensionVariables';
+import { localize } from '../src/localize';
+import { FunctionAppProvider } from '../src/tree/FunctionAppProvider';
+import * as fsUtil from '../src/utils/fs';
+
+// This ensures that this will only run on nightly cron builds
```
StephenWeatherford: You're being too technology-specific here.  This determination should be based on something more directly tied to the intention, e.g. a TEST_TYPE environment variable or ENABLE_LONG_RUNNING_TESTS or something.  Then in travis.yml you can set that up if it's a cron build.  That'll make it much easier to enable when not on travis.

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,71 @@
...
+            await vscode.commands.executeCommand('azureFunctions.deleteFunctionApp');
+            const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+            const deletedApp: Site | undefined = await client.webApps.get(resourceName, resourceName);
+            assert.ifError(deletedApp); // if app was deleted, get() returns null.  assert.ifError throws if the value passed is not null/undefined
+        });
+    });
+} else {
+    const skippingTests: string = localize('skippingTests', 'Skipping Azure tests for a non-cron build');
+    // tslint:disable-next-line:no-console
+    console.log(skippingTests);
```
StephenWeatherford: use "this.skip()"

test/createAzureResource.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as vscode from 'vscode';
+import { AzureTreeDataProvider, DialogResponses, TestAzureAccount, TestUserInput } from 'vscode-azureextensionui';
+import { ext } from '../src/extensionVariables';
+import { localize } from '../src/localize';
+import { FunctionAppProvider } from '../src/tree/FunctionAppProvider';
+import * as fsUtil from '../src/utils/fs';
+
+// This ensures that this will only run on nightly cron builds
+if (process.env.TRAVIS_EVENT_TYPE === 'cron') {
```
StephenWeatherford: Much better would be to either:

1) create longRunningTest()/longRunningTest() functions that wrap suite()/test() and do the skip automatically

longRunningSuite("Create Azure Resources", () => {

2) use JavaScript decorators to indicate long-running suites/tests (I've never tried this in JavaScript, but you can find info here: https://github.com/tc39/proposal-decorators)

@longRunning
suite("Create Azure Resources", () => ...

test/_init.test.ts
StephenWeatherford: Not sure that mine's better, though.

test/_init.test.ts
StephenWeatherford: It's not per test file.

StephenWeatherford: And doesn't matter where it appears.

nturinski: Just to confirm, any suiteSetup() call will be executed before any test is run in all of the test files within the directory?

StephenWeatherford: Mocha documentation is horrible, but see https://github.com/Microsoft/vscode-docker/blob/145935f31ac58f2e409db23214e8280cde67845c/test/global.test.ts#L48 for example. suite()/test() etc. just register the tests they represent. Tests are not run until all top-level stuff is registered. So it doesn't matter when/where it's found, it will be run first, and you can have multiple of them.

test/createAzureResource.test.ts
StephenWeatherford: Example of 1: https://github.com/Microsoft/vscode-docker/blob/145935f31ac58f2e409db23214e8280cde67845c/test/buildAndRun.test.ts#L115

test/_init.test.ts
EricJizbaMSFT: I just want to clarify that "top-level" is important in what Stephen is saying. For example:
```
suiteSetup1(...)

suite('example' () => {
    
    suiteSetup2(...)
})
```

suiteSetup1 will be run before everything. suiteSetup2 will only be run before the tests inside 'example'

Also don't necessarily use my existing test code as an example. I don't know if I knew that when originally writing these tests.

test/_init.test.ts
EricJizbaMSFT: I think I prefer Nathan's approach for the following reasons:
1. It doesn't require changes to production code
1. It automatically initializes/disposes a few things for us (i.e. extensionContext and outputChannel) rather than having to mock that stuff

test/createAzureResource.test.ts
```diff
@@ -16,51 +16,56 @@ import { localize } from '../src/localize';
 import { FunctionAppProvider } from '../src/tree/FunctionAppProvider';
 import * as fsUtil from '../src/utils/fs';
 
+suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
+    this.timeout(120 * 1000); // two minutes should be enough time
```
StephenWeatherford: And 640K ought to be enough for anybody... :-)

test/createAzureResource.test.ts
```diff
@@ -16,51 +16,56 @@ import { localize } from '../src/localize';
...
 
-        suiteTeardown(async () => {
-            const client: ResourceManagementClient = getResourceManagementClient(testAccount);
-            ext.outputChannel.appendLine(`Deleting resource group "${resourceName}..."`);
-            await client.resourceGroups.deleteMethod(resourceName);
-            ext.outputChannel.appendLine(`Resource group "${resourceName}" deleted.`);
-            ext.tree.dispose();
-        });
+    suiteTeardown(async function (this: IHookCallbackContext): Promise<void> {
+        if (process.env.TRAVIS_EVENT_TYPE !== 'cron') {
```
StephenWeatherford: ENABLE_LONG_RUNNING_TESTS?

tools/JsonCli/src/obj/Debug/netcoreapp2.0/Microsoft.TemplateEngine.JsonCli.AssemblyInfo.cs
```diff
@@ -0,0 +1,23 @@
+//------------------------------------------------------------------------------
```
StephenWeatherford: Are these files intentional?

Issue #483, opened 2018-07-25
Restarting log stream after disconnect results in multiple streams with the same name


pcnate:
while streaming logs, the dropdown selection becomes polluted with old Log Streams.

pcnate:
sometimes the log becomes disconnected, sometimes during a deploy. Start streaming does not check for existing "session?" in the dropdown under output.

![azure-functions-log-screenshot](https://user-images.githubusercontent.com/621940/43330865-ff4645ec-9189-11e8-899b-f40a364ba4bd.png)

ubikusss:
Did a few tests. Normally, when you try to start the log stream and there is a log stream already active, you get an information message. 
Still, if you deploy and start the stream log again you will get a second log stream in the output window.
Thanks!

pcnate:
I can't verify but I think a few times the log stream became disconnected without a deploy and created another log stream in the output window. I think the solution would be to always check for an existing log stream output and either remove the old one and create a new log stream output or reconnect to the old one.

EricJizbaMSFT:
Fixed in https://github.com/Microsoft/vscode-azurefunctions/pull/501

PR #501, opened 2018-08-08
Fix for https://github.com/Microsoft/vscode-azurefunctions/issues/483

ubikusss:


src/extension.ts
```diff
@@ -62,6 +62,9 @@ export function activate(context: vscode.ExtensionContext): void {
     ext.outputChannel = outputChannel;
     context.subscriptions.push(outputChannel);
 
+    // tslint:disable-next-line:prefer-const
+    let logsOutputChannels: { [channelName: string]: { channelObj: vscode.OutputChannel } } = {};
```
EricJizbaMSFT: If you're opposed to the `prefer-const` tslint rule I'd be curious to hear why, but otherwise please just use `const` instead of `let` rather than disabling the rule

src/commands/logstream/startStreamingLogs.ts
```diff
@@ -12,15 +12,15 @@ import { localize } from '../../localize';
 import { FunctionAppTreeItem } from '../../tree/FunctionAppTreeItem';
 import { ILogStreamTreeItem } from './ILogStreamTreeItem';
 
-export async function startStreamingLogs(context: vscode.ExtensionContext, tree: AzureTreeDataProvider, node?: IAzureNode<ILogStreamTreeItem>): Promise<void> {
+export async function startStreamingLogs(context: vscode.ExtensionContext, tree: AzureTreeDataProvider, logsOutputChannels: { [channelName: string]: { channelObj: vscode.OutputChannel } }, node?: IAzureNode<ILogStreamTreeItem>): Promise<void> {
```
EricJizbaMSFT: You can just declare `logsOutputChannels` above this function rather than passing it in. That'll keep it scoped to this file and make extension.ts a little less cluttered

src/commands/logstream/startStreamingLogs.ts
```diff
@@ -38,12 +38,12 @@ export async function startStreamingLogs(context: vscode.ExtensionContext, tree:
             await treeItem.client.updateLogsConfig(logsConfig);
         }
 
-        if (!treeItem.logStreamOutputChannel) {
+        if (logsOutputChannels[treeItem.logStreamLabel] == null) {
```
EricJizbaMSFT: Wouldn't this be undefined instead of null? In any case, I prefer `!logsOutputChannels[treeItem.logStreamLabel]` to cover all bases

EricJizbaMSFT: Also you should change the type of `channelObj` to `vscode.OutputChannel | undefined`.

The linter doesn't like `!<vscode.OutputChannel>` since theoretically that would always be defined/true. However, it makes sense to have `!<vscode.OutputChannel | undefined>` because it could be true or false

src/extension.ts
ubikusss: Changed to `const`. I was thinking that it's going to be reassigned but I see it works.

src/commands/logstream/startStreamingLogs.ts
ubikusss: I tried that but I got some errors where the channel was disabled, In the `extension.ts` scope never happened. We need the extension scope for this to work. Kept it the same for now.

src/commands/logstream/startStreamingLogs.ts
ubikusss: Although counter intuitive, actually, `==null` covers both cases (still, I changed it to: `!logsOutputChannels[treeItem.logStreamLabel]`):

_Debug console output:_
`logsOutputChannels[treeItem.logStreamLabel]
undefined
(logsOutputChannels[treeItem.logStreamLabel] == null)
true
(logsOutputChannels[treeItem.logStreamLabel] == undefined)
true`

Also, did not use `vscode.OutputChannel | undefined` because `appservice.startStreamingLogs` does not accept it:
`export declare function startStreamingLogs(client: SiteClient, outputChannel: vscode.OutputChannel, path?: string): Promise<ILogStream>;`

Let me know if you have another idea, I am open :)

ubikusss: Changed back to `logsOutputChannels[treeItem.logStreamLabel] == null` due to the reason you mentioned.

EricJizbaMSFT: That's an interesting point about `== null`. However, we prefer `===` instead of `==`, which would have different behavior. We even have a "triple-equals" tslint rule specified [here](https://github.com/Microsoft/vscode-azurefunctions/blob/master/tslint.json#L99), but I guess we currently have `allow-null-check` specified. I think it's weird to allow that for just one counterintuitive edge case, so we will probably disallow it going forward.

Here's what I suggest to make `vscode.OutputChannel | undefined` work:
```typescript
let outputChannel: vscode.OutputChannel | undefined = logsOutputChannels[treeItem.logStreamLabel].channelObj;
if (!outputChannel) {
    outputChannel = vscode.window.createOutputChannel(localize('logStreamLabel', '{0} - Log Stream', treeItem.logStreamLabel));
    context.subscriptions.push(outputChannel);
    logsOutputChannels[treeItem.logStreamLabel] = { channelObj: outputChannel };
}
treeItem.logStream = await appservice.startStreamingLogs(treeItem.client, outputChannel, treeItem.logStreamPath);
```

src/extension.ts
```diff
@@ -62,6 +62,8 @@ export function activate(context: vscode.ExtensionContext): void {
     ext.outputChannel = outputChannel;
     context.subscriptions.push(outputChannel);
 
+    const logsOutputChannels: { [channelName: string]: { channelObj: vscode.OutputChannel } } = {};
```
EricJizbaMSFT: I still think this should be `vscode.OutputChannel | undefined` as described in my other comment, but I also don't think `channelObj` is necessary. You can just do: `{ [channelName: string]: vscode.OutputChannel | undefined }`

src/commands/logstream/startStreamingLogs.ts
EricJizbaMSFT: Sounds fine

src/commands/logstream/startStreamingLogs.ts
ubikusss: Makes total sense, thanks!

ubikusss:
Made the changes. Thank you very much Eric, most helpful!

PR #506, opened 2018-08-13
Ensure C# runtime logic is always run

### LINKED ISSUE #490: C# create IoThub triggered function through API won't set azureFunctions.deploySubpath
If I invoke the API directly in this way:
vscode.commands.executeCommand('azureFunctions.createNewProject', azureFunctionsFolder,
          'C#','beta', false, 'Azure.Function.CSharp.IotHubTrigger.2.x', // other functions settings

setting.json file is as follow:
 
{
  "azureFunctions.projectRuntime": "beta",
  "azureFunctions.projectLanguage": "C#",
  "azureFunctions.templateFilter": "All"
}

Compared with function creation in UI, it does not include the setting of 
"azureFunctions.deploySubpath": "bin/Release/netstandard2.0/publish"

EricJizbaMSFT:
`getRuntime` has some important logic in the CSharpProjectCreator, but it isn't always run. I've separated that logic out into it's own function and added some related tests.

Fixes #490

### DIFF
--- a/src/commands/createNewProject/CSharpProjectCreator.ts
+++ b/src/commands/createNewProject/CSharpProjectCreator.ts
@@ -25,6 +25,8 @@ export class CSharpProjectCreator extends ProjectCreatorBase {
     private _debugSubpath: string;
     private _runtime: ProjectRuntime;
 
+    private _detectedRuntime: boolean = false;
+
     public async addNonVSCodeFiles(): Promise<void> {
         await dotnetUtils.validateDotnetInstalled();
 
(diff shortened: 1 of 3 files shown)


src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -25,6 +25,8 @@ export class CSharpProjectCreator extends ProjectCreatorBase {
     private _debugSubpath: string;
     private _runtime: ProjectRuntime;
 
+    private _detectedRuntime: boolean = false;
```
nturinski: nit: _detectedRuntime sounds like it'd hold the runtime to me.  What about _hasDetectedRuntime so we know it's a state?

src/commands/createNewProject/CSharpProjectCreator.ts
```diff
@@ -36,9 +38,23 @@ export class CSharpProjectCreator extends ProjectCreatorBase {
         this._runtime = await tryGetLocalRuntimeVersion() || await promptForProjectRuntime(this.ui);
         const identity: string = `Microsoft.AzureFunctions.ProjectTemplate.CSharp.${this._runtime === ProjectRuntime.one ? '1' : '2'}.x`;
         await executeDotnetTemplateCommand(this._runtime, this.functionAppPath, 'create', '--identity', identity, '--arg:name', projectName);
+
+        if (!this._detectedRuntime) {
+            await this.detectRuntime();
```
nturinski: Do we need to call this here and on line 49?

EricJizbaMSFT: Yes that's actually the main fix for the bug. There's several different entry points for C# projects and not all of them were calling `getRuntime`. I refactored most of the logic into the separate `detectRuntime` function so that I could call it twice

PR #510, opened 2018-08-14
Include backup templates with the extension

### LINKED ISSUE #462: Include backup templates with extension
Repro steps:
Tried to create and azure function with C# and reported 504 Gateway Timeout. There's a problem with the region, I'm from Chile and when I tried to create with Javascript and installed the npm package reported the same 504 error with azureedge.net, but I have a Linux virtual machine connected to West US, connected via SSH and tried to download the final manualy and was OK. To problem may be with some Internet Providers from Chile and the CDN.

Action: azureFunctions.createFunction
Error type: Error
Error Message: Command "dotnet restore --packages . --source https://www.myget.org/F/azure-appservice/api/v3/index.json --source https://api.nuget.org/v3/index.json" failed with exit code "1":
  Restoring packages for C:\Users\psanchez\AppData\Local\Temp\7a84b73ba0\packages.csproj...
  Failed to download package 'microsoft.azurefunctions.projecttemplates.2.0.0-beta-10167' from 'https://www.myget.org/F/azure-appservice/api/v3/flatcontainer/microsoft.azurefunctions.projecttemplates/2.0.0-beta-10167/microsoft.azurefunctions.projecttemplates.2.0.0-beta-10167.nupkg'.
  Response status code does not indicate success: 504 (Gateway Timeout).
  Failed to download package 'microsoft.azurefunctions.projecttemplates.2.0.0-beta-10167' from 'https://www.myget.org/F/azure-appservice/api/v3/flatcontainer/microsoft.azurefunctions.projecttemplates/2.0.0-beta-10167/microsoft.azurefunctions.projecttemplates.2.0.0-beta-10167.nupkg'.
  Response status code does not indicate success: 504 (Gateway Timeout).
  Failed to download package 'azure.functions.templates.2.0.0-beta-10167' from 'https://www.myget.org/F/azure-appservice/api/v3/flatcontainer/azure.functions.templates/2.0.0-beta-10167/azure.functions.templates.2.0.0-beta-10167.nupkg'.
  Response status code does not indicate success: 504 (Gateway Timeout).
C:\Program Files\dotnet\sdk\2.1.200\NuGet.targets(114,5): error : Failed to download package 'microsoft.azurefunctions.projecttemplates.2.0.0-beta-10167' from 'https://www.myget.org/F/azure-appservi

EricJizbaMSFT:
We've had enough users run into problems downloading and using templates from the cli feed that we want to include them with the extension. I've split it up into two commits so that you could see just the code changes by themselves: https://github.com/Microsoft/vscode-azurefunctions/commit/000bad339ecd4baa7866ce11e0a62b3e694da2be

Fixes #462


src/templates/DotnetTemplateRetriever.ts
```diff
@@ -48,13 +49,22 @@ export class DotnetTemplateRetriever extends TemplateRetriever {
         const itemFilePath: string = getDotnetItemTemplatePath(runtime);
         await downloadFile(cliFeedJson.releases[templateVersion].itemTemplates, itemFilePath);
 
-        this._rawTemplates = <object[]>JSON.parse(await executeDotnetTemplateCommand(runtime, undefined, 'list'));
-        return parseDotnetTemplates(this._rawTemplates, runtime);
+        return await this.parseTemplates(runtime);
+    }
+
+    protected async getTemplatesFromBackup(runtime: ProjectRuntime): Promise<IFunctionTemplate[]> {
+        await fse.copy(ext.context.asAbsolutePath(path.join('resources', 'backupDotnetTemplates')), getDotnetTemplatesPath(), { overwrite: true, recursive: false });
```
StephenWeatherford: create a getDotnetTemplatesBackupPath()?

StephenWeatherford: You're overwriting a location inside the installation location of the extension?  Or am I missing something?  That doesn't sound like a good idea.

EricJizbaMSFT: In terms of `getDotnetTemplatesBackupPath`, not a huge fan of creating simple functions that are only used once.

And yep - that's exactly what I'm doing. It's the same location we use when downloading these nuget packages directly from the cli-feed. If you're worried about overwriting, I'm doing it because the packages we downloaded from the cli-feed might be corrupt (otherwise why would we have gotten to the 'backup' case?)

If you're worried about using the installation location of the extension as a storage place in the first place, I'm doing it because we have to reference the nuget package every time the user creates a template. We can't just download the nuget package once and throw it away (otherwise I would've used the temp folder). `ExtensionContext` provides a workspace-specific storage path, but I'm pretty sure you're supposed to use the extension's install path for 'global' storage

StephenWeatherford: In general it's a no-no because users don't necessarily have write access to where apps are installed. But I believe these aren't stored in program files.

StephenWeatherford: > In terms of getDotnetTemplatesBackupPath, not a huge fan of creating simple functions that are only used once.

It's more about discoverability and organization, since the other locations are discoverable from where they're grouped in those simple functions.

StephenWeatherford: ... but just consider it a suggestion

EricJizbaMSFT: Yeah these are the install locations:
* Windows `%USERPROFILE%\.vscode\extensions`
* macOS `~/.vscode/extensions`
* Linux `~/.vscode/extensions`

I'm open to storing this somewhere else if you have a better idea. Although I will say that another benefit of storing these in the extension's folder itself is that it will be cleaned up when un-installing the extension.

PR #512, opened 2018-08-15
extract java part into parseJavaTemplates.ts

### LINKED ISSUE #505: Move parseJavaTemplates to its own file
Rather than keeping it in parseScriptTemplates.ts

jdneo:
resolve #505 

Also did little changes of the method: `parseScriptTemplate()` - extract two method from it:
- parseCommonSettingsMap()
- parseUserPromptedSettings()

to make the `parseJavaTemplates` has its own version of `parseJavaTemplate()`


EricJizbaMSFT:
I don't think I was clear about what I wanted for #505. I've renamed it from `Refactor parseJavaTemplates to its own file` to `Move parseJavaTemplates to its own file`. The whole reason for doing this separately is so that it's easier to review (and easier to see changes in the commit history). I'm not saying you can't refactor. I'm just saying that moving code or renaming files should always be done on its own.

In terms of the refactor, what's the reason for `parseJavaTemplates` to have its own version of `parseJavaTemplate`? It looks almost exactly the same and I'd like to avoid copied code

jdneo:
I see. Update the code that leave the `parseScriptTemplate()` as it is. Just changed some comments.

PR #525, opened 2018-08-23
Python Project Creating/Local Debugging

### LINKED ISSUE #214: Support creating Python on Linux Consumption

nturinski:
Fixes #214 

Made some assumptions for Linux, but still need to do testing.

Requires that the experimental Python extension is installed (supposed to release in August) and `python -m pip install --pre ptvsd` is run WHILE running in the virtual environment.

EDIT: More work for deploy is required, but I was thinking that could be another PR.


src/commands/createNewProject/createNewProject.ts
```diff
@@ -45,7 +46,8 @@ export async function createNewProject(
             const languagePicks: QuickPickItem[] = [
                 { label: ProjectLanguage.JavaScript, description: '' },
                 { label: ProjectLanguage.CSharp, description: '' },
-                { label: ProjectLanguage.Java, description: '' }
+                { label: ProjectLanguage.Java, description: '' },
+                { label: ProjectLanguage.Python, description: '(Preview)' }
```
EricJizbaMSFT: Please remove 'Python' from this quick pick until we have full support and we know we're ready to release

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+                    host: 'localhost',
+                    preLaunchTask: funcHostTaskId
+                }
+            ]
+        };
+    }
+
+    public async addNonVSCodeFiles(): Promise<void> {
+        // need a custom message for warning as this is a dependency
+        if (await validateFuncCoreToolsInstalled(true)) {
```
EricJizbaMSFT: I don't think `validateFuncCoreToolsInstalled` will work as it is. For example, the message currently says the following:
> You must have the Azure Functions Core Tools installed to debug your local functions.

Which is not true for Python (since you need it for more things). Also we should make sure the dialog is modal since this blocks project creation.

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+                    presentation: {
+                        reveal: 'always'
+                    },
+                    options: {
+                        env: {
+                            // tslint:disable-next-line:no-invalid-template-strings
+                            'languageWorkers:python:arguments': '-m ptvsd --server --port 9091 --file'
+                        }
+                    },
+                    problemMatcher: []
```
EricJizbaMSFT: Don't you need the problemMatcher?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    public async addNonVSCodeFiles(): Promise<void> {
+        // need a custom message for warning as this is a dependency
+        if (await validateFuncCoreToolsInstalled(true)) {
+            await this.validatePythonVersion();
+            await this.ensureVirtualEnviornment();
+            await this.createPythonProject();
+        } else {
+            throw new UserCancelledError();
```
EricJizbaMSFT: I don't quite follow why this case should be a UserCancelledError?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+                }
+            ]
+        };
+    }
+
+    private async validatePythonVersion(): Promise<void> {
+        const minReqVersion: string = '3.6.0';
+        let incompatiableVersion: boolean = false;
+        const pythonVersionIncompatiable: string = localize('pythonVersionIncompatiable', 'Current version of Python is incompatiable with Azure Functions Core Tools. Python 3.6 is required.');
+        try {
```
EricJizbaMSFT: Please pull out the copied code into a separate function that you call for each possible python alias

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+                    problemMatcher: []
+                }
+            ]
+        };
+    }
+
+    private async validatePythonVersion(): Promise<void> {
+        const minReqVersion: string = '3.6.0';
+        let incompatiableVersion: boolean = false;
+        const pythonVersionIncompatiable: string = localize('pythonVersionIncompatiable', 'Current version of Python is incompatiable with Azure Functions Core Tools. Python 3.6 is required.');
```
EricJizbaMSFT: Reference `minReqVersion` in the string rather than hard-coding '3.6'. Otherwise we might forget to update the message even if we change `minReqVersion`

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,186 @@
...
+            // if there is no venv path, prompt user and recursively call function again
+            await this.ensureVirtualEnviornment();
+            return await this.getVenvActivatePath(platform);
+        }
+    }
+
+    private async createPythonProject(): Promise<void> {
+        switch (process.platform) {
+            case Platform.Windows:
+                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `${await this.getVenvActivatePath(Platform.Windows)} && func init ./ --worker-runtime python`);
```
EricJizbaMSFT: Shouldn't this be `|` instead of `&&`?

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: Yeah, my comment was to start a conversation about this issue.  Do we want to change `validateFuncCoreToolsInstalled` to allow custom messages or just strip the message out entirely and let the function that calls it handle it?

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: Yeah, could you help me with that? 👼

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: No, I think cpUtils spawns a bash terminal so `|` doesn't work even in Windows.

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: The prompt for the Functions installation is a forced modal prompt.  But if they hit Cancel, it wasn't being handled as a `UserCancelledError`.

nturinski: Refer to this: https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/pickFuncProcess.ts#L19

EricJizbaMSFT: >  But if they hit Cancel, it wasn't being handled as a UserCancelledError.

Then we should fix that?

I honestly don't get why pickFuncProcess has a `UserCancelledError` either (Yes I realize the irony since I'm the one that wrote that code haha)

nturinski: I think it's because originally when we wrote the function, we didn't account for having a `forcePrompt` property.  So at that time, we didn't want a cancel to throw an error and cancel the entire operation.  I guess the real fix would be to throw a `UserCancelledError` only in the forcePrompt scenario (which I will fix)

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+import { cpUtils } from "../../utils/cpUtils";
+import { funcHostProblemMatcher, funcHostTaskId } from "./IProjectCreator";
+import { ScriptProjectCreatorBase } from './ScriptProjectCreatorBase';
+
+export class PythonProjectCreator extends ScriptProjectCreatorBase {
+    public readonly templateFilter: TemplateFilter = TemplateFilter.Verified;
+    private readonly pythonVenvPathSetting: string = 'python.venvRelativePath';
+    private pythonAlias: string;
+    public getLaunchJson(): {} {
+        // currently the Python extension launches Windows with type: python and MacOS with type: pythonExperimental
```
EricJizbaMSFT: Can you file a bug to track this. I'm assuming we'll have to change this soon?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+                    port: 9091,
+                    host: 'localhost',
+                    preLaunchTask: funcHostTaskId
+                }
+            ]
+        };
+    }
+
+    public async addNonVSCodeFiles(): Promise<void> {
+        const funcCoreRequired: string = localize('funcCoreRequired', 'You must have the Azure Functions Core Tools installed to create, debug, and local Python functions.');
```
EricJizbaMSFT: > to create, debug, and local Python functions

Are you missing a word there?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+        if (!await validateFuncCoreToolsInstalled(true /* forcePrompt */, funcCoreRequired)) {
+            throw new UserCancelledError();
+        }
+        await this.validatePythonVersion();
+        await this.setVirtualEnviornment();
```
EricJizbaMSFT: At some point, can you make sure to test this scenario:
1. Create a python project
1. Push it to GitHub
1. Clone it to a fresh location on your local computer
1. Run project

I'm just curious to see if all the virtual environment and launch/tasks stuff works from a fresh clone. Also try creating a project on one OS and then cloning it to a different OS. I don't think you necessarily need to do all of this for this PR, but it definitely needs to be done before we release.

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    private async getVenvActivatePath(platform: Platform): Promise<string> {
+        const venvPath: string | undefined = getFuncExtensionSetting(this.pythonVenvPathSetting, this.functionAppPath);
+        if (venvPath) {
+            switch (platform) {
+                case Platform.Windows:
+                    return path.join('.', venvPath, 'Scripts', 'activate');
+                case Platform.MacOS:
+                default:
+                    // assuming OSX and Linux use the same path -- need to some testing
```
EricJizbaMSFT: Don't leave this comment in the code. Either test it before merging or file an issue on GitHub

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+        return {
+            version: '2.0.0',
+            tasks: [
+                {
+                    label: localize('azFunc.runFuncHost', 'Run Functions Host'),
+                    identifier: funcHostTaskId,
+                    type: 'shell',
+                    dependsOn: 'build',
+                    osx: {
+                        command: `source ${await this.getVenvActivatePath(Platform.MacOS)} && func start host`
```
EricJizbaMSFT: See the variable substitution section here: https://code.visualstudio.com/docs/editor/tasks#_variable-substitution

I believe you can do `${config:azureFunctions.python.venvRelativePath}` rather than hard-coding the path

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,164 @@
...
+            // if there is no venv path, prompt user and recursively call function again
+            await this.setVirtualEnviornment();
+            return await this.getVenvActivatePath(platform);
+        }
+    }
+
+    private async createPythonProject(): Promise<void> {
+        switch (process.platform) {
+            case Platform.Windows:
+                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `${await this.getVenvActivatePath(Platform.Windows)} && func init ./ --worker-runtime python`);
```
EricJizbaMSFT: nit: can you make a constant for `func init ./ --worker-runtime python` since it's used twice

src/constants.ts
```diff
@@ -65,3 +65,9 @@ export enum ScmType {
     LocalGit = 'LocalGit',
     GitHub = 'GitHub'
 }
+
+export enum PythonAlias {
```
EricJizbaMSFT: nit: This is only used in the PythonProjectCreator file, right? Seems like that's a better place to put this

test/createNewProject.test.ts
```diff
@@ -116,6 +117,7 @@ suite('Create New Project Tests', async function (this: ISuiteCallbackContext):
     const pythonProject: string = 'PythonProject';
     test(pythonProject, async () => {
         const projectPath: string = path.join(testFolderPath, pythonProject);
+        await installFuncCoreTools(PackageManager.npm);
```
EricJizbaMSFT: I would actually suggest you comment this test out (and undo your changes to travis.yml) and file an issue to fix separately

The python experience is new and might be changing in the next week or two anyways. Sometimes it's easier to let it 'bake' before making tests work

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: So this is kind of tricky because the path changes based on the OS.  The `azureFunctions.python.venvRelativePath` setting is only for locating the root of the virtual environment.  The `getVenvActivatePath` then uses the root to output the path of the activate script based on which OS is passed in.

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -30,7 +30,7 @@ export async function validateFuncCoreToolsInstalled(forcePrompt: boolean = fals
                 installed = true;
             } else {
                 const items: MessageItem[] = [];
-                const message: string = localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to debug your local functions.');
+                const message: string = customMessage ? customMessage : localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to debug your local functions.');
```
nturinski: @EricJizbaMSFT Are you okay with this solution??

package.json
```diff
@@ -591,6 +591,10 @@
                             "OpenInCurrentWindow"
                         ],
                         "description": "%azFunc.projectOpenBehaviorDescription%"
+                    },
+                    "azureFunctions.python.venvRelativePath": {
```
EricJizbaMSFT: Can we change this to `venvName` since that's what it is now?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,170 @@
...
+        const newButton: MessageItem = { title: 'Create New' };
+        const existingButton: MessageItem = { title: 'Use Existing' };
+        const venvRequired: string = localize('venvRequired', 'You must be running in a Python virtual environment to create a Python Function project. Create a new one or use an existing?');
+        const input: MessageItem = await ext.ui.showWarningMessage(venvRequired, { modal: true }, newButton, existingButton);
+        if (input === newButton) {
+            // if there is no func_env, create one as it's required for Python functions
+            await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, this.pythonAlias, '-m', 'venv', 'func_env');
+            pythonVenvPath = funcEnv;
+        } else {
+            const options: OpenDialogOptions = {
```
EricJizbaMSFT: Should this still be a browse dialog or just an input box where they enter the name of the venv?

If we had an input box, I feel like it might say this:
> Enter the name of the Python virtual environment required for this project. It will be created if it does not already exist.

You might not even need the warning above if you do this. And obviously the input box should default to a value of 'func_env'

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    private async createPythonProject(): Promise<void> {
+        const funcInitPython: string = 'func init ./ --worker-runtime python';
+        switch (process.platform) {
+            case Platform.Windows:
+                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `${await this.getVenvActivatePath(Platform.Windows)} && ${funcInitPython}`);
+                break;
+            case Platform.MacOS:
+            default:
+                // assuming OSX and Linux use the same path -- need to some testing
```
EricJizbaMSFT: nit: Remove this comment

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: Well, I'm not sure how Python users use virtual environments, but the folder that contains the venv doesn't need to be within the same root folder of the project.

If we give someone the option to use their own venv, I'd imagine it wouldn't be within their Python project (since they are creating a new project at this moment).

I don't know how common it would be, but I guess someone could link it to an existing venv that is already set up to use Python Functions.

package.json
nturinski: Hmm, it could still be a path.  For example, if I link it to a venv outside of my project, the setting will look like:

`..\\pythonProject\\func_env`

Maybe we should talk about whether or not we should support existing venvs.

src/commands/createNewProject/PythonProjectCreator.ts
EricJizbaMSFT: Okay well I think the existing 'browse' option is confusing (it's not clear if I'm supposed to select the folder containing the 'activate' command or the root folder of the venv). I'm also not sure what the use case is for an existing venv (especially one outside your repo). I would rather simplify our user experience than cater to an edge-case that we don't understand. Plus those edge-case users have a workaround (manually edit their tasks.json)

What if we remove all venvRelativePath stuff from this PR and create a separate issue to debate/discuss/address the "existing venv" scenario you're talking about?

EricJizbaMSFT: (But obviously keep a branch with your existing venvRelativePath stuff just in case we end up going with it)

nturinski: Mmkay.  I'm okay with that.

nturinski:
Ah yes, thanks for the reminder.

It's been a long journey...

Issue #517, opened 2018-08-17
Find out a way to better guide users if maven is not in path 


jdneo:
We got feedback that a lot of users get errors because that Maven is not in their environment path. Need to find a way for the users to better troubleshooting.

jdneo:
Addressed in #530

PR #530, opened 2018-08-28
Add guide for users to set maven path.

Flanker32:
Add guide for users to set maven path, for issue #517


src/utils/mavenUtils.ts
```diff
@@ -9,13 +9,22 @@ import { TelemetryProperties } from "vscode-azureextensionui";
 import * as xml2js from 'xml2js';
 import { localize } from '../localize';
 import { cpUtils } from './cpUtils';
+import { ext } from '../extensionVariables';
+import { DialogResponses } from 'vscode-azureextensionui';
+
```
jdneo: Remove this empty line

src/utils/mavenUtils.ts
```diff
@@ -9,13 +9,22 @@ import { TelemetryProperties } from "vscode-azureextensionui";
...
+import opn = require("opn");
 
 export namespace mavenUtils {
     const mvnCommand: string = 'mvn';
     export async function validateMavenInstalled(workingDirectory: string | undefined): Promise<void> {
         try {
             await cpUtils.executeCommand(undefined, workingDirectory, mvnCommand, '--version');
         } catch (error) {
+            const message: string = localize('azFunc.mvnNotFound', 'Failed to find "maven", please set the maven path');
+            const result: vscode.MessageItem = await ext.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
```
jdneo: We need to make some action when the user selects `DialogResponses.dontWarnAgain`

src/utils/mavenUtils.ts
```diff
@@ -9,13 +9,22 @@ import { TelemetryProperties } from "vscode-azureextensionui";
...
+            if (result === DialogResponses.learnMore) {
+                await opn('https://aka.ms/azurefunction_maven');
```
jdneo: I suppose you will create a short-url later, right?

docs/azurefunction_maven.md
```diff
@@ -0,0 +1,25 @@
+# Config Maven for VS Code
```
jdneo: Nice document. I think we can also reference it in the README: https://github.com/Microsoft/vscode-azurefunctions#java

src/utils/mavenUtils.ts
Flanker32: I plan to throw an Error no matter user selects learnMore or dontWarnAgain, in order to interrupt the execution of function which invoke validateMavenInstalled, so i didn't add another if but just throw an Error behind it.

package.nls.json
```diff
@@ -32,6 +32,7 @@
     "azFunc.deploySubpathDescription": "The default subpath of a workspace folder to use when deploying.",
     "azFunc.showCoreToolsWarningDescription": "Show a warning if your installed version of Azure Functions Core Tools is out-of-date.",
     "azFunc.show64BitWarningDescription": "Show a warning to install a 64-bit version of the Azure Functions Core Tools when you create a .NET Framework project.",
+    "azFunc.showMavenPathWarningDescription": "Show a warning to set maven path when you create java functions",
```
jdneo: `Show a warning when Maven path is not set correctly.`

Flanker32:
test this please

docs/azurefunction_maven.md
```diff
@@ -0,0 +1,25 @@
+# Configure Maven for VS Code
```
jdneo: Hmm, maybe the name here still needs refinement. I think it could be better if we change it to `Configure Maven for Java Azure Functions Development`

src/utils/mavenUtils.ts
```diff
@@ -16,6 +20,16 @@ export namespace mavenUtils {
         try {
             await cpUtils.executeCommand(undefined, workingDirectory, mvnCommand, '--version');
         } catch (error) {
+            const settingKey: string = 'showMavenPathWarning';
+            if (getFuncExtensionSetting<boolean>(settingKey)) {
+                const message: string = localize('azFunc.mvnNotFound', 'Failed to find "maven", please ensure that the maven bin directory is in your system path.');
+                const result: vscode.MessageItem = await ext.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
+                if (result === DialogResponses.learnMore) {
+                    await opn('https://aka.ms/azurefunction_maven');
```
jdneo: Don't forget to update the short url after PR merged. Currently it is pointing to your own repo.

jdneo:
@Flanker32 The PR looks good to me, except for some wording concerns.

@EricJizbaMSFT please also take a review when you have time, thanks! 😄

docs/azurefunction_maven.md
```diff
@@ -0,0 +1,25 @@
...
+
+Maven is a software project management and comprehension tool, which could manage a project's build, reporting and documentation from a central piece of information. The VS Code Azure Function Plugin use maven to generate Java Function project. Please ensure you have already download maven and set the PATH correctlly before you create Java Functions.
```
nturinski: Could you make the following changes?

1. VS Code Azure Functions Extension (versus Plugin)
1. download => downloaded
1 . correctlly => correctly

Flanker32: Done~

docs/azurefunction_maven.md
```diff
@@ -0,0 +1,25 @@
...
+    * Unix-based operating systems (Linux, Solaris and Mac OS X)
+
+       `export PATH=/usr/local/apache-maven-3.x.y/bin:$PATH`
+    * Windows
+
+      `set PATH="c:\program files\apache-maven-3.x.y\bin";%PATH%`
+
+4. Run `mvn --version` to verify that it is correctly installed.
+
+> NOTE: Please restart vscode to make the path effective when you finished the steps above.
```
SummerSun: finish, not finished

src/utils/mavenUtils.ts
```diff
@@ -16,6 +20,16 @@ export namespace mavenUtils {
         try {
             await cpUtils.executeCommand(undefined, workingDirectory, mvnCommand, '--version');
         } catch (error) {
+            const settingKey: string = 'showMavenPathWarning';
+            if (getFuncExtensionSetting<boolean>(settingKey)) {
+                const message: string = localize('azFunc.mvnNotFound', 'Failed to find "maven", please ensure that the maven bin directory is in your system path.');
```
EricJizbaMSFT: I don't think a "Don't warn again" button is appropriate in this case:
1. This message is very similar to the error you throw on line 33, but better. I would rather have this show every time than that error
1. This is really an error, not a warning. We don't hide this kind of error from the user (aka an error that blocks them from continuing)

jdneo: That's a valid concern. Also I've noticed that we will have two toasts here:
- One is for trouble shotting
- Another is the original one `Failed to find "maven", please ensure ...`

@EricJizbaMSFT Is it possible that we can only have one toast here?

EricJizbaMSFT: Well one option is not to throw an error. But obviously then you don't get it in telemetry, etc. 

That's why we have the `actionContext.suppressErrorDisplay` property. Here's an example: https://github.com/Microsoft/vscode-azurefunctions/blob/b56a2b0ccddac99993891204250fc17f6898ad16/src/commands/createNewProject/validateFunctionProjects.ts#L33

src/commands/createFunction/JavaFunctionCreator.ts
```diff
@@ -24,11 +24,13 @@ export class JavaFunctionCreator extends FunctionCreatorBase {
...
     private _functionName: string;
     private _telemetryProperties: TelemetryProperties;
+    private _actionContext: IActionContext;
 
-    constructor(functionAppPath: string, template: IFunctionTemplate, outputChannel: OutputChannel, telemetryProperties: TelemetryProperties) {
+    constructor(functionAppPath: string, template: IFunctionTemplate, outputChannel: OutputChannel, actionContext: IActionContext) {
         super(functionAppPath, template);
         this._outputChannel = outputChannel;
-        this._telemetryProperties = telemetryProperties;
+        this._telemetryProperties = actionContext.properties;
```
jdneo: No need to cache this since we have `IActionContext` now

src/utils/mavenUtils.ts
Flanker32: @EricJizbaMSFT Thanks for your advice, and please have a review.

Issue #516, opened 2018-08-16
Some of the annotation parameter value will be empty when add a new Java function


jdneo:
For example, when the user wants to add a new EventHubTrigger function, he will be asked to fill the value of `consumerGroup`. In some cases, the user may enter `$Default`.

Then it will be parsed into a Maven command in the following command:
```mvn azure-functions:add -B "-Dfunctions.package=com.function" "-Dfunctions.name=EventHubTriggerJava3" "-Dfunctions.template=EventHubTrigger" "-Dconnection=AzureWebJobsStorage" "-DeventHubName=samples-workitems" "-DconsumerGroup=\$Default"```

In Windows it will be fine. But in Linux / Mac, the shell will try to interpret dollar signs, which may result an empty string.

A solution could be we check the system platform to determine which one to use between `'` and `"`.
- Windows -> `"`
- Others -> `'`

PR #533, opened 2018-08-29
Use different quotes for issue516

Flanker32:
Use different quotation mark in different system for issue #516

### DIFF
--- a/src/utils/mavenUtils.ts
+++ b/src/utils/mavenUtils.ts
@@ -9,9 +9,11 @@ import { TelemetryProperties } from "vscode-azureextensionui";
 import * as xml2js from 'xml2js';
 import { localize } from '../localize';
 import { cpUtils } from './cpUtils';
+import { Platform } from '../constants';
 
 export namespace mavenUtils {
     const mvnCommand: string = 'mvn';
+    const quotationMark: string = process.platform == Platform.Windows ? "\"" : "'";
     export async function validateMavenInstalled(workingDirectory: string | undefined): Promise<void> {
         try {
             await cpUtils.executeCommand(undefined, workingDirectory, mvnCommand, '--version');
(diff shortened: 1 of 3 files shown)


src/utils/mavenUtils.ts
```diff
@@ -64,4 +66,11 @@ export namespace mavenUtils {
         }
         return result.cmdOutput;
     }
+
+    function replaceParameterQuotationMark(input: string): string {
+        if (input.startsWith('\'') || input.startsWith('"')) {
```
jdneo: What if the input does not end with `'`/`"`?

src/utils/mavenUtils.ts
```diff
@@ -64,4 +66,11 @@ export namespace mavenUtils {
         }
         return result.cmdOutput;
     }
+
+    function replaceParameterQuotationMark(input: string): string {
```
EricJizbaMSFT: I would rather create a function that formats the arg _before_ you pass it in rather than try to change it afterwards. If you format it correctly from the start, you don't have to worry about all the weird edge cases. @jdneo mentioned one edge case, but here's another: JavaProjectCreator uses the quotes like this `-DgroupId="${groupId}"` instead of this `"-DgroupId=${groupId}"`

I suggest you create a function like this:
```typescript
function formatMavenArg(key: string, value: string): string {
    return `-${key}=${quotationMark}${value}${quotationMark}`
}
```

which you leverage in [JavaProjectCreator](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createNewProject/JavaProjectCreator.ts) and [JavaFunctionCreator](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createFunction/JavaFunctionCreator.ts)

PR #539, opened 2018-08-30
Fix maven arg being passed incorrectly

EricJizbaMSFT:
@Flanker32 @jdneo FYI - I didn't notice this until after I had merged this PR https://github.com/Microsoft/vscode-azurefunctions/pull/533


jdneo:
Thank you @EricJizbaMSFT 

@Flanker32 Let's involve the test team to verify if we have any more regressions.

Flanker32:
@EricJizbaMSFT  Thanks a lot. 
@jdneo  Got it, i'll make the plan later.

PR #549, opened 2018-08-31
Run 'func extensions install' before debug/deploy

### LINKED ISSUE #382: Func extensions install
Functions now has capabilities that let the user run `func extension install` at the base of a function project. This will inspect the function.json files and install the relevant extensions. Would be a good to have an experience for the same in VS Code. Perhaps, an option in the **Command Palette** ? @fiveisprime 

Note: This applies to all languages other than C# class library (.net)

EricJizbaMSFT:
Users need to run `func extensions install` before debugging or deploying on JavaScript projects. This was always the case for some triggers (that we didn't "officially" support in VS Code), but it wasn't required on blob/queue triggers until the most recent functions release. Hence why we're adding this now.

As part of this change, I made the C# 'preDeploy' logic more generic so that it would also cover the JavaScript `func extensions install` case. I also changed the C# logic so that it won't run for local git deployments, since that should be covered by kudu.

Fixes #382

### DIFF
--- a/src/commands/createNewProject/JavaScriptProjectCreator.ts
+++ b/src/commands/createNewProject/JavaScriptProjectCreator.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ProjectRuntime, TemplateFilter } from "../../constants";
+import { installExtensionsId, ProjectRuntime, TemplateFilter } from "../../constants";
 import { localize } from "../../localize";
 import { funcHostProblemMatcher, funcHostTaskId, funcHostTaskLabel } from "./IProjectCreator";
 import { ITaskOptions } from "./ITasksJson";
@@ -16,6 +16,7 @@ export class JavaScriptProjectCreator extends ScriptProjectCreatorBase {
     public readonly templateFilter: TemplateFilter = TemplateFilter.Verified;
 
     public readonly functionsWorkerRuntime: string | undefined = 'node';
+    public preDeployTask: string = installExtensionsId;
 
     public getLaunchJson(): {} {
         return {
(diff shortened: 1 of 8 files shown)


src/commands/createNewProject/JavaScriptProjectCreator.ts
```diff
@@ -56,7 +57,17 @@ export class JavaScriptProjectCreator extends ScriptProjectCreatorBase {
                     },
                     problemMatcher: [
                         funcHostProblemMatcher
-                    ]
+                    ],
+                    dependsOn: installExtensionsId
```
StephenWeatherford: Do you need to consider if its empty?

src/commands/deploy.ts
```diff
@@ -214,39 +212,82 @@ async function verifyRuntimeIsCompatible(localRuntime: ProjectRuntime, ui: IAzur
...
+
+        if (taskName) {
+            // We only need to warn if they have the setting defined
+            ext.outputChannel.appendLine(localize('ignoringPreDeployTask', 'WARNING: Ignoring preDeployTask "{0}" for non-zip deploy.', taskName));
+        }
+
+        return;
+    }
+
+    let isTaskNameDefined: boolean = true;
```
StephenWeatherford: why not let isTaskNameDefined = !!taskName;
?

src/commands/deploy.ts
```diff
@@ -214,39 +212,82 @@ async function verifyRuntimeIsCompatible(localRuntime: ProjectRuntime, ui: IAzur
...
+        isTaskNameDefined = false;
+        switch (language) {
+            case ProjectLanguage.CSharp:
+                taskName = publishTaskId;
+                break;
+            case ProjectLanguage.JavaScript:
+                taskName = installExtensionsId;
+                break;
+            default:
+                return; // preDeployTask not needed
```
StephenWeatherford: Need telemetry for this?

src/commands/deploy.ts
EricJizbaMSFT: Honestly I don't know what to call this variable. It's really "is task name defined is VS Code settings" or "is task name originally defined"

I can't do `!!taskName` because it will always be defined by the time I end up using `isTaskNameDefined`

StephenWeatherford: How about isTaskNameDefinedInSettings?

Not that it matters, but that doesn't make sense to me.  Your logic sets isTaskNameDefined=false if (!taskName).  That's equivalent to isTaskNameDefined = !!taskName, meaning if taskName is not undefined and not empty string.

EricJizbaMSFT: Ohhh yes I could do that. I thought you were saying get rid of the `isTaskNameDefined` variable and just do `!!taskName` later on. I get you now

src/commands/deploy.ts
EricJizbaMSFT: Don't think so. Also it can be inferred from our existing telemetry (we already have the language and we're tracking the 'preDeployTask' if it's defined in settings).

PR #551, opened 2018-08-31
Nat/create with rg

### LINKED ISSUE #471: in API, support the input of resource group in createFunctionApp()
Currently, the API of create Function App is as follow:

const functionAppId: string = await vscode.commands.executeCommand<string>('azureFunctions.createFunctionApp', '00000000-0000-0000-0000-000000000000');

Could you provide one more input parameter of resource group name?

nturinski:
Fixes #471 
Allow resourceGroup to be a programmatic property for Function apps

### DIFF
--- a/src/commands/createFunctionApp.ts
+++ b/src/commands/createFunctionApp.ts
@@ -16,6 +16,6 @@ export async function createFunctionApp(actionContext: IActionContext, tree: Azu
         node = subscription;
     }
 
-    const funcAppNode: IAzureNode = await node.createChild(actionContext, resourceGroup);
+    const funcAppNode: IAzureNode = await node.createChild({ actionContext, resourceGroup });
     return funcAppNode.id;
 }
--- a/src/commands/createFunctionApp.ts
+++ b/src/commands/createFunctionApp.ts
@@ -6,14 +6,14 @@
 import { AzureTreeDataProvider, IActionContext, IAzureNode, IAzureParentNode } from 'vscode-azureextensionui';
 import { nodeUtils } from '../utils/nodeUtils';
 
-export async function createFunctionApp(actionContext: IActionContext, tree: AzureTreeDataProvider, arg?: IAzureParentNode | string): Promise<string> {
+export async function createFunctionApp(actionContext: IActionContext, tree: AzureTreeDataProvider, subscription?: IAzureParentNode | string, resourecGroup?: IAzureParentNode | string): Promise<string> {
     let node: IAzureParentNode;
-    if (typeof arg === 'string') {
-        node = await nodeUtils.getSubscriptionNode(tree, arg);
-    } else if (!arg) {
+    if (typeof subscription === 'string') {
+        node = await nodeUtils.getSubscriptionNode(tree, subscription);
+    } else if (!subscription) {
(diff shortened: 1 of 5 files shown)


src/commands/createFunctionApp.ts
```diff
@@ -4,16 +4,23 @@
...
         node = <IAzureParentNode>await tree.showNodePicker(AzureTreeDataProvider.subscriptionContextValue);
     } else {
-        node = arg;
+        node = subscription;
+    }
+    if (resourceGroup) {
+        if (typeof resourceGroup !== 'string') {
+            resourceGroup = resourceGroup.treeItem.client.resourceGroup;
+        }
+        actionContext.properties.resourceGroup = resourceGroup;
```
EricJizbaMSFT: `actionContext.properties` is used for telemetry and we don't want to collect this in telemetry

We'll have to find a different way to pass the resource group along

src/commands/createFunctionApp.ts
```diff
@@ -4,16 +4,23 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureTreeDataProvider, IActionContext, IAzureNode, IAzureParentNode } from 'vscode-azureextensionui';
+import { FunctionAppTreeItem } from '../tree/FunctionAppTreeItem';
 import { nodeUtils } from '../utils/nodeUtils';
 
-export async function createFunctionApp(actionContext: IActionContext, tree: AzureTreeDataProvider, arg?: IAzureParentNode | string): Promise<string> {
+export async function createFunctionApp(actionContext: IActionContext, tree: AzureTreeDataProvider, subscription?: IAzureParentNode | string, resourceGroup?: IAzureParentNode<FunctionAppTreeItem> | string): Promise<string> {
```
EricJizbaMSFT: When will resourceGroup ever be `IAzureParentNode<FunctionAppTreeItem>`? Shouldn't it just be string?

src/commands/createFunctionApp.ts
nturinski: There are UserSettings here:
https://github.com/Microsoft/vscode-azurefunctions/blob/56a097178f1deac96f69cac174be26f2c2899825/src/commands/createFunction/createFunction.ts#L150

But it seems like it's used for appSettings which get passed into the ActionContext.  Should we just create a new property?

src/commands/createFunctionApp.ts
nturinski: I wasn't sure how it was going to be used, but I thought a user could just pass another function app and use the same resource group as the one passed.

src/commands/createFunctionApp.ts
```diff
@@ -6,16 +6,16 @@
...
+        node = await nodeUtils.getSubscriptionNode(tree, subscription);
+    } else if (!subscription) {
         node = <IAzureParentNode>await tree.showNodePicker(AzureTreeDataProvider.subscriptionContextValue);
     } else {
-        node = arg;
+        node = subscription;
     }
 
-    const funcAppNode: IAzureNode = await node.createChild(actionContext);
+    const funcAppNode: IAzureNode = await node.createChild(actionContext, resourceGroup);
```
EricJizbaMSFT: createChild only supports one `userOptions` parameter as of right now. You can combine them into one object: `{ actionContext, resourceGroup }`

PR #569, opened 2018-09-06
Fix "func extensions install" on v1

EricJizbaMSFT:
We recently added support to run "func extensions install" before debug/deploy, but that only applies for v2. You get this error if you try to run it on v1:

screen shot 2018-09-06 at 2 55 36 pm

### DIFF
--- a/src/commands/createNewProject/JavaScriptProjectCreator.ts
+++ b/src/commands/createNewProject/JavaScriptProjectCreator.ts
@@ -26,7 +26,6 @@ export class JavaScriptProjectCreator extends ScriptProjectCreatorBase {
                     type: 'node',
                     request: 'attach',
                     port: 5858,
-                    protocol: 'inspector',
                     preLaunchTask: funcHostTaskId
                 }
             ]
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
Also removed `protocol: 'inspector',` per discussion with Matt. It doesn't seem to work for me on v1 with JS (although not sure why no one has reported the issue) and VS Code is smart enough to pick the protocol itself

src/commands/createNewProject/JavaScriptProjectCreator.ts
```diff
@@ -36,40 +34,50 @@ export class JavaScriptProjectCreator extends ScriptProjectCreatorBase {
 
     public getTasksJson(runtime: string): {} {
         let options: ITaskOptions | undefined;
+        // tslint:disable-next-line:no-any
+        const funcTask: any = {
+            label: funcHostTaskLabel,
+            identifier: funcHostTaskId,
```
nturinski: Are we going to get rid of `identifier`?

EricJizbaMSFT: Eventually, but it's not doing any harm so I'd rather just leave it for this release

PR #558, opened 2018-09-04
Refactor pickProcess to leverage tasks api

EricJizbaMSFT:
We've had a whole host of issues related to pickProcess for C# debugging. I was able to simplify this logic by leveraging VS Code's tasks api that was released a few months ago rather than searching the entire list of processes myself. The VS Code tasks api actually gives me the pid of the task, which is great.

Unfortunately it's never that simple, though 😅 On Windows, it gives me the pid of the parent PowerShell process, not the actual func process. In order to find the child process, I decided to leverage the [windows-process-tree](https://github.com/Microsoft/vscode-windows-process-tree) module that's created for and shipped with VS Code, which is a native windows module and avoids a lot of the problems we originally ran into with wmic (which most other modules like 'ps-node' use).

Fixes #436
Fixes #518 
Fixes #534 
Fixes #437

PR #573, opened 2018-09-07
Make pickFuncProcess more robust to timing issues

EricJizbaMSFT:
As described in this PR: https://github.com/Microsoft/vscode-azurefunctions/pull/558, we have to get the inner 'func' process on Windows instead of attaching to the parent PowerShell process. Well it turns out there's a delay between when the PowerShell process is created and when it has the child process running the func cli. This change prevents us from attaching to the PowerShell process and getting "unverified" breakpoints.

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -114,7 +114,7 @@ async function getInnermostWindowsPid(pid: string, timeoutInSeconds: number, tim
         if (psTree.name.toLowerCase().includes('func')) {
             return psTree.pid.toString();
         } else {
-            await delay(1000);
+            await delay(500);
         }
     }
 
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -27,10 +27,19 @@ export async function pickFuncProcess(actionContext: IActionContext): Promise<st
         throw new Error(localize('noFuncTask', 'Failed to find task with label "{0}".', funcHostTaskLabel));
     }
 
-    const pid: string = await startFuncTask(funcTask, actionContext);
+    const settingKey: string = 'pickProcessTimeout';
+    const settingValue: number | undefined = getFuncExtensionSetting<number>(settingKey);
+    const timeoutInSeconds: number = Number(settingValue);
+    if (isNaN(timeoutInSeconds)) {
+        throw new Error(localize('invalidSettingValue', 'The setting "{0}" must be a number, but instead found "{1}".', settingKey, settingValue));
+    }
+    actionContext.properties.timeoutInSeconds = timeoutInSeconds.toString();
+    const timeoutError: Error = new Error(localize('failedToFindFuncHost', 'Failed to detect running Functions host within "{0}" seconds. You may want to adjust the "{1}" setting.', timeoutInSeconds, `${extensionPrefix}.${settingKey}`));
+
+    const pid: string = await startFuncTask(funcTask, timeoutInSeconds, timeoutError);
     // On Mac/Linux we can leverage the pid of the task directly.
     // On Windows, the pid of the task corresponds to the parent PowerShell process and we have to drill down to get the actual func process
-    return isWindows ? await getInnermostWindowsPid(pid) : pid;
+    return isWindows ? await getInnermostWindowsPid(pid, timeoutInSeconds, timeoutError) : pid;
 }
 
 function isFuncTask(task: Task): boolean {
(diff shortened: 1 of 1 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -27,10 +27,19 @@ export async function pickFuncProcess(actionContext: IActionContext): Promise<st
...
+        throw new Error(localize('invalidSettingValue', 'The setting "{0}" must be a number, but instead found "{1}".', settingKey, settingValue));
+    }
+    actionContext.properties.timeoutInSeconds = timeoutInSeconds.toString();
+    const timeoutError: Error = new Error(localize('failedToFindFuncHost', 'Failed to detect running Functions host within "{0}" seconds. You may want to adjust the "{1}" setting.', timeoutInSeconds, `${extensionPrefix}.${settingKey}`));
+
+    const pid: string = await startFuncTask(funcTask, timeoutInSeconds, timeoutError);
     // On Mac/Linux we can leverage the pid of the task directly.
     // On Windows, the pid of the task corresponds to the parent PowerShell process and we have to drill down to get the actual func process
-    return isWindows ? await getInnermostWindowsPid(pid) : pid;
+    return isWindows ? await getInnermostWindowsPid(pid, timeoutInSeconds, timeoutError) : pid;
```
nturinski: nit: Is there any reason to do `await startFuncTask()` before returning?  Seems like you could just do:
`return isWindows ? await getInnermostWindowsPid(pid, timeoutInSeconds, timeoutError) : await startFuncTask(funcTask, timeoutInSeconds, timeoutError);`

src/commands/pickFuncProcess.ts
```diff
@@ -83,32 +84,41 @@ async function startFuncTask(funcTask: Task, actionContext: IActionContext): Pro
...
     if (!windowsProcessTree) {
         throw new Error(localize('noWindowsProcessTree', 'Failed to find dependency "{0}".', moduleName));
     }
 
-    let psTree: IProcessTreeNode | undefined = await new Promise<IProcessTreeNode | undefined>((resolve: (p: IProcessTreeNode | undefined) => void): void => {
-        windowsProcessTree.getProcessTree(Number(pid), resolve);
-    });
+    const maxTime: number = Date.now() + timeoutInSeconds * 1000;
+    while (Date.now() < maxTime) {
+        let psTree: IProcessTreeNode | undefined = await new Promise<IProcessTreeNode | undefined>((resolve: (p: IProcessTreeNode | undefined) => void): void => {
```
nturinski: Should this be part of the `while` loop?  Doesn't seem like we'd have to grab the powershell process again after the delay.

src/commands/pickFuncProcess.ts
```diff
@@ -122,3 +132,7 @@ interface IProcessTreeNode {
 interface IWindowsProcessTree {
     getProcessTree(rootPid: number, callback: (tree: IProcessTreeNode | undefined) => void): void;
 }
+
+async function delay(ms: number): Promise<void> {
```
nturinski: Delay is a pretty generic function.  Should we move it to somewhere more common?

src/commands/pickFuncProcess.ts
EricJizbaMSFT: startFuncTask does a few things, including _starting the task_ (as the name describes) and returning the pid. We need that original pid to find the _inner_ pid

src/commands/pickFuncProcess.ts
EricJizbaMSFT: It's not grabbing the powershell process - it's grabbing the whole tree of parent/child processes

nturinski: Well, either way, should we be grabbing the whole tree over and over again per while loop?

src/commands/pickFuncProcess.ts
EricJizbaMSFT: Yeah it's pretty generic (I mostly copied it from a util function Stephen had).

I can go either way on this one. On one hand, I don't like how the 'utils' folder is just a pile of random functions, so I try not to put stuff in there until we're actually using it multiple times. On the other hand, I feel like that could lead to more duplicated work because the next person is less likely to find this function since it's not in utils. Thoughts? In any case, I think I'll just leave as is for this PR.

nturinski: Always could wait until someone accidentally duplicates it and move it then.  I agree it's less discoverable, but I'm sure one of us will remember come PR time... 😅

src/commands/pickFuncProcess.ts
EricJizbaMSFT: I'm not sure how else I would do it? Plus 'windows-process-tree' is a native windows module and is very fast

nturinski: Couldn't you just grab it prior to entering the `while (Date.now() < maxTime)` loop (along with the error checking)?  I mean, I guess if it's fast, it doesn't really matter, just didn't seem necessary.

EricJizbaMSFT: Oh I see what you mean - and the answer is no I can't. The whole point of this PR is that the child processes haven't started yet and don't exist in the tree yet. The tree is different each time.

src/commands/pickFuncProcess.ts
StephenWeatherford: I wouldn't mind seeing it in ui.  Maybe I'll add it when I'm doing other stuff there.

PR #574, opened 2018-09-08
Nturinski/run from zip

nturinski:
Not production ready, but wanted to get the PR opened for feedback and eyes.

Known issues: 
`func pack` and probably any `func` command won't work with fsPaths that contain spaces.  Will have to wrap them in quotes.


src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,198 @@
...
+                resolve();
+            }
+        });
+    });
+    const sasToken: string = blobService.generateSharedAccessSignature(containerName, blobName, <azureStorage.common.SharedAccessPolicy>{
+        AccessPolicy: {
+            Permissions: azureStorage.BlobUtilities.SharedAccessPermissions.READ + azureStorage.BlobUtilities.SharedAccessPermissions.LIST,
+            Start: azureStorage.date.secondsFromNow(-10),
+            // for clock desync
+            Expiry: azureStorage.date.daysFromNow(365),
```
EricJizbaMSFT: I remember having a discussion about this time. Is this the same thing Kudu does for Windows Run From Package?

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,198 @@
...
+        }
+    } catch (error) {
+        throw error;
+    }
+    return { zipFilePath: zipFilePath, createdZip: createdZip };
+}
+
+function randomFileName(): string {
+    // tslint:disable-next-line:insecure-random
+    return Math.random().toString(36).replace(/[^a-z]+/g, '').substr(0, 10);
```
EricJizbaMSFT: Don't use this. Use fsUtils.getRandomHexString

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,198 @@
...
+            // for clock desync
+            Expiry: azureStorage.date.daysFromNow(365),
+            ResourceTypes: azureStorage.BlobUtilities.BlobContainerPublicAccessType.BLOB
+        }
+    });
+
+    return blobService.getUrl(containerName, blobName, sasToken, true);
+}
+
+async function zipDirectory(fsPath: string, zipFileName?: string, globPattern: string = '**/*', ignorePattern?: string | string[]): Promise<{ zipFilePath: string, createdZip: boolean }> {
```
EricJizbaMSFT: You copied this from the shared package, right? That makes me think `runFromPackageDeploy` would fit better in the shared package

src/commands/deploy.ts
```diff
@@ -238,6 +249,8 @@ async function runPreDeployTask(deployFsPath: string, telemetryProperties: Telem
                     taskName = installExtensionsId;
                 }
                 break;
+            case ProjectLanguage.Python:
+                taskName = funcPack;
```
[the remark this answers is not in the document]
nturinski: Classic switch case bonanza...

src/commands/runFromPackageDeploy.ts
nturinski: I think that we discussed this back when I was implementing this.  If I recall correctly, we didn't want it in the shared package because no other extension would need `runFromPackageDeploy`.  Should we expose the `zipDirectory` function instead?

EricJizbaMSFT: I guess you said `func pack` zips for us, so I guess this comment thread no longer applies

nturinski: Well, we only run it for Python so we actually still need it for node.

EricJizbaMSFT: I guess I don't care too much. It seems wrong to expose `zipDirectory` from the "app service" package, so I'd rather move the deploy code over to the shared package or leave the copied zip code here as you have it.

Also definitely focus on Python. I don't care too much if JavaScript doesn't work.

package.nls.json
```diff
@@ -40,6 +40,7 @@
     "azFunc.deleteProxy": "Delete Proxy...",
     "azFunc.pickProcessTimeoutDescription": "The timeout (in seconds) to be used when searching for the Azure Functions host process. Since a build is required every time you F5, you may need to adjust this based on how long your build takes.",
     "azFunc.showFuncInstallationDescription": "Show a warning to install Azure Functions Core Tools CLI when you create a new project if the CLI is not installed.",
+    "azFunc.deployStorageAccountDescription": "Default Azure Storage Account to host package files deployed to Linux Consumption Function Apps.",
```
EricJizbaMSFT: This is the id, correct? Change to "Default Azure Storage Account **_id_**"

src/commands/deploy.ts
```diff
@@ -133,17 +143,21 @@ export async function deploy(ui: IAzureUserInput, actionContext: IActionContext,
     const children: IAzureNode[] = await node.getCachedChildren();
     const functionsNode: IAzureParentNode<FunctionsTreeItem> = <IAzureParentNode<FunctionsTreeItem>>children.find((n: IAzureNode) => n.treeItem instanceof FunctionsTreeItem);
     await node.treeDataProvider.refresh(functionsNode);
+    if (isLinuxConsumptionPlan) {
+        // We can't get the function triggers due to no Kudu, so end here
+        return;
```
EricJizbaMSFT: Remove this check for now. I'm hoping we'll be able to list functions

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,212 @@
...
+import * as fsUtil from '../utils/fs';
+
+/**
+ * Method of deployment that is only intended to be used for Linux Consumption Function apps.
+ * To deploy with Run from Package on a Windows plan, create the app setting "WEBSITE_RUN_FROM_ZIP" and set it to "1".
+ * Then deploy via "zipdeploy" as usual.
+ */
+
+export async function runFromPackageDeploy(actionContext: IActionContext, node: IAzureParentNode<FunctionAppTreeItem>, fsPath: string): Promise<void> {
+    const language: ProjectLanguage = await getProjectLanguage(fsPath, ext.ui);
```
EricJizbaMSFT: no need to get language again. Just pass it in as a parameter to runFromPackageDeploy

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,212 @@
...
+            const appSettings: StringDictionary = await client.listApplicationSettings();
+            if (appSettings.properties) {
+                const WEBSITE_RUN_FROM_ZIP: string = 'WEBSITE_RUN_FROM_ZIP';
+                appSettings.properties[WEBSITE_RUN_FROM_ZIP] = blobUrl;
+            } else {
+                throw new ArgumentError(appSettings);
+            }
+            await client.updateApplicationSettings(appSettings);
+        } catch (error) {
+            // tslint:disable-next-line:no-unsafe-any
```
EricJizbaMSFT: Remove this catch block

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,212 @@
...
+            } else {
+                throw error;
+            }
+        } finally {
+            // clean up zip file
+            if (createdZip) {
+                const cleaningZip: string = localize('cleaningZip', 'Cleaning zip package...');
+                p.report({ message: cleaningZip });
+                ext.outputChannel.appendLine(cleaningZip);
+                await new Promise((resolve: () => void, reject: (err: Error) => void): void => {
```
EricJizbaMSFT: Use fse.remove

src/commands/runFromPackageDeploy.ts
```diff
@@ -0,0 +1,212 @@
...
+    if (!result.keys || result.keys.length === 0) {
+        throw new ArgumentError(result);
+    } else {
+        key = result.keys[0].value;
+    }
+
+    if (name !== undefined && key !== undefined) {
+        return azureStorage.createBlobService(name, key);
+    } else {
+        throw new ArgumentError({ name: name, key: key });
```
EricJizbaMSFT: nit: just do `throw new ArgumentError(result)`

PR #578, opened 2018-09-12
Contribute func-watch problemMatcher

EricJizbaMSFT:
I'm basically shifting the implementation of the problem matcher from each user's project to our extension. This means we can change/improve it in the future without having to modify the user's project.

Unfortunately, the func cli is very annoying to parse, so I only added support for one error (although it's somewhat common). Now users will see this in the problems view:
screen shot 2018-09-12 at 8 38 30 am

### DIFF
--- a/package.json
+++ b/package.json
@@ -477,6 +477,35 @@
                 "url": "http://json.schemastore.org/proxies"
             }
         ],
+        "problemPatterns": [
+            {
+                "name": "func",
+                "kind": "file",
+                "regexp": "^.*Missing.*AzureWebJobsStorage.*(local.settings.json).*$",
+                "file": 1,
+                "message": 0
+            }
+        ],
+        "problemMatchers": [
+            {
+                "name": "func-watch",
+                "label": "%azFunc.problemMatchers.funcWatch.label%",
+                "owner": "Azure Functions",
+                "source": "func",
+                "applyTo": "allDocuments",
+                "fileLocation": [
+                    "relative",
+                    "${workspaceFolder}"
+                ],
+                "pattern": "$func",
+                "background": {
+                    "activeOnStart": true,
+                    "beginsPattern": "^.*(Job host stopped|signaling restart).*$",
+                    "endsPattern": "^.*(?:Application|Job host) started.*$"
+                },
+                "severity": "error"
+            }
+        ],
         "configuration": [
             {
                 "title": "Azure Functions",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -477,6 +477,35 @@
                 "url": "http://json.schemastore.org/proxies"
             }
         ],
+        "problemPatterns": [
+            {
+                "name": "func",
+                "kind": "file",
+                "regexp": "^.*Missing.*AzureWebJobsStorage.*(local.settings.json).*$",
```
StephenWeatherford: Consider more general, e.g.:

Missing value for \w+ in ([\w\.]+\w)

package.json
```diff
@@ -477,6 +477,35 @@
                 "url": "http://json.schemastore.org/proxies"
             }
         ],
+        "problemPatterns": [
```
StephenWeatherford: What happens with current projects that have a conflicting problem matcher defined?

package.json
EricJizbaMSFT: I don't think that regexp applies though. Like I mentioned, the func cli has REALLY inconsistent formatting for errors. I don't want to accidentally pick up things that aren't errors (or that are warnings instead of errors)

I picked one of the most common errors and made it work for that, but we'll have to file an issue on the func cli if we want to actually have a generic regexp

package.json
EricJizbaMSFT: Here's what an old project looks like:
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Run Functions Host",
      "type": "shell",
      "command": "func host start",
      "isBackground": true,
      "problemMatcher": [
        {
          "owner": "azureFunctions",
          "pattern": [
            {
              "regexp": "\\b\\B",
              "file": 1,
              "location": 2,
              "message": 3
            }
          ],
          "background": {
            "activeOnStart": true,
            "beginsPattern": "^.*Stopping host.*",
            "endsPattern": "^.*Job host started.*"
          }
        }
      ]
    }
  ]
}
```

Here's what a new project looks like:
```json
{
  "version": "2.0.0",
  "tasks": [
    {
      "label": "Run Functions Host",
      "type": "shell",
      "command": "func host start",
      "isBackground": true,
      "problemMatcher": "$func-watch"
    }
  ]
}
```
Old projects will continue to use that "hard-coded" problem matcher as if nothing had changed. Only new projects will benefit from "func-watch"

Issue #562, opened 2018-09-05
Change runtime version from 'beta' to '~2'


asavaritayal:
For v2 runtime GA, change `beta` to `~2` in the runtime selection dropdown. 

/cc @fiveisprime @jeffhollan

StephenWeatherford:
@EricJizbaMSFT

EricJizbaMSFT:
Thanks for the ping @StephenWeatherford - but just FYI to @asavaritayal: We've discussed this process before and the most important thing is that you create an issue _in our repo_. Thank you for doing that in this case :)

I can think of too many times when people missed an email or pinged the wrong person, but I can't think of any instance where an issues filed on GitHub (in the correct repo) was lost. We regularly triage open issues and I much prefer to rely on that process. (Plus I'm subscribed to the repo and get the notification anyways)

EricJizbaMSFT:
@asavaritayal we check the user's remote runtime version when deploying and I vaguely remember we would get "latest" returned sometimes. Is that still a valid version setting? If so - what will it mean going forward? ~1 or ~2?

jeffhollan:
@paulbatum 

I believe `latest` is still a valid setting though one that will go through a bit of a shift.  `latest` today is synonymous with ~1, but once ~2 becomes GA `latest` will then be ~2.  We actually need to alert users of this setting and generally don't recommend any version numbers that aren't more specific.

`beta` is another value that will be very common.  It today is same as ~2.  I suppose in future `beta` may be a version slightly higher than ~2 but should be in the same family from a template standpoint.  However related to this issue I think the only setting that would map to ~1 after GA would be something that starts with 1 or ~1

PR #581, opened 2018-09-12
Change beta to ~2

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/562


package.json
```diff
@@ -546,6 +546,7 @@
                         "type": "string",
                         "enum": [
                             "~1",
+                            "~2",
                             "beta"
```
StephenWeatherford: beta's in the enum here, but it's missing from runtime.md and api.md and elsewhere...?

EricJizbaMSFT: Yeah a lot of people will have it in their settings for VS Code. I figure we leave it for a release or two before removing it. Added descriptions so that it will now say "Deprecated":

screen shot 2018-09-12 at 1 32 11 pm

src/templates/FunctionTemplates.ts
```diff
@@ -26,7 +26,7 @@ export class FunctionTemplates {
         this.copyCSharpSettingsFromJS();
     }
 
-    public async getTemplates(language: string, runtime: string = ProjectRuntime.one, functionAppPath: string, templateFilter?: string, telemetryProperties?: TelemetryProperties): Promise<IFunctionTemplate[]> {
+    public async getTemplates(language: string, runtime: string = ProjectRuntime.v1, functionAppPath: string, templateFilter?: string, telemetryProperties?: TelemetryProperties): Promise<IFunctionTemplate[]> {
```
nturinski: Should we default to v2 now?

src/templates/TemplateRetriever.ts
```diff
@@ -92,10 +92,10 @@ export abstract class TemplateRetriever {
 
     protected getBackupVersion(runtime: ProjectRuntime): string {
```
nturinski: This isn't needed anymore, right?

src/templates/FunctionTemplates.ts
EricJizbaMSFT: Turns out we don't need a default value for this

src/templates/TemplateRetriever.ts
EricJizbaMSFT: Yeah it still is. This is the version of the backup templates shipped with the extension.

PR #582, opened 2018-09-12
Install ptvsd and other local debug fixes

nturinski:
Installs ptvsd (no longer need pre)
Ignores output when looking for python alias
Adds "func_env" & ."python_packages" to .gitignore (as this should not be checked in)

I don't think there are dev dependencies for python unless you make a different requirement.txt.  func_env is packed in with the .zip file created by `func pack` so it's probably necessary to run on the remote.

### DIFF
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -176,12 +176,24 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
         // .gitignore is created by `func init`
         const gitignorePath: string = path.join(this.functionAppPath, gitignoreFileName);
         if (await fse.pathExists(gitignorePath)) {
-            ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', this.funcEnv));
+            const pythonPackages: string = 'python_packages';
+            let writeFile: boolean = false;
             let gitignoreContents: string = (await fse.readFile(gitignorePath)).toString();
             // the func_env and ._python_packages are recreated and should not be checked in
-            gitignoreContents = gitignoreContents.concat(`${os.EOL}${this.funcEnv}`);
-            gitignoreContents = gitignoreContents.concat(`${os.EOL}.python_packages`);
-            await fse.writeFile(gitignorePath, gitignoreContents);
+            if (!gitignoreContents.includes(this.funcEnv)) {
+                ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', this.funcEnv));
+                gitignoreContents = gitignoreContents.concat(`${os.EOL}${this.funcEnv}`);
+                writeFile = true;
+            }
+            if (!gitignoreContents.includes(pythonPackages)) {
+                ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', pythonPackages));
+                gitignoreContents = gitignoreContents.concat(`${os.EOL}${pythonPackages}`);
+                writeFile = true;
+            }
+
+            if (writeFile) {
+                await fse.writeFile(gitignorePath, gitignoreContents);
+            }
         }
     }
 }
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -176,7 +176,7 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
         // .gitignore is created by `func init`
         const gitignorePath: string = path.join(this.functionAppPath, gitignoreFileName);
         if (await fse.pathExists(gitignorePath)) {
-            const pythonPackages: string = 'python_packages';
+            const pythonPackages: string = '.python_packages';
             let writeFile: boolean = false;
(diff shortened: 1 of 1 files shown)


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -159,12 +159,22 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
         const funcInitPython: string = 'func init ./ --worker-runtime python';
         switch (process.platform) {
             case Platform.Windows:
-                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `${await this.getVenvActivatePath(Platform.Windows)} && ${funcInitPython}`);
+                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `${await this.getVenvActivatePath(Platform.Windows)} && ${funcInitPython} && pip install ptvsd`);
                 break;
             case Platform.MacOS:
             default:
-                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `source ${await this.getVenvActivatePath(Platform.MacOS)} && ${funcInitPython}`);
+                await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `source ${await this.getVenvActivatePath(Platform.MacOS)} && ${funcInitPython} && pip install ptvsd`);
```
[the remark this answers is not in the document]
EricJizbaMSFT: Maybe "pip install ptvsd" should be a part of creating the virtual environment, rather than in "requirements.txt". We will have to add logic to re-create the venv if it's not there anyways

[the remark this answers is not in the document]
nturinski: Yeah... I was hoping you wouldn't say that.  I put it in here because it already had switch cases based on Platform, but you're right regarding comment 2.

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -166,5 +173,15 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                 await cpUtils.executeCommand(ext.outputChannel, this.functionAppPath, `source ${await this.getVenvActivatePath(Platform.MacOS)} && ${funcInitPython}`);
                 break;
         }
+        // .gitignore is created by `func init`
+        const gitignorePath: string = path.join(this.functionAppPath, gitignoreFileName);
+        if (await fse.pathExists(gitignorePath)) {
+            ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', this.funcEnv));
+            let gitignoreContents: string = (await fse.readFile(gitignorePath)).toString();
+            // the func_env and ._python_packages are recreated and should not be checked in
+            gitignoreContents = gitignoreContents.concat(`${os.EOL}${this.funcEnv}`);
```
EricJizbaMSFT: Eventually the func cli might do this for us. So you should do some checking like `if (!gitignoreContents.includes(this.funcEnv))`. No need to edit the file if it already has those lines.

nturinski: I did a bit more checking.  Probably is fine, but will give you a chance to verify if you don't like it

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -176,12 +176,24 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
...
-            gitignoreContents = gitignoreContents.concat(`${os.EOL}${this.funcEnv}`);
-            gitignoreContents = gitignoreContents.concat(`${os.EOL}.python_packages`);
-            await fse.writeFile(gitignorePath, gitignoreContents);
+            if (!gitignoreContents.includes(this.funcEnv)) {
+                ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', this.funcEnv));
+                gitignoreContents = gitignoreContents.concat(`${os.EOL}${this.funcEnv}`);
+                writeFile = true;
+            }
+            if (!gitignoreContents.includes(pythonPackages)) {
+                ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', pythonPackages));
```
EricJizbaMSFT: the 'gitAddFunc_Env' key has to be different inside localize

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -176,12 +176,24 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
         // .gitignore is created by `func init`
         const gitignorePath: string = path.join(this.functionAppPath, gitignoreFileName);
         if (await fse.pathExists(gitignorePath)) {
-            ext.outputChannel.appendLine(localize('gitAddFunc_Env', 'Adding "{0}" to .gitignore...', this.funcEnv));
+            const pythonPackages: string = 'python_packages';
```
EricJizbaMSFT: Shouldn't this be '.' at the beginning?

src/commands/createNewProject/PythonProjectCreator.ts
nturinski: .... thank you for being my monitor.

EricJizbaMSFT: monitor

Looks good now

Issue #584, opened 2018-09-12
Python projects need to run `func extensions install` before debugging


EricJizbaMSFT:
I see this for blob triggers unless I run `func extensions install`
```
[9/12/18 10:45:53 PM] The following 1 functions are in error:
[9/12/18 10:45:53 PM] BlobTriggerPython: The binding type(s) 'blobTrigger' are not registered. Please ensure the type is correct and the binding extension is installed.
```

@asavaritayal is this right? Should we be running `func extensions install`?

asavaritayal:
Correct, this applies to all function apps running on V2. Any bindings other than HTTP and timer will require a `func extensions install` before debugging.

EricJizbaMSFT:
My understanding was this wasn't required for C# or Java function apps. Only JavaScript - and apparently also Python.

For Python, do we need to run this before debug _and_ deploy? Or will `func pack` take care of the deploy scenario for us?

Issue #583, opened 2018-09-12
New python project should have empty AzureWebJobsStorage on Mac


EricJizbaMSFT:
In local.settings.json, I see this after creating a Python project:
"AzureWebJobsStorage": "{AzureWebJobsStorage}",

But I get errors like this:
```
[9/12/18 10:42:23 PM] Error indexing method 'Functions.BlobTriggerPython'
[9/12/18 10:42:23 PM] Microsoft.Azure.WebJobs.Host: Error indexing method 'Functions.BlobTriggerPython'. Microsoft.WindowsAzure.Storage: Settings must be of the form "name=value".
[9/12/18 10:42:23 PM] Function 'Functions.BlobTriggerPython' failed indexing and will be disabled.
[9/12/18 10:42:23 PM] Found the following functions:
[9/12/18 10:42:23 PM] Host.Functions.TimerTriggerPython
[9/12/18 10:42:23 PM] Host.Functions.QueueTriggerPython
[9/12/18 10:42:23 PM] Host.Functions.HttpTriggerPython
[9/12/18 10:42:23 PM]
[9/12/18 10:42:23 PM] Host initialized (540ms)
[9/12/18 10:42:23 PM] The listener for function 'Functions.TimerTriggerPython' was unable to start.
[9/12/18 10:42:23 PM] The listener for function 'Functions.TimerTriggerPython' was unable to start. Microsoft.WindowsAzure.Storage: Settings must be of the form "name=value".
```

Maybe it's trying to use the local storage emulator? Which doesn't work on non-windows... Anyways, it works if I change it to a connection string.

PR #585, opened 2018-09-12
Add python verified templates

EricJizbaMSFT:
And Cosmos DB trigger to v2 JavaScript

HttpTriggers are the only ones that actually work out of the box (at least on mac), but I filed the relevant bugs, so I feel comfortable putting these in verified:
https://github.com/Microsoft/vscode-azurefunctions/issues/584
https://github.com/Microsoft/vscode-azurefunctions/issues/583

Otherwise they're working for me.


nturinski:
Just to confirm, we still need to run `func extension install` before, right?

EricJizbaMSFT:
As discussed offline - yes that work still needs to be done

PR #588, opened 2018-09-13
Include func extensions install in Run Functions Host task

### LINKED ISSUE #584: Python projects need to run `func extensions install` before debugging

nturinski:
Fixes #584

### DIFF
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -65,13 +65,13 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                     type: 'shell',
                     dependsOn: 'build',
                     osx: {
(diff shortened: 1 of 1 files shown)


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -65,13 +65,13 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                     type: 'shell',
                     dependsOn: 'build',
```
EricJizbaMSFT: Can you remove this build thing since we decided it probs doesn't do anything

Issue #500, opened 2018-08-08
[Suggestion] Open file after downloading settings locally


v-dantch:
Repro Steps:
1. Right click 'Application Settings' node -> select 'Download Remote Settings...'.
2. Check the state from 'OUTPUT'.

Actual:
No way to open the downloaded settings locally in extension.
![suggestion](https://user-images.githubusercontent.com/34729092/43826220-56e0d726-9ae6-11e8-94cc-69dd132a711a.png)

Suggestion:
Add a link to open the downloaded settings locally.

PR #595, opened 2018-09-13
Show local settings file after downloading

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/500

### DIFF
--- a/src/commands/appSettings/downloadAppSettings.ts
+++ b/src/commands/appSettings/downloadAppSettings.ts
@@ -60,5 +60,6 @@ export async function downloadAppSettings(node?: IAzureNode): Promise<void> {
         }
     });
 
-    await vscode.window.showTextDocument(await vscode.workspace.openTextDocument(localSettingsUri));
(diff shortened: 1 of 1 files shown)


src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -59,4 +59,6 @@ export async function downloadAppSettings(node?: IAzureNode): Promise<void> {
             }
         }
     });
+
+    await vscode.window.showTextDocument(await vscode.workspace.openTextDocument(localSettingsUri));
```
PrashanthCorp: I'm not a fan of having multiple uses of await in a single expression. await tends to abstract away the fact that a separate context is created for evaluating/resolving that promise. That works fine in a single context. However, it can lead to errors like the ones mentioned here: 
https://dev.to/antogarand/javascript-async-maths-is-hard-5b6i and here : https://www.reddit.com/r/javascript/comments/8lih82/javascript_operator_causing_race_conditions_when/?st=jm13b2wo&sh=8abe1c71

I know this goes against your policy of single use expressions, but you could consider this.

PR #601, opened 2018-09-13
 Don't use storage emulator on mac/linux

EricJizbaMSFT:
By default, the python project created with 'func init' is setting azureWebJobsStorage to use the local storage emulator, but that's not available on mac/linux. Instead we set to empty string.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/583


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -189,5 +191,14 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                 await fse.writeFile(gitignorePath, gitignoreContents);
             }
         }
+
```
StephenWeatherford: HOw about a comment saying why?

PR #605, opened 2018-09-17
Use admin api to get function keys

### LINKED ISSUE #567: Fail to display the object result after deploying to function app
Repro Steps:
1. Create a new project with C#.
2. Create a new function with HTTP trigger.
3. Debug the .CS file.
4. Click 'Deploy to Function App...' to deploy the file to function app.
5. Copy the function URL and check the result.

Expect:
The object result displays well.

Actual:
The object result does not diaplay.
actual

EricJizbaMSFT:
Fixes #567 

Based on the discussion here: https://github.com/Azure/azure-functions-host/issues/3411, the kudu api is deprecated and the ARM api is not ready. However, the admin api works

Relies on this PR in shared package: https://github.com/Microsoft/vscode-azuretools/pull/264


src/tree/FunctionTreeItem.ts
```diff
@@ -73,32 +71,42 @@ export class FunctionTreeItem implements ILogStreamTreeItem {
...
 
     public async initializeTriggerUrl(): Promise<void> {
-        const kuduClient: KuduClient = await getKuduClient(this.client);
-        const functionSecrets: FunctionSecrets = await kuduClient.functionModel.getSecrets(this._name);
-        if (functionSecrets.triggerUrl === undefined) {
-            throw new ArgumentError(functionSecrets);
+        const triggerUrl: URL = new URL(`${this.client.defaultHostUrl}/api/${this._name}`);
+        const key: string | undefined = await this.getKey();
+        if (key) {
+            // tslint:disable-next-line:no-backbone-get-set-outside-model
```
StephenWeatherford: Given that you're not using backbone/, you should probably just disable this rule

src/tree/FunctionTreeItem.ts
```diff
@@ -73,32 +71,42 @@ export class FunctionTreeItem implements ILogStreamTreeItem {
...
+            case HttpAuthLevel.anonymous:
+            default:
+                return undefined;
         }
 
-        this._triggerUrl = triggerUrl.toString();
+        const data: string = await functionsAdminRequest(this.client, urlPath);
+        try {
+            // tslint:disable-next-line:no-unsafe-any
+            const result: string = JSON.parse(data).value;
```
StephenWeatherford: I had more like with:

const result = <string>JSON.parse(data).value

may depend on your tslint settings though

Issue #502, opened 2018-08-09
Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again.


ranbirpaul:
Repro steps:
<Enter steps to reproduce issue>

Action: azureFunctions.deploy
Error type: Error
Error Message: Unable to write to Workspace Settings because no workspace is opened. Please open a workspace first and try again.

Version: 0.10.0
OS: win32

JohnReeson:
You need to add your function app folder to workspace first. 

![1](https://user-images.githubusercontent.com/35160013/43973185-5d4ec62a-9d09-11e8-91f5-86a7e51cbfec.png)

EricJizbaMSFT:
Offhand I'm not sure what settings we're trying to write while deploying. As @JohnReeson mentioned you can just open the folder as a workaround. However, we can investigate getting rid of this error completely

JohnReeson:
![12](https://user-images.githubusercontent.com/35160013/44307129-69991d00-a3cf-11e8-9837-e02fa3fe17ee.png)

@EricJizbaMSFT  It seems to write azure-functions-related settings(in ~/.vscode/settings.json), same as when we initialize a new function project. Looks like any function related operation requires this workspace settings and we may get the prompt when no workspace is opened(no settings available).

PR #606, opened 2018-09-17
Force users to have workspace open when deploying

EricJizbaMSFT:
We check several settings when deploying and we can't read/write those if the folder isn't open in the workspace

Here's the new message:
screen shot 2018-09-17 at 2 28 58 pm

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/502

### DIFF
--- a/src/utils/workspace.ts
+++ b/src/utils/workspace.ts
@@ -92,6 +92,7 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
     } else {
         if (message) {
             const open: vscode.MessageItem = { title: localize('open', 'Open Folder') };
+            // No need to check result. Open/Cancel are the only possibilities and Cancel will throw a UserCancelledError
             await ext.ui.showWarningMessage(message, { modal: true }, open);
         }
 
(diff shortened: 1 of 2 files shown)


src/utils/workspace.ts
```diff
@@ -80,16 +80,21 @@ const projectOpenBehaviorSetting: string = 'projectOpenBehavior';
...
     const folder: vscode.WorkspaceFolder | undefined = openFolders.find((f: vscode.WorkspaceFolder): boolean => {
-        return fsUtils.isPathEqual(f.uri.fsPath, fsPath);
+        return fsUtils.isPathEqual(f.uri.fsPath, fsPath) || (allowSubFolder && fsUtils.isSubpath(f.uri.fsPath, fsPath));
     });
 
     if (folder) {
         actionContext.properties.openBehavior = 'AlreadyOpen';
     } else {
+        if (message) {
+            const open: vscode.MessageItem = { title: localize('open', 'Open Folder') };
```
StephenWeatherford: Am I missing where you're checking for 'Open Folder' being selected?

EricJizbaMSFT: Nope. 'Open Folder' is the only option and ext.ui handles the 'cancel' case with a UserCancelledError. I agree it looks a little weird, but it definitely works.

StephenWeatherford: Might be worth a comment.

PR #609, opened 2018-09-18
Clarify preview support for Python

EricJizbaMSFT:
Per the Functions team, they're not doing a "full" preview of Python in Azure quite yet, so they want our support under a feature flag for the ignite release.

I also added "Linux Preview" in the explorer like so (Approved by Mr Matt):
screen shot 2018-09-18 at 9 42 28 am
screen shot 2018-09-18 at 9 42 13 am

### DIFF
--- a/src/tree/FunctionAppProvider.ts
+++ b/src/tree/FunctionAppProvider.ts
@@ -4,8 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementClient } from 'azure-arm-website';
-import { Site, WebAppCollection } from "azure-arm-website/lib/models";
-import { OutputChannel } from "vscode";
+import { AppServicePlan, Site, WebAppCollection } from "azure-arm-website/lib/models";
 import { createFunctionApp, IAppCreateOptions, SiteClient } from 'vscode-azureappservice';
 import { addExtensionUserAgent, IActionContext, IAzureNode, IAzureTreeItem, IChildProvider, parseError } from 'vscode-azureextensionui';
 import { ProjectLanguage, projectLanguageSetting, ProjectRuntime, projectRuntimeSetting } from '../constants';
@@ -20,11 +19,6 @@ export class FunctionAppProvider implements IChildProvider {
     public readonly childTypeLabel: string = localize('azFunc.FunctionApp', 'Function App');
 
     private _nextLink: string | undefined;
-    private readonly _outputChannel: OutputChannel;
-
-    public constructor(outputChannel: OutputChannel) {
-        this._outputChannel = outputChannel;
-    }
 
     public hasMoreChildren(): boolean {
         return this._nextLink !== undefined;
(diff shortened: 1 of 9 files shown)


src/tree/FunctionAppProvider.ts
```diff
@@ -56,18 +50,20 @@ export class FunctionAppProvider implements IChildProvider {
...
         const treeItems: IAzureTreeItem[] = [];
-        for (const site of webAppCollection) {
+        await Promise.all(webAppCollection.map(async (site: Site) => {
             try {
                 const siteClient: SiteClient = new SiteClient(site, node);
                 if (siteClient.isFunctionApp) {
-                    treeItems.push(new FunctionAppTreeItem(siteClient, this._outputChannel));
+                    const asp: AppServicePlan = await siteClient.getAppServicePlan();
+                    const isLinuxPreview: boolean = siteClient.kind.toLowerCase().includes('linux') && !!asp.sku && !!asp.sku.tier && asp.sku.tier.toLowerCase() === 'dynamic';
+                    treeItems.push(new FunctionAppTreeItem(siteClient, isLinuxPreview));
```
StephenWeatherford: I'm confused on the link between python and Linux.  You say python requires Linux, but isn't Linux used elsewhere?

EricJizbaMSFT: There are two separate things in preview right now:
1. Python
1. Linux consumption

Python only works on Linux. Other project languages _can_ be created on Linux, but they already work on Windows so we don't support their _creation_ from our extension (but we do support displaying them in the tree).

src/tree/FunctionAppTreeItem.ts
```diff
@@ -23,15 +23,15 @@ export class FunctionAppTreeItem implements ILogStreamTreeItem, IAzureParentTree
     private readonly _functionsTreeItem: FunctionsTreeItem;
     private readonly _appSettingsTreeItem: AppSettingsTreeItem;
     private readonly _proxiesTreeItem: ProxiesTreeItem;
-    private readonly _outputChannel: OutputChannel;
+    private readonly _isLinuxPreview: boolean;
 
-    public constructor(client: SiteClient, outputChannel: OutputChannel) {
+    public constructor(client: SiteClient, isLinuxPreview: boolean) {
```
StephenWeatherford: I'd rather see 

public constructor(client: SiteClient, options: { isLinuxPreview: Boolean})

That'd be much easier to understand.

src/tree/FunctionAppProvider.ts
StephenWeatherford: Some comments in the code would be helpful.

src/tree/FunctionAppProvider.ts
```diff
@@ -56,18 +50,20 @@ export class FunctionAppProvider implements IChildProvider {
         this._nextLink = webAppCollection.nextLink;
 
         const treeItems: IAzureTreeItem[] = [];
-        for (const site of webAppCollection) {
+        await Promise.all(webAppCollection.map(async (site: Site) => {
```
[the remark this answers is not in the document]
EricJizbaMSFT: Seems like a risky change right before a release. Will look into later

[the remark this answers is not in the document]
EricJizbaMSFT: (We use Promise.all in several places, so I would want to change all at once)

Issue #563, opened 2018-09-05
Change npm install command to use latest tag


fiveisprime:
As part of GA, v2 will be moved to latest so we'll need to remove the core tag from the npm install command when installing the functions tools.

EricJizbaMSFT:
Going forward it should be 'latest' for 'v2' and 'v1' for 'v1' per mr matt

PR #610, opened 2018-09-18
Dynamically retrieve appropriate dist-tag for npm

EricJizbaMSFT:
The tags are changing when v2 GA's soon. We could hard-code the new strings and try to time our release perfectly with the func cli, but I'd rather dynamically check the dist-tags.

Tested on mac/windows. I'm not worried about Linux since we don't support install in that scenario (Linux requires sudo so we ignore it).

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/563

### DIFF
--- a/src/funcCoreTools/getNpmDistTag.ts
+++ b/src/funcCoreTools/getNpmDistTag.ts
@@ -1,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line:no-require-imports
+import request = require('request-promise');
+import { ProjectRuntime } from '../constants';
+import { localize } from '../localize';
+
+const npmRegistryUri: string = 'https://aka.ms/W2mvv3';
+
+export interface INpmDistTag { tag: string; value: string; }
+
+export async function getNpmDistTag(runtime: ProjectRuntime): Promise<INpmDistTag> {
+    const tags: { [key: string]: string } = <{ [key: string]: string }>JSON.parse(await <Thenable<string>>request(npmRegistryUri));
+    for (const key of Object.keys(tags)) {
+        if ((runtime === ProjectRuntime.v1 && tags[key].startsWith('1')) ||
+            (runtime === ProjectRuntime.v2 && tags[key].startsWith('2'))) {
+            return { tag: key, value: tags[key] };
+        }
+    }
+
+    throw new Error(localize('noDistTag', 'Failed to retrieve NPM tag for runtime "{0}".', runtime));
+}
--- a/src/funcCoreTools/installFuncCoreTools.ts
+++ b/src/funcCoreTools/installFuncCoreTools.ts
@@ -8,6 +8,7 @@ import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { promptForProjectRuntime } from "../ProjectSettings";
 import { cpUtils } from '../utils/cpUtils';
+import { getNpmDistTag, INpmDistTag } from './getNpmDistTag';
 
 export async function installFuncCoreTools(packageManager: PackageManager): Promise<void> {
     let runtime: ProjectRuntime;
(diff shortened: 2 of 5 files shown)


src/funcCoreTools/getNpmDistTag.ts
```diff
@@ -0,0 +1,25 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+// tslint:disable-next-line:no-require-imports
+import request = require('request-promise');
```
StephenWeatherford: Just FYI, in some scenarios I had issues with lint using request-promise because it uses bluebird.  In docker I switched to request-native, which is the same without bluebird.  You may not hit this, though, I think it was related to my trying to strongly type something from request.

EricJizbaMSFT: No I hit it, too. I worked around it by casting to `<Thenable<string>>` below.

I thought about moving to the request-native one you mentioned (after I saw your PR in docker), but I'm actually hoping we can switch to [request-light](https://www.npmjs.com/package/request-light), which is catered to VS Code extensions and supports proxies. I filed an issue here to get more info on the package: https://github.com/Microsoft/node-request-light/issues/4

src/funcCoreTools/getNpmDistTag.ts
```diff
@@ -0,0 +1,25 @@
...
+import { localize } from '../localize';
+
+const npmRegistryUri: string = 'https://aka.ms/W2mvv3';
+
+export interface INpmDistTag { tag: string; value: string; }
+
+export async function getNpmDistTag(runtime: ProjectRuntime): Promise<INpmDistTag> {
+    const tags: { [key: string]: string } = <{ [key: string]: string }>JSON.parse(await <Thenable<string>>request(npmRegistryUri));
+    for (const key of Object.keys(tags)) {
+        if ((runtime === ProjectRuntime.v1 && tags[key].startsWith('1')) ||
```
StephenWeatherford: You're probably gonna have to revamp this later...

src/funcCoreTools/installFuncCoreTools.ts
```diff
@@ -20,16 +21,8 @@ export async function installFuncCoreTools(packageManager: PackageManager): Prom
     ext.outputChannel.show();
     switch (packageManager) {
         case PackageManager.npm:
-            switch (runtime) {
-                case ProjectRuntime.v1:
-                    await cpUtils.executeCommand(ext.outputChannel, undefined, 'npm', 'install', '-g', funcPackageName);
-                    break;
-                case ProjectRuntime.v2:
-                    await cpUtils.executeCommand(ext.outputChannel, undefined, 'npm', 'install', '-g', `${funcPackageName}@core`, '--unsafe-perm', 'true');
```
StephenWeatherford: --unsafe-perm no longer needed?

src/funcCoreTools/getNpmDistTag.ts
StephenWeatherford: Cool, thanks

src/funcCoreTools/installOrUpdateFuncCoreTools.ts
```diff
@@ -15,8 +15,7 @@ import { funcToolsInstalled } from './validateFuncCoreToolsInstalled';
 export async function installOrUpdateFuncCoreTools(): Promise<void> {
     const isFuncInstalled: boolean = await funcToolsInstalled();
     const packageManager: PackageManager | undefined = await getFuncPackageManager(isFuncInstalled);
-    // tslint:disable-next-line:strict-boolean-expressions
-    if (!packageManager) {
+    if (packageManager === undefined) {
```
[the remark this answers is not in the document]
StephenWeatherford: How was it a bug?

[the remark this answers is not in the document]
EricJizbaMSFT: The enum for this one is just:
```
export enum PackageManager {
    npm,
    brew
}
```

Which meant 'npm' boiled down to `0` which returns the wrong thing with `!packageManager`. This is the whole reason strict-boolean-expressions checks against enums

src/funcCoreTools/getNpmDistTag.ts
EricJizbaMSFT: Why? You mean when there's a v3?

PR #611, opened 2018-09-18
Add separate tree item for invalid function

### LINKED ISSUE #497: Error Message: Failed to parse function.json: Expected "bindings" element of type "Array".
Repro steps:
<Enter steps to reproduce issue>

Action: azureFunctions.deploy
Error type: Error
Error Message: Failed to parse function.json: Expected "bindings" element of type "Array".

Version: 0.10.0
OS: win32

EricJizbaMSFT:
We print out functions after a user deploys and a single invalid function results in an error that makes it seem like the entire deploy failed. Instead, we will leverage the 'InvalidTreeItem' logic when listing functions so that those get filtered out.

Here is what it looks like (Not great on dark theme, but the grayscale is accurate compared to the "not great" view of the colored icon itself):
screen shot 2018-09-18 at 2 11 22 pm
screen shot 2018-09-18 at 2 11 37 pm

Fixes #497


src/tree/FunctionsTreeItem.ts
```diff
@@ -45,14 +46,24 @@ export class FunctionsTreeItem implements IAzureParentTreeItem {
...
+            try {
+                const treeItem: FunctionTreeItem = new FunctionTreeItem(this._client, fe, this._outputChannel);
+                if (treeItem.config.isHttpTrigger) {
+                    // We want to cache the trigger url so that it is instantaneously copied when the user performs the copy action
+                    // (Otherwise there might be a second or two delay which could lead to confusion)
+                    await treeItem.initializeTriggerUrl();
+                }
+                treeItems.push(treeItem);
+            } catch (error) {
+                if (fe.id) {
```
StephenWeatherford: handle !fe.id
?

StephenWeatherford: e.g. (fe.id ? getFunctionNameFromId(fe.id) : '(invalid ID)')?

EricJizbaMSFT: I'll handle in a separate PR

PR #614, opened 2018-09-19
Prep for 0.11.0 release

EricJizbaMSFT:


README.md
```diff
@@ -64,7 +67,7 @@ npm i -g azure-functions-core-tools
 To install v2:
 
 ```bash
-npm i -g azure-functions-core-tools@core --unsafe-perm true
+npm i -g azure-functions-core-tools@core
```
EricJizbaMSFT: `--unsafe-perm true` is only necessary when running as sudo and we already have the note a few lines above that says:
 > **NOTE**: npm can be used on all platforms. On unix platforms, you may need to specify `--unsafe-perm` if you are running npm with sudo. That's due to npm behavior of post install script.

README.md
```diff
@@ -19,6 +19,7 @@ Install the prerequisites for your desired language:
 * [JavaScript](#javascript)
 * [C#](#c)
 * [Java](#java)
+* [Python](#python)
```
StephenWeatherford: [Python (Preview)]
?

README.md
```diff
@@ -53,7 +54,9 @@ Install the prerequisites for your desired language:
 
 ### Windows
 
-Both v1 and v2 of the runtime can be installed on Windows. v2 is cross platform, but is still in preview and not recommended for production use. v1 is Windows-only, but is approved for production use.
+Both v1 and v2 of the runtime can be installed on Windows. [See here](https://docs.microsoft.com/azure/azure-functions/functions-versions) for more information on each version.
+
+> **IMPORTANT**: v2 is becoming generally available (GA) soon (as of 9/19/2018). After GA, the "core" tag will no longer be required for v2, and a "v1" tag will be introduced for v1.
```
StephenWeatherford: Rather than talking about tags, I wonder if it would be better to be explicitly about the npm install instructions after GA.

Issue #615, opened 2018-09-20
[Discussion] "Deploy to Function App" on a sub folder for Python


EricJizbaMSFT:
Since we're running `func pack` and looking for a specific zip file, I'm not sure it makes sense to allow users to "Deploy to Function App" on a sub folder (and actually respect that sub folder)

cc @davidebbo started this discussion

PR #620, opened 2018-09-20
Give "azureFunctions.deploySubpath" setting more precedence

EricJizbaMSFT:
We've had multiple customers fail to deploy because they selected "Deploy to Function App" from the wrong sub folder of a project. This hopefully fixes that in two ways:

1. Use "azureFunctions.deploySubpath" instead of the folder they select. Users will get this warning:
screen shot 2018-09-20 at 10 08 17 am

1. If there is only one workspace folder open and the deploySubpath setting is not an empty string - just use that path without prompting the user at all.

C# already has "deploySubpath" set on new projects. I modified Python and JavaScript to also set it.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/615


fiveisprime:
I love this fix, but I'm not sold on the dialog text. The bit about the deploy subpath is more distracting than helpful. Maybe reword it as 

> Deploying "." instead of selected folder "directory". Use `azureFunctions.deploySubpath` to override this behavior.

EricJizbaMSFT:
You got it boss:
screen shot 2018-09-20 at 11 05 37 am

src/commands/deploy.ts
```diff
@@ -159,15 +151,54 @@ async function listHttpTriggerUrls(node: IAzureParentNode): Promise<void> {
...
+}
+
 /**
  * Appends the deploySubpath setting if the target path matches the root of a workspace folder
  * If the targetPath is a sub folder instead of the root, leave the targetPath as-is and assume they want that exact folder used
  */
-function appendDeploySubpathSetting(targetPath: string): string {
-    if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.some((folder: vscode.WorkspaceFolder) => isPathEqual(folder.uri.fsPath, targetPath))) {
+async function appendDeploySubpathSetting(targetPath: string): Promise<string> {
+    if (vscode.workspace.workspaceFolders) {
```
StephenWeatherford: You had another PR to error if no workspaceFolders, right?

EricJizbaMSFT:
@fiveisprime updated dialog based on offline discussion with Stephen
screen shot 2018-09-20 at 11 49 12 am

"override" is confusing because "azureFunctions.deploySubpath" is already overriding the selected folder. "change" sounds better to us. Let us know if you disagree

Issue #591, opened 2018-09-13
Throw error if the user is trying to create a function using python 3.7


asavaritayal:
Python 3.7 is not yet supported for Azure Functions.

EricJizbaMSFT:
@asavaritayal I'm reluctant to hard-code a max version in our extension since (I assume) you will eventually support 3.7. At that point, we would have to ship an update just to unblock 3.7 users.

The func cli will have some of it's own error handling around this, right? That might be enough...

EricJizbaMSFT:
Tested on 3.7. Users will see this error message:
![screen shot 2018-09-20 at 1 45 19 pm](https://user-images.githubusercontent.com/11282622/45846017-7ca75180-bcdb-11e8-9e78-d82f71cdeed3.png)

And this is the full output:
```
Running command: "python3 -m venv func_env"...
Finished running command: "python3 -m venv func_env".
Running command: "source func_env/bin/activate && pip install ptvsd "...
Collecting ptvsd
  Using cached https://files.pythonhosted.org/packages/ab/96/d454d2d7da6b2410f4d74d651f447161d8ae2404fcba4454d683eceb4e09/ptvsd-4.1.2.zip
Installing collected packages: ptvsd
  Running setup.py install for ptvsd: started
    Running setup.py install for ptvsd: finished with status 'done'
Successfully installed ptvsd-4.1.2
You are using pip version 10.0.1, however version 18.0 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Finished running command: "source func_env/bin/activate && pip install ptvsd ".
Running command: "source func_env/bin/activate && func init ./ --worker-runtime python "...
Python 3.6 is required. Current python version is 'Python 3.7.0


'
Error: Failed to run "source func_env/bin/activate && func init ./ --worker-runtime python" command. Check output window for more details.
```

This line makes it pretty obvious: "Python 3.6 is required. Current python version is 'Python 3.7.0", but I admit it's annoying to parse through the output. @jinglouMSFT @fiveisprime throughts? Should we hardcode the maximum version of 3.7 in our own validation or is this a good enough experience?

PR #622, opened 2018-09-20
Improve Python version parsing

EricJizbaMSFT:
This doesn't happen consistently for me, but sometimes the `python --version` command includes more information than just "Python 3.7.0". For example:
screen shot 2018-09-20 at 1 54 05 pm

Plus I just trust a regex more than I trust the substring logic


EricJizbaMSFT:
Also fixes https://github.com/Microsoft/vscode-azurefunctions/issues/591 - discussed with @jinglouMSFT offline.

PR #635, opened 2018-09-21
 Better error message when failing to list functions on linux

EricJizbaMSFT:
There's a bug in the ARM api. Hopefully they fix it soon, but I think it's worth it to show a slightly better error message since we try to list functions after every deploy.

From this:
screen shot 2018-09-21 at 10 48 29 am

To this:
screen shot 2018-09-21 at 10 45 04 am

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -130,24 +130,31 @@ export async function deploy(ui: IAzureUserInput, actionContext: IActionContext,
         }
     });
 
-    await listHttpTriggerUrls(node);
+    await listHttpTriggerUrls(node, actionContext);
 }
 
-async function listHttpTriggerUrls(node: IAzureParentNode): Promise<void> {
-    const children: IAzureNode[] = await node.getCachedChildren();
-    const functionsNode: IAzureParentNode<FunctionsTreeItem> = <IAzureParentNode<FunctionsTreeItem>>children.find((n: IAzureNode) => n.treeItem instanceof FunctionsTreeItem);
-    await node.treeDataProvider.refresh(functionsNode);
-    const functions: IAzureNode[] = await functionsNode.getCachedChildren();
-    const anonFunctions: IAzureNode<FunctionTreeItem>[] = <IAzureNode<FunctionTreeItem>[]>functions.filter((f: IAzureNode) => f.treeItem instanceof FunctionTreeItem && f.treeItem.config.isHttpTrigger && f.treeItem.config.authLevel === HttpAuthLevel.anonymous);
-    if (anonFunctions.length > 0) {
-        ext.outputChannel.appendLine(localize('anonymousFunctionUrls', 'HTTP Trigger Urls:'));
-        for (const func of anonFunctions) {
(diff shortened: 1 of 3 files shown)


PrashanthCorp:
Can the user act on this to remove the cause of error? If yes, maybe a line on a workaround might help. Else, this looks like the best that can be done, and I'll approve this.

EricJizbaMSFT:
The tree will still have that error, though:
screen shot 2018-09-21 at 11 26 23 am

src/commands/deploy.ts
```diff
@@ -130,24 +130,31 @@ export async function deploy(ui: IAzureUserInput, actionContext: IActionContext,
...
+            }
         }
-    }
 
-    if (functions.find((f: IAzureNode) => f.treeItem instanceof FunctionTreeItem && f.treeItem.config.isHttpTrigger && f.treeItem.config.authLevel !== HttpAuthLevel.anonymous)) {
-        ext.outputChannel.appendLine(localize('nonAnonymousWarning', 'WARNING: Some http trigger urls cannot be displayed in the output window because they require an authentication token. Instead, you may copy them from the Azure Functions explorer.'));
+        if (functions.find((f: IAzureNode) => f.treeItem instanceof FunctionTreeItem && f.treeItem.config.isHttpTrigger && f.treeItem.config.authLevel !== HttpAuthLevel.anonymous)) {
+            ext.outputChannel.appendLine(localize('nonAnonymousWarning', 'WARNING: Some http trigger urls cannot be displayed in the output window because they require an authentication token. Instead, you may copy them from the Azure Functions explorer.'));
+        }
+    } catch (error) {
```
StephenWeatherford: I'd rather see this confined to a smaller area and maybe to a specific error code.

src/commands/deploy.ts
```diff
@@ -130,24 +130,31 @@ export async function deploy(ui: IAzureUserInput, actionContext: IActionContext,
...
+        // suppress error notification and instead display a warning in the output. We don't want it to seem like the deployment failed.
+        actionContext.suppressErrorDisplay = true;
+        ext.outputChannel.appendLine(localize('failedToList', 'WARNING: Deployment succeeded, but failed to list http trigger urls.'));
+        throw error;
```
PrashanthCorp: parseError?

PrashanthCorp: Scratch this. You're not using the message field anywhere, letting VSCode handle it would not require a parseError here.

EricJizbaMSFT: I'm just re-throwing this so that it will be caught in telemetry (it's not displayed since 'suppressErrorDisplay' is set to true above). No need to parse or do anything else

src/commands/deploy.ts
EricJizbaMSFT: Discussed offline. This is confined to the `listHttpTriggerUrls` function, which only occurs after a deployment has already succeeded. No matter the error - we don't want "listing the triggers" to make it seem like the deploy itself has failed

PR #663, opened 2018-09-25
Improve tests

EricJizbaMSFT:
Mostly based off of @StephenWeatherford's stuff in Docker here: https://github.com/Microsoft/vscode-docker/pull/387. Includes the following:
1. Fix template tests so that we guarantee it's testing against cli feed _and_ backup templates (current code was just testing against cli feed twice)
1. Remove a little bit of duplicate code in suiteSetup
1. Turn off telemetry for tests
1. Allow running a subset of tests with MOCHA_grep (super excited about this one - thanks Stephen)


EricJizbaMSFT:
Couldn't wait for MOCHA_grep haha

Issue #639, opened 2018-09-22
Creating a queue triggered Function after debug results in extensions.csproj not getting updated


davidebbo:
This is probably a side effect of #638.

Steps:
- New app
- Add http trigger
- Debug it
- Add queue trigger

Result: extensions.csproj is not updated, so it's broken.

Issue #638, opened 2018-09-22
Stop Debugging doesn't cause func.exe to stop


davidebbo:
After debugging and choosing Stop Debugging, func.exe is still running. There is no simple gesture to stop it other than Ctrl-C out of it or kill the process.

By comparison, in VS, Stop Debugging stops func.exe (if it had started it, not if you attached to it post creation).

PR #671, opened 2018-09-28
Stop func host task after debug session ends

EricJizbaMSFT:
The bug was originally filed for Python, but this helps all languages.

Also removed 'identifier' from tasks.json since it's now deprecated.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/638
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/639


julielerman:
can this be what is creating so many issues with me right now trying to debug with latest extensio (.11), tools, cli (2.0.3) and webjobs (3.0.1 cosmosdb) installed? I'm on VSCode 1.28.1. Also: .NET Command Line Tools (2.2.100-preview2-009404)
image

Issue #649, opened 2018-09-23
C# debugging attaches to wrong process when func installed through npm on Mac


EricJizbaMSFT:
Repro steps:
1. Install the func cli through npm (not brew) on mac
1. Create a new C# project with an HttpTrigger
1. Set a breakpoint in HttpTrigger
1. Start debugging
1. Hit the HttpTrigger

Expected: breakpoint is hit
Actual: breakpoint is never hit

This does _not_ repro on Windows. It also does _not_ repro if you installed the func cli through brew (which is the default install method on a mac).

PR #675, opened 2018-10-01
Get child process if func installed with npm on mac

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/649

See that issue for more details


src/commands/pickFuncProcess.ts
```diff
@@ -106,6 +105,26 @@ async function startFuncTask(funcTask: Task, timeoutInSeconds: number, timeoutEr
...
+/**
+ * Gets the innermost child pid for a unix process. This is only really necessary if the user installs 'func' with a tool like 'npm', which uses a wrapper around the main func exe
+ */
+async function getInnermostUnixPid(pid: string): Promise<string> {
+    return await new Promise<string>((resolve: (pid: string) => void, reject: (e: Error) => void): void => {
+        unixPsTree(parseInt(pid), (error: Error | undefined, children: unixPsTree.PS[]) => {
+            if (error) {
+                reject(error);
+            } else {
+                const child: unixPsTree.PS | undefined = children.pop();
```
nturinski: Would this get the innermost?  Isn't this only getting the first child?

EricJizbaMSFT: I mean 'pop' is removing the last element from the array. I'm assuming the last element is the innermost child, but I suppose it might not be. That being said - I'm not sure how I can test that since I can only reproduce a case where 'func' is the first (and only) child.

Let me know if you think it's worth it and I'll try to repro a case with multiple nested children

nturinski: Nah, I think if it becomes a problem, we can resolve it then.

PR #678, opened 2018-10-01
Fix getNpmDistTag for v1

EricJizbaMSFT:
We thought the func team was going to add a 'v1' tag [here](http://registry.npmjs.org/-/package/azure-functions-core-tools/dist-tags), but instead this is what the api is returning:
```json
{
  "latest": "2.0.3",
  "core": "2.0.3"
}
```

The only consequence for users right now is that they're not getting this warning:
screen shot 2018-10-01 at 3 48 51 pm

The [install instructions](https://github.com/Azure/azure-functions-core-tools#windows) say to just use `npm i -g azure-functions-core-tools@1` which is easy enough for us to do (and won't ever change in the future since it's not really a 'tag', just the major version).

### DIFF
--- a/src/funcCoreTools/getNpmDistTag.ts
+++ b/src/funcCoreTools/getNpmDistTag.ts
@@ -31,10 +31,10 @@ export async function getNpmDistTag(runtime: ProjectRuntime): Promise<INpmDistTa
             throw new RangeError(localize('invalidRuntime', 'Invalid runtime "{0}".', runtime));
     }
 
-    const filteredVersions: string[] = Object.keys(packageMetadata.versions).filter((v: string) => v.startsWith(majorVersion));
-    if (filteredVersions.length < 1) {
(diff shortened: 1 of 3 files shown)


src/funcCoreTools/getNpmDistTag.ts
```diff
@@ -5,21 +5,36 @@
...
+            break;
+        case ProjectRuntime.v2:
+            majorVersion = '2';
+            break;
+        default:
+            throw new RangeError(localize('invalidRuntime', 'Invalid runtime "{0}".', runtime));
     }
 
-    throw new Error(localize('noDistTag', 'Failed to retrieve NPM tag for runtime "{0}".', runtime));
+    const filteredVersions: string[] = Object.keys(packageMetadata.versions).filter((v: string) => v.startsWith(majorVersion));
```
StephenWeatherford: You're going be sorry when v10 comes out :-)

PR #683, opened 2018-10-02
 Add pip install command to python projects before debugging

### LINKED ISSUE #666: Add `pip install` step to `tasks.json`
## Currently, we only activate the virtual environment

```json
"osx": {
        "command": "func extensions install && . func_env/bin/activate && func host start"
      },
```

## Would be great if we can also install the dependencies when starting the functions host

```json
"osx": {
        "command": "func extensions install && . func_env/bin/activate && pip install -r requirements.txt && func host start"
      },
```

EricJizbaMSFT:
Fixes #666 

I tested deploying and `func pack` already handles the requirements in that case.

### DIFF
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -96,13 +96,13 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                     identifier: funcHostTaskId,
                     type: 'shell',
                     osx: {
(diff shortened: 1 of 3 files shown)


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -224,14 +225,12 @@ async function validatePythonAlias(pyAlias: PythonAlias): Promise<string | undef
     }
 }
 
-function convertToVenvCommand(command: string, platform: NodeJS.Platform, separator?: string): string {
+function getVenvActivateCommand(platform: NodeJS.Platform): string {
```
nturinski: I kind of liked having the `convertToVenvCommand`.  I'm not sure it's immediately obvious why you're using `joinCommands` and then `getVenvActivateCommand.

PR #689, opened 2018-10-03
Refactor to use new tree

EricJizbaMSFT:
Dependent on this PR: https://github.com/Microsoft/vscode-azuretools/pull/278


src/tree/FunctionAppTreeItem.ts
```diff
@@ -3,47 +3,51 @@
...
-        this._functionsTreeItem = new FunctionsTreeItem(client);
-        this._appSettingsTreeItem = new AppSettingsTreeItem(client);
-        this._proxiesTreeItem = new ProxiesTreeItem(client);
+        this._functionsTreeItem = new FunctionsTreeItem(this);
+        this._appSettingsTreeItem = new AppSettingsTreeItem(this);
+        this._proxiesTreeItem = new ProxiesTreeItem(this);
         this._isLinuxPreview = isLinuxPreview;
     }
 
+    public get root(): ISiteTreeRoot {
```
StephenWeatherford: Doesn't this conflict with the existing root property?

EricJizbaMSFT: Nope just overrides it

StephenWeatherford: an @override doc comment tag or //override comment would be helpful.

PR #695, opened 2018-10-04
Check for C#Script && v1 runtime to enable WEBSITE_RUN_FROM_PACKAGE

### LINKED ISSUE #684: Creating function app on v1 for CSX should not use "Run From Package"
Copied from email conversation:
> For CSX, we do a NuGet restore after deployment. Hence, when deploying CSX from VS Code, the user’s site won’t work since running from package makes the filesystem read only, and the restore fails. They would have been fine if they performed a NuGet restore themselves, but we don’t give any tooling for this. This is true whether on 1.x or 2.x, but for 2.x we’re comfortable making the breaking change to encourage customers to move away from CSX.

nturinski:
Fixes #684 

Just check if the project is NOT and the runtime is NOT v1.  If it is, then we set `WEBSITE_RUN_FROM_PACKAGE`.  Otherwise, don't.

### DIFF
--- a/src/tree/FunctionAppProvider.ts
+++ b/src/tree/FunctionAppProvider.ts
@@ -86,6 +86,8 @@ export class FunctionAppProvider implements IChildProvider {
             createOptions.runtime = 'python';
         } else {
             createOptions.os = 'windows';
+            // need to check if a project is both a C# Script and running v1
+            // https://github.com/Microsoft/vscode-azurefunctions/issues/684
             if (language !== ProjectLanguage.CSharpScript || runtime !== ProjectRuntime.v1) {
                 // WEBSITE_RUN_FROM_PACKAGE has several benefits, so make that the default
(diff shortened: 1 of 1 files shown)


src/tree/FunctionAppProvider.ts
```diff
@@ -86,9 +86,11 @@ export class FunctionAppProvider implements IChildProvider {
             createOptions.runtime = 'python';
         } else {
             createOptions.os = 'windows';
-            // WEBSITE_RUN_FROM_PACKAGE has several benefits, so make that the default
-            // https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package
-            functionAppSettings.WEBSITE_RUN_FROM_PACKAGE = '1';
+            if (language !== ProjectLanguage.CSharpScript || runtime !== ProjectRuntime.v1) {
```
EricJizbaMSFT: Please add a comment or just link to the original issue - otherwise people won't know why CSharpScript is different

PR #698, opened 2018-10-08
Check for websitecontent setting prior to deploy

### LINKED ISSUE #625: Error when publishing a Python app on Linux Consumption
Environment: 
	• MacOS
	• Azure Functions extension 0.11.0 with azureFunctions.enablePython==true
	• Core Tools 2.0.2.0

Repro steps:

1. Create new Python project worked great
2. Create new HTTP triggered function worked great
3. F5 experience was perfect!
4. Publishing to a Linux consumption app in Azure failed terribly

     - On publish, I picked '+Create new app' and gave it the relevant details including app name, resource group name, storage account name, region ,etc. 
     - The app was created but portal threw the following errors. Doesn't look like the functions were indexed. 

image

It seems the the publish is setting the following app settings:

- WEBSITE_CONTENTAZUREFILECONNECTIONSTRING
- WEBSITE_CONTENTSHARE

These are not required for consumption and cause an error in the publishing experience.

nturinski:
Fixes #625 

Out of the 3 scenarios, I believe the only case we have to do anything is with Linux Consumption.


src/commands/deploy.ts
```diff
@@ -82,6 +82,22 @@ export async function deploy(this: IActionContext, target?: vscode.Uri | string
         throw new Error(localize('pythonNotAvailableOnWindows', 'Python projects are not supported on Windows Function apps.  Deploy to a Linux Consumption app.'));
     }
 
+    // we need this check due to this issue: https://github.com/Microsoft/vscode-azurefunctions/issues/625
```
EricJizbaMSFT: Can you move this to it's own function? The 'deploy' function is already too long

src/constants.ts
```diff
@@ -68,3 +68,6 @@ export enum ScmType {
 export const publishTaskId: string = 'publish';
 export const installExtensionsId: string = 'installExtensions';
 export const funcPackId: string = 'funcPack';
+
+export const WEBSITE_CONTENTAZUREFILECONNECTIONSTRING: string = 'WEBSITE_CONTENTAZUREFILECONNECTIONSTRING';
```
EricJizbaMSFT: Why are these in the constants file? I've mentioned this before - but I don't like throwing every constant in here unless it's actually used by multiple files or if it's a notable/important constant.

src/commands/deploy.ts
```diff
@@ -82,6 +82,22 @@ export async function deploy(this: IActionContext, target?: vscode.Uri | string
...
+    if (node.isLinuxPreview) {
+        const applicationSettings: StringDictionary = await client.listApplicationSettings();
+        if (applicationSettings.properties && (applicationSettings.properties[WEBSITE_CONTENTAZUREFILECONNECTIONSTRING] || applicationSettings.properties[WEBSITE_CONTENTSHARE])) {
```
EricJizbaMSFT: Add telemetry so we know how often this happens

src/commands/deploy.ts
```diff
@@ -82,6 +82,22 @@ export async function deploy(this: IActionContext, target?: vscode.Uri | string
...
+            await ext.ui.showWarningMessage(
+                localize('noDeployWithWebsiteContent', 'App settings "{0}" and "{1}" are required for Portal editing, but prevent deployment.  Delete these settings and deploy?', WEBSITE_CONTENTAZUREFILECONNECTIONSTRING, WEBSITE_CONTENTSHARE),
```
EricJizbaMSFT: Screenshot of the message? I'm guessing it looks pretty ugly... 😬

nturinski: screen shot 2018-10-11 at 12 59 20 pm

src/commands/deploy.ts
```diff
@@ -351,3 +351,28 @@ async function runPreDeployTask(deployFsPath: string, telemetryProperties: Telem
...
+
+async function verifyWebContentSettings(node: FunctionAppTreeItem, telemetryProperties: TelemetryProperties): Promise<void> {
+    if (node.isLinuxPreview) {
+        // we need this check due to this issue: https://github.com/Microsoft/vscode-azurefunctions/issues/625
+        const client: SiteClient = node.root.client;
+        const applicationSettings: StringDictionary = await client.listApplicationSettings();
+        const WEBSITE_CONTENTAZUREFILECONNECTIONSTRING: string = 'WEBSITE_CONTENTAZUREFILECONNECTIONSTRING';
+        const WEBSITE_CONTENTSHARE: string = 'WEBSITE_CONTENTSHARE';
+        if (applicationSettings.properties && (applicationSettings.properties[WEBSITE_CONTENTAZUREFILECONNECTIONSTRING] || applicationSettings.properties[WEBSITE_CONTENTSHARE])) {
+            telemetryProperties.webContentSettingsRemoved = 'false ';
```
EricJizbaMSFT: nit: 'false ' should be 'false' (get rid of extra space)

src/commands/deploy.ts
```diff
@@ -351,3 +351,28 @@ async function runPreDeployTask(deployFsPath: string, telemetryProperties: Telem
...
+                { modal: true },
+                DialogResponses.yes,
+                DialogResponses.cancel
+            );
+            delete applicationSettings.properties[WEBSITE_CONTENTAZUREFILECONNECTIONSTRING];
+            delete applicationSettings.properties[WEBSITE_CONTENTSHARE];
+            telemetryProperties.webContentSettingsRemoved = 'true';
+            await client.updateApplicationSettings(applicationSettings);
+            // if the user cancels the deployment, the app settings node doesn't reflect the deleted settings
+            await node.refresh();
```
EricJizbaMSFT: You can just call `node. _appSettingsTreeItem.refresh()` which would be faster since it doesn't have to refresh other stuff

(Of course you'd have to make that tree item public)

nturinski: I'm not super concerned about this-- it seems like something that won't come up very often.

PR #699, opened 2018-10-09
Exclude obj and bin folders when creating a non C# project

### LINKED ISSUE #658: Don't show C# related files when using other runtimes
Got a bunch of feedback around this at Ignite. Folks were confused to see the `extensions.csproj` file, the `/bin` and the `/obj` directories when developing functions using node or python. We should treat these files as the functions runtime implementation details and not surface them as part of the user files in VS Code. 

/cc @fabiocav

nturinski:
Fixes #658 

This is a temporary solution since ultimatey the func cli shouldn't have a dependency on dotnet.

If somebody knows how to ONLY get the workspace settings, please let me know.  Every time I get the projectConfiguration settings, it pulls down the default settings along with them (which includes all the constants that I defined).  I don't think there is any "harm" in just writing them in the workspace settings, but it looks messy.  That being said, it is safer to do that because it definitely will not alter any of the user's settings (in case they enabled `**/.git` to be included in their workspace settings, though this is a new project so that technically shouldn't happen).

### DIFF
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as fse from 'fs-extra';
-import { ProgressLocation, QuickPickItem, QuickPickOptions, window } from 'vscode';
+import { ProgressLocation, QuickPickItem, QuickPickOptions, Uri, window, workspace, WorkspaceConfiguration } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
-import { ProjectLanguage, projectLanguageSetting, ProjectRuntime } from '../../constants';
+import { DefaultFilesExcluded, ProjectLanguage, projectLanguageSetting, ProjectRuntime } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { validateFuncCoreToolsInstalled } from '../../funcCoreTools/validateFuncCoreToolsInstalled';
 import { localize } from '../../localize';
@@ -69,6 +69,10 @@ export async function createNewProject(
         await projectCreator.addNonVSCodeFiles();
 
         await initProjectForVSCode(actionContext, functionAppPath, language, runtime, projectCreator);
+        // Functions has a dependency on C# that creates build artifacts: https://github.com/Microsoft/vscode-azurefunctions/issues/658
+        if (language !== ProjectLanguage.CSharp) {
+            await addToFilesExclude(['bin', 'obj'], functionAppPath);
+        }
 
         if (await gitUtils.isGitInstalled(functionAppPath) && !await gitUtils.isInsideRepo(functionAppPath)) {
             await gitUtils.gitInit(ext.outputChannel, functionAppPath);
@@ -105,4 +109,26 @@ export function getProjectCreator(language: string, functionAppPath: string, act
         default:
             return new ScriptProjectCreatorBase(functionAppPath, actionContext.properties);
     }
+}
+
+async function addToFilesExclude(filesToExclude: string | string[], fsPath: string): Promise<void> {
+    const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration('files', Uri.file(fsPath));
+    // tslint:disable:no-any no-unsafe-any
+    let excludedFiles: any | undefined = projectConfiguration.get('exclude');
+    if (excludedFiles) {
+        for (const key of Object.keys(DefaultFilesExcluded)) {
+            delete excludedFiles[DefaultFilesExcluded[key]];
+        }
+    } else {
+        excludedFiles = {};
+    }
+
+    if (filesToExclude.length > 0) {
+        for (const file of filesToExclude) {
+            excludedFiles[file] = true;
+        }
+    } else {
+        excludedFiles.filesToExclude = true;
+    }
+    await projectConfiguration.update('exclude', excludedFiles, false);
 }
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -6,7 +6,7 @@
 import * as fse from 'fs-extra';
 import { ProgressLocation, QuickPickItem, QuickPickOptions, Uri, window, workspace, WorkspaceConfiguration } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
-import { DefaultFilesExcluded, ProjectLanguage, projectLanguageSetting, ProjectRuntime } from '../../constants';
+import { ProjectLanguage, projectLanguageSetting, ProjectRuntime } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { validateFuncCoreToolsInstalled } from '../../funcCoreTools/validateFuncCoreToolsInstalled';
 import { localize } from '../../localize';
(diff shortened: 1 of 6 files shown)


src/commands/createNewProject/createNewProject.ts
```diff
@@ -69,6 +69,10 @@ export async function createNewProject(
         await projectCreator.addNonVSCodeFiles();
 
         await initProjectForVSCode(actionContext, functionAppPath, language, runtime, projectCreator);
+        // Functions has a dependency on C# that creates build artifacts: https://github.com/Microsoft/vscode-azurefunctions/issues/658
+        if (language !== ProjectLanguage.CSharp) {
```
StephenWeatherford: I thought this was for C#, not for everything but C#...?

src/commands/createNewProject/createNewProject.ts
```diff
@@ -106,3 +110,25 @@ export function getProjectCreator(language: string, functionAppPath: string, act
             return new ScriptProjectCreatorBase(functionAppPath, actionContext.properties);
     }
 }
+
+async function addToFilesExclude(filesToExclude: string | string[], fsPath: string): Promise<void> {
+    const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration('files', Uri.file(fsPath));
+    // tslint:disable:no-any no-unsafe-any
+    let excludedFiles: any | undefined = projectConfiguration.get('exclude');
```
StephenWeatherford: Intellisense shows that vscode has 'files.exclude' typed to 'object' so you could use {}|undefined

src/commands/createNewProject/createNewProject.ts
```diff
@@ -106,3 +110,25 @@ export function getProjectCreator(language: string, functionAppPath: string, act
...
+    if (excludedFiles) {
+        for (const key of Object.keys(DefaultFilesExcluded)) {
```
StephenWeatherford: I don't understand why you're doing this.

src/commands/createNewProject/createNewProject.ts
```diff
@@ -106,3 +110,25 @@ export function getProjectCreator(language: string, functionAppPath: string, act
...
+    } else {
+        excludedFiles = {};
+    }
+
+    if (filesToExclude.length > 0) {
+        for (const file of filesToExclude) {
+            excludedFiles[file] = true;
+        }
+    } else {
+        excludedFiles.filesToExclude = true;
```
StephenWeatherford: Huh?

src/commands/createNewProject/createNewProject.ts
nturinski: We want to hide these folders for non-C# projects.  The func cli would create a `bin` and `obj` folder in Node and Python projects which was confusing users at Ignite.

src/commands/createNewProject/createNewProject.ts
nturinski: If I use the `{}` type, it will throw a type definition warning that `property` does not exist (obj or bin in this case)

src/commands/createNewProject/createNewProject.ts
nturinski: I wanted to make the function a bit more versatile than just adding `obj` and `bin` into the files.exclude property, but wanted to allow users the option to either input just one directory as a string or multiple directories as a string[].

src/commands/createNewProject/createNewProject.ts
nturinski: This is to handle the case where it is a string[] or a string.

StephenWeatherford:
Take a look at WorkspaceConfiguration.inspect.

src/commands/createNewProject/createNewProject.ts
StephenWeatherford: Then you should more strongly define it: {bin?: boolean, obj?: boolean}.

Also makes interpreting the code easier.

src/commands/createNewProject/createNewProject.ts
StephenWeatherford: No, I don't understand the code.

src/commands/createNewProject/createNewProject.ts
StephenWeatherford: You're setting it to:
  "files.exclude": {
    filesToExclude: true
  }

??

src/commands/createNewProject/createNewProject.ts
EricJizbaMSFT: I try not to have language-specific code in this file. Please move this into a language-specific projectCreator.

Also double check the languages when you move it into a project creator. This really only needs to happen for languages that explicitly have the 'func extensions install' task. For example - Java doesn't use that task and thus doesn't need this.

src/commands/createNewProject/createNewProject.ts
```diff
@@ -106,3 +110,20 @@ export function getProjectCreator(language: string, functionAppPath: string, act
             return new ScriptProjectCreatorBase(functionAppPath, actionContext.properties);
     }
 }
+
+async function addToFilesExclude(filesToExclude: string | string[], fsPath: string): Promise<void> {
+    const exclude: string = 'exclude';
+    const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration('files', Uri.file(fsPath));
```
EricJizbaMSFT: If you're editing settings, it should be done in the `writeVSCodeSettings` function in `initProjectForVSCode`:
https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createNewProject/initProjectForVSCode.ts#L86

If you're not familiar with 'init project', take a look at the "Create new project from func cli" test case in our OneNote.

nturinski:
One issue I thought of with this refactor:

1. Have a project opened that include `files.exclude` workspace settings
1. Create a new project (either JS or Python)
1. Browse to a new folder
1. When the project is created, it inherits the settings of the previous project.

src/commands/createNewProject/initProjectForVSCode.ts
```diff
@@ -98,6 +99,10 @@ async function writeVSCodeSettings(projectCreator: ProjectCreatorBase, vscodePat
                 data[`${extensionPrefix}.${preDeployTaskSetting}`] = projectCreator.preDeployTask;
             }
 
+            if (projectCreator.excludedFiles) {
+                data[filesExcludeSetting] = addToFilesExcludeSetting(projectCreator.excludedFiles, vscodePath);
```
EricJizbaMSFT: Don't pass in `vscodePath`. Pass in `data`, which has already been read from the correct settings.json file

src/commands/createNewProject/initProjectForVSCode.ts
```diff
@@ -124,6 +128,20 @@ async function writeExtensionRecommendations(projectCreator: ProjectCreatorBase,
     );
 }
 
+function addToFilesExcludeSetting(filesToExclude: string | string[], data: {}): { [key: string]: boolean } {
+    // tslint:disable-next-line:no-unsafe-any
```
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah because data is type `{}` so `data[filesExcludeSetting]` has type any. Technically the user could have invalid settings where `data[filesExcludeSetting]` was the wrong type, but I'm not too worried about it.

Issue #334, opened 2018-04-23
Refactor funcStagingPortalTemplateData tests


nturinski:
With the new CDN for the templates, there are currently no "staging" templates to test.  We should change the tests to retrieve the "pre-release v.x' templates and test against those if the Function team confirms that those tags will be used asthe equivalence of staging templates.

PR #718, opened 2018-10-12
Run tests against staging templates

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/334

### DIFF
--- a/test/createFunction/FunctionTesterBase.ts
+++ b/test/createFunction/FunctionTesterBase.ts
@@ -15,7 +15,7 @@ import { ext } from '../../src/extensionVariables';
 import { getGlobalFuncExtensionSetting, updateGlobalSetting } from '../../src/ProjectSettings';
 import { FunctionTemplates } from '../../src/templates/FunctionTemplates';
 import * as fsUtil from '../../src/utils/fs';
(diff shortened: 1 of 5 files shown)


test/createFunction/FunctionTesterBase.ts
```diff
@@ -62,6 +62,7 @@ export abstract class FunctionTesterBase implements vscode.Disposable {
     public async testCreateFunction(templateName: string, ...inputs: (string | undefined)[]): Promise<void> {
         await this.testCreateFunctionInternal(latestTemplates, this.funcPortalTestFolder, templateName, inputs.slice());
         await this.testCreateFunctionInternal(backupTemplates, this.backupTestFolder, templateName, inputs.slice());
+        await this.testCreateFunctionInternal(stagingTemplates, this.funcStagingPortalTestFolder, templateName, inputs.slice());
```
nturinski: nit: Do we need to call `inputs.slice()` on 3 separate occasions?

EricJizbaMSFT: `inputs.slice()` is an easy way to create a copy of the array. The tests will call `pop` on each array to make sure the inputs are actually used so we need three separate copies

Issue #640, opened 2018-09-22
VS Code should not add AzureWebJobsDashboard App Setting


davidebbo:
It's an obsolete setting that we're moving away from. Even though it doesn't technically break things, it can have a perf impact.

EricJizbaMSFT:
This one can be removed from everything? Linux and windows?

davidebbo:
It can be removed in all **v2** scenarios, regardless of platform. But it is still needed in v1 (not sure if you support v1 at all).

PR #731, opened 2018-10-17
Move app settings logic to this repo

EricJizbaMSFT:
Based on this PR in shared package: https://github.com/Microsoft/vscode-azuretools/pull/288

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/640


src/tree/FunctionAppProvider.ts
```diff
@@ -121,3 +116,59 @@ async function getDefaultRuntime(actionContext: IActionContext): Promise<Project
...
+        appSettings.push({
+            name: 'AzureWebJobsDashboard',
+            value: context.storageConnectionString
+        });
+    }
+
+    // These settings only apply for Windows https://github.com/Microsoft/vscode-azurefunctions/issues/625
+    if (context.os === 'windows') {
+        appSettings.push({
+            name: 'WEBSITE_CONTENTAZUREFILECONNECTIONSTRING',
```
nturinski: Huh, maybe time to make a constant for these.  I can think of somewhere else that they're used 👼

EricJizbaMSFT: Lol yes (after the release)

PR #732, opened 2018-10-17
Local git deploy path check

nturinski:


src/commands/deploy.ts
```diff
@@ -178,6 +182,11 @@ async function getDeployFsPath(): Promise<string> {
     return await workspaceUtil.selectWorkspaceFolder(ext.ui, workspaceMessage, (f: vscode.WorkspaceFolder) => getFuncExtensionSetting(deploySubpathSetting, f.uri.fsPath));
 }
 
+async function removeSubDeployFromFsPath(deployFsPath: string): Promise<string> {
```
EricJizbaMSFT: I'm not totally sold on this logic. Seems kinda hacky. Completely different approach: What if `ensureFolderIsOpen` returns the folder path and then you just use that folder path instead of deployFsPath for local git?

nturinski: Agreed.  I was looking for something just like this.

Should I rename the function to include AndReturn or do you think it'll be obvious enough?

src/utils/workspace.ts
```diff
@@ -154,4 +154,5 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
             throw new Error(message);
         }
     }
+    return fsPath;
```
EricJizbaMSFT: fsPath is the same thing passed in. You have to actually return `folder.fsPath` or whatever

nturinski: Oh shoot, of course, I got rid of the subDeploy setting earlier >< Thanks for catching that

nturinski:
Okay so I'm not sure about this fix and wanted your opinion on it--

Technically we only need to return if the folder is already opened.  Every other case we should throw an error after opening the folder in VS Code.  So I don't THINK it'll ever reach `uri.fsPath`, but it'll have the path if anything changes.

Part of the reason I did that was to not have to have it return <string|undefined>, since it shouldn't ever return undefined (since it SHOULD throw that error) but I'd totally understand about wanting to strictly define this.

src/utils/workspace.ts
```diff
@@ -147,13 +147,18 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
...
+            if (openBehavior === OpenBehavior.OpenInNewWindow) {
+                // extension host will not be restarted since a new instance of VS Code is opened, but we don't want deploy to continue
+                throw new UserCancelledError();
+            }
         }
 
         if (message) {
             // After we've opened the folder, throw the error message for the sake of telemetry
             actionContext.suppressErrorDisplay = true;
             throw new Error(message);
```
EricJizbaMSFT: I feel like your new logic is correct and this old logic that threw `new Error(message)` is incorrect. Right? Can we get rid of that?

nturinski: Well, that message would contain a message about requiring the user to open the folder to deploy which was sent to telemetry.  We'd be losing that for cases 1 and 2.

Issue #740, opened 2018-10-18
Init python project for VS Code fails when venv doesn't exist


EricJizbaMSFT:
Repro steps:
1. Create a python project from the cli with a venv _not_ named "func_env"
```
python -m venv env1
. env1/bin/activate
func init --worker-runtime python
func new --template HttpTrigger --name HttpTrigger1
```
(NOTE: The activate command is different on windows: `env1/Scripts/activate`)
1. Open VS Code
1. Say yes when you get this prompt:
![image](https://user-images.githubusercontent.com/11282622/47176180-c890ea80-d2c9-11e8-9910-30dbb3dfe8d4.png)

Expected: it succeeds:
Actual: 
```
Running command: "func_env\Scripts\activate && pip install ptvsd "...
The system cannot find the path specified.
Error: Failed to run "func_env\Scripts\activate && pip install ptvsd" command. Check output window for more details.
```

PR #742, opened 2018-10-18
Handle existing python virtual environments

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/740

We still default to 'func_env' for new projects - but this improves the logic when detecting projects created from the func cli as described in the above issue. I introduced a `azureFunctions.pythonVenv` setting that is referenced directly from the "tasks.json" file.

### DIFF
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -7,9 +7,9 @@ import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
 import * as semver from 'semver';
-import { MessageItem, window } from 'vscode';
-import { DialogResponses, parseError, UserCancelledError } from 'vscode-azureextensionui';
-import { funcPackId, gitignoreFileName, isWindows, localSettingsFileName, Platform, ProjectRuntime, TemplateFilter } from "../../constants";
+import { MessageItem, QuickPickItem, window } from 'vscode';
+import { IAzureQuickPickOptions, parseError, UserCancelledError } from 'vscode-azureextensionui';
+import { extensionPrefix, funcPackId, gitignoreFileName, isWindows, localSettingsFileName, Platform, ProjectRuntime, TemplateFilter } from "../../constants";
 import { ext } from '../../extensionVariables';
 import { funcHostCommand, funcHostTaskLabel } from "../../funcCoreTools/funcHostTask";
 import { validateFuncCoreToolsInstalled } from '../../funcCoreTools/validateFuncCoreToolsInstalled';
@@ -20,13 +20,14 @@ import * as fsUtil from '../../utils/fs';
 import { funcWatchProblemMatcher } from "./IProjectCreator";
 import { ScriptProjectCreatorBase } from './ScriptProjectCreatorBase';
 
-export const funcEnvName: string = 'func_env';
-
 export enum PythonAlias {
     python = 'python',
     python3 = 'python3',
     py = 'py'
 }
+
+export const pythonVenvSetting: string = 'pythonVenv';
+const fullPythonVenvSetting: string = `${extensionPrefix}.${pythonVenvSetting}`;
 
 const minPythonVersion: string = '3.6.0';
 const minPythonVersionLabel: string = '3.6.x'; // Use invalid semver as the label to make it more clear that any patch version is allowed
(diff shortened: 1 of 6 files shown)


src/commands/createNewProject/validateFunctionProjects.ts
```diff
@@ -152,19 +152,20 @@ async function promptToUpdateDebugConfiguration(fsPath: string): Promise<boolean
...
-        if (!await fse.pathExists(path.join(folderPath, funcEnvName))) {
+        const venvName: string | undefined = getFuncExtensionSetting(pythonVenvSetting, folderPath);
+        if (venvName && !await fse.pathExists(path.join(folderPath, venvName))) {
             actionContext.properties.pythonVenvExists = 'false';
 
             const settingKey: string = 'showPythonVenvWarning';
             if (getFuncExtensionSetting<boolean>(settingKey)) {
                 const createVenv: vscode.MessageItem = { title: localize('createVenv', 'Create virtual environment') };
-                const message: string = localize('uninitializedWarning', 'Failed to find Python virtual environment, which is required to debug and deploy your Azure Functions project.');
+                const message: string = localize('uninitializedWarning', 'Failed to find Python virtual environment "{0}", which is required to debug and deploy your Azure Functions project.', venvName);
```
nturinski: What will the venvName be if there is no pythonVenv setting?  Undefined?

src/commands/createNewProject/validateFunctionProjects.ts
```diff
@@ -152,19 +152,20 @@ async function promptToUpdateDebugConfiguration(fsPath: string): Promise<boolean
 
 async function verifyPythonVenv(projectLanguage: string | undefined, folderPath: string, actionContext: IActionContext): Promise<void> {
     if (projectLanguage === ProjectLanguage.Python) {
-        if (!await fse.pathExists(path.join(folderPath, funcEnvName))) {
+        const venvName: string | undefined = getFuncExtensionSetting(pythonVenvSetting, folderPath);
+        if (venvName && !await fse.pathExists(path.join(folderPath, venvName))) {
```
nturinski: Can't you just use `ensureVenv()`?

EricJizbaMSFT: Nah
1. That iterates through all possibilities - this is faster and only checks one
1. This logic has a message more specific to this scenario

src/commands/createNewProject/validateFunctionProjects.ts
EricJizbaMSFT: We won't get here if thats the case. `if (venvName && ` above

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -158,7 +157,7 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                 }
             }
 
-            esnureInGitIgnore(funcEnvName);
+            esnureInGitIgnore(venvName);
```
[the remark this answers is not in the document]
EricJizbaMSFT: Lol no probs me

src/commands/createNewProject/validateFunctionProjects.ts
nturinski: Oh right, sorry.  My actual question is do we not prompt the user to create a venv if there's no setting detected?

EricJizbaMSFT: Correct we do not prompt. Seems fair to me, but let me know if you disagree

nturinski: Well, when a user creates via the Func CLI, they definitely won't have the pythonVenv setting.  Since at that time, the user can select a `projectLanguage`, they will be able to hit this check.  

`ensureVenv()` is only invoked at project creation so the only way they would be able to get to the Quickpick of selecting a venv is if they created a new project in a folder that already has venvs, right?

It seems like this check wouldn't catch the case of creating via Func CLI then, it'd only help if they cloned a project that already has the proper workspace setting.

EricJizbaMSFT: If the user creates through func cli - they will hit this dialog:
image

Which creates tasks.json/settings.json/all that jazz including executing `ensureVenv`

`verifyPythonVenv` is only meant to handle the case where they clone a repo already configured for VS Code (but that doesn't have the venv since it's not checked in to git)

nturinski: Ah, I rest all further pressing the witness, your honor

PR #743, opened 2018-10-19
Add host attribute

nturinski:


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -113,7 +113,7 @@ export class PythonProjectCreator extends ScriptProjectCreatorBase {
                     },
                     options: {
                         env: {
-                            languageWorkers__python__arguments: '-m ptvsd --server --port 9091 --file'
+                            languageWorkers__python__arguments: '-m ptvsd --server --port 9091 --host localhost --file'
```
EricJizbaMSFT: The func cli seems to be using `127.0.0.1` instead of `localhost`. I'm not sure if there's any real difference, but I think it's safer to be consistent

PR #754, opened 2018-10-22
Fix C# project creation logic to handle 'netcoreapp'

EricJizbaMSFT:
Builds started failing over the weekend: https://travis-ci.org/Microsoft/vscode-azurefunctions/builds/444489354

Looks like C# projects used to default to 'netstandard2.0' and now they default to 'netcoreapp2.1'. I took a look at the list of possibilities: https://docs.microsoft.com/en-us/nuget/reference/target-frameworks and my new regex should handle the old value (netstandard) the new value (netcoreapp) and 'netcore' (not sure if that's actually possible, but it's fair to assume it would be ProjectRuntime.v2)

### DIFF
--- a/test/createNewProject.test.ts
+++ b/test/createNewProject.test.ts
@@ -66,7 +66,7 @@ suite('Create New Project Tests', async function (this: ISuiteCallbackContext): 
         await validateVSCodeProjectFiles(projectPath);
         const projectName: string = path.basename(projectPath);
         assert.equal(await fse.pathExists(path.join(projectPath, `${projectName}.csproj`)), true, 'csproj does not exist');
-        await validateSetting(projectPath, `${extensionPrefix}.${deploySubpathSetting}`, 'bin/Release/netstandard2.0/publish');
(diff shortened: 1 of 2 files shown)


test/createNewProject.test.ts
```diff
@@ -66,7 +66,7 @@ suite('Create New Project Tests', async function (this: ISuiteCallbackContext):
         await validateVSCodeProjectFiles(projectPath);
         const projectName: string = path.basename(projectPath);
         assert.equal(await fse.pathExists(path.join(projectPath, `${projectName}.csproj`)), true, 'csproj does not exist');
-        await validateSetting(projectPath, `${extensionPrefix}.${deploySubpathSetting}`, 'bin/Release/netstandard2.0/publish');
+        await validateSetting(projectPath, `${extensionPrefix}.${deploySubpathSetting}`, 'bin/Release/netcoreapp2.1/publish');
```
nturinski: Should we validate what version of .NET they're running before setting their deployPath?  This could a tricky problem for users to fix if they're unaware of the setting.

nturinski: Never mind, this is a test file.

PR #778, opened 2018-10-31
Finish up pipelines setup

EricJizbaMSFT:
Change badge on readme

Add "ComponentGovernance" task. Rather than manually register every third party tool we use, It automatically does it for us.


EricJizbaMSFT:
I cleaned up a few more things I noticed as I was setting up the first _other_ repo (app service).  Should be done with that cleanup now.

PR #779, opened 2018-11-03
"Add to workspace" with no folders opens in current window

### LINKED ISSUE #752: "Add to workspace" with no folders open should just "Open in current window"
Repro steps:
1. Close all folders in VS Code
1. Create a new project
1. Select "Add to workspace" when prompted

Expected:
Folder is open in VS Code like normal:
screen shot 2018-10-22 at 10 22 13 am

Actual:
Untitled "multi-root" workspace with only one folder:
screen shot 2018-10-22 at 10 21 41 am

alexweininger:
Fixes #752 

image
Create new project -> language -> **add to workspace**

Now it opens the project folder in the current window instead of adding it as a workspace folder in an untitled workspace.

image

`if (openFolders.length)` is what I used to check if no folders are open. Is this implicit coercion good practice? I didn't see any issues with using it since I think `openFolders` will always be an array based on its declaration:

```typescript
const openFolders: vscode.WorkspaceFolder[] = vscode.workspace.workspaceFolders || [];
```

Thanks!

### DIFF
--- a/src/utils/workspace.ts
+++ b/src/utils/workspace.ts
@@ -143,12 +143,13 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
         }
 
         const uri: vscode.Uri = vscode.Uri.file(fsPath);
+
+        if (!openFolders.length && openBehavior === OpenBehavior.AddToWorkspace) {
+            openBehavior = OpenBehavior.OpenInCurrentWindow;
+        }
+
         if (openBehavior === OpenBehavior.AddToWorkspace) {
-            if (openFolders.length) {
-                vscode.workspace.updateWorkspaceFolders(openFolders.length, 0, { uri: uri });
-            } else {
-                await vscode.commands.executeCommand('vscode.openFolder', uri);
-            }
+            vscode.workspace.updateWorkspaceFolders(openFolders.length, 0, { uri: uri });
         } else {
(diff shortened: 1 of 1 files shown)


src/utils/workspace.ts
```diff
@@ -144,7 +144,11 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
 
         const uri: vscode.Uri = vscode.Uri.file(fsPath);
         if (openBehavior === OpenBehavior.AddToWorkspace) {
-            vscode.workspace.updateWorkspaceFolders(openFolders.length, 0, { uri: uri });
+            if (openFolders.length) {
```
EricJizbaMSFT: Yeah `if (openFolders.length)` is fine

src/utils/workspace.ts
```diff
@@ -144,7 +144,11 @@ export async function ensureFolderIsOpen(fsPath: string, actionContext: IActionC
...
+                vscode.workspace.updateWorkspaceFolders(openFolders.length, 0, { uri: uri });
+            } else {
+                await vscode.commands.executeCommand('vscode.openFolder', uri);
```
EricJizbaMSFT: nit: We try to avoid copied code as much as possible. I suggest you change the value of openBehavior before this if/else so that it will automatically call the existing code to 'OpenInCurrentWindow'

PR #787, opened 2018-11-05
Fix template source logic for C# tests

EricJizbaMSFT:
So I realized that the C# tests weren't actually running tests based on the different template sources since they overwrite the nuget package in the same location. The key fix was to change the location from:
```
path.join('resources', 'dotnetTemplates')
```
to
```
path.join('resources', 'dotnetTemplates', ext.templateSource || '')
```

The rest of the changes are based on that and a little bit of refactor. There should be no functional changes.


src/templates/TemplateProvider.ts
```diff
@@ -105,30 +105,33 @@ export async function getFunctionTemplates(source?: 'backup' | 'cliFeed' | 'stag
...
                 this.properties.runtime = runtime;
                 this.properties.templateType = templateRetriever.templateType;
-                const useStagingTemplates: boolean = source === 'stagingCliFeed';
-                const templateVersion: string | undefined = await tryGetTemplateVersionSetting(this, cliFeedJson, runtime, useStagingTemplates);
+                const templateVersion: string | undefined = await tryGetTemplateVersionSetting(this, cliFeedJson, runtime);
                 let templates: IFunctionTemplate[] | undefined;
 
                 // 1. Use the cached templates if they match templateVersion
-                if (!source && ext.context.globalState.get(templateRetriever.getCacheKey(TemplateRetriever.templateVersionKey, runtime)) === templateVersion) {
+                // tslint:disable-next-line:strict-boolean-expressions
```
nturinski: Annoying that we have to disable this rule every check 😐

Issue #753, opened 2018-10-22
Use the python version active in the current python extension


qubitron:
The functions extension should use the Python version selected using the interpreter selector specified by the user.

Per:
https://code.visualstudio.com/docs/python/environments

Use the python.pythonPath setting to get the python interpreter currently selected in the extension.

EricJizbaMSFT:
We can definitely respect the "python.pythonPath" setting if it's at the "user" level.

However, it seems like that setting is usually created at the workspace level. That won't help us since we need to select the interpreter _before_ creating a python project (aka in an empty folder). We run `func init` to create the project and that command fails if it's not in a venv or if the version is wrong. @qubitron how do you suggest we leverage that setting?

brettcannon:
@EricJizbaMSFT so is the issue that `func init` provides no way to specify the interpreter and so you need to make sure the environment is already activated? Otherwise I'm not quite following what the issue of a workspace setting is as it sounds like it already needs to exist before `func init` runs.

brettcannon:
Right, I remember finding that a bit odd about the Functions CLI the last time I played with it because I believe it then created another directory to put the files into instead of doing it into the current directory as I expected.

So one way to get around this is if you simply send the command to the terminal then the Python extension -- if already running -- will activate any chosen virtual environment automatically in the terminal, so that could be taken care of for you.

PR #795, opened 2018-11-07
Improve python alias selection

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/753

~~Now users will always get a prompt like this~~:
screen shot 2018-11-07 at 10 16 39 am

~~Before they had no idea or control over which alias/version we used. This was especially problematic if we were defaulting to a 3.7 version of python which doesn't actually work with the func cli. (I still don't want to block them from using 3.7 because I don't like hard-coding a max version that could change in the future)~~

Edit: See updated comment with new experience

### DIFF
--- a/src/commands/createNewProject/PythonProjectCreator.ts
+++ b/src/commands/createNewProject/PythonProjectCreator.ts
@@ -16,6 +16,7 @@ import { funcHostCommand, funcHostTaskLabel } from "../../funcCoreTools/funcHost
 import { validateFuncCoreToolsInstalled } from '../../funcCoreTools/validateFuncCoreToolsInstalled';
 import { azureWebJobsStorageKey, getLocalSettings, ILocalAppSettings } from '../../LocalAppSettings';
 import { localize } from "../../localize";
+import { getGlobalFuncExtensionSetting } from '../../ProjectSettings';
 import { cpUtils } from "../../utils/cpUtils";
 import * as fsUtil from '../../utils/fs';
 import { funcWatchProblemMatcher } from "./ProjectCreatorBase";
@@ -25,6 +26,7 @@ export const pythonVenvSetting: string = 'pythonVenv';
 const fullPythonVenvSetting: string = `${extensionPrefix}.${pythonVenvSetting}`;
 
 const minPythonVersion: string = '3.6.0';
+const maxPythonVersion: string = '3.7.0';
 const minPythonVersionLabel: string = '3.6.x'; // Use invalid semver as the label to make it more clear that any patch version is allowed
 
 export class PythonProjectCreator extends ScriptProjectCreatorBase {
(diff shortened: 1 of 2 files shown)


src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -287,23 +285,23 @@ function getVenvActivateCommand(venvName: string, platform: NodeJS.Platform): st
 }
 
 async function getPythonAlias(): Promise<string> {
-    for (const key of Object.keys(PythonAlias)) {
-        const alias: PythonAlias = <PythonAlias>PythonAlias[key];
-        const errorMessage: string | undefined = await validatePythonAlias(alias);
+    const aliasesToTry: string[] = ['python3.6', 'py -3.6', 'python3', 'python', 'py'];
+    const globalPythonPathSetting: string | undefined = getGlobalFuncExtensionSetting('pythonPath', 'python');
+    if (globalPythonPathSetting) {
+        aliasesToTry.unshift(globalPythonPathSetting);
```
nturinski: Is this why you got rid of the PythonAlias enum?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -287,23 +285,23 @@ function getVenvActivateCommand(venvName: string, platform: NodeJS.Platform): st
...
-    const result: MessageItem | undefined = await window.showErrorMessage(pythonMsg, { modal: true }, enterPython);
-    if (!result) {
-        throw new UserCancelledError();
-    } else {
-        const placeHolder: string = localize('pyAliasPlaceholder', 'Enter the Python alias (if its in your PATH) or the full path to your Python executable.');
-        return await ext.ui.showInputBox({ placeHolder, validateInput: validatePythonAlias });
-    }
+    const prompt: string = localize('pyAliasPlaceholder', 'Enter the alias or full path of the Python "{0}" executable to use.', minPythonVersionLabel);
+    // Don't validate max version when prompting (because the Functions team will assumably support 3.7+ at some point and we don't want to block people from using that)
+    return await ext.ui.showInputBox({ prompt, validateInput: validatePythonAlias });
```
nturinski: Should we validate their alias before returning it?

src/commands/createNewProject/PythonProjectCreator.ts
```diff
@@ -8,29 +8,25 @@ import * as fse from 'fs-extra';
...
-    python = 'python',
-    python3 = 'python3',
-    py = 'py'
-}
-
 export const pythonVenvSetting: string = 'pythonVenv';
 const fullPythonVenvSetting: string = `${extensionPrefix}.${pythonVenvSetting}`;
 
 const minPythonVersion: string = '3.6.0';
+const maxPythonVersion: string = '3.7.0';
```
nturinski: Any worry about having to manually update the `maxPythonVersion`?

src/commands/createNewProject/PythonProjectCreator.ts
EricJizbaMSFT: Yeah I mean I also just find it easier to iterate through an array of strings rather than having to do the `Object.keys` thing.

src/commands/createNewProject/PythonProjectCreator.ts
EricJizbaMSFT: It already has `validateInput: validatePythonAlias`

src/commands/createNewProject/PythonProjectCreator.ts
EricJizbaMSFT: Yes. I very much dislike hard-coding a max version. But we only use it when silently picking the python alias. I intentionally am _not_ using it if we prompt the user for the python alias because I don't want to block people on this

Issue #552, opened 2018-09-03
Error exporting Remote Settings from VsCode


ScoutmanPt:
Action: azureFunctions.appSettings.download
Error type: SyntaxError
Error Message: Unexpected token / in JSON at position 284

Version: 0.10.0
OS: win32

PR #801, opened 2018-11-08
Allow user to overwrite app settings in case of download error

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/552

### DIFF
--- a/src/LocalAppSettings.ts
+++ b/src/LocalAppSettings.ts
@@ -5,7 +5,7 @@
 
 import * as fse from 'fs-extra';
 import * as vscode from 'vscode';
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
So it's the _local_ file that failed to parse, which could easily happen based on the user editing it. Here's a scenario I could see happening:
1. I really messed up my local.settings.json file somehow and I don't know how to fix it
1. I decide to just start from scratch by downloading my remote settings
1. I get a prompt saying "Failed to parse local settings. Overwrite?" and I say yes because I don't need my trash local settings anymore

PR #806, opened 2018-11-12
Update to latest UI package

EricJizbaMSFT:
There was a breaking change where we have to change the order of how we call `registerUIExtensionVariables`. Depends on https://github.com/Microsoft/vscode-azuretools/pull/313


EricJizbaMSFT:
ping @nturinski this one is kinda blocking functions work for me

PR #813, opened 2018-11-19
Add telemetry on whether func is installed

EricJizbaMSFT:
There has been talk of taking a harder dependency on the func cli. One reason I don't want to do that is because I think there's a solid group of people that still create/deploy without having the func cli on their machine. This adds telemetry to validate that assumption.

### DIFF
--- a/src/funcCoreTools/getLocalFuncCoreToolsVersion.ts
+++ b/src/funcCoreTools/getLocalFuncCoreToolsVersion.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as semver from 'semver';
+import { IActionContext } from 'vscode-azureextensionui';
 import { ext } from '../extensionVariables';
 import { cpUtils } from '../utils/cpUtils';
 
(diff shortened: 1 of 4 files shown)


src/funcCoreTools/getLocalFuncCoreToolsVersion.ts
```diff
@@ -26,3 +27,15 @@ export async function getLocalFuncCoreToolsVersion(): Promise<string | null> {
...
     }
 }
+
+export function addLocalFuncTelemetry(actionContext: IActionContext): void {
+    actionContext.properties.funcCliVersion = 'unknown';
+
+    // tslint:disable-next-line:no-floating-promises
+    getLocalFuncCoreToolsVersion().then((version: string) => {
+        // tslint:disable-next-line:strict-boolean-expressions
+        actionContext.properties.funcCliVersion = version || 'none';
```
StephenWeatherford: There's no guarantee that the property will get set in time to be included in telemetry...

EricJizbaMSFT: I don't think it'll happen very often, but that's why I set `funcCliVersion` to 'unknown' a few lines up (so that I could at least differentiate that case)

PR #837, opened 2018-11-29
Clean up generated tasks.json

EricJizbaMSFT:
There was some unnecessary code here. Specifically:
1. presentation.reveal already defaults to 'always'
1. CSharpScriptProjectCreator doesn't need to re-implement getTasksJson because it's the same as ScriptProjectCreatorBase
1. Update docs/debugConfig.md to reflect what the tasks should actually look like these days


nturinski:
It's not even time for spring cleaning!

PR #838, opened 2018-11-30
Add swap slot and create slot

EricJizbaMSFT:
Depends on https://github.com/Microsoft/vscode-azuretools/pull/346

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/831
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/832

### DIFF
--- a/package.json
+++ b/package.json
@@ -493,6 +493,14 @@
                 {
                     "command": "azureFunctions.debugFunctionAppOnAzure",
                     "when": "config.azureFunctions.enableRemoteDebugging == true"
+                },
+                {
+                    "command": "azureFunctions.createSlot",
+                    "when": "config.azureFunctions.enableSlots == true"
+                },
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -239,6 +241,16 @@
                 "command": "azureFunctions.installOrUpdateFuncCoreTools",
                 "title": "%azFunc.installOrUpdateFuncCoreTools%",
                 "category": "Azure Functions"
+            },
+            {
+                "command": "azureFunctions.createSlot",
+                "title": "%azFunc.createSlot%",
+                "category": "Azure Functions"
```
StephenWeatherford: Do you care that these will show in command palette even with enableSlots=false?

EricJizbaMSFT: Yeah I do. Fixed thanks

PR #842, opened 2018-11-30
Implement deployments node

nturinski:
Everything seems to be working alright, but ran into a couple of issues:
- Deployments Open in Portal doesn't work correctly.  Doesn't seem like Functions have an equivalent to "Deployment Center" (so I removed it from the context menu)
- When configuring a Function App to GitHub, the logs indicate that kudu is trying to build via run-from-zip
image


q
```diff
@@ -0,0 +1,3090 @@
+[33mcommit c32bc61c60fc34f113018879b52d42cae842b301[m[33m ([m[1;36mHEAD -> [m[1;32mnat/deployments[m[33m, [m[1;31morigin/nat/deployments[m[33m)[m
```
EricJizbaMSFT: ?

package-lock.json
```diff
@@ -18,7 +18,7 @@
         },
         "@types/events": {
             "version": "1.2.0",
-            "resolved": "http://registry.npmjs.org/@types/events/-/events-1.2.0.tgz",
```
EricJizbaMSFT: What version of npm are you on? I'm on `6.4.1` and it doesn't make these changes...

src/tree/SlotTreeItemBase.ts
```diff
@@ -3,7 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AppSettingsTreeItem, AppSettingTreeItem, deleteSite, ISiteTreeRoot, SiteClient } from 'vscode-azureappservice';
+// tslint:disable-next-line:no-submodule-imports
+import * as WebSiteModels from 'azure-arm-website/lib/models';
```
EricJizbaMSFT: nit: `import { WebSiteManagementModels } from 'azure-arm-website';` without disabling the lint rule

package-lock.json
EricJizbaMSFT: Well we're gonna have to figure this out on Monday otherwise it'll be super annoying if each of our machines changes this back and forth lol

nturinski: Well, we should be secure with our http protocol anyway 😜

PR #853, opened 2018-12-06
Update to gulp 4 and clean up pipelines

EricJizbaMSFT:
- Update to gulp 4 to get rid of some npm audit stuff
    - `gulp.task` is [deprecated](https://github.com/gulpjs/gulp/blob/master/docs/getting-started/3-creating-tasks.md), so switch to functions
    - Leverage `ts-node` so that we can use typescript for gulpfile
- Clean up pipelines
    - Use `common/test.yml` for all OS's. I think it's easier to read the steps if they're in the same file. Plus there's less copied code.
    - Always use Node 8. Again - this is just to keep our builds consistent (previously I think macOS was defaulting to Node 6)


EricJizbaMSFT:
@StephenWeatherford FYI I've been working on doing this in the docker repo as well. I think it will help a few things there

Issue #852, opened 2018-12-06
Issue publishing function


bpware:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: azureFunctions.deploy
Error type: Error
Error Message: Failed to deploy. Pre-deploy task "installExtensions" failed with exit code "127".

Version: 0.13.0
OS: darwin

EricJizbaMSFT:
@bpware are you just using Http and timer triggers? If so you can probably remove `installExtensions` as a preDeployTask from your settings.json or tasks.json. Let me know if you have any issues doing that.

If you have triggers of different types (blob, queue, etc.) you will need to check the output of `installExtensions` and fix any errors otherwise those won't work

Issue #862, opened 2018-12-14
Failed to deploy. Pre-deploy task "installExtensions" failed with exit code "1"


thunem:
Repro steps:
Trying to deploy function from vscode with azure funztion extension installed.

Action: azureFunctions.deploy
Error type: Error
Error Message: Failed to deploy. Pre-deploy task "installExtensions" failed with exit code "1".

Version: 0.13.0
OS: win32

I've tried to remove 
  "azureFunctions.preDeployTask": "installExtensions",
from settings.json without any luck

found one more in task.json .. that made it work :)

thunem:
Remove installextensions from settings.json AND task.json and it will deploy :)

Issue #859, opened 2018-12-12
Unable to deploy through VS Code - node.js


pink-cupcakes:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: azureFunctions.deploy
Error type: Error
Error Message: Failed to deploy. Pre-deploy task "installExtensions" failed with exit code "2".

Version: 0.13.0
OS: win32

EricJizbaMSFT:
@pink-cupcakes closing as duplicate of https://github.com/Microsoft/vscode-azurefunctions/issues/852. Please feel free to answer my questions in this comment https://github.com/Microsoft/vscode-azurefunctions/issues/852#issuecomment-446697967 so we can get more info on this situation

PR #864, opened 2018-12-14
Let people deploy even if pre-deploy task fails

EricJizbaMSFT:
Before we released 0.13.0, pre deploy tasks would silently fail. We added an error message, but that's blocking some people from deploying (enough for a patch release IMO and Matt agrees). Matt came up with the "Deploy Anyway" button (similar to how you can select "Debug Anyway" if you have build errors). Only 13% of people create triggers that need to run `func extensions install`, so it's unnecessary to block in most cases.

I also wanted to make it easier to remove a `preDeployTask`, so I got rid of the logic that was setting defaults. It will still warn the user, but now we respect if they set `preDeployTask` to nothing.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/852

Related https://github.com/Microsoft/vscode-azurefunctions/issues/859
Related https://github.com/Microsoft/vscode-azurefunctions/issues/862

screen shot 2018-12-14 at 2 09 50 pm

### DIFF
--- a/src/commands/deploy.ts
+++ b/src/commands/deploy.ts
@@ -12,7 +12,7 @@ import * as path from 'path';
 import * as vscode from 'vscode';
 import * as appservice from 'vscode-azureappservice';
 import { AzureTreeItem, DialogResponses, IActionContext, IAzureUserInput, TelemetryProperties, UserCancelledError } from 'vscode-azureextensionui';
-import { deploySubpathSetting, extensionPrefix, funcPackId, installExtensionsId, preDeployTaskSetting, ProjectLanguage, ProjectRuntime, publishTaskId, ScmType } from '../constants';
+import { deploySubpathSetting, extensionPrefix, funcPackId, installExtensionsId, preDeployTaskSetting, ProjectLanguage, ProjectRuntime, publishTaskId, ScmType, settingsFileName, vscodeFolderName } from '../constants';
 import { ArgumentError } from '../errors';
 import { ext } from '../extensionVariables';
 import { addLocalFuncTelemetry } from '../funcCoreTools/getLocalFuncCoreToolsVersion';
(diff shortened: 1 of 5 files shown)


src/commands/deploy.ts
```diff
@@ -287,78 +287,99 @@ async function runPreDeployTask(deployFsPath: string, telemetryProperties: Telem
...
+            const fullMessage: string = getFullPreDeployMessage(messageLines);
+            throw new Error(fullMessage);
         } else {
-            messageLines.push(localize('noPreDeployTaskWarning', 'WARNING: Did not find preDeploy task "{0}". The deployment will continue, but the selected folder may not reflect your latest changes.', taskName));
-            messageLines.push(localize('howToAddPreDeploy', 'In order to ensure that you always deploy your latest changes, add a preDeploy task with the following steps:'));
+            const recommendedTaskName: string | undefined = getRecommendedTaskName(language, runtime);
+            if (recommendedTaskName) {
+                messageLines.push(localize('noPreDeployTaskWarning', 'WARNING: Did not find recommended preDeploy task "{0}". The deployment will continue, but the selected folder may not reflect your latest changes.', recommendedTaskName));
+                messageLines.push(localize('howToAddPreDeploy', 'In order to ensure that you always deploy your latest changes, add a preDeploy task with the following steps:'));
+                const fullMessage: string = getFullPreDeployMessage(messageLines);
```
nturinski: Why have a function to add these lines when you push the previous two lines into the array above?  I don't see `getFullPreDeployMessage` called anywhere else and it seems like these will always be called together.

EricJizbaMSFT: `getFullPreDeployMessage` is also called a few lines up. Just search for the function name

jsandersrocks: This is good, however you must deploy the extensions for Blob Trigger and others or the runtime will fail.  Creating an azure function from scratch Azure Blob trigger and deploying per the document will result in failure   alias jsanders

EricJizbaMSFT: @jsandersrocks I don't quite understand what you're saying here. If you have an issue or would like a change in behavior please file a new issue and we can discuss further there: https://github.com/Microsoft/vscode-azurefunctions/issues/new

PR #908, opened 2019-01-10
Obfuscate app settings

EricJizbaMSFT:
Similar to https://github.com/Microsoft/vscode-azureappservice/pull/771. This was a breaking change in the app service package so I have to implement it before I can take any other changes in that package


package.nls.json
```diff
@@ -60,5 +60,6 @@
     "azFunc.disconnectRepo": "Disconnect from Repo...",
     "azFunc.createSlot": "Create Slot...",
     "azFunc.swapSlot": "Swap Slot...",
-    "azFunc.advancedCreationDescription": "Enables advanced creation of function apps, which bypasses several defaults and prompts instead."
+    "azFunc.advancedCreationDescription": "Enables advanced creation of function apps, which bypasses several defaults and prompts instead.",
```
StephenWeatherford: nit: "bypasses several defaults and prompts"... by then my mind assumes prompts is a noun and has to reread the sentence.  suggest: "restores several prompts instead of using defaults"??  There's probably something better.

src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -57,6 +58,11 @@ export async function downloadAppSettings(node?: AppSettingsTreeItem): Promise<v
         }
     });
 
-    const doc: vscode.TextDocument = await vscode.workspace.openTextDocument(localSettingsUri);
-    await vscode.window.showTextDocument(doc);
+    const result: string | undefined = await vscode.window.showInformationMessage(localize('downloadedSettings', `Successfully downloaded settings from "{0}".`, client.fullName), localize('openFile', 'Open File'));
+    if (!result) {
+        throw new UserCancelledError();
```
StephenWeatherford: Throwing cancel doesn't seem right - didn't it just succeed?

package.nls.json
EricJizbaMSFT: I see what you mean about the noun. Messed with the message a bit

Issue #935, opened 2019-01-25
Bump minimum version of vscode engine to 1.31.0


EricJizbaMSFT:
Due to issues listed in https://github.com/Microsoft/vscode-azurefunctions/issues/822

I don't want to do it yet because it causes issues with the tests

Issue #933, opened 2019-01-24
Fix preDeployTask name detection for provided task


EricJizbaMSFT:
Related to https://github.com/Microsoft/vscode-azurefunctions/issues/822

Since we're providing `func extensions install`, the task name isn't being detected correctly

Issue #932, opened 2019-01-24
Update `verifyDebugConfigIsValid` to work with new debug config


EricJizbaMSFT:
Related to https://github.com/Microsoft/vscode-azurefunctions/issues/822

`verifyDebugConfigIsValid` and this doc will no longer be entirely accurate:
https://github.com/Microsoft/vscode-azurefunctions/blob/master/docs/debugConfig.md

PR #936, opened 2019-01-25
Provide func tasks programmatically

### LINKED ISSUE #821: Refactor Python launch.json/tasks.json to leverage getRemoteLauncherCommand api
As of today we install `ptvsd` into the user's project and hard-code the launcher command. Ideally we should use the `getRemoteLauncherCommand` command provided here: https://github.com/Microsoft/vscode-python/issues/3121 ~~(not quite released yet)~~, which has the following benefits:
1. No need to install ptvsd
1. We automatically get updates to the launcher command if/when they happen

### LINKED ISSUE #822: Provide `func host start` as a task
...rather than hard-coding it in every user's project. It provides the following benefits:
1. We can update the task under the covers as necessary without having to change a user's project
1. We write less stuff to a user's `launch.json` and `tasks.json`

There's two issues in VS Code that affect this experience:
https://github.com/Microsoft/vscode/issues/54397
https://github.com/Microsoft/vscode/issues/63874

EricJizbaMSFT:
By providing func tasks programmatically, we get the following benefits:
1. We can call the python extension api to get the debugger command instead of installing ptvsd into a user's project. This protects us from any breaking changes in ptvsd
1. We can fix things under the covers without having to manually change a user's project
1. We don't have to write as much stuff to the user's launch.json/tasks.json

I'm somewhat hesitant to merge this PR to master because it relies on a few fixes only in insiders, but I think it's worth it to give it some bake time since it affects a high priority scenario (debug) within our extension.

Fixes #822 
Fixes #821

I've also filed a few issues for remaining work: #932 #933 #935

### DIFF
--- a/src/debug/FuncTaskProvider.ts
+++ b/src/debug/FuncTaskProvider.ts
@@ -3,9 +3,9 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { CancellationToken, ProcessExecution, ShellExecution, Task, TaskProvider, workspace, WorkspaceFolder } from 'vscode';
+import { CancellationToken, ShellExecution, Task, TaskProvider, workspace, WorkspaceFolder } from 'vscode';
 import { isFunctionProject } from '../commands/createNewProject/isFunctionProject';
-import { extInstallCommand, func, funcHostStartCommand, funcWatchProblemMatcher, hostStartCommand, ProjectLanguage, projectLanguageSetting } from '../constants';
+import { extInstallCommand, func, funcExtInstallCommand, funcHostStartCommand, funcWatchProblemMatcher, hostStartCommand, ProjectLanguage, projectLanguageSetting } from '../constants';
 import { getFuncExtensionSetting } from '../ProjectSettings';
 import { FuncDebugProviderBase } from './FuncDebugProviderBase';
 import { JavaDebugProvider } from './JavaDebugProvider';
@@ -86,6 +86,6 @@ function getExtensionInstallTask(folder: WorkspaceFolder): Task {
         folder,
         extInstallCommand,
         func,
-        new ProcessExecution(func, ['extensions', 'install'])
+        new ShellExecution(funcExtInstallCommand)
     );
 }
--- a/src/debug/FuncTaskProvider.ts
+++ b/src/debug/FuncTaskProvider.ts
@@ -1,0 +1,91 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { CancellationToken, ProcessExecution, ShellExecution, Task, TaskProvider, workspace, WorkspaceFolder } from 'vscode';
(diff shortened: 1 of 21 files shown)


EricJizbaMSFT:
ping

src/debug/FuncTaskProvider.ts
```diff
@@ -0,0 +1,91 @@
...
+import { JavaDebugProvider } from './JavaDebugProvider';
+import { NodeDebugProvider } from './NodeDebugProvider';
+import { PythonDebugProvider } from './PythonDebugProvider';
+
+export class FuncTaskProvider implements TaskProvider {
+    private readonly _nodeDebugProvider: NodeDebugProvider;
+    private readonly _pythonDebugProvider: PythonDebugProvider;
+    private readonly _javaDebugProvider: JavaDebugProvider;
+
+    constructor(nodeDebugProvider: NodeDebugProvider, pythonDebugProvider: PythonDebugProvider, javaDebugProvider: JavaDebugProvider) {
```
nturinski: Why have this be part of the constructor rather than just export it from the respective files?  It doesn't seem like there are any dynamic properties in the constructors for any of the DebugProviders.

src/debug/PythonDebugProvider.ts
```diff
@@ -0,0 +1,56 @@
...
+    protected readonly debugConfig: DebugConfiguration = pythonDebugConfig;
+
+    public async getShellExecution(folder: WorkspaceFolder): Promise<ShellExecution> {
+        let command: string = funcHostStartCommand;
+        const port: number = this.getDebugPort(folder);
+        const venvName: string | undefined = getFuncExtensionSetting<string>('pythonVenv', folder.uri.fsPath);
+        if (venvName) {
+            command = convertToVenvCommand(venvName, process.platform, command);
+        }
+        const options: ShellExecutionOptions = { env: { languageWorkers__python__arguments: await getPythonCommand(localhost, port) } };
```
nturinski: Was the Python extension supposed to handle activating the virtual environment as well?  I thought that was part of the reason we wanted them to provide an API.

src/extension.ts
```diff
@@ -129,6 +134,15 @@ export async function activate(context: vscode.ExtensionContext): Promise<AzureE
...
         registerFuncHostTaskEvents();
+
+        const nodeDebugProvider: NodeDebugProvider = new NodeDebugProvider();
+        const pythonDebugProvider: PythonDebugProvider = new PythonDebugProvider();
+        const javaDebugProvider: JavaDebugProvider = new JavaDebugProvider();
+        // These don't actually overwrite "node", "python", etc. - they just add to it
+        context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('node', nodeDebugProvider));
+        context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('python', pythonDebugProvider));
+        context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('java', javaDebugProvider));
+        context.subscriptions.push(vscode.workspace.registerTaskProvider(func, new FuncTaskProvider(nodeDebugProvider, pythonDebugProvider, javaDebugProvider)));
```
nturinski: Is the C# launch too simple to justify creating a FuncTaskProvider for it?

EricJizbaMSFT:
Discussed offline. Might add more stuff for the C# scenario in the future, but not needed right now.

Issue #780, opened 2018-11-04
"disabled" is no longer valid for v2 function.json


jeffhollan:
Issue Type: <b>Bug</b>

For v2 templates, having "disabled" in the function.json is no longer the way to disable. Disabling requires setting an environment variable as [documented here](https://docs.microsoft.com/en-us/azure/azure-functions/disable-function#functions-2x---all-languages)

Extension version: 0.12.0
VS Code version: Code - Insiders 1.29.0-insider (d43f6cb9947c57a3254b556e69aeacc2b2025184, 2018-11-02T15:50:03.156Z)
OS version: Darwin x64 18.2.0

PR #955, opened 2019-02-04
Update disabled state logic for v2

EricJizbaMSFT:
v1 and v2 of the functions runtime disable functions in different ways. Update logic to reflect that. And clean up code a bit - I think this was originally written before `refreshImpl` or `description` existed

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/780

### DIFF
--- a/src/tree/FunctionTreeItem.ts
+++ b/src/tree/FunctionTreeItem.ts
@@ -8,10 +8,12 @@ import { URL } from 'url';
 import { ProgressLocation, window } from 'vscode';
 import { functionsAdminRequest, ISiteTreeRoot } from 'vscode-azureappservice';
 import { AzureParentTreeItem, AzureTreeItem, DialogResponses } from 'vscode-azureextensionui';
+import { ProjectRuntime } from '../constants';
 import { ArgumentError } from '../errors';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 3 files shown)


src/tree/FunctionTreeItem.ts
```diff
@@ -107,6 +116,40 @@ export class FunctionTreeItem extends AzureTreeItem<ISiteTreeRoot> {
...
+        if (projectRuntime === ProjectRuntime.v1) {
+            this._disabled = this.config.disabled;
+        } else {
+            const key: string = `AzureWebJobs.${this._name}.Disabled`;
+            /**
+             * The docs only officially mentioned 'true' and 'false', but here is what I found:
+             * The following resulted in a disabled function:
+             * true, tRue, 1
+             * The following resulted in an enabled function:
+             * false, fAlse, 0, fdsaf, 2, undefined
```
nturinski: Did somebody have a stroke while typing this?

EricJizbaMSFT: Lol what problem do you have with it?

EricJizbaMSFT: :) Exactly gotta test some edge cases - specifically a random word other than 'true' or 'false'

PR #957, opened 2019-02-04
Prompt user to install func cli before debugging

### LINKED ISSUE #358: Prompt for Function Core Tools installation before debugging
We only have a fix for C#, but ideally we should be able to do this for every language

EricJizbaMSFT:
C# was already handled in `pickFuncProcess`, this leverages `FuncDebugProviderBase` to add the logic to other languages. Also removes prompt after creating project

Fixes #358


EricJizbaMSFT:
Correct

PR #965, opened 2019-02-06
Remove black box from Java deploy and use common settings

EricJizbaMSFT:
Java projects have unique logic to run a preDeployTask and pick the folder to deploy which is pretty black-box-y. Instead, it should use the `preDeployTask` and `deploySubpath` settings. These settings let the user customize their project as they see fit and makes the experience consistent across all languages. Since this could break people, I added a [wiki](https://github.com/Microsoft/vscode-azurefunctions/wiki/Breaking-Changes-to-Java-Deploy-Configuration) and notification to update their project (similar to other breaking changes in the past).

The current release seems like the best time to do this since I already refactored how the tasks/launch works in [this PR](https://github.com/Microsoft/vscode-azurefunctions/pull/936) and we might as well change it all up at once. Plus it'd be nice to get this done before Java goes GA in the platform (no idea when that is).

### DIFF
--- a/src/commands/createNewProject/validateFunctionProjects.ts
+++ b/src/commands/createNewProject/validateFunctionProjects.ts
@@ -110,6 +110,7 @@ async function verifyJSDebugConfigIsValid(projectLanguage: string | undefined, f
                 const message: string = localize('uninitializedWarning', 'Your debug configuration is out of date and may not work with the latest version of the Azure Functions Core Tools.');
                 const learnMoreLink: string = 'https://aka.ms/AA1vrxa';
                 if (await promptToUpdateProject(folderPath, settingKey, message, learnMoreLink)) {
+                    actionContext.suppressErrorDisplay = false;
                     await initProjectForVSCode(actionContext, folderPath, projectLanguage);
                     actionContext.properties.updatedDebugConfig = 'true';
                 }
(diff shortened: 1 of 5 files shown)


src/commands/createNewProject/validateFunctionProjects.ts
```diff
@@ -79,7 +80,7 @@ function isInitializedProject(folderPath: string): boolean {
  * JavaScript debugging in the func cli had breaking changes in v2.0.1-beta.30 (~6/2018). This verifies users are up-to-date with the latest working debug config.
  * See https://aka.ms/AA1vrxa for more info
  */
-async function verifyDebugConfigIsValid(projectLanguage: string | undefined, folderPath: string, actionContext: IActionContext): Promise<void> {
+async function verifyJSDebugConfigIsValid(projectLanguage: string | undefined, folderPath: string, actionContext: IActionContext): Promise<void> {
```
nturinski: I see why you're doing this, but it feels kind of messy to me to verify that the projectLanguage is the respective language when the functions are named for each language.  Why not just have the switch case within a general `verifyDebugConfigIsValid` function?

nturinski: On second thought, I guess there's a lot of custom logic.

PR #973, opened 2019-02-07
Use platform specific "path.join" for pip install command

EricJizbaMSFT:
This is what the task looks like after my change:
```
{
  "label": "pipInstall",
  "type": "shell",
  "osx": {
    "command": "${config:azureFunctions.pythonVenv}/bin/python -m pip install -r requirements.txt"
  },
  "windows": {
    "command": "${config:azureFunctions.pythonVenv}\\Scripts\\python -m pip install -r requirements.txt"
  },
  "linux": {
    "command": "${config:azureFunctions.pythonVenv}/bin/python -m pip install -r requirements.txt"
  },
  "problemMatcher": []
}
```
Previously, the path separator would work on the _original_ OS, but not if someone cloned their repo to a different OS.


EricJizbaMSFT:
Turns out there's a little more to this. I'll re-open when I finish that

PR #978, opened 2019-02-08
Fix and consolidate venvUtils for all terminals/OS's

### LINKED ISSUE #814: Don't install pylint by default in new functions environments
I noticed that pylint is installed as part of initializing your new environment, this was done as a fix to #631, but this isn't the right solution to this problem as we are also planning to move away from pylint when the language server becomes the default.

EricJizbaMSFT:
I ran into two bugs while testing the release candidate:
1. The tasks.json we generate is meant to work if someone pushes their project to GitHub and clones it on a different OS. That wasn't the case for our "pipInstall" task because it always used the path separator from the _original_ OS
1. Windows users with their terminal set to bash would fail to debug

In order to fix, I ended up doing quite a bit of refactoring of venvUtils and added several tests to cover all the weird edge cases. Here's the list of terminals: https://code.visualstudio.com/docs/editor/integrated-terminal#_configuration

I also removed the logic to install "pylint" by default. It's just another command that could fail and it slows down project creation. We originally did it because the Python extension was super annoying and prompted you to install pylint every time, but now they at least have a "Don't show again" button. Fixes #814

### DIFF
--- a/src/utils/venvUtils.ts
+++ b/src/utils/venvUtils.ts
@@ -3,57 +3,106 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as fse from 'fs-extra';
 import * as path from 'path';
-import { workspace, WorkspaceConfiguration, WorkspaceFolder } from 'vscode';
-import { Platform } from "../constants";
+import { workspace, WorkspaceConfiguration } from 'vscode';
+import { Platform, pythonVenvSetting } from "../constants";
 import { ext } from '../extensionVariables';
 import { getFuncExtensionSetting } from '../ProjectSettings';
 import { cpUtils } from './cpUtils';
 
 export namespace venvUtils {
-    const bashAndCmdSeparator: string = ' && ';
-    const powerShellSeparator: string = ' ; ';
-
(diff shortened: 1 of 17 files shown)


src/utils/venvUtils.ts
```diff
@@ -3,57 +3,106 @@
...
-    export function convertToVenvTask(folder: WorkspaceFolder, ...commands: string[]): string {
-        const venvName: string | undefined = getFuncExtensionSetting<string>('pythonVenv', folder.uri.fsPath);
+    export function convertToVenvCommand(command: string, folderPath: string, platform: NodeJS.Platform = process.platform): string {
+        const terminal: Terminal = getTerminal(platform);
+        const venvName: string | undefined = getFuncExtensionSetting<string>(pythonVenvSetting, folderPath);
         if (venvName) {
-            commands.unshift(getVenvActivateCommand(venvName));
+            return joinCommands(terminal, getVenvActivateCommand(venvName, terminal, platform), command);
+        } else {
+            return command;
```
nturinski: Should you silently return the command without the venv activation?  Won't this break users?

EricJizbaMSFT: It's not our requirement that they use a venv, it's the func cli's requirement. I'm not sure if they'll always have that requirement so I don't want to hard code an error into our code.

Issue #951, opened 2019-02-02
'Open in portal' doesn't work on Linux


v-dantch:
**OS:** Linux
**Node.js:** V 10.15.0
**Build Version**:[20190201.1](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=2674)

**Repro Steps:**
Right click one function app --> Select 'Open in Portal'.

**Expect:**
'Open in Portal' works well.

**Actual:**
'Open in portal' doesn't work.

**More Info:**
This issue only reproduces on Linux.

PR #979, opened 2019-02-08
Update to dev 0.1.7 to get opn/clipboardy fix

StephenWeatherford:
Depends on https://github.com/Microsoft/vscode-azuretools/pull/403

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/951

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -642,6 +642,12 @@
             "integrity": "sha1-p5SvDAWrF1KEbudTofIRoFugxE8=",
             "dev": true
         },
+        "array-filter": {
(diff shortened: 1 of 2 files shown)


StephenWeatherford:
> So the CopyWebpackPlugin problem only affected executables, right? Just making sure this isn't broken, too:
> https://github.com/Microsoft/vscode-azurefunctions/blob/master/webpack.config.js#L36

Yep, we can leave those as they are.

EricJizbaMSFT:
Thanks!

PR #993, opened 2019-02-14
add function app actions

v-wuzhai:
This test case contains 5 actions like below:

- Create a function app: createfunctionapp
- Stop a function app: stopfunctionapp
- Start a function app: startfunctionapp 
- Restart a function app: restartfunctionapp
- Delete a function app: deletefunctionapp

### DIFF
--- a/package.json
+++ b/package.json
@@ -734,6 +734,11 @@
                         "type": "string",
                         "description": "%azFunc.deploySubpathDescription%"
                     },
+                    "azureFunctions.projectSubpath": {
+                        "scope": "resource",
+                        "type": "string",
+                        "description": "%azFunc.projectSubpathDescription%"
+                    },
                     "azureFunctions.showCoreToolsWarning": {
                         "type": "boolean",
                         "description": "%azFunc.showCoreToolsWarningDescription%",
--- a/package.nls.json
+++ b/package.nls.json
@@ -31,6 +31,7 @@
     "azFunc.appSettings.toggleSlotSetting": "Toggle as Slot Setting",
     "azFunc.pickProcess": "Pick Process",
     "azFunc.deploySubpathDescription": "The default subpath of a workspace folder to use when deploying. If set, you will not be prompted for the folder path when deploying.",
+    "azFunc.projectSubpathDescription": "The default subpath of a workspace folder to use for project operations. This is only necessary if you have multiple projects in one workspace. See https://aka.ms/AA4nmfy for more information.",
     "azFunc.showCoreToolsWarningDescription": "Show a warning if your installed version of Azure Functions Core Tools is out-of-date.",
     "azFunc.showMultiCoreToolsWarningDescription": "Show a warning if multiple installs of Azure Functions Core Tools are detected.",
     "azFunc.show64BitWarningDescription": "Show a warning to install a 64-bit version of the Azure Functions Core Tools when you create a .NET Framework project.",
--- a/src/commands/createNewProject/verifyIsProject.ts
+++ b/src/commands/createNewProject/verifyIsProject.ts
@@ -5,11 +5,15 @@
 
 import * as fse from 'fs-extra';
 import * as path from 'path';
-import { DialogResponses, IActionContext } from 'vscode-azureextensionui';
+import { MessageItem } from 'vscode';
+import { DialogResponses, IActionContext, IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { hostFileName } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
+import { getFuncExtensionSetting, updateWorkspaceSetting } from '../../ProjectSettings';
 import { createNewProject } from './createNewProject';
+
+const projectSubpathKey: string = 'projectSubpath';
 
 // Use 'host.json' as an indicator that this is a functions project
 export async function isFunctionProject(folderPath: string): Promise<boolean> {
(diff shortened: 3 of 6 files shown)


test/functionapp.test.ts
```diff
@@ -0,0 +1,96 @@
...
+                console.log(`Resource group "${resourceGroup}" deleted.`);
+            } else {
+                // If the test failed, the resource group might not actually exist
+                console.log(`Ignoring resource group "${resourceGroup}" because it does not exist.`);
+            }
+        }
+    });
+    test('createFunctionApp', async function (this: IHookCallbackContext): Promise<void> {
+        this.timeout(5 * 60 * 1000);
+        resourceName = fsUtil.getRandomHexString();
```
nturinski: You need to do `getRandomHexString().toLowerCase();` because storage accounts cannot contain upper case chars

test/functionapp.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        resourceGroupsToDelete.push(resourceName);
+        const testInputs: string[] = [resourceName, '$(plus) Create new resource group', resourceName, '$(plus) Create new storage account', resourceName, 'East US'];
+        ext.ui = new TestUserInput(testInputs);
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp', testAccount.getSubscriptionId());
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
+        assert.ok(createdApp);
+    });
+    test('stopFunctionApp', async function (this: IHookCallbackContext): Promise<void> {
+        this.timeout(5 * 60 * 1000);
```
nturinski: I don't think you need a timeout for each individual test case.

test/functionapp.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        createdApp1 = await client.webApps.get(resourceName, resourceName);
+        assert.equal(createdApp1.state, 'Stopped', 'Function app State is incorrect.');
+        //'Restart' is disabled after stopping a function app in Azure portal.
+        ext.ui = new TestUserInput([resourceName, resourceName]);
+        await vscode.commands.executeCommand('azureFunctions.restartFunctionApp');
+        client = getWebsiteManagementClient(testAccount);
+        createdApp1 = await client.webApps.get(resourceName, resourceName);
+        assert.equal(createdApp1.state, 'Running', 'Function app State is incorrect.');
+    });
+    test('deleteFunctionApp', async function (this: IHookCallbackContext): Promise<void> {
```
nturinski: It seems like you kind of duplicated our test case already implemented here: 
https://github.com/Microsoft/vscode-azurefunctions/blob/master/test/createAzureResource.test.ts#L56-L71

Since this is a refactor, could you delete the old test?  We don't need to create/delete a resource twice, but I understand that you needed to create a resource to run the start/stop/restart commands on.

test/createAzureResource.test.ts
```diff
@@ -53,23 +53,6 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
-        resourceGroupsToDelete.push(resourceName);
-
-        const testInputs: string[] = [resourceName, '$(plus) Create new resource group', resourceName, '$(plus) Create new storage account', resourceName, 'West US'];
-        ext.ui = new TestUserInput(testInputs);
-        await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
-        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
-        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
-        assert.ok(createdApp);
-
-        ext.ui = new TestUserInput([resourceName, DialogResponses.deleteResponse.title, DialogResponses.yes.title]);
```
nturinski: Need to get rid of the reference to DialogResponse at the top.  It causes a build error.

test/functionapp.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    test('createFunctionApp', async () => {
+        resourceGroupsToDelete.push(resourceName);
+        const testInputs: string[] = [resourceName, '$(plus) Create new resource group', resourceName, '$(plus) Create new storage account', resourceName, 'East US'];
+        ext.ui = new TestUserInput(testInputs);
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp', testAccount.getSubscriptionId());
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
+        assert.ok(createdApp);
+    });
+    test('stopFunctionApp', async () => {
```
nturinski: nit: Can you have a break (enter/return) between tests?  It makes the readability much better :)

test/functionapp.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        ext.ui = new TestUserInput([resourceName]);
+        await vscode.commands.executeCommand('azureFunctions.stopFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp1: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
+        assert.equal(createdApp1.state, 'Stopped', 'Function app State is incorrect.');
```
nturinski: Could you include what the expected app state is?  Also, "App" should be capitalized and "state" should not be capitalized.

test/functionapp.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp', testAccount.getSubscriptionId());
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
+        assert.ok(createdApp);
+    });
+    test('stopFunctionApp', async () => {
+        ext.ui = new TestUserInput([resourceName]);
+        await vscode.commands.executeCommand('azureFunctions.stopFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp1: WebSiteManagementModels.Site = await client.webApps.get(resourceName, resourceName);
```
nturinski: Just call it createdApp (this applies to all of the tests.)  They aren't being numbered anyway and they're all within their own scope, so it's okay to reuse the variable name :)

PR #1051, opened 2019-03-03
Fix typo in Auth Token binding resource string

markekraus:
`you` -> `your`


EricJizbaMSFT:
Thanks for the contribution @markekraus, but we don't manually create or edit these files. They're generated elsewhere and we just copy/paste them here. Your fix would be overwritten the next time we do that. Plus these are only our "backup" templates. Most of the time we pull the latest and greatest templates directly to a user's machine.

I believe this is the ultimate source repo for the strings (not 100% sure) and I would suggest filing an issue/PR there: https://github.com/Azure/azure-functions-ux/blob/dev/server/Resources/Resources.resx

markekraus:
@EricJizbaMSFT ah.. I will try that repo. Thanks for pointing me in the right direction!

EricJizbaMSFT:
Closing this as it should be handled in other repo first

Issue #986, opened 2019-02-11
Show better progress when creating a function


EricJizbaMSFT:
There were a few places where nothing seemed to be happening and I wasn't sure what was going on.

This one reproed for me on windows, but not every time:
1. Create js project
1. Add blob trigger

There was a delay before "index.js" popped up of around 10ish seconds? Idk seems like we can just show progress while creating the function

This one only repros for me on Java, but wouldn't hurt other languages:
1. Create a project
1. Select add function
1. Select workspace folder

Before the quick pick with functions shows up, there was a notable delay. We should do that thing where the quick pick shows progress

EricJizbaMSFT:
Also if the user creates a storage account or Azure resource during this flow, we don't handle that well.

PR #1067, opened 2019-03-06
Show better progress when creating a function

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/986

### DIFF
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -6,7 +6,7 @@
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import { isString } from 'util';
-import { InputBoxOptions, MessageItem, QuickPickItem, Uri, window, workspace } from 'vscode';
+import { InputBoxOptions, MessageItem, ProgressLocation, QuickPickItem, Uri, window, workspace } from 'vscode';
 import { callWithTelemetryAndErrorHandling, DialogResponses, IActionContext, IAzureQuickPickItem, TelemetryProperties } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -203,17 +206,18 @@ async function promptForTemplate(functionAppPath: string, language: ProjectLangu
...
-        picks = picks.concat([
-            { label: localize('selectRuntime', '$(gear) Change project runtime'), description: localize('currentRuntime', 'Current: {0}', runtime), data: runtimePickId, suppressPersistence: true },
-            { label: localize('selectLanguage', '$(gear) Change project language'), description: localize('currentLanguage', 'Current: {0}', language), data: languagePickId, suppressPersistence: true },
-            { label: localize('selectFilter', '$(gear) Change template filter'), description: localize('currentFilter', 'Current: {0}', templateFilter), data: filterPickId, suppressPersistence: true }
-        ]);
-
-        const placeHolder: string = templates.length > 0 ? localize('azFunc.selectFuncTemplate', 'Select a function template') : localize('azFunc.noTemplatesFound', 'No templates found. Change your settings to view more templates');
-        const result: IFunctionTemplate | string = (await ext.ui.showQuickPick(picks, { placeHolder })).data;
+        const picksTask: Promise<IAzureQuickPickItem<IFunctionTemplate | string>[]> = ext.templateProviderTask.then(async templateProvider => {
```
nturinski: Why `.then` this when there's a `while` loop wrapped around this rather than just using an `await`?

EricJizbaMSFT: If I did await here, the quick pick wouldn't have the little progress bar. We need to pass in the full promise to `showQuickPick` without awaiting

PR #1069, opened 2019-03-06
Use vscode.open instead of opn

EricJizbaMSFT:
Related to https://github.com/Microsoft/vscode-azuretools/issues/406

Tested on all 3 OS's.

The promise returned by 'opn' often didn't resolve until after the browser was opened, which is why we didn't wait a bunch of the time. With the VS Code command, it's fine to await it.


src/utils/openInBrowser.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+
+export async function openInBrowser(url: string): Promise<void> {
```
StephenWeatherford: Consider adding an assert that the URL starts with https

EricJizbaMSFT: Are you worried about http vs https? If so, we already have a tslint rule "no-http-string"

Or are you worried about non-http/https? If so, I'd prefer not to restrict this. I figure VS Code already has validation under the covers and people might use this for other urls in the future. I've renamed from `openInBrowser` to `openUrl` to reflect that

Issue #1081, opened 2019-03-12
Multiple Functions projects on the same Workspace


galvesribeiro:
Hello!

I got that the extension now uses VSCode new workspace feature rather than plain old .sln files. 

The problem is, VSCode settings are now embedded on a per project basis on its own `.vscode` directory making things harder to maintain.

I'm trying to have the following workspace to work:

![image](https://user-images.githubusercontent.com/4714040/54166116-684e4100-4442-11e9-873a-825fc44f499c.png)

The idea is to have multiple Functions apps under the same workspace using a single `.code-workspace` file and start/build everything when I press F5 (yeah, I've changed the Host port on each one and also added the `compounds` section as the picture show).

The problem is that if I do that, build fail miserably since it tried to build the same project multiple times and the files are in use. I guess MSBuild is not seeing each one of those projects as different ones.

I'm trying to make the same behaviour for both the developers who have VS2017 on Windows using the `.sln` as the ones on Mac OSX using VSCode and its workspace files.

Can someone shed a light on how can I achieve that?

Thanks!

EricJizbaMSFT:
Interesting... I looked into this and couldn't get it working with the existing extension. We'll have to make changes on our side

galvesribeiro:
Yeah @EricJizbaMSFT 

This is what we would expect from the extensions that adopt multi-rooted workspaces.

For now we would keep the project directory structure as before, having a `.sln`, and all projects under `src/` and `test/` so we keep VS2017 working and then open every function project on its own VSCode window and build/run that individually along with any solution-wide project dependencies.

Looking forward to have the fix at least on Insider track.

Thanks for the reply!

EricJizbaMSFT:
@galvesribeiro I submitted a PR to fix this scenario. Here's our instructions for installing an insider extension: https://aka.ms/azCodeInsiders. I'll let you know once it's in master, or you can try a build from my branch: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=4041

I created a [sample repo](https://github.com/EricJizbaMSFT/MultiFuncApp) that was working for me. [This commit](https://github.com/EricJizbaMSFT/MultiFuncApp/commit/df9dfa8b1ccbf1d6577e4963835481aa36043da2) in particular will probably be most helpful for configuring your project.

galvesribeiro:
Awesome @EricJizbaMSFT! 

I'll try it first thing in the morning and let you know. 

Thank you very much for the help! :)

[...]

PR #1085, opened 2019-03-13
Support debugging multiple function apps at once

EricJizbaMSFT:
We generally work with multi-root workspace, but not if you have two function apps open at once. This PR fixes debugging so that you can actually run both at the same time. Here's a sample project: https://github.com/EricJizbaMSFT/MultiFuncApp

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1081

### DIFF
--- a/src/funcCoreTools/funcHostTask.ts
+++ b/src/funcCoreTools/funcHostTask.ts
@@ -5,61 +5,76 @@
 
 import * as vscode from 'vscode';
 import { IActionContext, registerEvent } from 'vscode-azureextensionui';
-import { hostStartCommand } from '../constants';
 import { localize } from '../localize';
-
-let isFuncHostRunning: boolean = false;
 
 // The name of the task before we started providing it in FuncTaskProvider.ts
 export const oldFuncHostNameRegEx: RegExp = /run\s*functions\s*host/i;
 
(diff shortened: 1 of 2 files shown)


src/funcCoreTools/funcHostTask.ts
```diff
@@ -5,61 +5,76 @@
...
+                        resolve();
+                        listener.dispose();
+                    }
+                });
+
+                const timeoutInSeconds: number = 30;
+                const timeoutError: Error = new Error(localize('failedToFindFuncHost', 'Failed to stop previous running Functions host within "{0}" seconds. Make sure the task has stopped before you debug again.', timeoutInSeconds));
+                setTimeout(() => { reject(timeoutError); }, timeoutInSeconds * 1000);
+            });
+            stopFuncHostPromiseMap.set(debugSession.workspaceFolder, stopFuncHostPromise);
```
nturinski: What's the point of setting this here if you just await the stopFuncHostPromise right after??

EricJizbaMSFT: `stopFuncHostPromiseMap` is used outside this function (in `stopFuncHost` above), which could be called outside this code path

Issue #1013, opened 2019-02-21
Add ability to mark an app setting as a "slot setting" for a slot


EricJizbaMSFT:
Same as app service issue: https://github.com/Microsoft/vscode-azureappservice/issues/796. Nathan already implemented in shared package, just need to register command in this extension

PR #1098, opened 2019-03-19
Add toggle setting to context menu

nturinski:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1013

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -944,7 +944,7 @@
                 },
                 "readable-stream": {
                     "version": "2.0.6",
-                    "resolved": "https://registry.npmjs.org/readable-stream/-/readable-stream-2.0.6.tgz",
+                    "resolved": "http://registry.npmjs.org/readable-stream/-/readable-stream-2.0.6.tgz",
                     "integrity": "sha1-j5A0HmilPMySh4jaz80Rs265t44=",
                     "requires": {
                         "core-util-is": "~1.0.0",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -5712,7 +5731,7 @@
         },
         "md5.js": {
             "version": "1.3.4",
-            "resolved": "https://registry.npmjs.org/md5.js/-/md5.js-1.3.4.tgz",
+            "resolved": "http://registry.npmjs.org/md5.js/-/md5.js-1.3.4.tgz",
```
EricJizbaMSFT: Can you do all that stuff to clean out your npm cache? So we don't get these http changes?

package.json
```diff
@@ -483,6 +489,11 @@
                     "when": "view == azureFunctionsExplorer && viewItem == applicationSettingItem",
                     "group": "1@3"
                 },
+                {
+                    "command": "azureFunctions.appSettings.toggleSlotSetting",
+                    "when": "view == azureFunctionsExplorer && viewItem == applicationSettingItem",
```
EricJizbaMSFT: I feel like we also want `config.azureFunctions.enableSlots == true` in this condition. What do you think?

nturinski: Yeah, that makes sense.  It'd be a pretty confusing setting if you didn't have slots enabled.  

That being said, I think the UI shows if a setting is a slot setting regardless of the `enableSlots` setting.  I'm not sure how they'd manage to do that-- maybe in the portal?

Issue #1058, opened 2019-03-05
Schedule for timer trigger


asavaritayal:
![image](https://user-images.githubusercontent.com/8435148/53792835-6dab0900-3ee1-11e9-80fa-9ea08d78b96e.png)

![image](https://user-images.githubusercontent.com/8435148/53792874-80bdd900-3ee1-11e9-976b-8be4f5362e48.png)

Two issues here: 

1. Helper link only appears when there is invalid input. Instead, we should surface it from the get go.
2. Link doesn't render correctly. Raw HTML syntax is visible. 

/cc @fiveisprime @gbowerman

EricJizbaMSFT:
So this string is defined here: https://github.com/Azure/azure-functions-templates/blob/20bde0b0ea46ebde088f257a333501e6dd89a3e3/Functions.Templates/Resources/Resources.resx#L982

But I believe that string is also displayed in the portal. Do we want to change it for both or special case Timer Trigger in our extension?

asavaritayal:
Re #1 - I think we want to display the string here: 
https://github.com/Azure/azure-functions-templates/blob/20bde0b0ea46ebde088f257a333501e6dd89a3e3/Functions.Templates/Resources/Resources.resx#L213

For #2, issue isn't really the string itself but that the HTML isn't rendering correctly. Is it possible for Code to make the links clickable?

EricJizbaMSFT:
Fixed the first part, which I think is the more important piece. I'll have to file an issue on VS Code for the second piece

EricJizbaMSFT:
Closing this as fixed since the description is way better at this point. Rendering the link would be more difficult and unclear how VS COde would support that. Wont-fixing that aspect

PR #1099, opened 2019-03-19
Show 'help' text when creating function

EricJizbaMSFT:
Before:
Screen Shot 2019-03-19 at 9 11 04 AM
Screen Shot 2019-03-19 at 9 11 16 AM

After:
Screen Shot 2019-03-19 at 9 09 30 AM
Screen Shot 2019-03-19 at 9 09 01 AM

Fixes most of https://github.com/Microsoft/vscode-azurefunctions/issues/1058

### DIFF
--- a/src/templates/parseScriptTemplates.ts
+++ b/src/templates/parseScriptTemplates.ts
@@ -33,6 +33,7 @@ interface IRawSetting {
     name: string;
     value: ValueType;
     label: string;
+    help?: string;
     defaultValue?: string;
     required: boolean;
     resource?: ResourceType;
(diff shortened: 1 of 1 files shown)


src/templates/parseScriptTemplates.ts
```diff
@@ -103,6 +104,7 @@ function parseScriptSetting(data: object, resources: IResources, variables: IVar
         name: getVariableValue(resources, variables, rawSetting.name),
         resourceType: rawSetting.resource,
         valueType: rawSetting.value,
+        description: rawSetting.help ? getResourceValue(resources, rawSetting.help) : undefined,
```
nturinski: Do we want to completely remove the description if there is no help dialogue?  Won't that get rid of whatever was normally there by default?

EricJizbaMSFT: Description was undefined in this object before my change, so I'm not removing anything

nturinski: Where was `Provide a 'Path'` coming from??

EricJizbaMSFT: https://github.com/Microsoft/vscode-azurefunctions/blob/ec9b71adbd5cb27d1206661837cbc7e72bef15d4/src/commands/createFunction/genericSteps/StringPromptStep.ts#L22

nturinski: Why not just set `rawSetting.description` instead of naming it `help`?  I get that it's more like a tooltip, but it ends up being equivalent to a quickpick's description property-- seems kinda confusing.

EricJizbaMSFT: I didn't choose the name `help`. This file is all about parsing templates that are defined externally and I have to use whatever structure that's given to us

PR #1100, opened 2019-03-19
Use localized strings when creating functions

EricJizbaMSFT:
The templates already have these strings localized, so might as well use them.

Screen Shot 2019-03-19 at 11 02 52 AM

Screen Shot 2019-03-19 at 11 03 21 AM

### DIFF
--- a/src/templates/ScriptTemplateRetriever.ts
+++ b/src/templates/ScriptTemplateRetriever.ts
@@ -98,12 +98,22 @@ export async function getResourcesPath(templatesPath: string, vscodeLang: string
         // Example: "en" for "english"
         const language: string = parts[0];
         // Example: "US" for "United States" (locale is optional)
-        // tslint:disable-next-line: strict-boolean-expressions
-        const locale: string | undefined = parts[1] || '[a-z]*';
+        let locale: string | undefined = parts[1];
 
         const files: string[] = await fse.readdir(folder);
-        const regExp: RegExp = new RegExp(`resources\\.${language}(-${locale})?\\.json`, 'i');
-        const matchingFile: string | undefined = files.find(f => regExp.test(f));
+        let matchingFile: string | undefined;
+        if (!locale) {
+            const regExp: RegExp = new RegExp(`resources\\.${language}\\.json`, 'i');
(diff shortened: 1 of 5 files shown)


src/templates/ScriptTemplateRetriever.ts
```diff
@@ -77,15 +78,42 @@ export class ScriptTemplateRetriever extends TemplateRetriever {
...
+    const folder: string = path.join(templatesPath, 'resources');
+
+    try {
+        // Example: "en-US"
+        const parts: string[] = vscodeLang.split('-');
+        // Example: "en" for "english"
+        const language: string = parts[0];
+        // Example: "US" for "United States" (locale is optional)
+        // tslint:disable-next-line: strict-boolean-expressions
+        const locale: string | undefined = parts[1] || '[a-z]*';
```
StephenWeatherford: Might be wise to allow uppercase here. I've frequently seen en-US for instance (https://tools.ietf.org/html/rfc5646).

StephenWeatherford: Ah, I see your case-intensitive modifier below...

src/templates/ScriptTemplateRetriever.ts
```diff
@@ -77,15 +78,42 @@ export class ScriptTemplateRetriever extends TemplateRetriever {
...
+
+        const files: string[] = await fse.readdir(folder);
+        const regExp: RegExp = new RegExp(`resources\\.${language}(-${locale})?\\.json`, 'i');
+        const matchingFile: string | undefined = files.find(f => regExp.test(f));
```
StephenWeatherford: How are the resources done?  Do you always have specific locales, e.g. en-us resources, or can you have just "en" resources?  If the vscode language is "en", and you have en resources and en-uk resources, en should be preferred.

EricJizbaMSFT: This is what the templates look like: https://github.com/Microsoft/vscode-azurefunctions/tree/master/resources/backupScriptTemplates/~2/resources

Not sure if that'll be very common, but fixed

Issue #1066, opened 2019-03-06
Improve error when failed to parse local.settings.json


maujy:
Repro steps:

1. Using VsCode Microsoft Azure to add Function App (Azure Cosmos DB trigger)
2.

Action: azureFunctions.createFunction
Error type: SyntaxError
Error Message: Unexpected token / in JSON at position 90

Version: 0.15.0
OS: win32
VS Code Version: 1.31.1

<details>
<summary>Call Stack</summary>

```
JSON.parse (<anonymous>)
dist/extension.bundle.js:9:117853
Generator.next (<anonymous>)
o (dist/extension.bundle.js:9:117002)
```

</details>

EricJizbaMSFT:
@maujy I think you have a syntax error in your local.settings.json file. Can you check that and try again? If that's the case, I'll repurpose this issue to improve the error message.

maujy:
Still has problem : (

local.setting.json:
{
  "IsEncrypted": false,
  "Values": {
    "FUNCTIONS_WORKER_RUNTIME": "python",
    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
    "CosmosConnectionString": "...",
    "LeaseSettings": "..."
  }
}



Action: azureFunctions.deploy
Error type: BadRequest
Error Message: The requested app service plan cannot be created in the current resource group because it is hosting Linux apps. Please choose a different resource group or create a new one.

Version: 0.15.0
OS: win32
VS Code Version: 1.31.1

<details>
<summary>Call Stack</summary>

```
module.exports.Ps.l.pipeline (dist/extension.bundle.js:277:2538438)
s (dist/extension.bundle.js:277:110820)
s (dist/extension.bundle.js:165:25754)
dist/extension.bundle.js:165:26994
o (dist/extension.bundle.js:223:7055)
dist/extension.bundle.js:165:24675
$.module.exports.t.requestLibrarySink.t.headersOnly.t.streamedResponse.t.body.function.t.body.pipe.a [as _callback] (dist/extension.bundle.js:1:578599)
$.module.exports.$.init.t._callback.t.callback.t.callback [as callback] (dist/extension.bundle.js:253:5461)
$.emit (events.js:182:13)
$.<anonymous> (dist/extension.bundle.js:253:18274)
$.emit (events.js:187:15)
IncomingMessage.<anonymous> (dist/extension.bundle.js:253:17116)
Object.onceWrapper (events.js:273:13)
[...]

PR #1106, opened 2019-03-20
Improve error when failed to parse local.settings.json

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1066

### DIFF
--- a/src/LocalAppSettings.ts
+++ b/src/LocalAppSettings.ts
@@ -80,12 +80,13 @@ export async function getLocalAppSettings(localSettingsPath: string, allowOverwr
                 return <ILocalAppSettings>JSON.parse(data);
             } catch (error) {
                 if (allowOverwrite) {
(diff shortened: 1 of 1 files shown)


src/LocalAppSettings.ts
```diff
@@ -80,12 +80,13 @@ export async function getLocalAppSettings(localSettingsPath: string, allowOverwr
                 return <ILocalAppSettings>JSON.parse(data);
             } catch (error) {
                 if (allowOverwrite) {
-                    const message: string = localize('failedToParse', 'Failed to parse local settings: {0}. Overwrite?', parseError(error).message);
+                    const message: string = localize('failedToParseWithOverwrite', 'Failed to parse "{0}": {1}. Overwrite?', localSettingsFileName, parseError(error).message);
```
nturinski: Pretty confident that a parse failure is the only error to show up here?

EricJizbaMSFT: The only code I'm calling within the `try` block is `JSON.parse` - so yes

Issue #904, opened 2019-01-10
v1 JS project uses v2 template 


JohnReeson:
host.json contains `"version": "2.0"` and local.settings.json contains `"FUNCTIONS_WORKER_RUNTIME": "node"` so I assume it consumes v2 template somehow.

![capture](https://user-images.githubusercontent.com/35160013/50958396-90c9c580-14fb-11e9-8f3a-aa36a8f5a857.JPG)

Extension Version: 0.13.1
CLI version 1.0.17
Node v6

EricJizbaMSFT:
`FUNCTIONS_WORKER_RUNTIME` shouldn't hurt anything on v1 and could be helpful if people ever upgrade to v2. The host.json probably shouldn't say 2.0, though. 

If you create a new function - what runtime is listed?
![screen shot 2019-01-11 at 1 39 00 pm](https://user-images.githubusercontent.com/11282622/51060956-5e28d700-15a6-11e9-8e88-8658337e7029.png)

If it says "~1" then that means you're still getting the v1 templates.

JohnReeson:
Thanks for the explanation, it's ~1 as expected. By template, I mean the project template i.e. `host.json` and `local.settings.json`.   `"version": "2.0"` is exactly what I am worrying about. Even though the v1 format configuration does work anyway, we can see the misleading prompt according to v2 schema. 

![12](https://user-images.githubusercontent.com/35160013/51069117-c8538180-1663-11e9-9fed-d90306cb899c.png)

PR #1107, opened 2019-03-20
Write runtime-specific host.json

EricJizbaMSFT:
The "host.json" file that we write should only have `"version": "2.0"` if the runtime is v2. Seems like a simple fix, but I had to refactor when we detect or prompt for runtime in order to do this.

Also moved a few things down into `ScriptProjectCreatorBase.ts` that were copied across Python/TypeScript/PowerShell/JavaScript.

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/904

### DIFF
--- a/src/commands/createNewProject/initProjectForVSCode.ts
+++ b/src/commands/createNewProject/initProjectForVSCode.ts
@@ -6,11 +6,10 @@
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import { IActionContext, TelemetryProperties } from 'vscode-azureextensionui';
-import { deploySubpathSetting, extensionPrefix, filesExcludeSetting, gitignoreFileName, preDeployTaskSetting, projectLanguageSetting, projectRuntimeSetting, templateFilterSetting } from '../../constants';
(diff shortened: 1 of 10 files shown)


src/commands/createNewProject/initProjectForVSCode.ts
```diff
@@ -89,12 +84,12 @@ async function writeDebugConfiguration(projectCreator: ProjectCreatorBase, vscod
     }
 }
 
-async function writeVSCodeSettings(projectCreator: ProjectCreatorBase, vscodePath: string, language: string, templateFilter: string): Promise<void> {
+async function writeVSCodeSettings(projectCreator: ProjectCreatorBase, vscodePath: string, language: string, templateFilter: string, runtime: ProjectRuntime): Promise<void> {
```
nturinski: nit: Can't you call `getRuntime()` from projectCreator rather than having to pass it in?

EricJizbaMSFT: I tried to use getRuntime as little as possible just because:
1. It's async
1. It will prompt if runtime hasn't been defined yet, and I want to make it somewhat clear when that prompt might happen

Issue #777, opened 2018-10-31
Alert users when multiple installations of core tools are present


fiveisprime:
This has come up a few times now where a user has the older v1 tools installed via npm then installs v2 using homebrew. In this case, whichever comes first in the path is used and it often creates a very confusing issue for the user.

We should be able to detect when multiple installations are available and let users know what they need to do to remedy the situation. It may require using `brew list` and `npm ls -g --depth 0` to avoid the path order.

PR #1108, opened 2019-03-20
 Alert users when multiple installs of func cli are present

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/777

See issue for more info. A few of the dialogs:
Screen Shot 2019-03-20 at 11 40 52 AM

Screen Shot 2019-03-20 at 11 44 44 AM

### DIFF
--- a/src/funcCoreTools/uninstallFuncCoreTools.ts
+++ b/src/funcCoreTools/uninstallFuncCoreTools.ts
@@ -3,28 +3,36 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { IAzureQuickPickItem } from 'vscode-azureextensionui';
 import { funcPackageName, PackageManager } from '../constants';
 import { ext } from '../extensionVariables';
 import { localize } from '../localize';
 import { cpUtils } from '../utils/cpUtils';
-import { getFuncPackageManager } from './getFuncPackageManager';
-import { funcToolsInstalled } from './validateFuncCoreToolsInstalled';
+import { getFuncPackageManagers } from './getFuncPackageManagers';
 
-export async function uninstallFuncCoreTools(): Promise<void> {
+export async function uninstallFuncCoreTools(packageManagers?: PackageManager[]): Promise<void> {
     ext.outputChannel.show();
-    if (await funcToolsInstalled()) {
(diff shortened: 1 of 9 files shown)


src/funcCoreTools/uninstallFuncCoreTools.ts
```diff
@@ -3,28 +3,36 @@
...
+        throw new Error(localize('notInstalled', 'Cannot uninstall Azure Functions Core Tools because it is not installed with brew or npm.'));
+    } else if (packageManagers.length === 1) {
+        packageManager = packageManagers[0];
     } else {
-        throw new Error(localize('notInstalled', 'Cannot uninstall Azure Functions Core Tools because it is not installed.'));
+        const placeHolder: string = localize('multipleInstalls', 'Multiple installs of the func cli detected. Select the one to uninstall');
+        const picks: IAzureQuickPickItem<PackageManager>[] = packageManagers.map(pm => { return { label: pm, data: pm }; });
+        packageManager = (await ext.ui.showQuickPick(picks, { placeHolder })).data;
+    }
+
```
PrashanthCorp: Since the wording of the prompt asks for which install to remove, I think there should should be a confirmation. 

A couple of questions: 
1.  The package manager enum shows only npm and brew. Are there only two possibilities? 
2. If so, the options ["Keep npm" and "Keep brew" ] or ["Uninstall npm" and "Uninstall brew"] would be less prone to human error (for all the users who don't fully read the prompt message :P)

nturinski: In this case, the command that was invoked was UninstallFuncCoreTools which is kind of the confirmation in itself.  We can assume that the user fully intends to uninstall by calling this command since it is only exposed via the command pallette. (Sorry for responding, just happened to approve this PR as soon as you commented 😆)

EricJizbaMSFT: 1. There's a few more, but those are the only two we support
1. To add to what Nathan said - it's pretty easy to re-install the func cli, so I'm not that worried about a confirmation message.

I'm okay with making the picks more explicit, but I don't like just "Uninstall npm" because that makes it seem like I'm uninstalling _npm_ instead of _func_. Changed to "Uninstall npm package"

Screen Shot 2019-03-20 at 2 57 11 PM

Issue #1012, opened 2019-02-21
Add "View commit in Github" to deployment item


EricJizbaMSFT:
Same as app service issue: https://github.com/Microsoft/vscode-azureappservice/issues/726. Nathan already implemented in shared package - just need to register command in this extension

PR #1109, opened 2019-03-20
Add viewCommitInGitHub

nturinski:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1012

### DIFF
--- a/src/commands/deployments/viewCommitInGitHub.ts
+++ b/src/commands/deployments/viewCommitInGitHub.ts
@@ -1,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DeploymentTreeItem } from "vscode-azureappservice";
+import { ext } from "../../extensionVariables";
+
+export async function viewCommitInGitHub(node?: DeploymentTreeItem): Promise<void> {
+    if (!node) {
+        node = <DeploymentTreeItem>(await ext.tree.showTreeItemPicker('deployment/github'));
+    }
+    await node.viewCommitInGitHub();
+}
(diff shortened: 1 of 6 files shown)


src/commands/deployments/viewCommitInGitHub.ts
```diff
@@ -0,0 +1,14 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { DeploymentTreeItem } from "vscode-azureappservice";
+import { ext } from "../../extensionVariables";
+
+export async function viewCommitInGitHub(node?: DeploymentTreeItem): Promise<void> {
+    if (!node) {
+        node = <DeploymentTreeItem>(await ext.tree.showTreeItemPicker('deployment/github'));
```
EricJizbaMSFT: Just to confirm - the casing here doesn't matter?

src/commands/deployments/viewCommitInGitHub.ts
```diff
@@ -0,0 +1,14 @@
...
+    }
+    await node.viewCommitInGitHub();
```
PrashanthCorp: I can't see where this is implemented. Any pointers?

src/commands/deployments/viewCommitInGitHub.ts
nturinski: The casing here does matter.  In the DeploymentTreeItem constructor, I made everything `toLowerCase()` to avoid confusion.  Maybe this should be done in the TreeItemPicker level?

src/commands/deployments/viewCommitInGitHub.ts
nturinski: https://github.com/Microsoft/vscode-azuretools/blob/v0.33.0-appservice/appservice/src/tree/DeploymentTreeItem.ts#L134-L144

src/commands/deployments/viewCommitInGitHub.ts
EricJizbaMSFT: Yeah TreeItemPicker should probably ignore case if it's not already. Not a big deal, though

PR #1122, opened 2019-03-26
"Basic" create function app

### LINKED ISSUE #1054: Basic creation support
Same as for app service, but creates plan and storage account without prompting.

EricJizbaMSFT:
Automatically creates storage account/resource group without prompting the user. Verified nightly build passes with my changes.

Fixes #1054

### DIFF
--- a/src/tree/FunctionAppProvider.ts
+++ b/src/tree/FunctionAppProvider.ts
@@ -5,8 +5,8 @@
 
 import { WebSiteManagementClient, WebSiteManagementModels } from 'azure-arm-website';
 import { AppKind, IAppServiceWizardContext, IAppSettingsContext, SiteClient, SiteCreateStep, SiteNameStep, SiteOSStep, SiteRuntimeStep, WebsiteOS } from 'vscode-azureappservice';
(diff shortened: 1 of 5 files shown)


src/tree/FunctionAppProvider.ts
```diff
@@ -83,47 +83,59 @@ export class FunctionAppProvider extends SubscriptionTreeItem {
...
+        }
+
+        executeSteps.push(new SiteCreateStep(async (context): Promise<WebSiteManagementModels.NameValuePair[]> => await createFunctionAppSettings(context, runtime, language)));
 
         const title: string = localize('functionAppCreatingTitle', 'Create new function app');
         const wizard: AzureWizard<IAppServiceWizardContext> = new AzureWizard(wizardContext, { promptSteps, executeSteps, title });
 
         await wizard.prompt(actionContext);
         showCreatingTreeItem(nonNullProp(wizardContext, 'newSiteName'));
+        if (!advancedCreation) {
```
nturinski: How come this can't be done within the if statement from lines 101-104?  Or in `setBasicCreateDefaults`?

EricJizbaMSFT: `wizardContext.relatedNameTask;` is based on the function app name they entered, so it only exists after we call `wizard.prompt`

PR #1127, opened 2019-03-27
Fix bug that happened in standup

EricJizbaMSFT:
The back button only works if you actually prompt when you return `true` in `shouldPrompt`. This step wasn't prompting when it should've been (we can at least prompt and let them select `+ Create new local app setting`


EricJizbaMSFT:
Good question. I'll look into it tomorrow

PR #1128, opened 2019-03-27
Add function app actions test and delete the repeating section

v-wuzhai:
Add a case functionapp.test.ts to test Function App, it contains 'Create/Stop/Start/Restart/Delete' function app. Delete the repeating section for 'Create and Delete Function App' from createAzureResource.test.ts.

Merge https://github.com/Microsoft/vscode-azurefunctions/pull/993 to this PR.


test/functionapp.test.ts
```diff
@@ -0,0 +1,121 @@
...
+                console.log(`Resource group "${resourceGroup}" deleted.`);
+            } else {
+                // If the test failed, the resource group might not actually exist
+                console.log(`Ignoring resource group "${resourceGroup}" because it does not exist.`);
+            }
+        }
+        ext.tree.dispose();
+    });
+
+    test('createFunctionApp (basic)', async () => {
```
nturinski: Could you capitalize Basic?

test/functionapp.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        const resourceName1: string = getRandomHexString().toLowerCase();
+        resourceGroupsToDelete.push(resourceName1);
+        const testInputs: string[] = [resourceName1, '$(plus) Create new resource group', resourceName1, '$(plus) Create new storage account', resourceName1, 'East US'];
```
nturinski: With the basic creation, you only need one input (the resource name).

test/functionapp.test.ts
```diff
@@ -0,0 +1,121 @@
...
+            } else {
+                // If the test failed, the resource group might not actually exist
+                console.log(`Ignoring resource group "${resourceGroup}" because it does not exist.`);
+            }
+        }
+        ext.tree.dispose();
+    });
+
+    test('createFunctionApp (basic)', async () => {
+        const resourceName1: string = getRandomHexString().toLowerCase();
```
nturinski: Naming the variable resourceName1 is kind of confusing since the other variable is just called resourceName.  Renaming them as resourceName1 and resourceName2 would be better.

nturinski: You should explicitly set the advancedCreation to `false` in this test.  If you're running this locally, it will have your current settings (which would be set to `true`)

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        ext.tree = new AzureTreeDataProvider(FunctionAppProvider, 'azureFunctions.startTesting', undefined, testAccount);
+    });
+
+    suiteTeardown(async function (this: IHookCallbackContext): Promise<void> {
+        if (!longRunningTestsEnabled) {
+            this.skip();
+        }
+        this.timeout(1200 * 1000);
+        await updateGlobalSetting(projectLanguageSetting, oldProjectLanguage);
+        await vscode.workspace.getConfiguration(extensionPrefix).update('advancedCreation', oldAdvancedCreationSetting, vscode.ConfigurationTarget.Global);
```
nturinski: Can't you use updateGlobalSetting here as well? I don't think that Azure App Service has that util function (where that example came from)

nturinski: As well as the other two times you call it in the tests

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+suite('Function App actions', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(1200 * 1000);
+    const resourceGroupsToDelete: string[] = [];
+    let oldAdvancedCreationSetting: boolean | undefined;
+    let oldProjectLanguage: ProjectLanguage | undefined;
+    const resourceName1: string = getRandomHexString().toLowerCase();
+    const testAccount: TestAzureAccount = new TestAzureAccount();
+
+    suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
+        if (!longRunningTestsEnabled) {
```
EricJizbaMSFT: There's a lot of duplicate code in this file and `createAzureResource.test.ts` (at least suiteSetup and suiteTeardown). Please add your tests to the other file instead of copying this code

test/functionapp.test.ts
EricJizbaMSFT: Use runWithFuncSetting like here: https://github.com/Microsoft/vscode-azurefunctions/blob/master/test/createAzureResource.test.ts#L89

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName2, resourceName2);
+        assert.ok(createdApp);
+    });
+
+    test('createFunctionApp (Advanced)', async () => {
+        resourceGroupsToDelete.push(resourceName1);
+        await vscode.workspace.getConfiguration(extensionPrefix).update('advancedCreation', true, vscode.ConfigurationTarget.Global);
+        const testInputs: string[] = [resourceName1, 'Linux', '.NET', '$(plus) Create new resource group', resourceName1, '$(plus) Create new storage account', resourceName1, 'West US'];
```
EricJizbaMSFT: Linux is still in preview in Azure. Please run the test against Windows.

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        await updateGlobalSetting(projectLanguageSetting, ProjectLanguage.JavaScript);
+        ext.ui = new TestUserInput([resourceName2]);
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName2, resourceName2);
+        assert.ok(createdApp);
+    });
+
+    test('createFunctionApp (Advanced)', async () => {
+        resourceGroupsToDelete.push(resourceName1);
```
EricJizbaMSFT: To more throughly test advancedCreation, you should use different names for the function app name, resource group name, and storage account name. Otherwise it's a little too similar to basic create

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName1, resourceName1);
+        assert.ok(createdApp);
+    });
+
+    test('stopFunctionApp', async () => {
+        ext.ui = new TestUserInput([resourceName1]);
+        await vscode.commands.executeCommand('azureFunctions.stopFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
```
EricJizbaMSFT: You repeated this line a lot:
```
const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
```

Just define `client` once in suiteSetup

EricJizbaMSFT: You need to verify the app is running before you call `azureFunctions.stopFunctionApp`

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName1, resourceName1);
+        assert.equal(createdApp.state, 'Stopped', `Function App state should be 'Stopped' rather than ${createdApp.state}.`);
+    });
+
+    test('startFunctionApp', async () => {
+        ext.ui = new TestUserInput([resourceName1]);
+        await vscode.commands.executeCommand('azureFunctions.startFunctionApp');
```
EricJizbaMSFT: You need to verify the app is stopped before you call `azureFunctions.startFunctionApp `

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName1, resourceName1);
+        assert.equal(createdApp.state, 'Running', `Function App state should be 'Running' rather than ${createdApp.state}.`);
+    });
+
+    test('restartFunctionApp', async () => {
+        let client: WebSiteManagementClient;
```
EricJizbaMSFT: Usually a restart occurs from a running state, not a stopped state. I suggest removing this logic to stop the function app and instead verifying that the app is "Running" before and after you call `'azureFunctions.restartFunctionApp'`

test/functionapp.test.ts
```diff
@@ -0,0 +1,125 @@
...
+        const testInputs: string[] = [resourceName1, 'Linux', '.NET', '$(plus) Create new resource group', resourceName1, '$(plus) Create new storage account', resourceName1, 'West US'];
+        ext.ui = new TestUserInput(testInputs);
+        await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+        const client: WebSiteManagementClient = getWebsiteManagementClient(testAccount);
+        const createdApp: WebSiteManagementModels.Site = await client.webApps.get(resourceName1, resourceName1);
+        assert.ok(createdApp);
+    });
+
+    test('stopFunctionApp', async () => {
+        ext.ui = new TestUserInput([resourceName1]);
```
EricJizbaMSFT: Please use the app created from the basic create test for your other tests, instead of the app created in the advanced create test. Basic create is the default and is higher priority.

test/functionapp.test.ts
v-wuzhai: Do you mean that add all our tests from functionapp.test.ts to your createAzureResource.test.ts? Such as stopFunctionApp, startFunctionApp and so on.

test/functionapp.test.ts
v-wuzhai: I am confused for this comment, if verifing the app is "Running" before and after we call 'azureFunctions.restartFunctionApp', It cannot  exclud this situation "The 'azureFunctions.restartFunctionApp' execution fails, but the result is still pass".

test/functionapp.test.ts
EricJizbaMSFT: Yes that is correct. We might rename `createAzureResource.test.ts` to be more generic, but I would prefer to do that after your PR because renaming the file can make it hard to review a PR.

test/functionapp.test.ts
EricJizbaMSFT: I agree it's not perfect, but my suggestion is simpler and covers the more common case. Your existing test is essentially testing "start" instead of "restart".

If you want the test to be completely accurate, you would poll for the status while `azureFunctions.restartFunctionApp` occurs to make sure it goes to 'Stopped' state in the middle. You're welcome to try that, but I think that code will be pretty complex and isn't worth it for now.

PR #1139, opened 2019-03-29
Fix TypeScript scriptPath on Windows

anthonychu:
`scriptPath` in function.json should always be in posix format

/cc @duncan


EricJizbaMSFT:
@anthonychu I have a guess - but can you confirm why this change in necessary?

anthonychu:
Becomes an issue when collaborating with folks on macOS/Linux or deploying to Linux. And `func new` generates posix paths here on all platforms so this would match that.

duncan:
It's all about the x-platform experience. In my case, I hit this faster than you might expect when I created a function using VSCode on Windows using WSL. `"scriptFile": "..\\dist\\..." means that the function won't run locally when in Bash. Of course, flipping the slashes around to `/` fixes everything.

Issue #924, opened 2019-01-22
Validate FUNCTIONS_WORKER_RUNTIME when deploying


EricJizbaMSFT:
i.e. if it's set to node, but they're deploying a python app

PR #1141, opened 2019-03-29
Verify language worker runtime on deploy

EricJizbaMSFT:
For example if they're trying to deploy a .NET project to a Node function app

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/924


EricJizbaMSFT:
ping

Issue #659, opened 2018-09-25
"Could not load template" after updating extension


phanthoa:
&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: azureFunctions.createNewProject
Error type: Error
Error Message: Command "dotnet "c:\Users\thoapv\.vscode\extensions\ms-azuretools.vscode-azurefunctions-0.10.2\resources\dotnetJsonCli\Microsoft.TemplateEngine.JsonCli.dll" --require "c:\Users\thoapv\.vscode\extensions\ms-azuretools.vscode-azurefunctions-0.10.2\resources\dotnetTemplates\itemTemplates-beta.nupkg" --require "c:\Users\thoapv\.vscode\extensions\ms-azuretools.vscode-azurefunctions-0.10.2\resources\dotnetTemplates\projectTemplates-beta.nupkg" --operation create --identity Microsoft.AzureFunctions.ProjectTemplate.CSharp.2.x --arg:name "TSDVTestFuncion" --arg:AzureFunctionsVersion v2" failed with exit code "2097158":
{
  "Message": "Could not load template",
  "Status": 2097158,
  "TemplateFullName": "Azure Functions",
  "ResultInfo": null,
  "OutputBaseDirectory": null
}

Version: 0.10.2
OS: win32

EricJizbaMSFT:
Sorry for the delayed response. Can you give us a little more info? For example:
1. Does this always happen for you?
1. Do the file paths listed in the "--require" above actually exist?

firebellys:
This started happening for me as well. It happens every time (repeatable) and I have all those required files. I've made dozens of functions on this machine recently too, so this is not expected.

&lt;Please be sure to remove any private information before submitting.&gt;

Repro steps:
&lt;Enter steps to reproduce issue&gt;

Action: azureFunctions.createFunction
Error type: Error
[...]

PR #1147, opened 2019-04-01
Use globalStoragePath for storing dotnet templates

EricJizbaMSFT:
`context.globalStoragePath` was added to VS Code [pretty recently](https://github.com/Microsoft/vscode/issues/66030). We should use that instead of storing directly in our extension folder.

Related: https://github.com/Microsoft/vscode-azurefunctions/issues/659


EricJizbaMSFT:
ping

PR #1151, opened 2019-04-02
Don't use webpack for local development

EricJizbaMSFT:
As discussed offline - getting tired of the delay when we F5 the extension. Main fix was to use an 
 environment variable in main.js

I also had to switch `getCoreNodeModule` back to a *.ts file so that it would work for both webpack and non-webpack. As far as I can tell, we always run `npm build` before webpack, so I don't think this comment applies anymore anyways:
> Since webpack shouldn't have to depend on npm build to transpile .ts -> .js, we keep this file in .js

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -15,6 +15,26 @@
             "outFiles": [
                 "${workspaceFolder}/**/*.js"
             ],
+            "preLaunchTask": "npm: compile",
+            "env": {
+                "AZCODE_IGNORE_BUNDLE": "1",
+                "DEBUGTELEMETRY": "1",
+                "NODE_DEBUG": ""
+            }
+        },
+        {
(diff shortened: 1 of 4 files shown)


StephenWeatherford:
> As discussed offline - getting tired of the delay when we F5 the extension. Main fix was to use an
> environment variable in main.js
> I also had to switch getCoreNodeModule back to a *.ts file so that it would work for both webpack and non-webpack. As far as I can tell, we always run npm build before webpack, so I don't think this comment applies anymore anyways:
> 
> Since webpack shouldn't have to depend on npm build to transpile .ts -> .js, we keep this file in .js

Right now npm run build is still used for building the test files, which could theoretically be fixed, but build is fast, I'm good with this.

StephenWeatherford:
Thanks!

Issue #1082, opened 2019-03-12
Enable PowerShell project


EricJizbaMSFT:
PowerShell support was added here: https://github.com/Microsoft/vscode-azurefunctions/pull/1075, but can't be enabled until vscode-powershell 1.12 and 2.0.0-preview.2 are released.

- [x] Add PowerShell and it's prerequisites to README
- [x] Add PowerShell to new project quick pick
- [x] Add templates to "Verified" category (should have same number as JavaScript)
- [x] Double check all "Verified" templates

PR #1158, opened 2019-04-05
Adding support in function app creation to enable Managed Dependencies for PowerShell.

Francisco-Gamino:
This PR contains the following changes:

* Enable Managed Dependencies for PowerShell function apps by default
* Removing feature flag for PowerShell
* Adding PowerShell (preview) to drop down list of available runtime workers

### Installation flow
At function app creation, Managed Dependencies will be enable by default for PowerShell. Managed dependencies enables modules to be automatically managed by the Functions service. Currently, only the Azure Az module is supported.

When creating a PowerShell function app, an entry will be added to host.json and a new requirements.psd1 file will be created. Then, when the function is started, the PowerShell worker will install the dependencies (from the PSGallery) defined in requirements.psd1 to LocalApplicationData\AzureFunctions\FunctionAppName\ManagedDependencies.

PowerShell (preview) now is an option available.

image

At app creation, Managed Dependencies for PowerShell is enabled by default.  

```
# The following files and folders are created after creating an HttpTrigger function.
.vscode
HttpTrigger
.funcignore
.gitignore
host.json
local.settings.json
profile.ps1
proxies.json
requirements.psd1


# Contents of host.json
{
  "version": "2.0",
  "managedDependency": {
    "enabled": true
  }
}

# Contents of requirements.psd1

# This file enables modules to be automatically managed by the Functions service.
# Only the Azure Az module is supported in preview.
# See https://aka.ms/functionsmanageddependency for additional information.
#
@{
    'Az' = '1.*'
}
```

Question:
* Where should I add tests?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -497,10 +497,10 @@
                             "dev": true,
                             "optional": true,
                             "requires": {
-                                "deep-extend": "^0.6.0",
-                                "ini": "~1.3.0",
-                                "minimist": "^1.2.0",
-                                "strip-json-comments": "~2.0.1"
+                                "deep-extend": "0.6.0",
+                                "ini": "1.3.5",
+                                "minimist": "1.2.0",
+                                "strip-json-comments": "2.0.1"
                             },
                             "dependencies": {
                                 "minimist": {
@@ -3903,8 +3903,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "delegates": "^1.0.0",
-                        "readable-stream": "^2.0.6"
+                        "delegates": "1.0.0",
+                        "readable-stream": "2.3.6"
                     }
                 },
                 "balanced-match": {
@@ -3986,7 +3986,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "minipass": "^2.2.1"
+                        "minipass": "2.2.4"
                     }
                 },
                 "fs.realpath": {
@@ -4001,14 +4001,14 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "aproba": "^1.0.3",
-                        "console-control-strings": "^1.0.0",
-                        "has-unicode": "^2.0.0",
-                        "object-assign": "^4.1.0",
-                        "signal-exit": "^3.0.0",
-                        "string-width": "^1.0.1",
-                        "strip-ansi": "^3.0.1",
-                        "wide-align": "^1.1.0"
+                        "aproba": "1.2.0",
+                        "console-control-strings": "1.1.0",
+                        "has-unicode": "2.0.1",
+                        "object-assign": "4.1.1",
+                        "signal-exit": "3.0.2",
+                        "string-width": "1.0.2",
+                        "strip-ansi": "3.0.1",
+                        "wide-align": "1.1.2"
                     }
                 },
                 "glob": {
@@ -4017,12 +4017,12 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "fs.realpath": "^1.0.0",
-                        "inflight": "^1.0.4",
-                        "inherits": "2",
-                        "minimatch": "^3.0.4",
-                        "once": "^1.3.0",
-                        "path-is-absolute": "^1.0.0"
+                        "fs.realpath": "1.0.0",
+                        "inflight": "1.0.6",
+                        "inherits": "2.0.3",
+                        "minimatch": "3.0.4",
+                        "once": "1.4.0",
+                        "path-is-absolute": "1.0.1"
                     }
                 },
                 "has-unicode": {
@@ -4037,7 +4037,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "safer-buffer": "^2.1.0"
+                        "safer-buffer": "2.1.2"
                     }
                 },
                 "ignore-walk": {
@@ -4046,7 +4046,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "minimatch": "^3.0.4"
+                        "minimatch": "3.0.4"
                     }
                 },
                 "inflight": {
@@ -4055,8 +4055,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "once": "^1.3.0",
-                        "wrappy": "1"
+                        "once": "1.4.0",
+                        "wrappy": "1.0.2"
                     }
                 },
                 "inherits": {
@@ -4077,7 +4077,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "number-is-nan": "^1.0.0"
+                        "number-is-nan": "1.0.1"
                     }
                 },
                 "isarray": {
@@ -4092,7 +4092,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "brace-expansion": "^1.1.7"
+                        "brace-expansion": "1.1.11"
                     }
                 },
                 "minimist": {
@@ -4107,8 +4107,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "safe-buffer": "^5.1.1",
-                        "yallist": "^3.0.0"
+                        "safe-buffer": "5.1.1",
+                        "yallist": "3.0.2"
                     }
                 },
                 "minizlib": {
@@ -4117,7 +4117,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "minipass": "^2.2.1"
+                        "minipass": "2.2.4"
                     }
                 },
                 "mkdirp": {
@@ -4141,9 +4141,9 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "debug": "^2.1.2",
-                        "iconv-lite": "^0.4.4",
-                        "sax": "^1.2.4"
+                        "debug": "2.6.9",
+                        "iconv-lite": "0.4.21",
+                        "sax": "1.2.4"
                     }
                 },
                 "node-pre-gyp": {
@@ -4152,16 +4152,16 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "detect-libc": "^1.0.2",
-                        "mkdirp": "^0.5.1",
-                        "needle": "^2.2.0",
-                        "nopt": "^4.0.1",
-                        "npm-packlist": "^1.1.6",
-                        "npmlog": "^4.0.2",
-                        "rc": "^1.1.7",
-                        "rimraf": "^2.6.1",
-                        "semver": "^5.3.0",
-                        "tar": "^4"
+                        "detect-libc": "1.0.3",
+                        "mkdirp": "0.5.1",
+                        "needle": "2.2.0",
+                        "nopt": "4.0.1",
+                        "npm-packlist": "1.1.10",
+                        "npmlog": "4.1.2",
+                        "rc": "1.2.7",
+                        "rimraf": "2.6.2",
+                        "semver": "5.5.0",
+                        "tar": "4.4.1"
                     }
                 },
                 "nopt": {
@@ -4170,8 +4170,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "abbrev": "1",
-                        "osenv": "^0.1.4"
+                        "abbrev": "1.1.1",
+                        "osenv": "0.1.5"
                     }
                 },
                 "npm-bundled": {
@@ -4186,8 +4186,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "ignore-walk": "^3.0.1",
-                        "npm-bundled": "^1.0.1"
+                        "ignore-walk": "3.0.1",
+                        "npm-bundled": "1.0.3"
                     }
                 },
                 "npmlog": {
@@ -4196,10 +4196,10 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "are-we-there-yet": "~1.1.2",
-                        "console-control-strings": "~1.1.0",
-                        "gauge": "~2.7.3",
-                        "set-blocking": "~2.0.0"
+                        "are-we-there-yet": "1.1.4",
+                        "console-control-strings": "1.1.0",
+                        "gauge": "2.7.4",
+                        "set-blocking": "2.0.0"
                     }
                 },
                 "number-is-nan": {
@@ -4220,7 +4220,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "wrappy": "1"
+                        "wrappy": "1.0.2"
                     }
                 },
                 "os-homedir": {
@@ -4241,8 +4241,8 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "os-homedir": "^1.0.0",
-                        "os-tmpdir": "^1.0.0"
+                        "os-homedir": "1.0.2",
+                        "os-tmpdir": "1.0.2"
                     }
                 },
                 "path-is-absolute": {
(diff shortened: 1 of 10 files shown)


Francisco-Gamino:
cc: @TylerLeonhardt @anirudhgarg @eamonoreilly .

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -156,4 +159,18 @@ export class ScriptProjectCreateStep extends ProjectCreateStepBase {
...
+                version: '2.0',
+                managedDependency: {
+                    enabled: true
+                }
+            };
+        } else if (runtime === ProjectRuntime.v2) {
+            return { version: '2.0' };
+        }
+        return {};
+    }
```
TylerLeonhardt: This could be refactored into:

```ts
private getHostContent(wizardContext: IProjectWizardContext, runtime: ProjectRuntime): object {
    if (runtime === ProjectRuntime.v1) {
        return {};
    }

    if (wizardContext.managedDependencies) {
        return {
            version: '2.0',
            managedDependency: {
               enabled: true
            }
        };
    }

    return { version: '2.0' };
}
```

It's a bit cleaner and checks less things. Also it future proofs other languages that support managedDeps.

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -6,7 +6,7 @@
 import * as fse from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import { gitignoreFileName, hostFileName, localSettingsFileName, ProjectRuntime, proxiesFileName } from '../../../constants';
+import { gitignoreFileName, hostFileName, localSettingsFileName, ProjectLanguage, ProjectRuntime, proxiesFileName } from '../../../constants';
```
TylerLeonhardt: If you follow my comment below, I don't think you need this anymore.

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -45,16 +44,18 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
         ];
 
-        if (getFuncExtensionSetting('enablePowerShell')) {
-            languagePicks.push({ label: ProjectLanguage.PowerShell, description: previewDescription });
-        }
-
         const options: QuickPickOptions = { placeHolder: localize('selectFuncTemplate', 'Select a language for your function project') };
         wizardContext.language = <ProjectLanguage>(await ext.ui.showQuickPick(languagePicks, options)).label;
+
+        if (wizardContext.language === ProjectLanguage.PowerShell) {
```
TylerLeonhardt: This is fine for now but we might want to refactor this into a "Languages support manangedDeps" array or something eventually.

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -45,16 +44,18 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
             { label: ProjectLanguage.JavaScript },
             { label: ProjectLanguage.TypeScript },
             { label: ProjectLanguage.CSharp },
+            { label: ProjectLanguage.PowerShell, description: previewDescription },
             { label: ProjectLanguage.Python, description: previewDescription },
             { label: ProjectLanguage.Java }
         ];
 
-        if (getFuncExtensionSetting('enablePowerShell')) {
```
TylerLeonhardt: Double check with @anirudhgarg  that we should be doing this yet. (cc @joeyaiello)

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -106,6 +106,9 @@ local.settings.json
 # TypeScript output
 dist
 out
+
+# Managed dependencies folder
+ManagedDependencies/
```
EricJizbaMSFT: Would this make sense to put in the `# Azure Functions artifacts` section above?

EricJizbaMSFT: This should also go in the `.funcignore` file - which applies when deploying

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
Francisco-Gamino: Thanks for the suggestion @TylerLeonhardt. Fixed.

src/commands/createNewProject/NewProjectLanguageStep.ts
Francisco-Gamino: I've checked with @anirudhgarg and @eamonoreilly, and given that the next release will be at the beginning of May, they are fine with the change. (cc @joeyaiello)

TylerLeonhardt:
@EricJizbaMSFT The design decision is that the vscode extension should prompt for enabling ManageDeps because the scope is still small and it only applies to Azure modules.

cc @joeyaiello & @eamonoreilly

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -45,16 +44,18 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
-            languagePicks.push({ label: ProjectLanguage.PowerShell, description: previewDescription });
-        }
-
         const options: QuickPickOptions = { placeHolder: localize('selectFuncTemplate', 'Select a language for your function project') };
         wizardContext.language = <ProjectLanguage>(await ext.ui.showQuickPick(languagePicks, options)).label;
+
+        if (wizardContext.language === ProjectLanguage.PowerShell) {
+            const managedDependenciesOptions: QuickPickOptions = { placeHolder: localize('enableManagedDependencies', 'Would you like to install the Azure modules and have these automatically managed in Azure?') };
+            wizardContext.managedDependencies = (await (ext.ui.showQuickPick([{ label: InstallManagedDependencies.Yes }, { label: InstallManagedDependencies.No }], managedDependenciesOptions))).label === InstallManagedDependencies.Yes;
```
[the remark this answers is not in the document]
TylerLeonhardt: So this is a matter of:

1. Adding a new file like `ManagedDependenciesStep.ts` under the `createNewProject` folder (looking at one of the other `Step`s for guidance
2. Then look at [LocalAppSettingListStep](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/commands/createFunction/genericSteps/LocalAppSettingListStep.ts#L44-L81) for guidance on how to make SubWizard

Is that a good summary?

src/commands/createNewProject/NewProjectLanguageStep.ts
EricJizbaMSFT: I prefer to keep master branch "release ready", so let's hold off on removing the feature flag. Plans change - we might release early or PowerShell support might release late. It's easy enough to remove the flag a day or two before release

Francisco-Gamino:
@EricJizbaMSFT 
Adding to @TylerLeonhardt comment on the design.

In the Func CLI, we have enable Managed Dependencies via a switch, which the user can specify during **func init**. If the user doesn't specified this switch, we prompt them. 

image


For the Azure Function Extensions, we would like a similar experience.

EricJizbaMSFT:
I'm not quite following. We really try to analyze each and every prompt for these core "getting started" scenarios because each one could lead to drop-off for new users. If the scope of the feature is so small, I would say a potentially confusing prompt isn't worth it. What are the downsides to enabling it by default?

Francisco-Gamino:
@EricJizbaMSFT: We have decided to turn the feature on by default. I will update the PR accordingly. Thanks.

src/constants.ts
```diff
@@ -57,6 +62,7 @@ export const settingsFileName: string = 'settings.json';
 export const vscodeFolderName: string = '.vscode';
 export const gitignoreFileName: string = '.gitignore';
 export const profileps1FileName: string = 'profile.ps1';
+export const requirementspsd1FileName: string = 'requirements.psd1';
```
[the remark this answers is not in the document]
Francisco-Gamino: I've moved both ```profileps1FileName and requirementspsd1FileName``` to ```PowerShellProjectCreateStep.ts```.

src/constants.ts
```diff
@@ -48,6 +48,11 @@ export enum Platform {
     Linux = 'linux'
 }
 
+export enum InstallManagedDependencies {
+    Yes = 'Yes',
+    No = 'No'
```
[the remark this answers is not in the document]
Francisco-Gamino: These are not longer needed.

src/commands/createNewProject/NewProjectLanguageStep.ts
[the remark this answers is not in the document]
TylerLeonhardt: SInce we're turning it on by default, we don't need this anymore :)

src/commands/createNewProject/NewProjectLanguageStep.ts
EricJizbaMSFT: As another reason to delay removal of the feature flag, we have [this issue](https://github.com/Microsoft/vscode-azurefunctions/issues/1082) tracking it and ideally all of the bullet points are done first

Francisco-Gamino: @EricJizbaMSFT : All the back end work is done. Managed Dependencies is already available in the Portal, and will be available in the next CLI release.

src/commands/createNewProject/NewProjectLanguageStep.ts
Francisco-Gamino: We are turning Managed Dependencies on by default, we no longer need this check nor the prompt.

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
Francisco-Gamino: We still need ProjectLanguage since Managed Dependencies is only supported in PowerShell.

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
Francisco-Gamino: We have changed the location where function app dependencies are installed to LocalApplicationData\AzureFunctions\FunctionAppName\ManagedDependencies, so we no longer need modify this file.

Francisco-Gamino: We no longer need to make this change. Please see above comment.

src/commands/createNewProject/NewProjectLanguageStep.ts
EricJizbaMSFT: @Francisco-Gamino the feature flag isn't related to managed dependencies, though. It's there because PowerShell is still in private preview in Azure (Correct? I don't see the runtime option when creating a function app) and because the items listed in [this issue](https://github.com/Microsoft/vscode-azurefunctions/issues/1082) have not been done yet.

Screen Shot 2019-04-10 at 9 48 53 AM

package-lock.json
```diff
@@ -4363,7 +4391,7 @@
                     "dev": true,
                     "optional": true,
                     "requires": {
-                        "string-width": "^1.0.2"
+                        "string-width": "1.0.2"
```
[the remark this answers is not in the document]
Francisco-Gamino: I am using 6.4.1
```
PS E:\GH\vscode-azurefunctions> npm --version
6.4.1
```

src/commands/createNewProject/NewProjectLanguageStep.ts
Francisco-Gamino: Thanks @EricJizbaMSFT for checking. PowerShell is supposed to be available for Public Preview in the first week of May. The Host (runtime), PowerShell worker and CLI changes are done. I am trying to finish the VSCode extension work as I am going on vacation starting 4/12. I will let Eamon and Anirud comment on the Portal changes.

@eamonoreilly @anirudhgarg: when is PowerShell going to be available as part of the runtime options?

Francisco-Gamino:
@EricJizbaMSFT @TylerLeonhardt : Could you please take another look? Thanks!

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -156,4 +156,22 @@ export class ScriptProjectCreateStep extends ProjectCreateStepBase {
             await fse.writeFile(funcIgnorePath, this.funcignore.sort().join(os.EOL));
         }
     }
+
+    private getHostContent(wizardContext: IProjectWizardContext, runtime: ProjectRuntime): object {
+        if (runtime === ProjectRuntime.v2) {
+            if (wizardContext.language === ProjectLanguage.PowerShell) {
```
TylerLeonhardt: We really shouldn't check the language... What we should have is either a Map or List of languages that _support_ managed dependencies.

I know that PowerShell is the only supported language, but it would be more readable and future proofing to invest in that now.

That said, if @EricJizbaMSFT is fine with this for now, I won't block on this.

However, I will block on at least adding a comment that says this is for managed dependencies.

EricJizbaMSFT: Yeah I would suggest you add a property to this class:
```
protected supportsManagedDependencies: boolean = false;
```

And then in `PowerShellProjectCreateStep.ts` overwrite it with:
```
protected supportsManagedDependencies: boolean = true;
```

src/commands/createNewProject/NewProjectLanguageStep.ts
EricJizbaMSFT: I'm happy to remove the feature flag myself closer to our release date if you'll be OOF. I want to make sure all the docs/tests are up to par with our other languages before that happens. We already have the above issue in our current milestone, so it won't be forgotten.

On the other hand I'm not as familiar with the managed dependency stuff so I would rather get this PR in before you're OOF.

src/commands/createNewProject/ProjectCreateStep/PowerShellProjectCreateStep.ts
```diff
@@ -41,5 +51,10 @@ export class PowerShellProjectCreateStep extends ScriptProjectCreateStep {
         if (await confirmOverwriteFile(profileps1Path)) {
             await fse.writeFile(profileps1Path, profileps1);
         }
+
+        const requirementspsd1Path: string = path.join(wizardContext.projectPath, requirementspsd1FileName);
+        if (await confirmOverwriteFile(requirementspsd1Path)) {
```
EricJizbaMSFT: We have some existing tests for creating a project. Just search the 'test' folder in this repo for `profile.ps1` and copy each line and replace it with `requirements.psd1`

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
Francisco-Gamino: @EricJizbaMSFT: Thanks for the suggestion. I've updated the code accordingly.

src/commands/createNewProject/ProjectCreateStep/PowerShellProjectCreateStep.ts
Francisco-Gamino: I have added a new check to ensure ```requirements.psd1``` gets created.

Francisco-Gamino:
@EricJizbaMSFT and @TylerLeonhardt : Could you please take another look? Thanks!

Francisco-Gamino:
@EricJizbaMSFT : I have pushed a new commit reverting back the PowerShell feature flag. Thanks for all your help!

Francisco-Gamino:
@EricJizbaMSFT: Yes please. Thanks again!

Issue #775, opened 2018-10-31
Improved Functions binding editing experience


fiveisprime:
Provide some kind of button (similar to the "add configuration" that's visible) to add a binding. Provide some intellisense for bindings; I believe there's a schema that we could leverage.

CC and thanks to @asavaritayal for the spec

EricJizbaMSFT:
VS Code issue to add floating buttons in the the editor: https://github.com/Microsoft/vscode/issues/70813

EricJizbaMSFT:
Closing this as fixed in https://github.com/Microsoft/vscode-azurefunctions/pull/1162

The entry points are not super obvious (see screenshots in PR), but that will be improved in https://github.com/Microsoft/vscode-azurefunctions/issues/1163

PR #1162, opened 2019-04-08
Allow user to add bindings

EricJizbaMSFT:
Current entry points are:
Screen Shot 2019-04-08 at 4 23 02 PM
Screen Shot 2019-04-08 at 4 23 09 PM

I also added preview support for a tree view of the local project, which IMO is a much more discoverable entry point. Unfortunately, I don't think this tree-view is fleshed-out enough to show by default, but I think it would be too big of a PR if I waited for that to happen. If user sets `azureFunctions.enableProjectTree` to "true", they will see this:
Screen Shot 2019-04-08 at 4 22 52 PM

Related: #775


src/commands/addBinding/BindingNameStep.ts
```diff
@@ -0,0 +1,56 @@
...
+export class BindingNameStep extends StringPromptStep {
+    private _functionConfig: FunctionConfig | undefined;
+
+    public async getDefaultValue(wizardContext: IBindingWizardContext): Promise<string | undefined> {
+        const defaultValue: string | undefined = await super.getDefaultValue(wizardContext);
+        if (defaultValue) {
+            let uniqueValue: string = defaultValue;
+
+            let count: number = 1;
+            while (count < 100) {
```
nturinski: I feel like 100 should be an optional parameter or something.  100 seems pretty arbitrary-- granted I'm not a Functions developer so I dunno if anyone would ever hit the limit 😋

src/commands/createFunction/genericSteps/StringPromptStep.ts
```diff
@@ -7,25 +7,33 @@ import { AzureWizardPromptStep } from "vscode-azureextensionui";
...
     }
 
-    public async prompt(wizardContext: IFunctionWizardContext): Promise<void> {
+    public async prompt(wizardContext: IBindingWizardContext): Promise<void> {
         wizardContext[this._setting.name] = await ext.ui.showInputBox({
             placeHolder: this._setting.label,
             prompt: this._setting.description || localize('stringSettingPrompt', 'Provide a \'{0}\'', this._setting.label),
-            validateInput: (s: string): string | undefined => this._setting.validateSetting(s),
-            value: this._setting.defaultValue
+            validateInput: async (s): Promise<string | undefined> => await this.validateInput(wizardContext, s),
```
nturinski: nit: No typedef for s

src/extension.ts
```diff
@@ -159,3 +164,26 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 // tslint:disable-next-line:no-empty
 export function deactivateInternal(): void {
 }
+
+/**
+ * This has some known limitations that will need to be addressed before removing the feature flag
+ * See https://github.com/Microsoft/vscode-azurefunctions/issues/1163
+ */
+async function getProjectTreeItems(context: vscode.ExtensionContext): Promise<AzureParentTreeItem[]> {
```
nturinski: Doesn't this deserve its own file?  Seems weird to be in the extension.ts

src/templates/parseScriptTemplates.ts
```diff
@@ -162,16 +176,19 @@ export function parseScriptTemplate(rawTemplate: IRawTemplate, resources: IResou
...
+                const bindingTemplate: IBindingTemplate | undefined = bindingTemplates.find(b => b.type === inBinding.type);
+                if (bindingTemplate) {
+                    const setting: IFunctionSetting | undefined = bindingTemplate.settings.find((bs: IFunctionSetting) => bs.name === settingName);
+                    if (setting) {
+                        const functionSpecificDefaultValue: string | undefined = inBinding[setting.name];
+                        if (functionSpecificDefaultValue) {
+                            // overwrite common default value with the function-specific default value
+                            setting.defaultValue = functionSpecificDefaultValue;
+                        }
+                        userPromptedSettings.push(setting);
```
nturinski: Pushing `undefined` is intentional here, right?

src/tree/FunctionTreeItem.ts
```diff
@@ -24,7 +24,7 @@ export class FunctionTreeItem extends AzureTreeItem<ISiteTreeRoot> {
     public readonly config: FunctionConfig;
 
     private readonly _name: string;
-    private _triggerUrl: string;
+    private _triggerUrl: string | undefined;
```
nturinski: I'm confused about this change.  Why not keep using `isHttpTrigger` to verify whether or not you can call triggerUrl?

src/tree/localProject/LocalFunctionsTreeItem.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as path from 'path';
+import { AzureParentTreeItem, AzureTreeItem, createTreeItemsWithErrorHandling } from 'vscode-azureextensionui';
+import { functionJsonFileName } from '../../constants';
+import { FunctionConfig } from '../../FunctionConfig';
+import { localize } from '../../localize';
+import { nodeUtils } from '../../utils/nodeUtils';
+import { IProjectRoot } from './IProjectRoot';
+import { LocalFunctionTreeItem } from './LocalFunctionTreeItem';
+
+export class LocalFunctionsTreeItem extends AzureParentTreeItem<IProjectRoot> {
```
nturinski: Were you going to create a different superclass for these `Local...TreeItems`?

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -0,0 +1,73 @@
...
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzureTreeItem<IProjectRoot>[]> {
+        return [this._localFunctionsTreeItem];
+    }
+
+    public isAncestorOfImpl(contextValue: string | RegExp): boolean {
+        return typeof contextValue === 'string' && /^azFuncLocal/.test(contextValue);
```
nturinski: Yay! RegExp actually came in handy!

src/commands/createFunction/genericSteps/StringPromptStep.ts
EricJizbaMSFT: Yeah I turned that rule off a while ago because it was too annoying: https://github.com/Microsoft/vscode-azurefunctions/blob/master/tslint.json#L78

This specific case is a little weird, but it's mostly for cases [like this](https://github.com/Microsoft/vscode-azurefunctions/blob/1200241b475bad976204892b68c3949f33a7db8b/src/debug/getFuncTaskCommand.ts#L22):
```typescript
const x: string[] = ['hello', 'world'];
x.find((s: string) => s === 'world'); // <-- gross. and obvious that s will be a string
x.find(s => s === 'world'); // <-- 😊
```

Let me know if you strongly feel we should turn that back on.

src/extension.ts
EricJizbaMSFT: Yeah I can move it

src/tree/localProject/LocalFunctionsTreeItem.ts
EricJizbaMSFT: Ideally yes - that was related to our discussion about `TreeItemPlus`. But it's too much work/refactoring that I don't want to do it before build

src/tree/localProject/LocalProjectTreeItem.ts
EricJizbaMSFT: Of course it did!

src/commands/addBinding/BindingNameStep.ts
EricJizbaMSFT: Oh it's super arbitrary. I'll make it a `const` and name it `maxTries`, but I'm not gonna do an optional parameter because no one is every gonna pass in a different value. The goal was just to pick a number that is so large it'll probably never be hit for this scenario but so that there's no possibility of an infinite loop if there's some sort of bug in `bindingExists`

src/templates/parseScriptTemplates.ts
EricJizbaMSFT: Uhhh no? I have `if (setting) {` a few lines up

src/tree/FunctionTreeItem.ts
EricJizbaMSFT: I changed `FunctionConfig` to be a lot more permissive since I was using that for local function.json files that the user can edit to a lot of weird things. As part of that, triggerUrl will be undefined if it's not an http trigger instead of throwing an error.

If I do the following, tslint will complain because `writeText` needs `string`, not `string | undefined`
if (node.config.isHttpTrigger) {
        await vscode.env.clipboard.writeText(node.triggerUrl);
}
I _could_ check `isHttpTrigger` _and_ `!node.triggerUrl`, but it was simpler to just check the second one

src/templates/parseScriptTemplates.ts
nturinski: Ah yes, my bad. Misread the curlies ➰

PR #1174, opened 2019-04-10
Improve warning when creating a function with no workspace open

EricJizbaMSFT:
Before:
Screen Shot 2019-04-10 at 11 32 56 AM

After:
Screen Shot 2019-04-10 at 11 31 36 AM

### DIFF
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -3,10 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { window, workspace, WorkspaceFolder } from 'vscode';
+import { commands, MessageItem, Uri, window, workspace, WorkspaceFolder } from 'vscode';
 import { AzureWizard, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { ProjectLanguage, ProjectRuntime } from '../../constants';
 import { NoWorkspaceError } from '../../errors';
+import { ext } from '../../extensionVariables';
 import { addLocalFuncTelemetry } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
 import { localize } from '../../localize';
 import { verifyAndPromptToCreateProject } from '../createNewProject/verifyIsProject';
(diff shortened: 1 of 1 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -54,3 +43,42 @@ export async function createFunction(
...
+    let folder: WorkspaceFolder | undefined;
+    if (!workspace.workspaceFolders || workspace.workspaceFolders.length === 0) {
+        const message: string = localize('noWorkspaceWarning', 'You must have a project open to create a function.');
+        const newProject: MessageItem = { title: localize('createNewProject', 'Create new project') };
+        const openExistingProject: MessageItem = { title: localize('openExistingProject', 'Open existing project') };
+        const result: MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, newProject, openExistingProject);
+
+        if (result === newProject) {
+            // don't wait
+            commands.executeCommand('azureFunctions.createNewProject');
```
PrashanthCorp: I'm wondering why you're using `executeCommand`  instead of calling the function (`createNewProject`) directly.

src/commands/createFunction/createFunction.ts
```diff
@@ -54,3 +43,42 @@ export async function createFunction(
...
+                canSelectMany: false,
+                openLabel: localize('open', 'Open')
+            });
+            // don't wait
+            commands.executeCommand('vscode.openFolder', uri[0]);
+            actionContext.properties.noWorkspaceResult = 'openExistingProject';
+        }
+
+        actionContext.suppressErrorDisplay = true;
+        throw new NoWorkspaceError();
```
PrashanthCorp: If you've opened  project would you still have to throw NoWorkspaceError?

EricJizbaMSFT: Opening a folder restarts the extension host, so they have to re-start creating a function. I'm just throwing the error so I can track it in telemetry

src/commands/createFunction/createFunction.ts
EricJizbaMSFT: Mostly I want a separate telemetry event instead of including it with the createFunction event

PR #1188, opened 2019-04-15
Make virtual environment optional when creating Python project

### LINKED ISSUE #1092: Standardize virtual env creation Python Functions
Python 3.7 should be supported soon and we might not want to create a virtual environment any more. See: https://github.com/Azure/azure-functions-core-tools/issues/923

EricJizbaMSFT:
The func cli simplified their requirements for a python project. They no longer require a set of default dependencies or a virtual environment. Because of that, I don't see any reason to require the func cli for project creation any more.

I added a setting to turn off venv creation and kept the default at `true` for a number of reasons:
1. We have existing logic to pick the python alias that is 3.6.x and use that for the venv. A project without a venv will use the version from the default `python` alias, which might not match 3.6.x and might fail to debug. This sounds bad and IMO is a regression since the old behavior would always create a working project.
1. A venv lets people manage dependencies on a project level instead of machine-wide. I think this is just good practice. For example, it helps teams collaborating through GitHub since we will prompt to re-create the venv and automatically run `pip install` based on the `requirements.txt` file that should be checked in to source control

User will see this prompt when creating a project:
Screen Shot 2019-04-15 at 1 46 03 PM
Screen Shot 2019-04-15 at 1 46 06 PM

Also added a python-specific gitignore.

Fixes #1092


src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
```diff
@@ -0,0 +1,108 @@
...
+    // tslint:disable-next-line: no-use-before-declare
+    protected gitignore: string = nodeGitignore;
+
+    constructor() {
+        super();
+        this.funcignore.push('*.js.map', '*.ts', 'tsconfig.json');
+    }
+}
+
+// https://raw.githubusercontent.com/github/gitignore/master/Node.gitignore
```
nturinski: What do you think about grabbing this dynamically?  I know that there are some downsides to it, but we wouldn't have to maintain this list.

EricJizbaMSFT: We would have to have a backup anyways since we've seen plenty of issues with corporate proxies, etc. where pulling something dynamically from the internet doesn't always work. Doesn't seem worth it to me - but you're welcome to file an issue and we can discuss further in triage if you think it's valuable.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
```diff
@@ -4,42 +4,69 @@
...
+const oldRequirements: string = `azure-functions==1.0.0b3
+azure-functions-worker==1.0.0b4
+grpcio==1.14.2
+grpcio-tools==1.14.2
+protobuf==3.7.1
+six==1.12.0
+`;
+
+export class PythonProjectCreateStep extends ScriptProjectCreateStep {
+    // tslint:disable-next-line: no-use-before-declare
```
nturinski: Why is this needed?? I must not be understanding this correctly, but I don't see it getting used before you've declared it.

EricJizbaMSFT: tslint wants `pythonGitignore` defined above `PythonProjectCreateStep` in this file. I think that's annoying because I care much more about `PythonProjectCreateStep`

EricJizbaMSFT:
Yeah honestly there's quite a few ways that users might run into problems if they don't create a virtual environment. For example, if the default `python` in their path is v2.7. I guess I'm okay with current experience since python is in preview and most users don't change settings anyways. Might consider fixing after build if a lot of people run into this

PR #1190, opened 2019-04-16
Add tests for deploying CSharp and JavaScript projects

v-wuzhai:
Deploy the CSharp and JavaScript projects to Function App (Basic).


package.json
```diff
@@ -939,6 +939,7 @@
         "azure-arm-storage": "^4.0.0",
         "azure-arm-website": "^5.3.0",
         "azure-storage": "^2.10.3",
+        "clipboardy": "^2.0.0",
```
nturinski: Since this is being used only in the test suites, this should be in the devDependecies

test/createAzureResource.test.ts
```diff
@@ -119,18 +145,57 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         assert.ifError(deletedApp);
     });
 
+    test('Deploy CSharp project (windows)', async () => {
+        await clearnewprojectfile(projectPath);
```
nturinski: Define `resourceName3` here since this is the only place it is being used.

test/createAzureResource.test.ts
```diff
@@ -154,3 +219,16 @@ function getWebsiteManagementClient(testAccount: TestAzureAccount): WebSiteManag
 function getResourceManagementClient(testAccount: TestAzureAccount): ResourceManagementClient {
     return new ResourceManagementClient(testAccount.getSubscriptionCredentials(), testAccount.getSubscriptionId());
 }
+
+async function clearnewprojectfile(projectpath: string): Promise<void> {
```
nturinski: Please camelCase function and variable names.
`clearNewProjectFile` and `projectPath`

test/createAzureResource.test.ts
```diff
@@ -58,14 +59,39 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         ext.tree.dispose();
     });
 
-    test('createFunctionApp (Basic)', async () => {
+    test('Create windows function app (Basic) and deploy JavaScript project', async () => {
```
nturinski: Make `CreateFunctionApp` and `Deploy Javascript project` tests two separate tests.

test/msdelay.ts
```diff
@@ -0,0 +1,8 @@
+/*---------------------------------------------------------------------------------------------
```
nturinski: Since we use delay in `pickFuncProcess.ts`, you should make a `delay.ts` file in the `util` folder, remove it from `pickFuncProcess.ts,` and then export the util file and use it in both your test and `pickFuncProcess`

test/createAzureResource.test.ts
```diff
@@ -58,14 +59,39 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+            });
         });
+
+        ext.ui = new TestUserInput([resourceName1, 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(5000);
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+
+        clipboard.writeSync(typeof (undefined));
```
nturinski: Could you leave a comment for the purpose of this line?

test/createAzureResource.test.ts
```diff
@@ -119,18 +145,57 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        await runWithFuncSetting('projectLanguage', ProjectLanguage.CSharp, async () => {
+            await runWithFuncSetting('advancedCreation', undefined, async () => {
+                ext.ui = new TestUserInput([resourceName3]);
+                await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+                const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceName3, resourceName3);
+                assert.ok(createdApp);
+                await runWithFuncSetting('projectRuntime', ProjectRuntime.v2, async () => {
+                    const templateId: string = 'Azure.Function.CSharp.HttpTrigger.2.x';
+                    const namespace: string = 'Company.Function';
+                    const accessrights: string = 'function';
```
nturinski: camelCase

test/createAzureResource.test.ts
```diff
@@ -58,14 +59,39 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        await runWithFuncSetting('projectLanguage', ProjectLanguage.JavaScript, async () => {
+            await runWithFuncSetting('advancedCreation', undefined, async () => {
+                ext.ui = new TestUserInput([resourceName1]);
+                await vscode.commands.executeCommand('azureFunctions.createFunctionApp');
+                const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceName1, resourceName1);
+                assert.ok(createdApp);
+                await runWithFuncSetting('projectRuntime', ProjectRuntime.v2, async () => {
+                    const templateId: string = 'HttpTrigger-JavaScript';
+                    const authLevel: string = 'function';
+                    await vscode.commands.executeCommand('azureFunctions.createNewProject', projectPath, ProjectLanguage.JavaScript, ProjectRuntime.v2, false, templateId, functionName, { aUtHLevel: authLevel });
```
nturinski: nit: aUtHLevel has very weird casing

test/createAzureResource.test.ts
```diff
@@ -58,14 +59,39 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(5000);
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+
+        clipboard.writeSync(typeof (undefined));
+        ext.ui = new TestUserInput([resourceName1, functionName]);
+        await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
+        assert.ok(clipboard.readSync() !== 'undefined', `The value of the copy function url should not be ${clipboard.readSync()}`);
+        assert.ok(clipboard.readSync().includes(resourceName1));
```
nturinski: Can you get what the exact url should be for the function trigger?  Is there a way to use the Azure webSiteClient to check it?

test/createAzureResource.test.ts
```diff
@@ -58,14 +57,74 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        ext.ui = new TestUserInput([resourceName1, functionName]);
+        await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
+        assert.ok(clipboard.readSync() !== 'undefined', `The value of the copy function url should not be ${clipboard.readSync()}`);
+
+        const result: string = execSync(`curl ${clipboard.readSync()}"&"name=${resourceName1}`).toString();
+        assert.equal(result, `Hello ${resourceName1}`, `The return value should be Hello ${resourceName1}`);
+    });
+
+    test('Deploy CSharp project (windows)', async () => {
+        const resourceName3: string = getRandomHexString().toLowerCase();
```
nturinski: Could you change this to resourceName2 and the one below to resourceName3?  It's a bit confusing that they are out of order.

test/createAzureResource.test.ts
```diff
@@ -58,14 +57,74 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        ext.ui = new TestUserInput([resourceName1, 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(5000);
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+        // Initialize the contents of the clipboard as "Undefined" and then get function URL by command
+        clipboard.writeSync(typeof (undefined));
+        ext.ui = new TestUserInput([resourceName1, functionName]);
+        await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
+        assert.ok(clipboard.readSync() !== 'undefined', `The value of the copy function url should not be ${clipboard.readSync()}`);
```
nturinski: Is this checking that `clibboard.readSync()` is undefined or "undefined" as in the string?  I'm not sure if `readSync` still returns a string even if the contents are `undefined`.

test/createAzureResource.test.ts
```diff
@@ -58,14 +57,74 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+
+        ext.ui = new TestUserInput([resourceName1, 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(5000);
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+        // Initialize the contents of the clipboard as "Undefined" and then get function URL by command
+        clipboard.writeSync(typeof (undefined));
+        ext.ui = new TestUserInput([resourceName1, functionName]);
+        await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
```
nturinski: Sorry to have back track (I know it must be frustrating) but I think that we shouldn't test `azureFunctions.copyFunctionUrl`.

In the code below, we are leveraging the vscode API so we are actually testing VS Code's functionality, not really the extension.

https://github.com/microsoft/vscode-azurefunctions/blob/15d06f939593cb58b217b6d6d3c3470e26493203/src/commands/copyFunctionUrl.ts#L16-L18

You should just access the `triggerUrl` property instead.  In the error message on L91, you should include the `triggerUrl` so we can see if it was properly generated in the error message in case of failure.

We don't want to use `clipboardy` because we have to install another package for one test case.  Also, there is no guarantee that `clipboardy` will always work and may cause false positives.

test/createAzureResource.test.ts
```diff
@@ -58,14 +57,74 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+                });
+            });
+        });
+
+        ext.ui = new TestUserInput([resourceName1, 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(5000);
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+        // Initialize the contents of the clipboard as "Undefined" and then get function URL by command
```
nturinski: As an overall comment, could you leave more new spaces between assertions?  That way it is very visually clear where the next portion of the test is and makes it easier for us to read 😃

test/createAzureResource.test.ts
v-wuzhai: Do you mean we shouldn't use "azureFunctions.copyFunctionUrl" command?
       If "YES", then whether we use the triggerUrl property to get the URL or not? How to use the triggerUrl? Can you give us some guidance?
       If "NO", then don't we need to verify 
"await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl')"? Do we need to get the URL via "await vscode.env.clipboard.readText()" for next step?

nturinski: Yes, we shouldn't use `azureFunctions.copyFunctionUrl`.  Instead, you should get the treeItem and access its `triggerUrl` property.  You should be able to use `ext.tree.findTreeItem` by using the full id.  I believe that the id can be constructed with the function app that you retrieve.

Then we can uninstall `clipboardy`.

test/createAzureResource.test.ts
```diff
@@ -6,30 +6,30 @@
...
 
 // tslint:disable-next-line:max-func-body-length
 suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Promise<void> {
     this.timeout(1200 * 1000);
     const resourceGroupsToDelete: string[] = [];
     const testAccount: TestAzureAccount = new TestAzureAccount();
-    let oldProjectLanguage: ProjectLanguage | undefined;
     let webSiteClient: WebSiteManagementClient;
     const resourceName1: string = getRandomHexString().toLowerCase();
+    const projectPath: string = path.join('D:', 'testOutput');
```
nturinski: This line is causing problems. While this will work for Windows, the D drive doesn't work with Linux and Mac.

image

I think you could use the `os` library for this.

`import * as os from 'os';`
And then call `os.tmpdir();`

There might be a better command, feel free to investigate 😃

test/createAzureResource.test.ts
```diff
@@ -59,26 +57,82 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+                    await validateProject(projectPath, getCSharpValidateOptions('testOutput', 'netcoreapp2.1'));
+                });
+            });
+        });
+
+        ext.ui = new TestUserInput([testAccount.getSubscriptionId(), resourceName2, 'Update remote runtime', 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(500);
+
+        //Verify the deployment result through triggerUrl
```
nturinski: nit: Space after //

test/createAzureResource.test.ts
```diff
@@ -59,26 +57,82 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        const fullId: string = <string>(getFunction[0].id);
+        const triggerUrl: string | undefined = (<FunctionTreeItem>await ext.tree.findTreeItem(fullId)).triggerUrl;
+        const result: string = execSync(`curl ${triggerUrl}"&"name=${resourceName2}`).toString();
+        assert.equal(result, `Hello, ${resourceName2}`, `The result should be "Hello, ${resourceName2}" rather than ${result} and the triggerUrl is ${triggerUrl}`);
     });
 
     test('createFunctionApp (Advanced)', async () => {
-        const resourceName2: string = getRandomHexString();
+        const resourceName3: string = getRandomHexString();
         const resourceGroupName: string = getRandomHexString();
```
nturinski: It doesn't seem like there's a need to have resourceGroupName and resourceName3.  In this test, just replace resourceGroupName with resourceName3.  I'd also just make it all lowercase so you can use it for the storageAccountName.

test/test.code-workspace
```diff
@@ -0,0 +1,10 @@
+{
+    "folders": [
+        {
+            "path": "D:/testOutput"
```
nturinski: This will have to change as well.  It might be harder since it's settings file, but I think you can do dynamic vscode settings.  Here are some references: https://code.visualstudio.com/docs/editor/variables-reference

test/createAzureResource.test.ts
v-wuzhai: The previous code is the same as your idea, but Eric added a comment to "createFunctionApp (Advanced)", he said it's a little too similar to basic create. You can see https://github.com/microsoft/vscode-azurefunctions/pull/1128#discussion_r273752258

test/createAzureResource.test.ts
```diff
@@ -6,30 +6,34 @@
...
 // tslint:disable-next-line:max-func-body-length
 suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Promise<void> {
     this.timeout(1200 * 1000);
     const resourceGroupsToDelete: string[] = [];
     const testAccount: TestAzureAccount = new TestAzureAccount();
-    let oldProjectLanguage: ProjectLanguage | undefined;
     let webSiteClient: WebSiteManagementClient;
     const resourceName1: string = getRandomHexString().toLowerCase();
+    // Get the *.code-workspace workspace file path
+    const projectPath: string = vscode.workspace.rootPath || os.tmpdir();
```
nturinski: Can you do the same thing as here?

https://github.com/microsoft/vscode-azurefunctions/blob/master/test/global.test.ts#L15-L23

test/createAzureResource.test.ts
```diff
@@ -6,30 +6,32 @@
...
     this.timeout(1200 * 1000);
     const resourceGroupsToDelete: string[] = [];
     const testAccount: TestAzureAccount = new TestAzureAccount();
-    let oldProjectLanguage: ProjectLanguage | undefined;
     let webSiteClient: WebSiteManagementClient;
     const resourceName1: string = getRandomHexString().toLowerCase();
+    // Get the *.code-workspace workspace file path
+    const projectPath: string = getTestRootFolder();
+    // tslint:disable-next-line: prefer-const
+    let context: IActionContext | undefined;
```
nturinski: I want to wait until Eric comes back from being OOF, but I think that this is not a good solution to not having the `IActionContext`.  I'll discuss with him offline and then update this comment.

v-wuzhai: I think the IActionContext does not actually sense, so handled it like this. But you should be right, so i wait for your reply. Thanks a lot.

test/createAzureResource.test.ts
```diff
@@ -59,26 +59,83 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+                });
+            });
         });
+
+        ext.ui = new TestUserInput([resourceName1, 'Deploy']);
+        await vscode.commands.executeCommand('azureFunctions.deploy');
+        await delay(500);
+
+        // Verify the deployment result through triggerUrl
+        const getFunction: WebSiteManagementModels.FunctionEnvelopeCollection = await webSiteClient.webApps.listFunctions(resourceName1, resourceName1);
```
EricJizbaMSFT: It looks like you and Nathan have gone back and forth on this, but I think it's better to use `'azureFunctions.copyFunctionUrl'` to get the url. It seems simpler and I think it tests more of the real-life scenario for a user. That being said, I agree with Nathan that we don't want you to install "clipboardy" so just use the `vscode.env.clipboard` apis like this:
```
await vscode.env.clipboard.writeText(''); // clear the clipboard
ext.ui = new TestUserInput(...);
await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
const triggerUrl = await vscode.env.clipboard.readText();
```

@nturinski are you fine with that? I feel like we might've discussed this offline, but I don't remember exactly what we talked about.

test/createAzureResource.test.ts
```diff
@@ -59,26 +59,83 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        assert.equal(getFunction[0].name, `${resourceName1}/${functionName}`);
+        const fullId: string = <string>(getFunction[0].id);
+        const triggerUrl: string | undefined = (<FunctionTreeItem>await ext.tree.findTreeItem(fullId, <IActionContext>context)).triggerUrl;
+        const result: string = execSync(`curl ${triggerUrl}"&"name=${resourceName1}`).toString();
```
EricJizbaMSFT: node.js already has ways to make a request without using a cli tool like `curl`. Take a look at something like this: https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/getCliFeedJson.ts#L45

test/global.test.ts
```diff
@@ -11,6 +11,42 @@ import * as vscode from 'vscode';
...
 import { ext, getRandomHexString, getTemplateProvider, parseError, TemplateProvider, TemplateSource, TestUserInput } from '../extension.bundle';
 
+export namespace constants {
+    export const testOutputName: string = 'testOutput';
+}
+
+// The root workspace folder that vscode is opened against for tests
+let testRootFolder: string;
+
+export function getTestRootFolder(): string {
```
EricJizbaMSFT: I'd prefer to move this into `createAzureResource.test.ts`. Mostly because that's the only place using it and `testRootFolder` is named very similar to `testFolderPath` below which is pretty confusing.

As a side note: `createAzureResource.test.ts` is getting kinda big and I want to split that into separate files eventually, but we should do that in a separate PR all by itself

test/createAzureResource.test.ts
nturinski: @EricJizbaMSFT  I'm good with that.  I think what we talked about was in the context of the Storage extension, but basically using copy was fine, but we don't need a test dedicated to the copy commands.

test/createAzureResource.test.ts
```diff
@@ -59,26 +58,71 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+                    const functionUrl: string = await vscode.env.clipboard.readText();
+                    const result: string = await getBody(functionUrl, resourceName1);
+                    assert.equal(result, `Hello ${resourceName1}`, `The result should be "Hello ${resourceName1}" rather than ${result} and the triggerUrl is ${functionUrl}`);
+                });
+            });
+        });
+    });
+
+    test('Deploy CSharp project (windows)', async () => {
+        const resourceName2: string = getRandomHexString().toLowerCase();
```
EricJizbaMSFT: This test and the javascript test have a lot of duplicate code. Can you create a function that looks similar to this:
```
async function testCreateProjectAndDeploy(createProjectInputs: string[], projectValidateOptions: IValidateProjectOptions, <any other parameters>): Promise<void> {
```
and then call that function from each test.

nturinski:
🍾 🎉

test/createAzureResource.test.ts
```diff
@@ -176,7 +174,7 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         await vscode.commands.executeCommand('azureFunctions.copyFunctionUrl');
         const functionUrl: string = await vscode.env.clipboard.readText();
         const result: string = await getBody(functionUrl, resourceName);
-        return [result, functionUrl];
+        assert.equal(result, expectResult, `The result should be "${expectResult}" rather than ${result} and the triggerUrl is ${functionUrl}`);
```
EricJizbaMSFT: No need to pass `expectResult`. This should just be:
```
assert.equal(result, `Hello, ${resourceName}`);
```

PR #1196, opened 2019-04-18
Make "Init for VS Code" less destructive

EricJizbaMSFT:
As opposed to overwriting the "tasks.json" and "launch.json" files if they already exist, we will just add our config to those files. I think this is especially necessary when users have their Functions project in a sub directory because that means they could have existing debug configs for other projects in other sub directories.

### DIFF
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -10,6 +10,7 @@ import { NoWorkspaceError } from '../../errors';
 import { ext } from '../../extensionVariables';
 import { addLocalFuncTelemetry } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
 import { localize } from '../../localize';
+import { getContainingWorkspace } from '../../utils/workspace';
 import { verifyInitForVSCode } from '../../vsCodeConfig/verifyInitForVSCode';
 import { verifyAndPromptToCreateProject } from '../createNewProject/verifyIsProject';
 import { FunctionListStep } from './FunctionListStep';
@@ -25,8 +26,12 @@ export async function createFunction(
     runtime?: ProjectRuntime): Promise<void> {
     addLocalFuncTelemetry(actionContext);
 
+    let workspaceFolder: WorkspaceFolder | undefined;
     if (workspacePath === undefined) {
-        workspacePath = await getWorkspacePath(actionContext);
+        workspaceFolder = await getWorkspacePath(actionContext);
+        workspacePath = workspaceFolder.uri.fsPath;
+    } else {
+        workspaceFolder = getContainingWorkspace(workspacePath);
     }
 
     const projectPath: string | undefined = await verifyAndPromptToCreateProject(actionContext, workspacePath);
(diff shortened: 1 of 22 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -36,15 +41,15 @@ export async function createFunction(
...
+    const wizardContext: IFunctionWizardContext = { actionContext, projectPath, workspacePath, workspaceFolder, runtime, language, functionName };
     const wizard: AzureWizard<IFunctionWizardContext> = new AzureWizard(wizardContext, {
         promptSteps: [await FunctionListStep.createFunctionListStep(wizardContext, { templateId, caseSensitiveFunctionSettings, isProjectWizard: false })]
     });
     await wizard.prompt(actionContext);
     await wizard.execute(actionContext);
 }
 
-async function getWorkspacePath(actionContext: IActionContext): Promise<string> {
+async function getWorkspacePath(actionContext: IActionContext): Promise<WorkspaceFolder> {
```
nturinski: nit: Function name is misleading now as it returns the whole workspace.

src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
```diff
@@ -68,36 +73,96 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
...
+        existingTasks = existingTasks || [];
+        // Remove tasks that match the ones we're about to add
+        existingTasks = existingTasks.filter(t1 => newTasks.find(t2 => {
+            return t1.type !== t2.type || t1.label !== t2.label || t1.command !== t2.command;
+        }));
+        existingTasks.push(...newTasks);
+        return existingTasks;
+    }
+
+    private async writeLaunchJson(wizardContext: IFunctionWizardContext, vscodePath: string, runtime: ProjectRuntime): Promise<void> {
```
nturinski: Why not just pass the `workspace` instead of the whole `wizardContext`?

src/vsCodeConfig/tasks.ts
```diff
@@ -0,0 +1,49 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { TaskDefinition, workspace, WorkspaceConfiguration, WorkspaceFolder } from "vscode";
```
nturinski: Why not just the vscode API directly?  Seems like you're only using the commands anyway.

EricJizbaMSFT: I mean it just simplifies my life a little bit. I get the tasks in multiple places and I had to do this each time:
```typescript
const config: WorkspaceConfiguration = workspace.getConfiguration('tasks', folder.uri);
// tslint:disable-next-line: strict-boolean-expressions
const tasks: IFuncTaskDefinition[] = config.get<IFuncTaskDefinition[]>('tasks') || [];
```

Now I can just do this:
```typescript
const tasks: ITask[] = getTasks(folder);
```

PR #1201, opened 2019-04-19
No longer recommend preDeployTask

EricJizbaMSFT:
We originally added this logic specifically for `func extensions install` because it was not an obvious task and because it was added after many users had already created projects without it. Now - that task isn't required. I don't think it's worth keeping the logic for the other tasks, since those are language-specific, pretty obvious, and the vast majority of projects have them by default anyways.

Screen Shot 2019-04-19 at 10 11 28 AM

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -80,7 +80,7 @@ export async function deploy(this: IActionContext, target?: vscode.Uri | string 
         this.properties.cancelStep = '';
     }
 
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
We already set up the preDeployTask when they create a new project or initialize it for use with VS Code. The only way the user will see this warning is if they manually removed that task from their project

nturinski:
Oh right, I forgot that's done on creation.  Okay, I take it back then!

Issue #205, opened 2018-02-05
Incorrect HTTPTrigger URLs in the console when custom route is defined 


simonaco:
Steps:
Define custom route in function.json
Publish function app to Azure using extension 
Notice URLs listed in the console are incorrect( https://sicotin-servelessapi2.azurewebsites.net/api/CreateHero instead of 
https://sicotin-servelessapi2.azurewebsites.net/api/hero )

<img width="1156" alt="screen shot 2018-02-05 at 07 26 07" src="https://user-images.githubusercontent.com/1573904/35804471-fde5cc3e-0a45-11e8-9f01-5bab6abb8f1f.png">

BadgerCode:
The same happens if you have a custom route prefix in the host.json file.

**host.json**
```json
{
    "http": {
        "routePrefix": ""
    }
}
```

**function.json**
```json
{
  "bindings": [
    {
      "authLevel": "anonymous",
      "type": "httpTrigger",
      "direction": "in",
      "route": "v1/lenny",
      "methods": [ "get" ],
      "name": "req"
    },
    {
      "type": "http",
      "direction": "out",
      "name": "res"
    }
  ],
  "disabled": false
}
```

Expected url: ```https://live-uksouth-lenny-today.azurewebsites.net/v1/lenny```
Actual url: ```https://live-uksouth-lenny-today.azurewebsites.net/api/v1-lenny```

EricJizbaMSFT:
Removing from future milestone. As a part of https://github.com/Azure/azure-functions-host/issues/3994, there will be a new ARM api that creates this url for us (instead of us having to manually parse out the route, etc.)

PR #1213, opened 2019-04-23
Fix http trigger url when custom route is defined

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/205

I was originally going to wait for the new ARM api which would handle this for us, but I'm not sure if that's released in all regions and the node sdk is being slow to update. Plus I'll probably want this logic for the "Local Project" tree view eventually anyways

### DIFF
--- a/src/tree/ProductionSlotTreeItem.ts
+++ b/src/tree/ProductionSlotTreeItem.ts
@@ -4,8 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { ISiteTreeRoot, SiteClient } from 'vscode-azureappservice';
-import { AzureParentTreeItem, AzureTreeItem } from 'vscode-azureextensionui';
+import { AzureTreeItem } from 'vscode-azureextensionui';
 import { getWorkspaceSetting } from '../vsCodeConfig/settings';
+import { FunctionAppProvider } from './FunctionAppProvider';
 import { SlotsTreeItem } from './SlotsTreeItem';
 import { SlotTreeItem } from './SlotTreeItem';
 import { SlotTreeItemBase } from './SlotTreeItemBase';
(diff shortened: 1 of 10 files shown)


src/tree/ProductionSlotTreeItem.ts
```diff
@@ -16,11 +17,17 @@ export class ProductionSlotTreeItem extends SlotTreeItemBase {
 
     private readonly _slotsTreeItem: SlotsTreeItem;
 
-    public constructor(parent: AzureParentTreeItem, client: SiteClient) {
+    private constructor(parent: FunctionAppProvider, client: SiteClient) {
         super(parent, client);
         this._slotsTreeItem = new SlotsTreeItem(this);
     }
 
+    public static async create(parent: FunctionAppProvider, client: SiteClient): Promise<ProductionSlotTreeItem> {
```
nturinski: I like this design pattern.  But is there a particular reason you're implementing this in this PR?

EricJizbaMSFT: Because of this change: https://github.com/Microsoft/vscode-azurefunctions/pull/1213/files#diff-c071ebec37986c057b8e2e46077ecbd7R72

I need to make sure refreshImpl happens when the tree item is first created

src/tree/SlotTreeItemBase.ts
```diff
@@ -65,6 +70,25 @@ export abstract class SlotTreeItemBase extends AzureParentTreeItem<ISiteTreeRoot
...
+        } catch {
+            // ignore and use default
+        }
+        // tslint:disable-next-line: strict-boolean-expressions
+        this.runtime = runtime || ProjectRuntime.v2;
+
+        // tslint:disable-next-line: no-any
+        let data: any;
+        try {
+            data = await this.root.client.kudu.functionModel.getHostSettings();
```
nturinski: Why are we special casing/expecting this to fail?

EricJizbaMSFT: It's very similar to the app service plan case you just fixed (where the user didn't have permissions). I don't want a failure for this to block deploy

PR #1214, opened 2019-04-23
Add option to 'Execute Function Now'

### LINKED ISSUE #402: Add option to 'Execute function now'
Per feedback collected by @fiveisprime - users would like a way to 'Execute function now'. The portal already has a 'run' button, so we could do something similar
screen shot 2018-05-31 at 9 28 43 am

This should be easy for things like timer triggers where there's no input needed - it might be a little harder for other types of triggers. The portal lets you add request body info, etc.:
screen shot 2018-05-31 at 9 31 25 am

EricJizbaMSFT:
This only applies to timer triggers since they don't require any input.

Also added "HTTP" or "Timer" to function node description so that I could hide context menu actions if they don't apply to those functions.
Screen Shot 2019-04-23 at 8 47 50 AM

Fixes #402


src/commands/copyFunctionUrl.ts
```diff
@@ -10,7 +10,7 @@ import { FunctionTreeItem } from '../tree/FunctionTreeItem';
 
 export async function copyFunctionUrl(node?: FunctionTreeItem): Promise<void> {
     if (!node) {
-        node = <FunctionTreeItem>await ext.tree.showTreeItemPicker([FunctionTreeItem.contextValue, FunctionTreeItem.readOnlyContextValue]);
+        node = <FunctionTreeItem>await ext.tree.showTreeItemPicker(/^azFuncFunctionHttp(ReadOnly|)$/i);
```
nturinski: Does copyFunctionUrl only work for ReadOnly functions?

EricJizbaMSFT: It works for non-readonly as well. `(ReadOnly|)` should handle that since there's an empty string after `|`

src/tree/SlotTreeItemBase.ts
```diff
@@ -106,8 +104,14 @@ export abstract class SlotTreeItemBase extends AzureParentTreeItem<ISiteTreeRoot
...
                 return this.deploymentsNode;
             default:
-                // DeploymentTreeItem.contextValue is a RegExp, but the passed in contextValue can be a string so check for a match
-                if (typeof expectedContextValue === 'string' && DeploymentTreeItem.contextValue.test(expectedContextValue)) { return this.deploymentsNode; }
+                if (typeof expectedContextValue === 'string') {
+                    // DeploymentTreeItem.contextValue is a RegExp, but the passed in contextValue can be a string so check for a match
+                    if (DeploymentTreeItem.contextValue.test(expectedContextValue)) {
+                        return this.deploymentsNode;
+                    }
+                } else if (expectedContextValue.source.includes(FunctionTreeItem.contextValueBase)) {
```
nturinski: Is this returning the parent? I'm not sure what the source is here.

src/commands/copyFunctionUrl.ts
nturinski: Is this for the Disabled case?

src/tree/SlotTreeItemBase.ts
EricJizbaMSFT: source is the RegExp as a string. Aka it's converting the regexp to a string and seeing if it has `azFuncFunction`

src/commands/copyFunctionUrl.ts
EricJizbaMSFT: Is what for the disabled case?

nturinski: Why have the RegExp to indicate that it is Read-Only or not?  Is it because you don't want copyFunctionUrl to appear in the context menu?  The only other label I can think of is (Disabled)

EricJizbaMSFT: Are you asking why I don't do `/^azFuncFunctionHttp/i`? No strong reason, but I prefer to be explicit. I ran into problems a few times when I didn't extend the RegExp all the way to the end of the string (aka when I didn't have `$` at the end) so this felt safer. Specifically the generic [functions node](https://github.com/Microsoft/vscode-azurefunctions/blob/master/src/tree/FunctionsTreeItem.ts#L16) has a context value of `azFuncFunctions` and I pretty much never meant to match that

Issue #854, opened 2018-12-08
Enable publishing to 'Premium (preview)' plan


asavaritayal:
The Azure Functions premium plan is a new hosting option for function apps that provides premium features like VNet connectivity, no cold start, and premium hardware, without having to compromise on things like latency or scale. More info here - https://github.com/Azure/Azure-Functions/blob/master/functions-premium-plan/overview.md

Using the Az Func Code extension, a user should be able to pick the premium hosting plan for their Function App.

#### Proposed workflow

If a user picks `+ Create a new Function App` during publishing, ask for a unique name and present with the following options for hosting:

```
Select a hosting plan
     - Consumption Plan
     - Premium Plan (preview)
```

The selection should always default to Consumption.

Corresponding CLI workflow is documented here - https://github.com/Azure/Azure-Functions/blob/master/functions-premium-plan/getting-started.md

More details **TBD**

/cc @jeffhollan

EricJizbaMSFT:
Should probably be done with https://github.com/Microsoft/vscode-azurefunctions/issues/811

EricJizbaMSFT:
https://azure.microsoft.com/en-us/blog/announcing-the-azure-functions-premium-plan-for-enterprise-serverless-workloads/

PR #1221, opened 2019-04-24
Allow users to select dedicated plan in advanced creation

EricJizbaMSFT:
...instead of always defaulting to consumption. Modeled on the same experience as the portal.

Relies on https://github.com/Microsoft/vscode-azuretools/pull/487

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/854

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9526,9 +9526,9 @@
             }
         },
         "vscode-azureappservice": {
-            "version": "0.37.1",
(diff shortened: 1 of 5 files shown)


nturinski:
My current change in `appservice` breaks the web app create scenario so you'll probably have to bump it to v0.38 or we can do two separate releases.

EricJizbaMSFT:
Prefer separate releases. We need to get the sorting stuff into app service repo without your refactoring anyways

Issue #1200, opened 2019-04-19
Additional runtime selection window popup when deploy a function app to Azure


lirenhe:
1. Create a workspace that contains 'Function' folder.
2.  Create a C# iothub triggered function inside this folder. 
3. Using the following API _vscode.commands.executeCommand('azureFunctions.createFunctionApp', subscriptionId, resourceGroupName) to deploy to Function App.  

Expect: During deployment, no need to select again for the runtime of the application
Actual: Need to select runtime again.

![new](https://user-images.githubusercontent.com/9100546/56416915-e82ebd00-62c4-11e9-8cfb-d8ac10baa762.PNG)

PR #1228, opened 2019-04-26
Be smarter about getting defaults for create new function app

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1200

### DIFF
--- a/src/vsCodeConfig/settings.ts
+++ b/src/vsCodeConfig/settings.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ConfigurationTarget, Uri, workspace, WorkspaceConfiguration } from "vscode";
+import { ConfigurationTarget, Uri, workspace, WorkspaceConfiguration, WorkspaceFolder } from "vscode";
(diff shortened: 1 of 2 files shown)


src/vsCodeConfig/settings.ts
```diff
@@ -43,6 +43,25 @@ export function getWorkspaceSetting<T>(key: string, fsPath?: string, prefix: str
...
+export function getWorkspaceSettingFromAnyFolder(key: string, prefix: string = extensionPrefix): string | undefined {
+    // tslint:disable-next-line: strict-boolean-expressions
+    const folders: WorkspaceFolder[] = workspace.workspaceFolders || [];
+    let result: string | undefined;
+    for (const folder of folders) {
+        const projectConfiguration: WorkspaceConfiguration = workspace.getConfiguration(prefix, folder.uri);
+        const folderResult: string | undefined = projectConfiguration.get<string>(key);
+        if (!result) {
+            result = folderResult;
+        } else if (folderResult && result !== folderResult) {
```
nturinski: So if you have two different results, you ignore both of them?

EricJizbaMSFT: There is no root - they're all at the same level. In most cases returning undefined means the user will be prompted for the value, which I think makes complete sense

Issue #1155, opened 2019-04-03
Include simple package.json file by default for plain JS


ColbyTresness:
We include a package.json file for TS, and we should include a very simple package.json file by default on function app creation. This is something we're planning on adding to the CLI experience as well.

fiveisprime:
Is this to increase visibility of where to run npm install or will this also include all the scripts like we do for TS?

ColbyTresness:
@mhoeger is planning on including some simple scripts, but not all those in the TS package.json. Primary purpose is showing people how to do the right things.

mhoeger:
https://github.com/Azure/azure-functions-core-tools/pull/1206

PR #1231, opened 2019-04-26
Add package.json to JavaScript projects

EricJizbaMSFT:
Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/1155

As part of this change, I also stopped writing down package-lock.json in the typescript scenario. I just don't think it's worth it. There's a lot of different versions of npm so it would change immediately for a lot of users anyways. The only previous concern was that people might get stuck in a bad situation if the latest version of `typescript` or `@azure/functions` has problems, but the user could easily change the version in their package.json - plus Microsoft owns both of those packages so I guess I'm fine taking that minimal risk.


EricJizbaMSFT:
They're trivial files and we don't want to take a dependency on the func cli when creating a project

Issue #976, opened 2019-02-08
Improve AzureWebJobsStorage experience for local debugging


chrisdias:
When I create a Cosmos trigger I'm asked a bunch of quick pick questions about the Cosmos database, collection, lease, etc.  

Then, I'm prompted in the lower right to select a Storage account to be set in `local.settings.json` for local debugging. 

![image](https://user-images.githubusercontent.com/1487073/52460609-04251e00-2b20-11e9-9f03-8b2348007169.png)

And when i choose "Set Storage Account" I'm sent back up to the quick pick lists.

Why does this prompt show up in the lower right instead of in the quick pick flow at the top/center?

EricJizbaMSFT:
Well...when this prompt was created the prompts _were_ at the top instead of as notifications in the bottom right 😂

Honestly I think we should have a larger discussion around this message. It's been there for a long time, but it's always been a little confusing. I don't think moving to a quick pick will solve that entirely. For example, here are a few things I think we could do:
1. If on Windows and storage emulator is installed, use that and don't prompt
    1. Question: Should we prompt users to install the emulator if it's not installed or keep using storage accounts in Azure as the default?
    1. Note to self: Need to check if the error makes sense if the emulator is installed, but not running
1. If `AzureWebJobsStorage` exists as an environment variable on the system, don't prompt. (Honestly this is what I usually do. It's so much easier to set this once per system instead of once per project)
    1. Note to self: Need to validate the connection string? Not sure if people will set and forget an environment variable and then be confused if they delete that account
1. Prompt right before debugging instead of after function creation. I was hesitant to make this a "required" dialog in the past (hence the "Skip for now" button)  because (a) you have other options as listed above and (b) you don't need it to deploy. But it seems much more reasonable to be "required" right before you debug

chrisdias:
I honestly have no idea why I'm being prompted for a storage account just for debugging :). Is there some way to make it all go away?

EricJizbaMSFT:
At the end of the day that's a question for the platform team. The current solution to "make it all go away" is to use the storage emulator, but that's only available on Windows.

Running the functions runtime locally is the same as running it in the cloud - which I always thought was one of its selling points. So if the platform needs Azure Storage when running in the cloud, it makes sense to me if it needs it locally. But again this is just my understanding and it'd be better to ask someone on the platform team

EricJizbaMSFT:
[...]

PR #1232, opened 2019-04-26
Improve AzureWebJobsStorage experience

EricJizbaMSFT:
Old dialog:
image

Here's what I changed to the dialog:
1. It's now modal
1. It's actually a part of the wizard
1. It allows for local emulator (and removed "Skip for Now" due to this change)

Screen Shot 2019-04-26 at 4 02 04 PM

I'm worried people will click "Use local emulator" and not actually install or run the emulator, so I also added logic to verify that before debugging:
Screen Shot 2019-04-26 at 4 02 14 PM

Fixes https://github.com/Microsoft/vscode-azurefunctions/issues/976

### DIFF
--- a/src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
+++ b/src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
@@ -14,7 +14,8 @@ export class AzureWebJobsStorageExecuteStep<T extends IAzureWebJobsStorageWizard
 
     public async execute(wizardContext: IAzureWebJobsStorageWizardContext): Promise<void> {
         let value: string;
-        if (wizardContext.azureWebJobsStorage === 'emulator') {
+        wizardContext.actionContext.properties.azureWebJobsStorageType = wizardContext.azureWebJobsStorageType;
+        if (wizardContext.azureWebJobsStorageType === 'emulator') {
             value = localEmulatorConnectionString;
         } else {
             value = (await getStorageConnectionString(<IStorageAccountWizardContext>wizardContext)).connectionString;
@@ -24,6 +25,6 @@ export class AzureWebJobsStorageExecuteStep<T extends IAzureWebJobsStorageWizard
     }
 
     public shouldExecute(wizardContext: IAzureWebJobsStorageWizardContext): boolean {
-        return !!wizardContext.azureWebJobsStorage;
+        return !!wizardContext.azureWebJobsStorageType;
     }
 }
--- a/src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
+++ b/src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
@@ -1,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep, IStorageAccountWizardContext } from 'vscode-azureextensionui';
+import { localEmulatorConnectionString } from '../../constants';
+import { azureWebJobsStorageKey, setLocalAppSetting } from '../../funcConfig/local.settings';
+import { getStorageConnectionString } from '../../utils/azure';
+import { IAzureWebJobsStorageWizardContext } from './IAzureWebJobsStorageWizardContext';
+
+export class AzureWebJobsStorageExecuteStep<T extends IAzureWebJobsStorageWizardContext> extends AzureWizardExecuteStep<T> {
+    public priority: number = 230;
+
+    public async execute(wizardContext: IAzureWebJobsStorageWizardContext): Promise<void> {
+        let value: string;
+        if (wizardContext.azureWebJobsStorage === 'emulator') {
+            value = localEmulatorConnectionString;
+        } else {
+            value = (await getStorageConnectionString(<IStorageAccountWizardContext>wizardContext)).connectionString;
+        }
+
+        await setLocalAppSetting(wizardContext.projectPath, azureWebJobsStorageKey, value, true /* suppressPrompt */);
+    }
+
+    public shouldExecute(wizardContext: IAzureWebJobsStorageWizardContext): boolean {
+        return !!wizardContext.azureWebJobsStorage;
+    }
+}
--- a/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
+++ b/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
@@ -18,18 +18,18 @@ export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardC
         const message: string = localize('selectAzureWebJobsStorage', 'AzureWebJobsStorage must be set in "{0}" to debug non-HTTP triggers locally.', localSettingsFileName);
         const result: MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, selectAccount, useEmulator);
         if (result === selectAccount) {
-            wizardContext.azureWebJobsStorage = 'azure';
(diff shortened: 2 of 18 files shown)


src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
```diff
@@ -0,0 +1,30 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep, IStorageAccountWizardContext } from 'vscode-azureextensionui';
+import { localEmulatorConnectionString } from '../../constants';
+import { azureWebJobsStorageKey, setLocalAppSetting } from '../../funcConfig/local.settings';
+import { getStorageConnectionString } from '../../utils/azure';
+import { IAzureWebJobsStorageWizardContext } from './IAzureWebJobsStorageWizardContext';
+
+export class AzureWebJobsStorageExecuteStep<T extends IAzureWebJobsStorageWizardContext> extends AzureWizardExecuteStep<T> {
+    public priority: number = 230;
```
nturinski: Where is this number coming from?  Is 230 arbitrary?

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -0,0 +1,61 @@
...
+    public async getSubWizard(wizardContext: T): Promise<IWizardOptions<T & ISubscriptionWizardContext> | undefined> {
+        if (wizardContext.azureWebJobsStorageType === 'azure') {
+            const promptSteps: AzureWizardPromptStep<T & ISubscriptionWizardContext>[] = [];
+
+            const subscriptionPromptStep: AzureWizardPromptStep<ISubscriptionWizardContext> | undefined = await ext.tree.getSubscriptionPromptStep(wizardContext);
+            if (subscriptionPromptStep) {
+                promptSteps.push(subscriptionPromptStep);
+            }
+
+            promptSteps.push(new StorageAccountListStep(
```
nturinski: This formatting is super wonky looking.

src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
EricJizbaMSFT: All execute steps need a priority to say when they execute
https://github.com/Microsoft/vscode-azuretools/pull/463
https://github.com/Microsoft/vscode-azuretools/pull/464
https://github.com/Microsoft/vscode-azurefunctions/pull/1143

src/commands/pickFuncProcess.ts
```diff
@@ -7,26 +7,25 @@ import * as unixPsTree from 'ps-tree';
...
 import { localize } from '../localize';
 import { getWindowsProcessTree, IProcessTreeNode, IWindowsProcessTree } from '../utils/windowsProcessTree';
-import { getDebugConfigs, isDebugConfigEqual } from '../vsCodeConfig/launch';
 import { getWorkspaceSetting } from '../vsCodeConfig/settings';
 
 export async function pickFuncProcess(this: IActionContext, debugConfig: vscode.DebugConfiguration): Promise<string | undefined> {
-    if (!await validateFuncCoreToolsInstalled()) {
+    const result: IPreDebugValidateResult = await preDebugValidate(debugConfig);
+    if (!result.shouldContinue) {
         throw new UserCancelledError();
```
nturinski: Since result.shouldContinue can be false without user input, should we throw an error that the user can action upon? (such as opening a workspace)

EricJizbaMSFT: How can it be false without user input?

nturinski: https://github.com/Microsoft/vscode-azurefunctions/blob/802c350b194543dfd35e3c02c0087e5bd06a224c/src/commands/createNewProject/verifyIsProject.ts#L28-L53

Doesn't this return `undefined` if there isn't a workspace opened?

src/debug/validatePreDebug.ts
```diff
@@ -0,0 +1,88 @@
...
+                    err ? reject(err) : resolve();
+                });
+            });
+        } catch (error) {
+            const message: string = localize('failedToConnectEmulator', 'Failed to verify "{0}" connection specified in "{1}". Is the local emulator installed and running?', azureWebJobsStorageKey, localSettingsFileName);
+            const learnMoreLink: string = isWindows ? 'https://aka.ms/AA4ym56' : 'https://aka.ms/AA4yef8';
+
+            const debugAnyway: vscode.MessageItem = { title: localize('debugAnyway', 'Debug anyway') };
+            try {
+                const result: vscode.MessageItem = await ext.ui.showWarningMessage(message, { learnMoreLink, modal: true }, debugAnyway);
```
nturinski: What other error are you expecting here?  It seems like it's just a dialog prompt, so I'm not sure why you have try/catch here.

src/funcConfig/local.settings.ts
```diff
@@ -21,53 +20,25 @@ export interface ILocalSettingsJson {
...
-            }
-        );
-
-        // tslint:disable-next-line:strict-boolean-expressions
-        settings.Values = settings.Values || {};
-        settings.Values[azureWebJobsStorageKey] = resourceResult.connectionString;
-        await fsUtil.writeFormattedJson(localSettingsPath, settings);
-    }
+    const settings: ILocalSettingsJson = await getLocalSettingsJson(path.join(projectPath, localSettingsFileName));
+    return settings.Values && settings.Values[azureWebJobsStorageKey];
```
nturinski: Is this returning a string or a boolean?  I thought `return settings.Values && settings.Values[azureWebJobsStorageKey];` would return `true` or `false.`

EricJizbaMSFT: Nah that returns a string. The build passed - do you not believe typescript? haha

src/commands/pickFuncProcess.ts
EricJizbaMSFT: You can't get to pickProcess if you don't have a workspace open since it only happens when you debug

src/funcConfig/local.settings.ts
nturinski: Haha, fair enough.  I tried it out to validate too-- that's super weird though.

src/debug/validatePreDebug.ts
EricJizbaMSFT: The catch block only swallows UserCancelledErrors - aka I want to return false instead of throwing a UserCancelledError. I can add a comment if that's confusing

nturinski: Eh, I kind of suspected as much, but I wanted to verify since it seemed like a weird pattern to me.  I'm a dummy who is easily confused, but I think a comment would be good.

src/utils/azure.ts
```diff
@@ -59,34 +59,23 @@ export interface IResourceResult {
...
-    const client: StorageManagementClient = createAzureClient(node.root, StorageManagementClient);
+export async function getStorageConnectionString(wizardContext: IStorageAccountWizardContext): Promise<IResourceResult> {
+    const client: StorageManagementClient = createAzureClient(wizardContext, StorageManagementClient);
     const storageAccount: StorageManagementModels.StorageAccount = <StorageManagementModels.StorageAccount>nonNullProp(wizardContext, 'storageAccount');
     const name: string = nonNullProp(storageAccount, 'name');
 
     const resourceGroup: string = getResourceGroupFromId(nonNullProp(storageAccount, 'id'));
     const result: StorageManagementModels.StorageAccountListKeysResult = await client.storageAccounts.listKeys(resourceGroup, name);
-    const key: string = nonNullProp(nonNullValue(nonNullProp(result, 'keys')[0], 'key[0]'), 'value');
+    const key: string = nonNullProp(nonNullValue(nonNullProp(result, 'keys')[0], 'keys[0]'), 'value');
```
nturinski: LOL yikes, this is a mess. No other comment.

PR #1237, opened 2019-04-29
Additional validation in verifyInitForVSCode

EricJizbaMSFT:
Before this PR: https://github.com/Microsoft/vscode-azurefunctions/pull/1183, there was a bug where the newly added settings were not returned properly in `verifyInitForVSCode`. However, that bug was hard to diagnose/solve because `nonNullValue` was not throwing an error for an empty string. If for some reason my above PR didn't fix all cases, at least now we'll get more accurate error message/call stack.


EricJizbaMSFT:
Are you talking about nonNull stuff? I say no - I've always wanted to avoid those packages becoming a dumping ground for random utils and this feels exactly like "dumpage" code.

Issue #1245, opened 2019-04-30
Failed to debug second time if using Azurite


EricJizbaMSFT:
Repro steps:
1. Create a javascript project
1. Add a timer trigger
1. Set `AzureWebJobsStorage` to  `UseDevelopmentStorage=true`
1. Make sure azurite is running (https://github.com/Azure/Azurite/blob/master/README.md#installation-and-usage)
1. Debug
1. Detach
1. Debug again

Expected:
It attaches

Actual:
It never attaches because the following is never printed: 
> Host lock lease acquired by instance ID '000000000000000000000000D21F4D53'.

EricJizbaMSFT:
https://github.com/Azure/Azurite/issues/37

JustinGrote:
This happens if you use any remote storage in my observation.

To fix, the problemMatcher for the task should match on the Job host starting, not on the lease request.

```
[5/3/2019 1:02:23 AM] Host initialized (632ms)
[5/3/2019 1:02:23 AM] Host started (644ms)
[5/3/2019 1:02:23 AM] Job host started
```

EricJizbaMSFT:
@JustinGrote we used to do that, but it doesn't work in all cases. See https://github.com/Microsoft/vscode-azurefunctions/issues/660 

Azurite specifically does not work as described in the above issue. If you're not seeing `Host lock lease acquired by instance ID` in a remote case, please file a separate issue with repro steps, project language, etc. and we can look into it.

EricJizbaMSFT:
Seems like both messages (`Job host start` and `Host lock lease acquired`) are not ideal, so I filed an issue on the func cli: https://github.com/Azure/azure-functions-host/issues/4384

JustinGrote:
I ended up just going back to Azure Storage Explorer due to a lot of issues
beyond that. Hopefully it gets working but for cross-plat development I
just point it to a "real" queue in my dev subscription for now as a
workaround.

EricJizbaMSFT:
I tried the latest version of Azurite and this seems to be working now

PR #1246, opened 2019-04-30
Remove emulator option for non-windows

EricJizbaMSFT:
The emulator story on non-windows is using Azurite - but that has some limitations (see https://github.com/Microsoft/vscode-azurefunctions/issues/1245) and probably isn't up-to-par quite yet

Since I'm not showing the emulator, I feel like I need to do a "Skip for now" button
Screen Shot 2019-04-30 at 2 30 16 PM

Screen Shot 2019-04-30 at 2 44 49 PM

### DIFF
--- a/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
+++ b/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
@@ -5,7 +5,7 @@
 
 import { MessageItem } from 'vscode';
 import { AzureWizardPromptStep, ISubscriptionWizardContext, IWizardOptions, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
-import { localSettingsFileName } from '../../constants';
+import { isWindows, localSettingsFileName } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
 import { IAzureWebJobsStorageWizardContext } from './IAzureWebJobsStorageWizardContext';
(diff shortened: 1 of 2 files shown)


src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -14,14 +14,26 @@ export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardC
...
+
+        const result: MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, ...buttons);
         if (result === selectAccount) {
             wizardContext.azureWebJobsStorageType = 'azure';
         } else if (result === useEmulator) {
             wizardContext.azureWebJobsStorageType = 'emulator';
         }
+
+        // tslint:disable-next-line: strict-boolean-expressions
+        wizardContext.actionContext.properties.azureWebJobsStorageType = wizardContext.azureWebJobsStorageType || 'skipForNow';
```
nturinski: You already have telemetry on the os right?

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -14,14 +14,26 @@ export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardC
...
 
         const message: string = localize('selectAzureWebJobsStorage', 'AzureWebJobsStorage must be set in "{0}" to debug non-HTTP triggers locally.', localSettingsFileName);
-        const result: MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, selectAccount, useEmulator);
+
+        const buttons: MessageItem[] = [selectAccount];
+        if (isWindows) {
+            // Only show on Windows until this is fixed: https://github.com/Microsoft/vscode-azurefunctions/issues/1245
+            buttons.push(useEmulator);
+        }
+        buttons.push(skipForNow);
```
nturinski: Should this only have `skipForNow` if it's not Windows?

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
EricJizbaMSFT: Yeah that part of the default stuff collected for all events

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
EricJizbaMSFT: I'm meh on that button, but figured I might as well add it and see what people do in the telemetry. We can remove it if no one picks it

JustinGrote:
FYI I've seen the above message with Azurite if AzureWebJobsDashboard is not specified. If you specify it with a blank "" it doesn't appear.

PR #1247, opened 2019-04-30
Add deploy to created function toast

nturinski:
Relies on this PR https://github.com/Microsoft/vscode-azuretools/pull/495

Doing this to include deploy as an actionable button after the resource is finished creating.

image


nturinski:
Nuts, forgot about deployment slots

package-lock.json
```diff
@@ -1476,7 +1476,7 @@
         },
         "azure-arm-cosmosdb": {
             "version": "1.1.2",
-            "resolved": "https://registry.npmjs.org/azure-arm-cosmosdb/-/azure-arm-cosmosdb-1.1.2.tgz",
+            "resolved": "http://registry.npmjs.org/azure-arm-cosmosdb/-/azure-arm-cosmosdb-1.1.2.tgz",
```
EricJizbaMSFT: nit: revert these changes and clean out your npm cache

src/commands/createSlot.ts
```diff
@@ -0,0 +1,20 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from 'vscode-azureextensionui';
+import { ext } from '../extensionVariables';
+import { SlotsTreeItem } from '../tree/SlotsTreeItem';
+import { SlotTreeItem } from '../tree/SlotTreeItem';
+
+export async function createSlot(context: IActionContext, node?: SlotsTreeItem, resourceGroup?: string): Promise<string> {
```
EricJizbaMSFT: Remove `resourceGroup` parameter. We don't support createSlot in [the api](https://github.com/microsoft/vscode-azurefunctions/wiki/Azure-Functions-API-(Preview)) so no need to add code that will never be used.

src/tree/SlotTreeItemBase.ts
```diff
@@ -166,4 +168,23 @@ export abstract class SlotTreeItemBase extends AzureParentTreeItem<ISiteTreeRoot
     public async deleteTreeItemImpl(): Promise<void> {
         await deleteSite(this.root.client);
     }
+
+    public showCreatedOutput(): void {
+        const slot: string = localize('slot', 'slot');
+        const functionApp: string = localize('functionApp', 'function app');
+        const createdNewSlotTree: string = localize('createSlotTree', 'Created new {0} "{1}": {2}', this.root.client.isSlot ? slot : functionApp, this.root.client.fullName, `https://${this.root.client.defaultHostName}`);
```
EricJizbaMSFT: So it's fine to insert things like a website name into a localized string since that won't be translated, but it's generally discouraged to insert things like "slot" which could be translated. For example, some languages might put the adjective after the noun like "Created slot new", which wouldn't work in your current setup. Just duplicate the string

PR #1250, opened 2019-05-01
Don't let one bad folder stop provideTasks

EricJizbaMSFT:
I'm seeing errors for `provideTasks` in telemetry where a folder path doesn't exist. Added a few extra checks when providing tasks since this can block debug/deploy if it fails.

### DIFF
--- a/src/debug/FuncTaskProvider.ts
+++ b/src/debug/FuncTaskProvider.ts
@@ -65,7 +65,8 @@ export class FuncTaskProvider implements TaskProvider {
                 }
 
                 if (!isNullOrUndefined(lastError)) {
-                    // throw the last error just for the sake of telemetry - shouldn't block providing tasks
+                    // throw the last error just for the sake of telemetry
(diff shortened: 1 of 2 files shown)


src/debug/FuncTaskProvider.ts
```diff
@@ -30,37 +31,48 @@ export class FuncTaskProvider implements TaskProvider {
...
-        const tasks: Task[] | undefined = await callWithTelemetryAndErrorHandling('provideTasks', async function (this: IActionContext): Promise<Task[]> {
+        await callWithTelemetryAndErrorHandling('provideTasks', async function (this: IActionContext): Promise<void> {
             this.properties.isActivationEvent = 'true';
             this.suppressErrorDisplay = true;
             this.suppressTelemetry = true;
 
-            const result: Task[] = [];
             if (workspace.workspaceFolders) {
+                // tslint:disable-next-line: no-any
+                let lastError: any;
```
StephenWeatherford: unknown

PR #1273, opened 2019-05-13
Move to latest tree that's friendlier to non-Azure scenarios

### LINKED ISSUE #1164: Project tree view doesn't detect new projects
Currently it only works if you open the extension with an existing project. It's not detecting new projects or changes to workspace folders

### LINKED ISSUE #1170: Project tree doesn't work great with tree item picker
It always shows the subscription and local project items. Ideally it should always use local project. Unfortunately I think I'll have to refactor the tree in the shared package to get this to work

EricJizbaMSFT:
Based on https://github.com/microsoft/vscode-azuretools/pull/499

The new tree is more generic and splits out Azure resources from non-Azure resources. This makes the project tree view in this repo much easier to implement. For example, this PR fixes #1164 to update the tree when the file system changes and fixes #1170 to support the tree item picker for local project commands.


src/commands/appSettings/downloadAppSettings.ts
```diff
@@ -18,7 +18,7 @@ import { getLocalSettingsFile } from "./getLocalSettingsFile";
 
 export async function downloadAppSettings(node?: AppSettingsTreeItem): Promise<void> {
     if (!node) {
-        node = <AppSettingsTreeItem>await ext.tree.showTreeItemPicker(AppSettingsTreeItem.contextValue);
+        node = await ext.tree.showTreeItemPicker<AppSettingsTreeItem>(AppSettingsTreeItem.contextValue);
```
nturinski: Is this just another way to cast?  Why change all the formatting?

EricJizbaMSFT: So that's actually a long story. I changed 
```
public showTreeItemPicker(...): Promise<AzExtTreeItem>;
```

to:
```
public showTreeItemPicker<T extends AzExtTreeItem>(...): Promise<T>;
```

because in Cosmos (the extension I initially used as a trial) typescript was able to infer the return type, which meant I didn't have to cast at all. Basically if `node` is type `AppSettingsTreeItem | undefined` it can infer that `T` should be `AppSettingsTreeItem` _or so I thought_. It turns out that only works if you have `strictNullChecks` off (which I guess Cosmos DB does). Aka it can only infer `T` if the type is exactly `AppSettingsTreeItem` and not `AppSettingsTreeItem | undefined`. So in functions, the inferal doesn't work, and now I have to specify a type for `T` which is why these changed.

src/tree/AzureAccountTreeItemWithProjects.ts
```diff
@@ -0,0 +1,83 @@
...
+import { LocalProjectTreeItem } from './localProject/LocalProjectTreeItem';
+import { isLocalTreeItem } from './localProject/LocalTreeItem';
+import { SubscriptionTreeItem } from './SubscriptionTreeItem';
+
+export const enableProjectTreeSetting: string = 'enableProjectTree';
+
+export class AzureAccountTreeItemWithProjects extends AzureAccountTreeItemBase {
+    private _projectDisposables: Disposable[] = [];
+
+    public constructor(testAccount?: TestAzureAccount) {
```
nturinski: Why are you passing a `TestAzureAccount` here?  Is this just so that you don't have to be logged in to use this?  I'm assuming that you're not actually expecting credentials to be in the process.env.

src/tree/AzureAccountTreeItemWithProjects.ts
```diff
@@ -0,0 +1,83 @@
...
+import { tryGetFunctionProjectRoot } from '../commands/createNewProject/verifyIsProject';
+import { hostFileName } from '../constants';
+import { localize } from '../localize';
+import { getWorkspaceSetting } from '../vsCodeConfig/settings';
+import { createRefreshFileWatcher } from './localProject/createRefreshFileWatcher';
+import { LocalProjectTreeItem } from './localProject/LocalProjectTreeItem';
+import { isLocalTreeItem } from './localProject/LocalTreeItem';
+import { SubscriptionTreeItem } from './SubscriptionTreeItem';
+
+export const enableProjectTreeSetting: string = 'enableProjectTree';
```
nturinski: It doesn't seem like any other file is using this so not sure why it's being exported.  Also if we did need to export it, aren't all the workspace settings in `constants.ts`?

src/tree/AzureAccountTreeItemWithProjects.ts
EricJizbaMSFT: This is the same `TestAzureAccount` that you implemented, just moved to a different place. It's used for the exact same reason (aka those long running tests). See [this PR](https://github.com/microsoft/vscode-azuretools/pull/499) for more info on the refactoring.

src/tree/AzureAccountTreeItemWithProjects.ts
EricJizbaMSFT: Workspace settings are only supposed to be in `constants.ts` if they're used in multiple files. I don't put _all_ settings in there because I don't want it to get cluttered.

I'll remove the `export`

EricJizbaMSFT:
Haha I'm gonna take care of updating the other repos (also app service should be simpler than functions)

PR #1288, opened 2019-05-20
Clean up tree naming

EricJizbaMSFT:
- Rename `nodeUtils` to `treeUtils` because "node" could mean either "TreeItem" or "node.js"
- Rename `azureFunctionsExplorer` to `azFuncTree` because it's shorter and more consistent with contextValue names (which all start with `azFunc`)


EricJizbaMSFT:
> Well, treeUtils also sounds like things related to the tree view itself. Is treeItemUtils too verbose?

I think it's totally fine for this file to have utils related to both tree items and the tree view itself

PR #1291, opened 2019-05-21
Updating Az module major version.

Francisco-Gamino:
Updating The Azure (Az) module major version to 2.

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/PowerShellProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/PowerShellProjectCreateStep.ts
@@ -41,7 +41,7 @@ const requirementspsd1: string = `# This file enables modules to be automaticall
 # See https://aka.ms/functionsmanageddependency for additional information.
 #
 @{
-    'Az' = '1.*'
+    'Az' = '2.*'
 }`;
 
 export class PowerShellProjectCreateStep extends ScriptProjectCreateStep {


EricJizbaMSFT:
@Francisco-Gamino can you provide a little more background for this change? Not that familiar with the Az PowerShell module and it's versioning/releases

For example - why is this changing now? Is 2.* something that recently GA'ed? Will 1.* continue to work?

Francisco-Gamino:
Hi @EricJizbaMSFT -- Azure provides a set of cmdlets to manage its services. These are released via the Az modules which is available at https://www.powershellgallery.com/packages/Az. The Azure PowerShell team just did a major release for Az, which is now version 2.1. The Az minor releases take place every couple of weeks and major releases are about every 5 months.

We would like to push out this change as soon as possible given that if customer/user creates a PowerShell function app, there will be installing Az 1.8 (since their requirements.psd1 file will have a content of 'Az' = '1.*', which is the latest version for major version 1). This version is already out of date. We would like function user to be in the latest Az major version 2. I hope this makes sense. Please let me know if you have any other questions. 

As a side note, we have already updated this major version in the Host for functions that get created via the Azure Portal, and the Core Tools for functions created via func.exe.

Francisco-Gamino:
> The PowerShell team plans to make breaking changes to the Az package about every 5 months?? That seems very frequent.
> 
> > As a side note, we have already updated this major version in the Host for functions
> 
> This is what I'm concerned about. Not as a blocker for this PR - but do we need to somehow notify users on version 1.* and update them? Will they be broken or will they continue to work?

Adding @anirudhgarg for FYI.

@EricJizbaMSFT : For customer that are on Az version 1, their function app will continue to work. 
If a customer wishes to move their function app to the next version, they can just update ```requirements.psd1``` with ```2.*```. 
This guidance will be post by our PM @eamonoreilly in Azure Service Updates.

PR #1366, opened 2019-07-02
Add new setting for application

v-wuzhai:
Add new setting for application


test/createAzureResource.test.ts
```diff
@@ -176,6 +187,19 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         const result: string = await getBody(functionUrl, resourceName);
         assert.equal(result, expectResult, `The result should be "${expectResult}" rather than ${result} and the triggerUrl is ${functionUrl}`);
     }
+
+    function applicationsettings(listApplicationSettings: { [propertyName: string]: string } | undefined, applicationSettingsKey: string): string | undefined {
```
nturinski: Are you calling this anywhere?  I'm not sure I understand what this is for.

test/createAzureResource.test.ts
```diff
@@ -120,6 +121,16 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
         assert.equal(createdApp.state, 'Running', `Function App state should be 'Running' rather than ${createdApp.state}.`);
     });
 
+    test('Add new setting', async () => {
+        const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceName1, resourceName1);
+        assert.ok(createdApp, `Function App ${resourceName1} not found`);
+        ext.ui = new TestUserInput([resourceName1, applicationSettings[0], applicationSettings[1]]);
+        await vscode.commands.executeCommand('azureFunctions.appSettings.add');
+        const listapplicationSettings: WebSiteManagementModels.StringDictionary = await webSiteClient.webApps.listApplicationSettings(resourceName1, resourceName1);
+        const applicationSettingsValue: string | undefined = applicationsettings(listapplicationSettings.properties, applicationSettings[0]);
```
nturinski: This is more complicated than what you need to do.  You should be able to call `listApplicationSettings` and then check the `.properties` object on the return value.  `properties` contains an object where the key is the setting name name and the value is the setting value.

Here is the structure of the returned object: 
image

test/createAzureResource.test.ts
```diff
@@ -120,6 +121,16 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        assert.equal(applicationSettingsValue, applicationSettings[1], `The expected setting value should be "${applicationSettings[1]}" rather than "${applicationSettingsValue}"`);
```
nturinski: You'll have to verify that appSettings.properties isn't undefined, but then you can just check `appSettings.properties.applicationSettings[0] === applicationSettings[1]`

test/createAzureResource.test.ts
```diff
@@ -25,6 +25,7 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
     const resourceName1: string = getRandomHexString().toLowerCase();
     // Get the *.code-workspace workspace file path
     const projectPath: string = getTestRootFolder();
+    const applicationSettings: string[] = ['FUNCTION_EXTENSION_URL', "https://github.com/microsoft/vscode-azurefunctions"];
```
nturinski: I don't think you need to define an applicationSettings array here.  I would just define a `const key: string` and a `const value: string`.

This isn't being used by any other tests (yet) so it doesn't need to be up here.  Though I assume that you're going to use these to test edit and delete next though?

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+    test('Add new setting', async () => {
+        let applicationSettingsValue: string | undefined;
+        const applicationSettings: { key: string; value: string } = { key: 'FUNCTION_EXTENSION_URL', value: "https://github.com/microsoft/vscode-azurefunctions" };
+        const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceName1, resourceName1);
+        assert.ok(createdApp, `Function App ${resourceName1} not found`);
+        ext.ui = new TestUserInput([resourceName1, applicationSettings.key, applicationSettings.value]);
+        await vscode.commands.executeCommand('azureFunctions.appSettings.add');
+        const listapplicationSettings: WebSiteManagementModels.StringDictionary = await webSiteClient.webApps.listApplicationSettings(resourceName1, resourceName1);
+        assert.ok(listapplicationSettings.properties, 'Failed to get application settings');
+        if (listapplicationSettings.properties !== undefined) {
```
nturinski: nit: `if (listApplicationSettings.properties)` will be sufficient.

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
     });
 
+    test('Add new setting', async () => {
+        let applicationSettingsValue: string | undefined;
+        const applicationSettings: { key: string; value: string } = { key: 'FUNCTION_EXTENSION_URL', value: "https://github.com/microsoft/vscode-azurefunctions" };
+        const createdApp: WebSiteManagementModels.Site = await webSiteClient.webApps.get(resourceName1, resourceName1);
+        assert.ok(createdApp, `Function App ${resourceName1} not found`);
+        ext.ui = new TestUserInput([resourceName1, applicationSettings.key, applicationSettings.value]);
+        await vscode.commands.executeCommand('azureFunctions.appSettings.add');
+        const listapplicationSettings: WebSiteManagementModels.StringDictionary = await webSiteClient.webApps.listApplicationSettings(resourceName1, resourceName1);
```
nturinski: camelCase the variable listApplicationSettings

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        assert.ok(listapplicationSettings.properties, 'Failed to get application settings');
```
nturinski: This assert isn't necessary since you'll be checking for it below.

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+        if (listapplicationSettings.properties !== undefined) {
+            if (listapplicationSettings.properties.FUNCTION_EXTENSION_URL !== undefined) {
```
nturinski: You don't need to check for this either.  Just assert that listApplicationSettings.properties[application.key] equals application.value

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+                applicationSettingsValue = listapplicationSettings.properties.FUNCTION_EXTENSION_URL;
+            } else {
+                assert.ok(listapplicationSettings.properties.FUNCTION_EXTENSION_URL, `The key "${applicationSettings.key}" by the application setting not found.`);
```
nturinski: Use the application.key that you set in the assert.

test/createAzureResource.test.ts
```diff
@@ -120,6 +120,25 @@ suite('Create Azure Resources', async function (this: ISuiteCallbackContext): Pr
...
+            }
+        }
+        assert.equal(applicationSettingsValue, applicationSettings.value, `The application setting for "${applicationSettings.key}=${applicationSettings.value}" not found`);
```
nturinski: Rather than the logic above, just do

`assert.equal(listApplicationSettings.properties[applicationSettings.key], applicationSettings.value)

v-wuzhai:
Merge to https://github.com/microsoft/vscode-azurefunctions/pull/1421

PR #1394, opened 2019-07-22
Python 3.6 Downloadable Version

askpatrickw:
The instructions say Python 3.6 is supported, but the finding an installable set of binaries for 3.6 is non-obvious.
This doc change makes it obvious.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ See [here](https://github.com/Azure/azure-functions-core-tools/blob/master/READM
 
 * [Python 3.6.x](https://www.python.org/downloads/)
 
+> **NOTE**: [Python 3.6.8](https://www.python.org/downloads/release/python-368/) is the last release of Python 3.6 with installable binaries for Windows.
+
 ### PowerShell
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -136,6 +136,8 @@ See [here](https://github.com/Azure/azure-functions-core-tools/blob/master/READM
 
 * [Python 3.6.x](https://www.python.org/downloads/)
 
+> **NOTE**: [Python 3.6.8](https://www.python.org/downloads/release/python-368/) is the last release of Python 3.6 with installable binaries for Windows.
```
EricJizbaMSFT: @askpatrickw do you have any docs that say this will be the last release of Python 3.6.x? Is it assumed since they moved on to 3.7? Also do we know _why_ 3.6.9 doesn't have installable binaries?

askpatrickw: I don't know _why_ exactly,  but this is what is stated in the release notes at the link I included.
https://www.python.org/downloads/release/python-368/

EricJizbaMSFT: Good enough for me - thanks!

Issue #1259, opened 2019-05-03
Pre debug validation doesn't happen if debug config type is `node2`


EricJizbaMSFT:
I guess sometimes the debug configuration we get inside `resolveDebugConfiguration` has a type of "node2" even when the type in the user's "launch.json" is "node". This results in the error:
```
Failed to find launch config matching name "Attach to Node Functions", request "attach", and type "node2".
```

Fortunately, we're not actually showing this error to the user and it doesn't block debugging. The only downside is we won't perform our pre-debug validation (aka that the func cli is installed and that the local emulator is running).

PR #1408, opened 2019-07-26
Fix pre-debug validation for node2

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1259

### DIFF
--- a/src/vsCodeConfig/launch.ts
+++ b/src/vsCodeConfig/launch.ts
@@ -28,7 +28,19 @@ export function updateLaunchVersion(folder: WorkspaceFolder, version: string): v
 }
 
 export function isDebugConfigEqual(c1: DebugConfiguration, c2: DebugConfiguration): boolean {
-    return c1.name === c2.name && c1.request === c2.request && c1.type === c2.type;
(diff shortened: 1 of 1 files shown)


src/vsCodeConfig/launch.ts
```diff
@@ -28,7 +28,19 @@ export function updateLaunchVersion(folder: WorkspaceFolder, version: string): v
...
+function isTypeEqual(type1: string, type2: string): boolean {
+    return type1 === type2 || (isNodeType(type1) && isNodeType(type2));
+}
+
+/**
+ * Special case node debug type because it can be either "node" or "node2"
+ * https://github.com/microsoft/vscode-azurefunctions/issues/1259
+ */
+function isNodeType(t: string): boolean {
+    return /^node2?$/i.test(t);
```
nturinski: Can't you just search for node instead of exclusively node2?  I'd expect if I had the "node" debug type for this to return true, but I don't think that'd be the case right now

EricJizbaMSFT: `?` is a quantifier meaning `2` is an optional character. This regexp matches both `node2` _and_ `node`

PR #1414, opened 2019-07-31
Clean up nightly tests

EricJizbaMSFT:
Recently fixed a bug when deploying to a linux dedicated plan (https://github.com/microsoft/vscode-azuretools/pull/545) and would like to add a nightly test to ensure it doesn't break again. Before I do that, I wanted to clean up the nightly tests a bit (splitting up into multiple files, better variable names, etc.). The only "real" change was I added a TypeScript version of the "Create Project And Deploy" test since it was so trivial.


test/nightly/global.nightly.test.ts
```diff
@@ -0,0 +1,54 @@
...
+import { TestAzureAccount } from 'vscode-azureextensiondev';
+import { AzExtTreeDataProvider, AzureAccountTreeItemWithProjects, createAzureClient, ext } from '../../extension.bundle';
+import { longRunningTestsEnabled } from '../global.test';
+
+export let testAccount: TestAzureAccount;
+export let testClient: WebSiteManagementClient;
+export const resourceGroupsToDelete: string[] = [];
+
+// Runs before all nightly tests
+suiteSetup(async function (this: IHookCallbackContext): Promise<void> {
```
nturinski: Do test files prefixed with global guarantee that it'll be run first?

EricJizbaMSFT: I don't think so. Honestly I don't entirely remember all the rules on this. Mostly I was copying the comment/pattern from `global.test.ts` [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/test/global.test.ts#L21)

I _believe_ the key part is `suiteSetup` outside of `suite` (which is how this file is set up) will run before any `suiteSetup` inside a `suite` (which is how most other test files are set up).

PR #1420, opened 2019-08-02
Add support for server-side build for Linux Function Apps

EricJizbaMSFT:
Depends on https://github.com/microsoft/vscode-azuretools/pull/549

~Currently setting is off by default. We can discuss and turn it on when we're ready~ Per offline discussion we want this on.

### DIFF
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/PythonInitVSCodeStep.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/PythonInitVSCodeStep.ts
@@ -20,7 +20,6 @@ export class PythonInitVSCodeStep extends ScriptInitVSCodeStep {
     protected async executeCore(context: IProjectWizardContext): Promise<void> {
         await super.executeCore(context);
 
-        const zipPath: string = this.setDeploySubpath(context, '.');
         this.settings.push({ key: 'scmDoBuildDuringDeployment', value: true });
 
         this._venvName = await getExistingVenv(context.projectPath);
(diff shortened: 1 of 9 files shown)


EricJizbaMSFT:
ping @nturinski

src/commands/initProjectForVSCode/InitVSCodeStep/PythonInitVSCodeStep.ts
```diff
@@ -7,29 +7,28 @@ import * as fse from 'fs-extra';
...
 
 export class PythonInitVSCodeStep extends ScriptInitVSCodeStep {
-    protected preDeployTask: string = packTaskName;
     private _venvName: string | undefined;
 
     protected async executeCore(context: IProjectWizardContext): Promise<void> {
         await super.executeCore(context);
 
-        const zipPath: string = this.setDeploySubpath(context, `${path.basename(context.projectPath)}.zip`);
+        this.settings.push({ key: 'scmDoBuildDuringDeployment', value: true });
```
nturinski: Isn't this always going to be true?  I thought while this was the default, we weren't going to force users to use it.

test/validateProject.ts
```diff
@@ -122,13 +122,13 @@ export function getFSharpValidateOptions(projectName: string, targetFramework: s
...
 
-export function getPythonValidateOptions(projectName: string, venvName: string): IValidateProjectOptions {
+export function getPythonValidateOptions(venvName: string): IValidateProjectOptions {
     return {
         expectedSettings: {
             projectLanguage: ProjectLanguage.Python,
             projectRuntime: ProjectRuntime.v2,
-            preDeployTask: 'func: pack',
-            deploySubpath: `${projectName}.zip`,
+            deploySubpath: '.',
```
nturinski: Does this mean we aren't supporting `func: pack`?

src/commands/initProjectForVSCode/InitVSCodeStep/PythonInitVSCodeStep.ts
EricJizbaMSFT: Yes it's always true, but this code only happens when creating a project. After the project is created, they can do whatever they want with their settings.

test/validateProject.ts
EricJizbaMSFT: Again, this is related to creating a project. People can still manually change to use `func: pack` _after_ the project is created if they want to.

PR #1421, opened 2019-08-05
Add the applicationSettings test cases

v-wuzhai:
This test case contains 4 actions like below:
•Add new setting
•Rename setting
•Edit setting
•Delete setting

### DIFF
--- a/test/nightly/applicationSettings.test.ts
+++ b/test/nightly/applicationSettings.test.ts
@@ -1,0 +1,75 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as assert from 'assert';
+import { WebSiteManagementModels as Models } from 'azure-arm-website';
+import { IHookCallbackContext, ISuiteCallbackContext } from 'mocha';
+import * as vscode from 'vscode';
+import { DialogResponses, getRandomHexString } from '../../extension.bundle';
+import { longRunningTestsEnabled, testUserInput } from '../global.test';
+import { resourceGroupsToDelete, testClient } from './global.nightly.test';
+
+suite('Application settings', async function (this: ISuiteCallbackContext): Promise<void> {
+    this.timeout(7 * 60 * 1000);
+
+    let resourceName: string;
+    let rename: string;
+    let edit: string;
+    const appSetting: { key: string; value: string } = {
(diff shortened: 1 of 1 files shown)


test/nightly/applicationSettings.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    test('Add new setting', async () => {
+        await testUserInput.runWithInputs([/create new function app/i, resourceName, 'Windows', '.NET', 'East US', appSetting.key, appSetting.value], async () => {
+            await vscode.commands.executeCommand('azureFunctions.appSettings.add');
+        });
+        await validateApplicationSetting(appSetting.key, appSetting.value);
+    });
+
+    test('Rename setting', async () => {
+        await validateApplicationSetting(appSetting.key, appSetting.value);
```
EricJizbaMSFT: Please keep in mind that each test you add:
1. Adds maintenance cost for the code
1. Increases the chance of false positives (for example the "Restart" test failed the other night [here](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8449&view=ms.vss-test-web.build-test-results-tab) and passed the next night with no changes on our end [here](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=8477))
1. Increases the nightly build time

I think the "Create Project and Deploy" test you added is by far the most valuable nightly test you've done. For the following reasons:
1. The scenario is high priority
1. The code underneath is complex and changes frequently
1. The feature covers multiple languages/operating systems and is difficult to test manually
1. We rely on several outside integrations that could break us (for example we don't own the templates)

By contrast, the app settings feature:
1. Is lower priority
1. Is pretty simple and doesn't change very often
1. Should be the same for all languages/operating systems
1. Doesn't really have any outside integrations (except Azure I guess)

I'm open to debate on this, but I'm leaning towards saying these tests aren't worth it. @nturinski feel free to jump in

nturinski: Eric brings up a lot of good points.  I think that it makes sense that we focus more on higher priorities test cases and all of its variations.

Originally, I thought it would make the most sense to have a greater breadth of test coverage and have automated tests for all the tests that the CTI team runs through, but as Eric pointed out, some of these tests are not very time consuming compared to creating and deploying resources.

I think it would make more sense to refocus our automated testing effort.  We should probably redo the automated test design document to cover more depth-- for instance, creating and deploying to more runtime languages with the function templates.

I'll talk with Eric offline and update the testing design documents so we can focus our efforts on higher priority scenarios.

nturinski:
Closing due to this testing scenario not being high priority.

Issue #1332, opened 2019-06-07
VS Code Azure Functions extension storage warning is hard to understand


gbowerman:
VS Code Azure Functions extension 0.17.1 (Windows)
The standard create Azure Functions flow in VS Code Azure Functions extension pops up a warning stating that AzureWebJobsStorage must be set in "local.settings.json" to debug non-HTTP triggers locally.

This is a problem for 2 reasons:
Why create a warning just for following the standard create flow. Instead make it an option, like "Would you like to set a storage account so you can debug this function locally?"
The warning message text is hard to understand. Why should a user trying to create a function know what AzureWebJobsStorage is, and why do they need to go edit local.settings.json?

![functionswarning](https://user-images.githubusercontent.com/4258913/59138243-ea1a1180-8940-11e9-9f94-95f90863674b.png)

PR #1426, opened 2019-08-06
Improve AzureWebJobsStorage experience

EricJizbaMSFT:
Two main improvements:
1. Delay prompting about AzureWebJobsStorage until user debugs, instead of right after creating a function. If a user never debugs, they will never see this warning. If a user _does_ debug without `AzureWebJobsStorage` configured correctly, they will always see this warning with the ability to fix. Unfortunately, this logic doesn't work for C# and Java
1. Make the warning more human-readable. I tried to avoid "AzureWebJobsStorage" and "local.settings.json" as words since they're very function-specific

Screen Shot 2019-08-06 at 10 50 50 AM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1332

### DIFF
--- a/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
+++ b/src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
@@ -5,25 +5,34 @@
 
 import { MessageItem } from 'vscode';
 import { AzureWizardPromptStep, ISubscriptionWizardContext, IWizardOptions, StorageAccountKind, StorageAccountListStep, StorageAccountPerformance, StorageAccountReplication } from 'vscode-azureextensionui';
-import { isWindows, localSettingsFileName } from '../../constants';
+import { isWindows } from '../../constants';
 import { ext } from '../../extensionVariables';
 import { localize } from '../../localize';
 import { IAzureWebJobsStorageWizardContext } from './IAzureWebJobsStorageWizardContext';
 
 export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardContext> extends AzureWizardPromptStep<T> {
+    private _suppressSkipForNow?: boolean;
+
+    public constructor(suppressSkipForNow?: boolean) {
+        super();
+        this._suppressSkipForNow = suppressSkipForNow;
+    }
+
     public async prompt(context: T): Promise<void> {
         const selectAccount: MessageItem = { title: localize('selectAzureAccount', 'Select storage account') };
         const useEmulator: MessageItem = { title: localize('userEmulator', 'Use local emulator') };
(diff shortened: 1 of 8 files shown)


src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -5,25 +5,34 @@
...
+        this._suppressSkipForNow = suppressSkipForNow;
+    }
+
     public async prompt(context: T): Promise<void> {
         const selectAccount: MessageItem = { title: localize('selectAzureAccount', 'Select storage account') };
         const useEmulator: MessageItem = { title: localize('userEmulator', 'Use local emulator') };
         const skipForNow: MessageItem = { title: localize('skipForNow', 'Skip for now') };
 
-        const message: string = localize('selectAzureWebJobsStorage', 'AzureWebJobsStorage must be set in "{0}" to debug non-HTTP triggers locally.', localSettingsFileName);
+        const message: string = localize('selectAzureWebJobsStorage', 'In order to debug, you must select a storage account for internal use by the Azure Functions runtime.');
```
nturinski: I'm a little concerned that removing the non-HTTP part might confuse users. Considering HTTP is the most popular trigger, I could see a situation where a user has been using HTTP triggers, decides to try a Timer trigger, and think that it's an error that this message has come up all of a sudden when they never needed storage to debug in the past.

src/debug/validatePreDebug.ts
```diff
@@ -4,34 +4,74 @@
...
-        if (projectPath) {
-            shouldContinue = await validateEmulatorIsRunning(projectPath);
+    let shouldContinue: boolean;
+    context.telemetry.properties.debugType = debugConfig.type;
+
+    try {
+        context.telemetry.properties.lastValidateStep = 'funcInstalled';
+        shouldContinue = await validateFuncCoreToolsInstalled();
+
+        if (shouldContinue) {
```
nturinski: I don't really understand why you have to return this value.  Either way, it seems like you exit the command (though for one instance, you return undefined).  But couldn't you just exit out of the function?

EricJizbaMSFT: Discussed offline

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
EricJizbaMSFT: I see where you're coming from, but I feel like there's going to be some confusion no matter what and thus I prioritized making the message as simple/straightforward as possible. For example, even if I say "non-HTTP" I feel like users would wonder "What does a timer trigger have to do with a storage account?". We can change based on feedback

Issue #660, opened 2018-09-25
Error while debugging Azure Functions python project


eduardomourar:
**OS**: MacOS v10.13.6
**Azure Functions Core Tools**: v2.0.3
**Function Runtime Version**: v2.0.12115.0

## Repro Steps:
1. Install a python 3.6 virtual environment using `pipenv`
2. Create Azure Functions python project
3. Create Function (HTTP trigger)
4. Put a breakpoint in a python file from project
5. Try debugging newly created function using VSCode

## Expect:
To start running Functions core tools with PTVSD and attach debugger to it.

## Actual:
Instead this is the error I get: **Failed to attach (connect ECONNREFUSED 127.0.0.1:9091)**.
Even with `--wait` parameter to ptvsd did not work.

## Workaround:
We need to ensure that we attach only when is ready. So by adding the following to project host.json:
```
"logging": {
    "logLevel": {
        "Host.General": "Information"
    }
}
```

And changing the [end pattern for func-watch matcher](https://github.com/Microsoft/vscode-azurefunctions/blob/ce4c9f9b47ea6df30627e6588d6fbd51b23d77b8/package.json#L510) to:
```
"endsPattern": "^.*Host lock lease acquired by instance ID.*$"
```

EricJizbaMSFT:
@asavaritayal @pragnagopa can you confirm - at what point is the language worker ready for us to attach the python debugger?

Is `Host lock lease acquired by instance ID` the signal from the log that its ready? Or is there something else that we should be waiting for?

pragnagopa:
Functions host launches the language worker process with the debug options for example here is the command for spawning java process where debug options are set in the appSetting languageWorkers:java:arguments

```
java  -jar -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 azure-functions-java-worker.jar --host 127.0.0.1 --port 4935 --workerId cbf948ff-015b-4f63-bc89-cc2a399e8098 --requestId dd38c7cc-745c-4427-9162-d074a2ba1f84 --grpcMaxMessageLength 134217728
```

Note:
 Environment variable **languageWorkers:java:arguments** 
 set to **-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005**

For python, it would be similar, languageWorkers:python:arguments needs to be set

pragnagopa:
Assuming you want to attach debugger after python process is launched, yes log statement
```
Host lock lease acquired by instance ID
```
[...]

PR #1428, opened 2019-08-06
Change background endspattern to stop at "Application Started"

JustinGrote:
This change will background the task when azure functions is ready to accept connections. This will save a few seconds between the task starting and the debug being available and speed up the developer feedback loop.

### DIFF
--- a/package.json
+++ b/package.json
@@ -721,7 +721,7 @@
                 "background": {
                     "activeOnStart": true,
                     "beginsPattern": "^.*(Job host stopped|signaling restart).*$",
-                    "endsPattern": "^.*Host lock lease acquired by instance ID.*$"
+                    "endsPattern": "^Application started.+$"
                 },
                 "severity": "error"
             }


PrashanthCorp:
Hey @JustinGrote , thanks for the PR, we appreciate it! Do you have a before-after gif, or something specific we can repro against to track the effect of this change over time?

package.json
```diff
@@ -721,7 +721,7 @@
                 "background": {
                     "activeOnStart": true,
                     "beginsPattern": "^.*(Job host stopped|signaling restart).*$",
-                    "endsPattern": "^.*Host lock lease acquired by instance ID.*$"
+                    "endsPattern": "^Application started.+$"
```
EricJizbaMSFT: @JustinGrote not sure if you remember, but we already discussed this [here](https://github.com/microsoft/vscode-azurefunctions/issues/1245#issuecomment-489176470).

This may work for some languages, but not all. Specifically Python would fail unless we waited for the lock lease message. See https://github.com/Microsoft/vscode-azurefunctions/issues/660 

I've asked the functions team to give us a debug-specific message that will work in all languages. See https://github.com/Azure/azure-functions-host/issues/4384

JustinGrote: The message changed so I thought this got resolved but apparently not. My bad :) guess this is stuck on that request.

EricJizbaMSFT: If you want to dig deeper and find out exactly when/why/how the message changed that would be great, but yeah at the moment I'm just waiting on that issue

Issue #589, opened 2018-09-13
Add log stream support for linux consumption apps


EricJizbaMSFT:
They don't have kudu, so log streaming currently isn't working

EricJizbaMSFT:
Per @asavaritayal:
> Instead of using the Kudu endpoint, launch the **Live Metrics Stream** view in the browser.

asavaritayal:
@EricJizbaMSFT do you have a rough estimate of when this might land?

EricJizbaMSFT:
@asavaritayal should be in our next release, which will probably be end of june-ish

jhodges10:
> @asavaritayal should be in our next release, which will probably be end of june-ish

Really need this for debugging. Currently trying to figure out why my function which runs locally won't work on Azure and it's rather impossible without a log stream :(

qalandarov:
@EricJizbaMSFT any updates on the progress?

PR #1433, opened 2019-08-07
"Support" log streaming for linux function apps

EricJizbaMSFT:
It's really just opening the Live Metrics stream in the portal. AFAIK there's no api for us to display this information directly in VS Code

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/589

### DIFF
--- a/src/utils/openUrl.ts
+++ b/src/utils/openUrl.ts
@@ -3,8 +3,14 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as vscode from 'vscode';
+// tslint:disable-next-line:no-require-imports
+import opn = require("opn");
 
 export async function openUrl(url: string): Promise<void> {
-    await vscode.env.openExternal(vscode.Uri.parse(url));
+    // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852
+    // Specifically, opening the Live Metrics Stream for Linux Function Apps doesn't work in this extension.
+    // await vscode.env.openExternal(vscode.Uri.parse(url));
+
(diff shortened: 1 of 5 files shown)


src/utils/openUrl.ts
```diff
@@ -3,8 +3,14 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import * as vscode from 'vscode';
+// tslint:disable-next-line:no-require-imports
+import opn = require("opn");
 
 export async function openUrl(url: string): Promise<void> {
-    await vscode.env.openExternal(vscode.Uri.parse(url));
+    // Using this functionality is blocked by https://github.com/Microsoft/vscode/issues/25852
```
nturinski: > Instead of referencing a VS Code bug, I would reference a bug in our repo. Specifically saying which scenario doesn't work with 'vscode.open' so that we will know what to test if/when we want to move back to that instead of opn

😅

EricJizbaMSFT: I did!
> Specifically, opening the Live Metrics Stream for Linux Function Apps doesn't work in this extension.

nturinski: Ohhh, that's what you meant.  I thought you meant to reference an issue in our Github repo. 👍

EricJizbaMSFT:
> Is there going to be a way to use this as a log streaming eventually? This is directing them to the portal to see their log stream, right?

Yeah it's directing them to their portal (which sucks) and it relies completely on app insights (which also sucks because that doesn't include things your print with `console.log`). Soooo yeah hopefully the Functions team eventually adds support for the same log streaming api that windows has

Issue #1344, opened 2019-06-14
Add remote debugging for node functions on Linux dedicated plans


mrcrane:
This is the first step to bring the App Services remote debugging experience to Functions.

Remote debugging requires two main pieces to be in place on the Azure side:
- The tunnel API needs to be available on Kudu.
- The runtime needs to launch in debug mode and configure the tunnel when the "remoteDebuggingEnabled" config is set.

The configuration piece has been taken care of here: Azure/azure-functions-docker#107

The Kudu tunnel API is currently only available on Linux dedicated plans. We have started the conversation about enabling the tunneling API for Linux consumption plans, but this is non-trivial work that will take some time.

This issue covers the work for Linux dedicated plans, which will give us a quick win and allow us to iron out all of the tooling issues. Once we have nailed down the experience with dedicated plans, we will be in a good place to add support for consumption plans when they become available.

mrcrane:
- The first step is to get the remote debug code into a place that it can be reused by functions.
  - See PR here: microsoft/vscode-azuretools#522
- After this we can update the functions extension to use the shared code
  - We will need to figure out how this will coexist with the existing, older Java remote debugging functionality. See: https://github.com/Microsoft/vscode-azurefunctions/wiki/Remote-Debug-Java-functions
  - For now we can just give it a separate experiment feature flag. The Java feature is experimental and is hidden behing the config `"azureFunctions.enableRemoteDebugging": true`. We can add a new flag such as `"azureFunctions.enableNodeRemoteDebugging": true` which enables this feature separately. This puts us in a place where we can test these features independently.
  - I will sync up with @fiveisprime to come up with a long term plan to merge with or replace the Java feature.

PR #1458, opened 2019-08-20
Add remote debugging support for Node.js on Linux App Service plans

mrcrane:
This will address microsoft/vscode-azurefunctions#1344 and it is the culmination of the refactoring work already done in microsoft/vscode-azuretools#522 and microsoft/vscode-azureappservice#1045

Given that the refactoring has already been done, this code is fairly straightforward. The main concern is around the fact that there is already a feature in place that supports Java remote debugging on Windows.

A few months ago (sorry for the delay) @fiveisprime @nturinski and I discussed this issue and we decided to go with 2 different remote debugging feature flags so that we can maintain both experiences for the time being. Since we expect this Node.js experience to be the main experience moving forward, I have reappropriated the `enableRemoteDebugging` flag (Python support is coming soon). Java debugging support will now be enabled using `enableJavaRemoteDebugging`. The settings page will now look like:

image

I'm happy to hear any feedback on how to make that experience less confusing.

More notes:
- Kudu is not available on Linux consumption, so we are still working with the Functions team to understand how this will work for that use case. Without Kudu we will likely need a new solution alongside this one.
- I have tested the Java remote debugging experience with this new code and it is still working
- Other than changing the feature flag, I have left the Java debugging experience untouched. It still uses the "Attach Debugger" wording unlike the "Start Remote Debugging" that we use for the App Service experience.
- In order to get Java debugging working with the new remote debugging code we would need the new tunneling API to be ported to Windows (it is currently Linux-only) and the Java runtime to respect the `remoteDebuggingEnabled` setting. This is feasible but probably low priority.


src/commands/remoteDebug/checkForRemoteDebugSupport.ts
```diff
@@ -0,0 +1,19 @@
...
+    // Read siteConfig.linuxFxVersion to determine debugging support
+    //   If the Function App is running on Windows, it will be empty
+    //   If the Function App is running on Linux consumption, it will be empty
+    //   If the Function App is running on a Linux App Service plan, it will contain Docker registry information, e.g. "DOCKER|repo.azurecr.io/image:tag"
+    //      The blessed Node.js Docker image will be something like "DOCKER|mcr.microsoft.com/azure-functions/node:2.0-node8-appservice"
+    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node')) {
+        return;
+    }
+
+    throw new Error('Azure Remote Debugging is currently only supported for Node.js Function Apps running on Linux App Service plans.');
```
EricJizbaMSFT: Based on the PR description - this only works on dedicated plans? I feel like that should be highlighted in the messaging since consumption is so much more popular

mrcrane: I looked into the conditional messaging approach but I wasn't able to find a good solution.

I went ahead and changed the messaging to be very clear, for this message and also for the description of the feature flag. Let me know what you think.

mrcrane:
I rebased on master and force-pushed to resolve merge conflicts.

EricJizbaMSFT:
Thanks!

PR #1488, opened 2019-09-12
Refactor remote debugging to support different languages in the future

mrcrane:
This PR will fail to build until it is updated with the changes in microsoft/vscode-azuretools#585 See that PR for more context.

This change is only here to accommodate the new language support that was added for App Service (see microsoft/vscode-azureappservice#1148). There is no functional difference, but it makes it very easy to support new languages in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -805,7 +805,7 @@
         },
         "@types/websocket": {
             "version": "0.0.37",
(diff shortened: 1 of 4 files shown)


EricJizbaMSFT:
Thanks!

PR #1506, opened 2019-09-23
Update azuretools package

nturinski:
cpUtils.executeCommand was intentionally left to pass in an outputChannel so there is logic prevent output being displayed to the channel.

### DIFF
--- a/src/commands/deploy/notifyDeployComplete.ts
+++ b/src/commands/deploy/notifyDeployComplete.ts
@@ -48,7 +48,7 @@ async function listHttpTriggerUrls(context: IActionContext, node: SlotTreeItemBa
         if (anonFunctions.length > 0) {
             ext.outputChannel.appendLog(localize('anonymousFunctionUrls', 'HTTP Trigger Urls:'));
             for (const func of anonFunctions) {
-                ext.outputChannel.appendLog(`  ${func.label}: ${func.triggerUrl}`);
+                ext.outputChannel.appendLine(`  ${func.label}: ${func.triggerUrl}`);
             }
         }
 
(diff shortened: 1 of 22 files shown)


src/commands/deploy/notifyDeployComplete.ts
```diff
@@ -46,19 +46,19 @@ async function listHttpTriggerUrls(context: IActionContext, node: SlotTreeItemBa
         const functions: AzExtTreeItem[] = await functionsNode.getCachedChildren(context);
         const anonFunctions: RemoteFunctionTreeItem[] = <RemoteFunctionTreeItem[]>functions.filter((f: AzureTreeItem) => f instanceof RemoteFunctionTreeItem && f.config.isHttpTrigger && f.config.authLevel === HttpAuthLevel.anonymous);
         if (anonFunctions.length > 0) {
-            ext.outputChannel.appendLine(localize('anonymousFunctionUrls', 'HTTP Trigger Urls:'));
+            ext.outputChannel.appendLog(localize('anonymousFunctionUrls', 'HTTP Trigger Urls:'));
             for (const func of anonFunctions) {
-                ext.outputChannel.appendLine(`  ${func.label}: ${func.triggerUrl}`);
+                ext.outputChannel.appendLog(`  ${func.label}: ${func.triggerUrl}`);
```
EricJizbaMSFT: I would use `appendLine` as it's listing out all the URLs. The timestamp is already displayed for the header above

nturinski:
`getDeployFsPath`, `createSiteAppSettings`, and `validateAppSettings` are all the other breaking changes.  I was going to update the deployFsPath in another PR, but the other two are not being used in this repo.

PR #1551, opened 2019-10-04
Update appservice breaking changes

nturinski:
appservice 0.47.0 changes `getDeployFsPath` to return the `originalDeployPath` that the user selected and the `effectiveDeployPath` where subpath settings have been appended.


src/commands/deploy/deploy.ts
```diff
@@ -102,6 +97,15 @@ export async function deploy(context: IActionContext, target?: vscode.Uri | stri
                     ext.outputChannel.appendLog(localize('stopFunctionApp', 'Stopping Function App: {0} ...', client.fullName));
                     await client.stop();
                 }
+                // preDeploy tasks are only required for zipdeploy so subpath may not exist
+                let deployFsPath: string = effectiveDeployFsPath;
+
+                if (siteConfig.scmType !== ScmType.None && effectiveDeployFsPath !== originalDeployFsPath) {
```
EricJizbaMSFT: There's already a variable `isZipDeploy` you can use here. Also I think your check for paths will work, but I would still prefer `isPathEqual` from pathUtils

Issue #1343, opened 2019-06-13
Bundle dependency not added when adding a binding to an existing function


fiveisprime:
1. Create a function app with an HTTP trigger
1. Right-click `function.json` and choose add binding
1. Choose "out", "Azure Queue Storage", then accept the defaults
1. Choose "AzureWebJobsStorage" for the storage account

Expected: `host.json` is updated with the bundle dependency the same way as when a new function is added.

Actual: the binding is added to `function.json`, but `host.json` is not updated.

EricJizbaMSFT:
As a manual workaround, users can add the "extensionBundle" property to their host.json like this:

```
{
  "version": "2.0",
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle",
    "version": "[1.*, 2.0.0)"
  }
}
```

See [here](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#extension-bundles) for more info

Issue #1202, opened 2019-04-19
Get default bundle range from cli-feed-v3.json


EricJizbaMSFT:
We don't want to write `[1.*, 2.0.0)` to every single project in case they add new templates depending on a later verision of the bundle. For example if they release templates relying on "1.1.0", we should be writing `[1.1.*, 2.0.0)`

PR #1556, opened 2019-10-08
Verify bundle in "Add Binding" feature

EricJizbaMSFT:
Previously we would only verify bundle in "Create function" feature. First commit just moved the code, [second commit](https://github.com/microsoft/vscode-azurefunctions/commit/b17a6a7b8c1ca87ebdf305d0956c35a8ab1fd814) actually changed it.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1343
Also fixes https://github.com/microsoft/vscode-azurefunctions/issues/1202 to get latest bundle range instead of hard-coded default.


EricJizbaMSFT:
Correct. If a users adds a function/binding other than http/timer, we need to make sure extensionBundle is defined.

Issue #1384, opened 2019-07-18
[Suggestion]Update the "Function App" to "slot" 


v-dantch:
**OS**: All
**Build Version:** [20190717.3](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=7757)
**Regression:** Not a Regression

**Repro Steps:** 
1. Create a new slot.
2. Check the context menu.

**Actual:**
Display "Delete Function App.../ Deploy to Function App...".

**Suggestion:**
Update the "Function App" to "slot".
![image](https://user-images.githubusercontent.com/34729092/61439432-1a0ef880-a974-11e9-9d7f-708e22c446f1.png)

PR #1559, opened 2019-10-08
Add slot-specific delete and deploy commands

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1384


src/commands/deploy/deploy.ts
```diff
@@ -21,7 +22,15 @@ import { notifyDeployComplete } from './notifyDeployComplete';
 import { runPreDeployTask } from './runPreDeployTask';
 import { verifyAppSettings } from './verifyAppSettings';
 
-export async function deploy(context: IActionContext, target?: vscode.Uri | string | SlotTreeItemBase, functionAppId?: string | {}): Promise<void> {
+export async function deployProductionSlot(context: IActionContext, target?: vscode.Uri | string | SlotTreeItemBase, functionAppId?: string | {}): Promise<void> {
+    await deploy(context, target, functionAppId, ProductionSlotTreeItem.contextValue);
+}
+
+export async function deploySlot(context: IActionContext, target?: vscode.Uri | string | SlotTreeItemBase, functionAppId?: string | {}): Promise<void> {
+    await deploy(context, target, functionAppId, SlotTreeItem.contextValue);
```
nturinski: I guess if they select an app that doesn't permissions for a slot, it'll just say no resources found?  I assume this would be a small use case since most people will be on Elastic anyway, but am wondering if App Service should do this too to support command palette for slots.

EricJizbaMSFT: Correct. I'll try to improve the error message in a separate PR

Issue #1548, opened 2019-10-04
Better handle function apps created before remote build was released


EricJizbaMSFT:
We need to better handle the situation when the user tries to deploy to a function app that doesn't support remote build (especially since that will be the default per https://github.com/microsoft/vscode-azurefunctions/issues/1541).

Right now they see "This operation is not supported by this app service plan."

fiveisprime:
How about..?

> The Function app you are deploying to doesn't support your current build configuration. Downgrade your build configuration or deploy to a newly created Function app.

Cancel | Learn More | Downgrade Configuration

In the "learn more" link mention that this was changed in August 2019 and provide a recommendation. I believe deploying to a new app would be best, but there are cases where a downgrade makes sense.

Should it also provide an option to create a new app that would dump them into the create experience? That seems like overkill to me..

EricJizbaMSFT:
Tweaked a bit:
> The selected Function App doesn\'t support your project\'s configuration. Deploy to a newer function app or downgrade your configuration.

A little shorter and honestly I never understood why this feature is called "build" in the first place. It's packaging up dependencies, not building anything.

PR #1560, opened 2019-10-08
Handle function apps that don't support remote build

EricJizbaMSFT:
Screen Shot 2019-10-08 at 4 15 02 PM

This will pop up for users with new projects and old function apps in Azure (created before Aug. 1st, 2019). I'll update the aka.ms link with more info closer to release date.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1548

### DIFF
--- a/src/commands/deploy/validateRemoteBuild.ts
+++ b/src/commands/deploy/validateRemoteBuild.ts
@@ -1,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { SiteClient } from 'vscode-azureappservice';
+import { IActionContext } from 'vscode-azureextensionui';
+import { deploySubpathSetting, packTaskName, preDeployTaskSetting, ProjectLanguage } from '../../constants';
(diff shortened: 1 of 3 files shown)


src/commands/deploy/validateRemoteBuild.ts
```diff
@@ -0,0 +1,39 @@
...
+import { tryGetFunctionProjectRoot } from '../createNewProject/verifyIsProject';
+import { ensureGitIgnoreContents } from '../initProjectForVSCode/InitVSCodeStep/PythonInitVSCodeStep';
+
+export async function validateRemoteBuild(context: IActionContext, client: SiteClient, workspacePath: string, language: ProjectLanguage): Promise<void> {
+    if (language === ProjectLanguage.Python && !client.kuduUrl) {
+        const remoteBuildKey: string = 'scmDoBuildDuringDeployment';
+
+        const remoteBuild: boolean | undefined = getWorkspaceSetting(remoteBuildKey, workspacePath);
+        if (remoteBuild) {
+            const message: string = localize('remoteBuildNotSupported', 'The selected Function App doesn\'t support your project\'s configuration. Deploy to a newer Function App or downgrade your config.');
```
nturinski: How does it know that the function app doesn't support remote build?  Does it only work with kudu?

EricJizbaMSFT: Yes it only works if it has kudu (I mean it's probably oryx, but we treat those the same [in SiteClient](https://github.com/microsoft/vscode-azuretools/blob/master/appservice/src/SiteClient.ts#L77))

PR #1578, opened 2019-10-14
Refactor a few functions that are getting long 👷🏻‍♂️🏗

EricJizbaMSFT:
- Move all node logic for deploy into it's own function (getDeployNode is meant to be analogous to getDeployFsPath)
- Move registering commands to its own file in the commands folder

Should be no functional difference


src/commands/deploy/deploy.ts
```diff
@@ -79,14 +51,14 @@ async function deploy(context: IActionContext, target: vscode.Uri | string | Slo
         throw new Error(localize('pythonNotAvailableOnWindows', 'Python projects are not supported on Windows Function Apps.  Deploy to a Linux Function App instead.'));
     }
 
-    await validateRemoteBuild(context, client, workspaceFolder.uri.fsPath, language);
+    await validateRemoteBuild(context, node.root.client, workspaceFolder.uri.fsPath, language);
```
nturinski: Why get rid of the client variable?  You call `node.root.client` enough times that it makes sense to keep the variable.

EricJizbaMSFT: Idk it just felt right - maybe because I was working on line length and that was an easy way to delete a line. No strong preference

PR #1587, opened 2019-10-16
Add basic test for "Add Binding" 🧪

EricJizbaMSFT:
As evidence these would've caught the recent bug, here's a failing build without the fix: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=11284&view=logs

### DIFF
--- a/test/addBinding.test.ts
+++ b/test/addBinding.test.ts
@@ -46,7 +46,7 @@ suite('Add Binding', async () => {
     });
 
     test('Tree', async () => {
-        const treeItem: AzExtTreeItem | undefined = await ext.tree.findTreeItem('/localProjecttestWorkspace/functions/HttpTriggerTest/bindings', createTestActionContext());
+        const treeItem: AzExtTreeItem | undefined = await ext.tree.findTreeItem(`/localProjecttestWorkspace/functions/${functionName}/bindings`, createTestActionContext());
         assert.ok(treeItem, 'Failed to find tree item');
         await validateAddBinding([treeItem], []);
     });
--- a/test/addBinding.test.ts
+++ b/test/addBinding.test.ts
@@ -1,0 +1,78 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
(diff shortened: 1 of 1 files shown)


test/addBinding.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        }
+        await validateAddBinding([], userInputs);
+    });
+
+    test('Uri', async () => {
+        await validateAddBinding([Uri.parse(functionJsonPath)], []);
+    });
+
+    test('Tree', async () => {
+        const treeItem: AzExtTreeItem | undefined = await ext.tree.findTreeItem('/localProjecttestWorkspace/functions/HttpTriggerTest/bindings', createTestActionContext());
```
nturinski: Why are you using a local project tree rather than the subscription item?

test/addBinding.test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    suiteTeardown(async () => {
+        const finalBindingsCount: number = await getBindingsCount();
+        assert.equal(finalBindingsCount, initialBindingsCount + 3, 'Not all expected bindings were added.');
+    });
+
+    test('Command Palette', async () => {
+        const userInputs: string[] = [functionName];
+        // https://github.com/microsoft/vscode-azurefunctions/issues/1586
+        const isLoggedOut: boolean = (await ext.azureAccountTreeItem.getCachedChildren(createTestActionContext())).some(s => /sign in/i.test(s.label));
```
nturinski: When would this happen? If the service principal was being funky?

test/addBinding.test.ts
nturinski: nit: I don't really think we'd likely change functionName from "HttpTriggerTest", but you could use the variable here instead of hardcoding it

test/addBinding.test.ts
EricJizbaMSFT: It happens for normal builds in Azure Dev Ops - we only use the service principal for nightly builds. Also if you're just running tests locally and you're logged out

test/addBinding.test.ts
EricJizbaMSFT: "Add Binding" is a local-only scenario. You can't add bindings in Azure after you've deployed

Issue #1489, opened 2019-09-12
Add support for Azure Functions 3.0


EricJizbaMSFT:
https://github.com/Azure/app-service-announcements/issues/200

EricJizbaMSFT:
_From @ColbyTresness in https://github.com/microsoft/vscode-azurefunctions/issues/1546_:
> Per email:
> 
> •	Whenever anyone opens the extension, check their core tools version
> •	If V3, use V3, if V1, use V1. Nothing complex here
> •	If V2, show info bubble about V3, with a forward link to a doc explaining how to update and the differences V3 provides. Until customer manually updates core tools version, use V2.
> •	Allow customers to preview V3 alongside V2 if they wish by leveraging dev containers. The doc referenced above should contain a link to instructions on how to set this up.

fiveisprime:
Regarding this point:

>  If V2, show info bubble about V3, with a forward link to a doc explaining how to update and the differences V3 provides. Until customer manually updates core tools version, use V2.

Let's skip the notification until we have compelling guidance on why customers should upgrade to v3.

EricJizbaMSFT:
Closing this as v3 support is basically done. Any remaining work should be covered by other issues:
https://github.com/microsoft/vscode-azurefunctions/labels/v3

PR #1606, opened 2019-10-21
Initial support for Azure Functions v3 Preview 🔍👀

EricJizbaMSFT:
Main changes for this PR:
1. Refactored versioning logic so that there was less hard-coded stuff
1. Went through all remaining instances of "1" or "2" and added a "3"
1. Renamed `ProjectRuntime` to `FuncVersion`. The old name doesn't make that much sense and conflicts with the definition of "runtime" related to node/.net/java
1. Added backup templates for v3

Related to https://github.com/microsoft/vscode-azurefunctions/issues/1489

Created some issues for remaining work: https://github.com/microsoft/vscode-azurefunctions/labels/v3


EricJizbaMSFT:
ping @nturinski

src/funcCoreTools/installOrUpdateFuncCoreTools.ts
```diff
@@ -31,12 +31,11 @@ export async function installOrUpdateFuncCoreTools(): Promise<void> {
             packageManager = (await ext.ui.showQuickPick(picks, { placeHolder })).data;
         }
 
-        let projectRuntime: ProjectRuntime | undefined = await tryGetLocalRuntimeVersion();
-        // tslint:disable-next-line:strict-boolean-expressions
-        if (!projectRuntime) {
-            projectRuntime = await promptForProjectRuntime(localize('selectLocalRuntime', 'Failed to detect local runtime automatically. Select your runtime to update'));
+        let version: FuncVersion | undefined = await tryGetLocalFuncVersion();
+        if (version === undefined) {
```
nturinski: Why break style all of a sudden and write `version === undefined`?

src/templates/CentralTemplateProvider.ts
```diff
@@ -37,44 +38,44 @@ export class CentralTemplateProvider {
...
 
-        let key: string = runtime;
+        let key: string = version;
         for (const provider of providers) {
             key += provider.templateType;
         }
 
-        context.telemetry.properties.runtime = runtime;
-        context.telemetry.properties.language = language;
+        context.telemetry.properties.projectRuntime = version;
```
nturinski: If you're changing the telemetry anyway, why leave the property name as projectRuntime?

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -88,15 +89,15 @@ export class LocalProjectTreeItem extends AzExtParentTreeItem implements Disposa
...
         const data: any = await fse.readJSON(path.join(this.projectPath, hostFileName));
-        return parseHostJson(data, runtime);
+        return parseHostJson(data, version);
     }
 
-    public async getRuntime(): Promise<ProjectRuntime> {
-        const rawSetting: string | undefined = getWorkspaceSetting(projectRuntimeSetting, this.workspacePath);
-        return nonNullValue(convertStringToRuntime(rawSetting), 'runtime');
+    public async getVersion(): Promise<FuncVersion> {
+        const rawSetting: string | undefined = getWorkspaceSetting(funcVersionSetting, this.workspacePath);
```
nturinski: How is it guaranteed that this is nonNull?

src/vsCodeConfig/verifyTargetFramework.ts
```diff
@@ -20,33 +20,37 @@ export async function verifyTargetFramework(projectLanguage: ProjectLanguage, fo
...
-                        if (tasksResult) {
-                            await tasksResult.update();
-                        }
-
-                        if (settingsResult) {
-                            await settingsResult.update();
-                        }
+            let targetFramework: string;
+            try {
+                targetFramework = await getTargetFramework(path.join(projectPath, projFileName));
```
nturinski: What's the expected error here?

src/funcCoreTools/installOrUpdateFuncCoreTools.ts
EricJizbaMSFT: I use `=== undefined` somewhat frequently for enums, so I don't think it's breaking style.

src/templates/CentralTemplateProvider.ts
EricJizbaMSFT: A bunch of other cases use `.projectRuntime` for telemetry and I didn't want to change them all. This was the only one using `.runtime` so I only changed it to be consistent

src/vsCodeConfig/verifyTargetFramework.ts
EricJizbaMSFT: A few possibilities: We fail to find the csproj, we can't read the csproj, we can't find the TargetFramework property

src/funcCoreTools/getBrewPackageName.ts
```diff
@@ -0,0 +1,21 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { funcPackageName } from '../constants';
+import { FuncVersion, getMajorVersion, isPreviewVersion } from '../FuncVersion';
+
+export function getBrewPackageName(version: FuncVersion): string {
+    let result: string = funcPackageName;
+    if (version !== FuncVersion.v2) {
```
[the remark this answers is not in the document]
EricJizbaMSFT: Can do

src/tree/localProject/LocalProjectTreeItem.ts
EricJizbaMSFT: Fair question - big picture I'm not too worried about it because I would see this error in telemetry if it was a problem, but I'll look into it after merging this PR (since the behavior didn't change as a part of this PR)

Issue #1604, opened 2019-10-21
Support .NET templates with only .NET Core 3.0 installed on machine


EricJizbaMSFT:
Right now we require 2.* to be installed because of [this tool](https://github.com/microsoft/vscode-azurefunctions/tree/master/tools/JsonCli)

PR #1612, opened 2019-10-23
Update JsonCli tool to work with .NET Core 3.0 🔧🆙

EricJizbaMSFT:
We use this tool to create C# functions, but it only works if they have a 2.x version of .NET Core on their machine. Now it should also work if they have 3.x.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1604


src/templates/dotnet/executeDotnetTemplateCommand.ts
```diff
@@ -36,3 +38,36 @@ export function getDotnetItemTemplatePath(runtime: ProjectRuntime): string {
...
+            versions += await cpUtils.executeCommand(undefined, workingDirectory, 'dotnet', '--list-sdks');
+        } catch {
+            // ignore
+        }
+
+        const majorVersions: string[] = ['3', '2'];
+        for (const majorVersion of majorVersions) {
+            const regExp: RegExp = new RegExp(`^\\s*${majorVersion}\\.`);
+            if (regExp.test(versions)) {
+                cachedFramework = `netcoreapp${majorVersion}.0`;
```
nturinski: What's the point of getting the sdk version if you just end up setting cachedFramework to either 3 or 2?

tools/JsonCli/.vscode/tasks.json
```diff
@@ -12,18 +12,36 @@
             "problemMatcher": "$msCompile"
         },
         {
-            "label": "publish",
+            "label": "publish2",
```
nturinski: I don't really like these labels as they seme more numerical rather than framework-specific.  What do you think about publish (2.0) and publish (3.0) or something like that?

tools/JsonCli/src/Microsoft.TemplateEngine.JsonCli.csproj
```diff
@@ -1,13 +1,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
-  <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <TargetFramework>netcoreapp2.0</TargetFramework>
-  </PropertyGroup>
+    <PropertyGroup>
+        <OutputType>Exe</OutputType>
+        <TargetFrameworks>netcoreapp2.0;netcoreapp3.0</TargetFrameworks>
```
nturinski: Just to be sure, but I'm assuming these changes are coming from something else that you're copying in, right?

src/templates/dotnet/executeDotnetTemplateCommand.ts
EricJizbaMSFT: I'm getting the sdk version to see if they have 3 or 2 on their machine. If they don't, it'll throw an error below and never set the cachedFramework

tools/JsonCli/.vscode/tasks.json
EricJizbaMSFT: No preference - this is an internal tool that we touch like once a year haha

src/templates/dotnet/executeDotnetTemplateCommand.ts
nturinski: `--version` doesn't get the major version? 😮

EricJizbaMSFT: Users can have multiple versions of the sdk installed on their machine and `dotnet --version` will only print out one version. Usually that's the latest sdk version, but users can configure it to be any version.

Here's example output:
```
> dotnet --version
2.1.202
> dotnet --list-sdks
1.1.14 [/usr/local/share/dotnet/sdk]
2.1.202 [/usr/local/share/dotnet/sdk]
```

tools/JsonCli/src/Microsoft.TemplateEngine.JsonCli.csproj
EricJizbaMSFT: Most of the changes, yes

Issue #1290, opened 2019-05-21
Default tasks.json does not work with nested folder structure


ahmedelnably:
Original issue https://github.com/Azure/azure-functions-core-tools/issues/1325

I have a function that sits somewhere under a src subfolder. The default tasks.json creates me the following task to start the function:

    {
      "type": "func",
      "dependsOn": "build",
      "options": {
        "cwd": "${workspaceFolder}/src/FuncClaims.Api/bin/Debug/netstandard2.0"
      },
      "command": "host start",
      "isBackground": true,
      "problemMatcher": "$func-watch"
    }
When attempting to build it, I get the following reported back:

Error: The func task detection didn't contribute a task for the following configuration:
{
    "type": "func",
    "dependsOn": "build",
    "options": {
        "cwd": "${workspaceFolder}/src/FuncClaims.Api/bin/Debug/netstandard2.0"
    },
    "command": "host start",
    "isBackground": true,
    "problemMatcher": "$func-watch"
}
The task will be ignored.
I found a older version of the task, and this does seem to work OK:

    {
      "label": "runFunctionsHost",
      "type": "shell",
      "dependsOn": "build",
      "options": {
        "cwd": "${workspaceFolder}/src/FuncClaims.Api/bin/Debug/netstandard2.0"
      },
      "command": "func host start",
      "isBackground": true,
      "presentation": {
        "reveal": "always"
      },
      "problemMatcher": "$func-watch"
    }

RadekTester:
So what's the current workaround for this?

EricJizbaMSFT:
@RadekTester the current workaround is to use the older version of the task mentioned above instead of the default task in the template

EricJizbaMSFT:
Actually you can use the existing task if you set the VS Code setting `azureFunctions.projectSubpath`

PR #1618, opened 2019-10-28
Use `resolveTask` to support non-standard tasks/projects 📋

EricJizbaMSFT:
VS Code [added support](https://code.visualstudio.com/updates/v1_37#_resolvetask-support) for the `resolveTask` method in 1.37. This lets us do two things:
1. Support tasks where the user has customized the command (previously `getFuncTaskCommand` was a workaround to support this - now I can remove that)
1. Support projects that are several sub folders down (aka too far down to be auto-detected). This fixes  https://github.com/microsoft/vscode-azurefunctions/issues/1290


src/debug/FuncTaskProvider.ts
```diff
@@ -73,20 +73,50 @@ export class FuncTaskProvider implements TaskProvider {
...
-        await callWithTelemetryAndErrorHandling('resolveTask', async (context: IActionContext) => {
+    public async resolveTask(task: Task, _token?: CancellationToken | undefined): Promise<Task | undefined> {
+        return await callWithTelemetryAndErrorHandling('resolveTask', async (context: IActionContext) => {
             context.telemetry.properties.isActivationEvent = 'true';
             context.errorHandling.suppressDisplay = true;
             context.telemetry.suppressIfSuccessful = true;
+
+            // tslint:disable-next-line: no-unsafe-any
+            const command: string | undefined = task.definition.command;
+            if (command && task.scope !== undefined && task.scope !== TaskScope.Global && task.scope !== TaskScope.Workspace) {
```
nturinski: Would it not be an error if we think there's a task to resolve, but it's undefined?

EricJizbaMSFT: We suppress errors (see above) so doesn't really matter if we throw an error here or not. resolveTask gets called a lot by VS Code and so we don't want to show errors when people might not even be trying to run a task from us

PR #1619, opened 2019-10-28
Fix version bump message for brew v3 ☕️

EricJizbaMSFT:
As a reminder, here's the relevant warning:
Screen Shot 2019-10-28 at 3 16 22 PM

I decided to give up on an aka.ms link for a few reasons:
1. If this fails for any reason, we suppress the error. The only downside is the user isn't prompted to update. (Although I added some extra logic just to make sure we're not recommending updating past a major version)
1. I think it's more likely that we forget to update an aka.ms link for this as opposed to the underlying link structure changing


nturinski:
Also brew is referring to 🍻 not ☕️

PR #1634, opened 2019-10-31
Don't prompt for func version as often

EricJizbaMSFT:
Talked to Matt and there are a few cases where we don't want to prompt for func version:
1. If you're creating a project, we auto detect the version from the func cli installed on your machine. If you don't have the func cli installed, this PR changes to automatically use v2 on mac/linux instead of prompting for v2 vs. v3 preview (v1 is windows-only).
1. If you try to debug without the func cli installed, we prompt you to install. This PR changes to use the version already set in your project

### DIFF
--- a/src/FuncVersion.ts
+++ b/src/FuncVersion.ts
@@ -18,14 +18,13 @@ export enum FuncVersion {
 export const latestGAVersion: FuncVersion = FuncVersion.v2;
 
 export async function promptForFuncVersion(message?: string): Promise<FuncVersion> {
-    const picks: IAzureQuickPickItem<FuncVersion | undefined>[] = [];
+    let picks: IAzureQuickPickItem<FuncVersion | undefined>[] = [
+        { label: 'Azure Functions v2', description: '(.NET Core)', data: FuncVersion.v2 },
+        { label: 'Azure Functions v3 Preview', description: '(.NET Core)', data: FuncVersion.v3 },
+        { label: 'Azure Functions v1', description: '(.NET Framework)', data: FuncVersion.v1 }
+    ];
 
-    picks.push({ label: 'Azure Functions v2', description: '(.NET Core)', data: FuncVersion.v2 });
-    picks.push({ label: 'Azure Functions v3 Preview', description: '(.NET Core)', data: FuncVersion.v3 });
-
-    if (isWindows) {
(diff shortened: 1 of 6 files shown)


src/FuncVersion.ts
```diff
@@ -18,14 +18,13 @@ export enum FuncVersion {
 export const latestGAVersion: FuncVersion = FuncVersion.v2;
 
 export async function promptForFuncVersion(message?: string): Promise<FuncVersion> {
-    const picks: IAzureQuickPickItem<FuncVersion | undefined>[] = [];
+    let picks: IAzureQuickPickItem<FuncVersion | undefined>[] = [
```
nturinski: I know why we list 2, 3, 1 but it still bothers me 🔢

src/commands/deploy/runPreDeployTask.ts
```diff
@@ -17,7 +17,7 @@ export async function runPreDeployTask(context: IActionContext, deployFsPath: st
     const preDeployTask: string | undefined = getWorkspaceSetting(preDeployTaskSetting, deployFsPath);
     if (preDeployTask && preDeployTask.startsWith('func:')) {
         const message: string = localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to run preDeployTask "{0}".', preDeployTask);
-        if (!await validateFuncCoreToolsInstalled(message)) {
+        if (!await validateFuncCoreToolsInstalled(message, deployFsPath)) {
```
nturinski: Would the workspace setting be in the deployFsPath or the workspace folder root?  I thought for functions projects, most of the settings would be in the root folder.

src/FuncVersion.ts
EricJizbaMSFT: Open to 2, 1, 3 if you like that better

src/commands/deploy/runPreDeployTask.ts
EricJizbaMSFT: You don't have to pass the exact workspace folder path when getting a setting. VS Code will automatically drill down to the folder containing that path

EricJizbaMSFT: Also to be clear - VS Code ignores "settings.json" files that are not at the root anyways

PR #1649, opened 2019-11-01
Prep for 0.20.0 release

EricJizbaMSFT:


CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
 
 # Change Log
 
+## 0.20.0 - 2019-11-04
+
+### Added
+- Select Node.js or Python runtime version when creating a Function App in Azure
+- Added support for [Azure Functions v3 Preview](https://aka.ms/AA6i3ev)
+  - Install v3 preview of the func cli from the command palette (brew and npm supported)
```
nturinski: This is weird because it's listing features and instructions in the same list.

CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
 
 # Change Log
 
+## 0.20.0 - 2019-11-04
+
+### Added
+- Select Node.js or Python runtime version when creating a Function App in Azure
+- Added support for [Azure Functions v3 Preview](https://aka.ms/AA6i3ev)
```
nturinski: Don't need added here, as the header is ###Added

EricJizbaMSFT: I have conflicting feelings about this, but I do this all the time. Most bullet points start with a verb and so if I remove "Added" it feels like "Support" should be a verb when actually it's a noun in this case

PR #1663, opened 2019-11-05
Update "Create New Project" tests 🧪 to work with v3

EricJizbaMSFT:
Tests were actually failing for backup templates, so I updated those.

Removed some tests for languages that were only experimental in v1 and not supported in v2/v3.


tslint.json
```diff
@@ -86,6 +86,7 @@
         ],
         "no-use-before-declare": false,
         "newline-per-chained-call": false,
-        "no-dynamic-delete": false
+        "no-dynamic-delete": false,
+        "trailing-comma": false
```
nturinski: Is there a way to move this to a test specific tslint.json?

EricJizbaMSFT: [Yes](https://github.com/microsoft/vscode-azurefunctions/blob/master/test/tslint.json), but I want to disable it for the whole repo. It's easier to copy/paste/rearrange lines if they all have a comma at the end

Issue #1658, opened 2019-11-05
Python 3.7 for local project


EricJizbaMSFT:
Today we default to 3.6 when creating a venv for new projects. We need to change the default logic to also support 3.7 and potentially prompt the user if we detect both.

Issue #1393, opened 2019-07-20
[Discussion] `azureFunctions.pythonVenv` vs. `python.pythonPath`


keitherskine:
I have a query about the version of Python used when I deploy an Azure Function from VS Code using the "up-arrow" in the Azure extension.  This is the "up-arrow" that is labelled "Deploy to Function App..." when you hover over it.

In my case, I am using a Python virtual environment that is set up for Python 3.6, although I also have Python 3.7 installed on my PC.  The name of the venv can be seen in the lower left corner of VS Code as, correctly, "3.6.6 64-bit ('azure': venv)".  Yet when I deploy the function, it appears VS Code is using the default Python version (3.7) on my PC.  Here is the start of the output:
```
> Executing task: func pack <

Deleting the old .python_packages directory
Python 3.6.x is recommended, and used in Azure Functions. You are using Python version Python 3.7.3.
Collecting azure-storage-blob==2.0.1 (from -r C:\Users\Keith\Documents\GitHub\MyAzureFunctionProj\requirements.txt (line 1))
```
As you can see, it warns about using version 3.7.3, even though I previously selected my 3.6 virtual environment.

In an attempt to fix this, I set the `azureFunctions.pythonVenv` settings to "C:\\Users\\Keith\\pyvenvs\\azure".  Here is the new output:
```
> Executing task: C:\Users\Keith\pyvenvs\azure\Scripts\activate ; func pack <

& C:/Users/Keith/pyvenvs/azure/Scripts/activate.ps1
Deleting the old .python_packages directory
Python 3.6.x is recommended, and used in Azure Functions. You are using Python version Python 3.7.3.
Collecting azure-storage-blob==2.0.1 (from -r C:\Users\Keith\Documents\GitHub\MyAzureFunctionProj\requirements.txt (line 1))
```
VS Code now appears to be trying to activate the Python 3.6 venv from `azureFunctions.pythonVenv`, yet still warns about using Python 3.7.  Is VS Code using the venv? Did the venv get activated?  Curious that both `activate.bat` and `Activate.ps1` seem to be being called.

This doesn't look right, although it has to be said it doesn't appear to be causing an issue.  Nevertheless, it would still be nice if the Azure extension used the venv defined in the `python.venvPath` setting rather than having a separate `azureFunctions.pythonVenv` setting.

EricJizbaMSFT:
[...]

PR #1670, opened 2019-11-08
Prompt user to pick python version when creating a project 🐍

EricJizbaMSFT:
The current experience will default to Python 3.6 if we find it on the machine and otherwise prompt the user to enter the python path. Instead, we will check for any supported version and prompt something like this:
Screen Shot 2019-11-07 at 4 01 17 PM

This will show every time users create a python project, which is kind of annoying from a "number of prompts" standpoint. To mitigate this, I moved the logic into wizard prompt steps (instead of `PythonProjectCreateStep`, an execute step) which means users get proper support for things like the back button. On the plus side, these prompts make it more obvious what python version is being used and how to skip creating a venv.

Validation if they try to manually enter a path:
Screen Shot 2019-11-07 at 4 16 10 PM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1658

### DIFF
--- a/src/commands/createNewProject/pythonSteps/IPythonVenvWizardContext.ts
+++ b/src/commands/createNewProject/pythonSteps/IPythonVenvWizardContext.ts
@@ -1,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "vscode-azureextensionui";
+
+export interface IPythonVenvWizardContext extends IActionContext {
+    projectPath: string;
+    pythonAlias?: string;
+    manuallyEnterAlias?: boolean;
+    useExistingVenv?: boolean;
+    venvName?: string;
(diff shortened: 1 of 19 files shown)


src/commands/createNewProject/pythonSteps/IPythonVenvWizardContext.ts
```diff
@@ -0,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IActionContext } from "vscode-azureextensionui";
+
+export interface IPythonVenvWizardContext extends IActionContext {
+    projectPath: string;
+    pythonAlias?: string;
```
nturinski: How about a setting to set this by default so that users can skip the python alias prompt if they want?

EricJizbaMSFT: Will wait for user feedback. We already have a few venv/python related settings and I don't feel like adding to the [confusion](https://github.com/microsoft/vscode-azurefunctions/issues/1393)

PR #1687, opened 2019-11-15
Better error message when resource not found 

EricJizbaMSFT:
There's an issue with the staging templates and the error was not very helpful. It was `Cannot read property 'match' of undefined`. Now the error will be `Resource "$timerTrigger_name_help" not found.`, which is much easier to diagnose.

Also turns out `defaultValue` can have "$" in the value and should not be using `getVariableValue`.

### DIFF
--- a/src/templates/script/parseScriptTemplates.ts
+++ b/src/templates/script/parseScriptTemplates.ts
@@ -6,6 +6,7 @@
 import { isString } from 'util';
 import { ProjectLanguage } from '../../constants';
 import { IFunctionBinding, ParsedFunctionJson } from '../../funcConfig/function';
+import { localize } from '../../localize';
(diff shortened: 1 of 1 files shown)


src/templates/script/parseScriptTemplates.ts
```diff
@@ -124,7 +126,7 @@ function parseScriptSetting(data: object, resources: IResources, variables: IVar
         resourceType: rawSetting.resource,
         valueType: rawSetting.value,
         description: rawSetting.help ? replaceHtmlLinkWithMarkdown(getResourceValue(resources, rawSetting.help)) : undefined,
-        defaultValue: rawSetting.defaultValue ? getVariableValue(resources, variables, rawSetting.defaultValue) : undefined,
+        defaultValue: rawSetting.defaultValue,
```
nturinski: Can you get rid of getVariableValue now? On L85?

Issue #1697, opened 2019-11-18
Update marketplace icon 


fiveisprime:
Use the new Functions icon

![image](https://user-images.githubusercontent.com/1186948/69075208-c7a71c80-09e5-11ea-9c88-51c84ccf7e10.png)

btardif:
you can get the icon here: https://cds-icons.azurewebsites.net/downloads/All_Icons/Compute/Icon-compute-29-Function-Apps.svg

EricJizbaMSFT:
@btardif do you have a "png" version of that icon - or a preferred tool for converting svgs to pngs? Seems like VS Code doesn't like svgs for the marketplace
![Screen Shot 2019-11-19 at 10 08 15 AM](https://user-images.githubusercontent.com/11282622/69173458-abbe7c00-0ab4-11ea-81e5-3f7aa13451aa.png)

PR #1703, opened 2019-11-20
Update marketplace icon

EricJizbaMSFT:
And use a better version of the svg for the tree

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1697


EricJizbaMSFT:
Also updated the gallery banner color because I didn't like the blue on blue. Went through the colors [here](https://github.com/microsoft/vscode/blob/master/extensions/theme-defaults/themes/dark_defaults.json) and picked the one I liked best.

Old color:
Screen Shot 2019-11-20 at 4 09 01 PM

New color:
Screen Shot 2019-11-20 at 4 03 07 PM

PR #1709, opened 2019-11-21
Only add bundles if extensions.csproj does not exist

### LINKED ISSUE #1698: Do not use bundles when extensions.csproj exists
Currently this extension is [checking for the existence](https://github.com/microsoft/vscode-azurefunctions/blob/9d9a73014c5edb17db523ba907ab017ff005dabb/src/utils/verifyExtensionBundle.ts#L59) of the extensions install command in `task.json` and `settings.json` to determine if bundles should be added to `host.json`.

It's probably more reliable to check for the existence of `extensions.csproj`. For instance, this workflow of installing extensions doesn't add the extensions install command in the vscode settings, so the VS Code extension keeps adding bundles back to host.json:

1. Create JS functions project using VS Code extension
1. `func extensions install -p Microsoft.Azure.WebJobs.Extensions.DurableTask -v 2.0.0`
1. Remove bundles from `host.json` if exists
1. Create a durable orchestration function using VS Code extension (this adds bundles into `host.json`)

/cc @ColbyTresness @soninaren

anthonychu:
Fixes #1698

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -43,6 +43,7 @@ export const settingsFileName: string = 'settings.json';
 export const vscodeFolderName: string = '.vscode';
 export const gitignoreFileName: string = '.gitignore';
 export const requirementsFileName: string = 'requirements.txt';
+export const extensionsCsprojFilename: string = 'extensions.csproj';
 
 export enum PackageManager {
     npm = 'npm',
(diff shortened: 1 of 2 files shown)


EricJizbaMSFT:
Closing in favor of discussion in #1698

PR #1745, opened 2019-12-09
Add @types/node to dev deps in package.json for TS projects

burkeholland:
If the user picks TypeScript, there is a 99% chance they are going to want at least the Node types as well. For instance, they can't use `process` without throwing an error if they don't have the types. I think it's safe to assume they need it and include by default.

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
@@ -21,7 +21,8 @@ export class TypeScriptProjectCreateStep extends JavaScriptProjectCreateStep {
 
     protected packageJsonDevDeps: { [key: string]: string } = {
         '@azure/functions': '^1.0.2-beta2',
-        typescript: '^3.3.3'
+        typescript: '^3.3.3',
+        '@types/node': '^12.12.15'
     };
 
     public async executeCore(context: IProjectWizardContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {


src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
```diff
@@ -21,7 +21,8 @@ export class TypeScriptProjectCreateStep extends JavaScriptProjectCreateStep {
 
     protected packageJsonDevDeps: { [key: string]: string } = {
         '@azure/functions': '^1.0.2-beta2',
-        typescript: '^3.3.3'
+        typescript: '^3.3.3',
+        '@types/node': '^12.12.15'
```
EricJizbaMSFT: The only problem I have with this is the version. People might not have node 12 and Azure Functions only [recently](https://github.com/Azure/azure-functions-nodejs-worker/issues/200) added some support for 12. I think ideally this version would be based on the version of Node installed on the user's machine, but I'm not sure the added complexity is worth it.

Are you aware of any popular project generators for typescript and do they include node types by default? If so, what version? I wonder if users would have more confusion/annoyance about missing types or types with the wrong version

burkeholland: This is a good question for @bnb. Tierney - what is your opinion on this? I'm trying to remove a friction point here, but I see the point Eric is making about Node versions. 

Thoughts?

bnb: Yeah, you probably wouldn’t want to get Node.js 12 types for a Node.js 10 function.

There’s also the additional challenge of The possibility of using Node.js 10 locally but Node.js 12 up in the deployment. Not sure if this matters for TS functions.

I’m not familiar enough with function development to know if this is possible in the sandbox, but could you just fetch the system version and then pass that as the value?

EricJizbaMSFT:
Talked with @fiveisprime. We decided it's easiest to just let people manage their own types for this. It's a bit annoying, but there's too many complications around the version. I'm open to changing this decision if we find a popular project generator we can model after or if more users file feedback

PR #1812, opened 2020-01-17
Simplify validation of language/version when deploying

### LINKED ISSUE #1662: Improve validation of mismatched runtime version when deploying
Related to v3 preview, there's a few things we might want to do:
1. Allow users to change their project version, instead of just the remote version
1. Validate app settings other than extension version? This might be more of a v1->v2 thing than v2->v3

EricJizbaMSFT:
In the past, we would validate the local language/version with the remote langauge/version when deploying and gave the user the option to update remote app settings. That gets complicated really quickly and we'd honestly prefer users just created a brand new app with the right settings. So instead we will do this:
1. If the versions don't match, warn the user and give them the option to "Deploy Anyway". Many apps will "just work" if there's a mismatch between v2 and v3
Screen Shot 2020-01-17 at 1 25 30 PM

1. If the languages don't match, throw an error. We think this is an unlikely scenario anyways (most people only use one language), and there's no point in "Deploy Anyway" because it won't work
Screen Shot 2020-01-17 at 1 31 50 PM

Fixes #1662


EricJizbaMSFT:
The first case applies for v1 as well.

The second case just deploys for v1. The app setting for language wasn't introduced until v2

PR #1826, opened 2020-01-23
Fix handling of old brew package tags

EricJizbaMSFT:
So I already moved `getBrewPackageName` to the new pattern (which always uses `@<major version>`). However, here are a few scenarios that are messed up because of this:
1. If you have old tag (i.e. "azure-functions-core-tools-v3-preview") installed, and run our command to uninstall the func cli it will fail because it tries to uninstall new tag (i.e. "azure-functions-core-tools@3") which doesn't exist on the machine
1. If you have old tag installed and cli team pushes an update, you will never be prompted to "Update to the latest version for the best experience" because it's looking for new tag only

The main fix is just to check the old tags in addition to the new tags. However, I also decided to have the "update" case uninstall the old tag (which will probably be deprecated soon? idk) and install the new tag.

### DIFF
--- a/src/funcCoreTools/getBrewPackageName.ts
+++ b/src/funcCoreTools/getBrewPackageName.ts
@@ -5,7 +5,37 @@
 
 import { funcPackageName } from '../constants';
 import { FuncVersion, getMajorVersion } from '../FuncVersion';
+import { cpUtils } from '../utils/cpUtils';
 
 export function getBrewPackageName(version: FuncVersion): string {
     return `${funcPackageName}@${getMajorVersion(version)}`;
+}
+
+export async function tryGetInstalledBrewPackageName(version: FuncVersion): Promise<string | undefined> {
+    const brewPackageName: string = getBrewPackageName(version);
+    if (await isBrewPackageInstalled(brewPackageName)) {
+        return brewPackageName;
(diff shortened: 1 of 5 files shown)


src/funcCoreTools/getBrewPackageName.ts
```diff
@@ -5,7 +5,37 @@
 
 import { funcPackageName } from '../constants';
 import { FuncVersion, getMajorVersion } from '../FuncVersion';
+import { cpUtils } from '../utils/cpUtils';
 
 export function getBrewPackageName(version: FuncVersion): string {
     return `${funcPackageName}@${getMajorVersion(version)}`;
 }
+
+export async function tryGetInstalledBrewPackageName(version: FuncVersion): Promise<string | undefined> {
```
nturinski: While `isBrewPackageInstalled` is generic-purpose, `tryGetInstalledBrewPackageName` is only specifically for Function Core Tools, right?  The naming is a little confusing to me.

EricJizbaMSFT: Huh yeah I guess `isBrewPackageInstalled` is generic, but it's not exported from this file and it's only really used for Function Core Tools

PR #1840, opened 2020-01-24
Update vm images

EricJizbaMSFT:
Got an email saying the images we're using will be removed in March


EricJizbaMSFT:
Assuming this is approved, I'll do the same in our other repos. Only real question is if we want to keep hard-coding a version or always use latest. I vote for latest, but not a strong opinion

PR #1860, opened 2020-02-03
Add `revealTreeItem` api

EricJizbaMSFT:


src/vscode-azurefunctions.api.d.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+export interface AzureFunctionsExtensionApi {
+    apiVersion: string;
+
+    revealTreeItem(resourceId: string): Promise<void>;
```
nturinski: Why does functions need an api?

EricJizbaMSFT: For the resource groups extension I started that we may or may not end up shipping. Either way I think 'reveal' is the most common way Azure extensions will interact with each other so I think this function is pretty harmless to add to an api

PR #1861, opened 2020-02-03
Use "func: extensions install" if extensions.csproj exists

EricJizbaMSFT:
When initializing for using for use with VS Code, previously we would add "func: extensions install" logic only if the user's func cli version didn't support bundles. However, there are still cases where people might not want to use bundles and could use "extensions.csproj" instead - now we will respect that file.

Related to https://github.com/microsoft/vscode-azurefunctions/issues/1698

### DIFF
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/ScriptInitVSCodeStep.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/ScriptInitVSCodeStep.ts
@@ -3,6 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as fse from 'fs-extra';
+import * as path from 'path';
 import * as semver from 'semver';
 import { TaskDefinition } from 'vscode';
 import { extInstallTaskName, func, funcWatchProblemMatcher, hostStartCommand } from '../../../constants';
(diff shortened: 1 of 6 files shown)


src/commands/initProjectForVSCode/InitVSCodeStep/ScriptInitVSCodeStep.ts
```diff
@@ -15,32 +17,36 @@ import { InitVSCodeStepBase } from './InitVSCodeStepBase';
...
             }
         ];
     }
 
     protected async executeCore(context: IProjectWizardContext): Promise<void> {
-        if (context.version === FuncVersion.v2) { // no need to check v1 or v3+
-            try {
+        try {
+            const extensionsCsprojPath: string = path.join(context.projectPath, 'extensions.csproj');
+            if (await fse.pathExists(extensionsCsprojPath)) {
```
nturinski: Should we alert or prompt the user telling them that csproj files aren't necessary anymore?  Not sure how much we want to get users off this path.

EricJizbaMSFT: There are still a few advanced scenarios where the csproj file is better than bundles, but yeah a prompt might make sense for the majority of users. I added `hasExtensionsCsproj` as a telemetry property which should help us decide if it's worth the work

Issue #1793, opened 2020-01-14
Strange interaction with "Bindings" nodes in Azure Function view...


weinand:
Interaction with "Bindings" nodes in Azure Function view is somewhat inconsistent and unexpected.

![2020-01-14_17-09-12](https://user-images.githubusercontent.com/1898161/72361066-255b9a80-36f1-11ea-983a-0f4dc4a596af.png)

- selecting the "res" and "req" leafs has no effect. This is unexpected. At least for the Azure subtree (green arrow) where a context menu "View properties" exists, I would expect that selecting the leafs shows the properties in the editor.
- in the "Local" subtree (red arrow), the leafs have no "View properties" action in the context menu. But there is an "Add Binding..." action on the "Bindings node.
- After adding a binding, the editor shows a json file and a new leafs appears under the "Bindings" node. But I could not find a way to make the json file show up again in the editor.

Issue #1862, opened 2020-02-03
Clicking leaf node should open right hand side


EricJizbaMSFT:
We're pretty inconsistent. For example "Deployments" has a button instead of just clicking the tree item

PR #1875, opened 2020-02-06
Remove binding tree items

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1793

The binding tree items were never that helpful and they just made the tree complicated/confusing/ugly. Main changes:
1. Got rid of binding nodes
1. Instead of right click menu for "View Properties" on function node, you just click the node (Related to https://github.com/microsoft/vscode-azurefunctions/issues/1862) and it opens either read-only function.json (if remote) or editable function.json (if local)
1. "Add Binding..." in local scenario was moved from "Bindings" node to Function node

### Remote
Screen Shot 2020-02-05 at 6 38 59 PM

### Local
Screen Shot 2020-02-05 at 6 38 10 PM
> NOTE: "Add Binding..." was on the "Bindings" node

### Remote
Screen Shot 2020-02-05 at 6 32 32 PM

### Local
Screen Shot 2020-02-05 at 6 32 20 PM


EricJizbaMSFT:
@fiveisprime can you take a look at this? Any worries users will be confused by the change in experience or just go for it?

fiveisprime:
I'm comfortable with this change. I don't know how helpful it was to have that in the first place so it should be fine.

I really like the click to open function.json, especially for local projects. That seems more helpful than having the bindings nodes.

:shipit:

Issue #1767, opened 2019-12-20
Automatically direct users to help page for 503 errors


EricJizbaMSFT:
We get a lot of people reporting 503 errors, which actually has a decent docs page here: https://docs.microsoft.com/azure/app-service/troubleshoot-http-502-http-503

Perhaps we can just automatically direct people there instead of having them report an issue on us

PR #1876, opened 2020-02-06
Update shared packages

EricJizbaMSFT:
Depends on https://github.com/microsoft/vscode-azuretools/pull/652

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1767
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1862


package-lock.json
```diff
@@ -9017,9 +9017,9 @@
             }
         },
         "vscode-azureextensionui": {
-            "version": "0.29.6",
-            "resolved": "https://registry.npmjs.org/vscode-azureextensionui/-/vscode-azureextensionui-0.29.6.tgz",
-            "integrity": "sha512-GKRNqdjHwu5qfxGD16Y64Z8t+nOQk4j9FKkipqgZGrnGfqO0v42PinbLWFT7WdD85ebPmmlpyNa4F/nDLV7oMQ==",
+            "version": "0.29.7",
```
nturinski: Odd that the vscode-azureappservice package didn't update.

package.json
```diff
@@ -640,11 +636,6 @@
                     "when": "view == azFuncTree && viewItem == deployment/github",
                     "group": "1@3"
                 },
-                {
```
nturinski: Wait, we're getting rid of the icon?  I thought we'd have both the double-click and the button.

package-lock.json
EricJizbaMSFT: I can't update it because I'm still waiting for https://github.com/microsoft/vscode-azuretools/pull/652 to be approved 😛

package.json
EricJizbaMSFT: https://github.com/microsoft/vscode-azurefunctions/issues/1862 is about becoming more consistent with VS Code patterns and I can't find any pattern that uses clicking a leaf node and a button for the same functionality

nturinski: I think people already have a hard enough time finding the deployment logs.  I think we need a way to make it obvious that a node can be double-clicked if we do this.

EricJizbaMSFT: First of all it's a single click, not a double click. Secondly, the argument made in https://github.com/microsoft/vscode-azurefunctions/issues/1793 was that people _did_ expect a single click to do something and it wasn't. That's how it works in the file view, git view, etc.

I agree we probably want to make deployment logs more prominent somehow, but that was not the goal of this change. The goal was to have more consistent behavior for leaf nodes. We can continue to discuss the other piece here: https://github.com/microsoft/vscode-azureappservice/issues/1251

Issue #1871, opened 2020-02-05
Flip Language/OS ordering of "Create Function App in Azure"


nigarfin:
We've found people are more interested in language than OS when it comes to creating Azure Functions. The current flow asks for OS first, and that limits the language options for users in the following step. Can we reverse the ordering so language is asked first then the user is prompted with OS?

For example, a Python dev selects Windows OS. In the language prompt, Python is not present.

PR #1890, opened 2020-02-12
Prompt for runtime before OS

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1871, which describes the reasoning pretty well


src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
```diff
@@ -48,10 +54,12 @@ export class FunctionAppRuntimeStep extends AzureWizardPromptStep<IFunctionAppWi
                 picks.unshift({ label: 'Node.js 12.x', data: 'node|12' });
             }
 
-            if (context.newSiteOS === WebsiteOS.linux) {
+            if (context.newSiteOS !== WebsiteOS.windows) {
```
nturinski: How would you get here with an OS already selected? Is there a programmatic setting?

EricJizbaMSFT: Basic create - for example, we default to linux for python projects, but still prompt them to see if they want 3.7.x or 3.6.x

Might be other cases, but that's the main one.

PR #1891, opened 2020-02-13
Fix tests on linux build machine

EricJizbaMSFT:
Same as https://github.com/microsoft/vscode-azureappservice/pull/1391


.azure-pipelines/linux/xvfb.init
```diff
@@ -22,7 +22,7 @@
 [ "${NETWORKING}" = "no" ] && exit 0
 
 PROG="/usr/bin/Xvfb"
-PROG_OPTIONS=":10 -ac"
+PROG_OPTIONS=":10 -ac -screen 0 1024x768x24"
```
nturinski: 🤷‍♀️

PR #1894, opened 2020-02-14
Support java on linux

### LINKED ISSUE #1601: Java on Linux
Add support for Java on Linux (default should still be Windows)

EricJizbaMSFT:
Relies on https://github.com/microsoft/vscode-azuretools/pull/659, where I refactored how stopping the app/syncing triggers works.

Fixes #1601

### DIFF
--- a/src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
+++ b/src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
@@ -18,8 +18,10 @@ export class FunctionAppRuntimeStep extends AzureWizardPromptStep<IFunctionAppWi
 
         if (/^python/i.test(context.newSiteRuntime)) {
             context.newSiteOS = WebsiteOS.linux;
-        } else if (/^(powershell|java)/i.test(context.newSiteRuntime)) {
+        } else if (/^powershell/i.test(context.newSiteRuntime)) {
             context.newSiteOS = WebsiteOS.windows;
+        } else if (/^java/i.test(context.newSiteRuntime) && context.version === FuncVersion.v2) {
+            context.newSiteOS = WebsiteOS.windows; // v1 doesn't support java at all. v2 is windows-only. v3+ supports both OS's.
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
```diff
@@ -50,7 +52,7 @@ export class FunctionAppRuntimeStep extends AzureWizardPromptStep<IFunctionAppWi
                 picks.unshift({ label: 'Node.js 8.x', data: 'node|8' });
             }
             picks.unshift({ label: 'Node.js 10.x', data: 'node|10' });
-            if (context.version === FuncVersion.v3) {
+            if (context.version !== FuncVersion.v2) {
                 picks.unshift({ label: 'Node.js 12.x', data: 'node|12' });
```
nturinski: Node.js 12 is available on v1?

src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
```diff
@@ -60,9 +62,13 @@ export class FunctionAppRuntimeStep extends AzureWizardPromptStep<IFunctionAppWi
             }
 
             if (context.newSiteOS !== WebsiteOS.linux) {
-                picks.push({ label: 'Java', data: 'java' });
                 picks.push({ label: 'PowerShell', data: 'powershell' });
             }
+
+            // v1 doesn't support java at all. v2 is windows-only. v3+ supports both OS's.
+            if (context.version !== FuncVersion.v2 || context.newSiteOS !== WebsiteOS.linux) {
```
nturinski: I assume that v1 doesn't support Linux at all, but I think people might find this comment/conditional confusing.

src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
EricJizbaMSFT: Nope. We're in an `else` case of `if (context.version === FuncVersion.v1) {` from above. I did this because if/when there's a v4, it's better to assume it will support Node 12

src/commands/createFunctionApp/FunctionAppRuntimeStep.ts
EricJizbaMSFT: I mean I also think this matrix is confusing... not sure how you want me to make it less confusing

nturinski: I probably should have looked it in all of its context rather than just the changes.  Since I assume this is also in the `else` case

EricJizbaMSFT: Yeah it is  ¯\\\_(ツ)\_/¯

Issue #1854, opened 2020-01-31
Support for Python 3.8


kulkarnisonia16:
Enable all scenarios for Python 3.8 similar to 3.7

EricJizbaMSFT:
@kulkarnisonia16 can you give us a timeline for this?

EricJizbaMSFT:
Aka when will this be available in the platform - and will it be v3-only or also v2?

kulkarnisonia16:
V3 only, Az cli/core-tools/Portal should all be able to light this by early March.

fiveisprime:
This should be landing in the core tools and platform on the 18th for us to test.

Issue #1752, opened 2019-12-13
Use availableStacks api to list runtimes


EricJizbaMSFT:
Need to investigate if the API works yet for functions. It didn't work back when we added support for python 3.6/3.7 and node 8/10, but I think they said that would happen soon

EricJizbaMSFT:
The API exists now, but doesn't work for our situation. It only lists the runtimes for the latest func version, but we need the runtimes _per_ version. For example, v3 supports node 10/12 and v2 supports node 8/10, but the API always lists node 10/12. I bugged the platform team about this, but it seemed unlikely to change soon - would have to dig out the email to find the exact response

divyagandhii:
> 
> 
> The API exists now, but doesn't work for our situation. It only lists the runtimes for the latest func version, but we need the runtimes _per_ version. For example, v3 supports node 10/12 and v2 supports node 8/10, but the API always lists node 10/12. I bugged the platform team about this, but it seemed unlikely to change soon - would have to dig out the email to find the exact response

Hey @EricJizbaMSFT , can you share why you need function versions per runtime?

divyagandhii:
cc @pragnagopa

EricJizbaMSFT:
@divyagandhi01 because the function app we create in Azure is based on the local version of the user's project. If they have a v1 project, we list Node 6. If they have a v2 project, we list Node 8 and 10. If they have a v3 project, we list Node 10 and 12. The matrix is listed [here](https://docs.microsoft.com/azure/azure-functions/functions-versions#languages) and we support the whole matrix

pragnagopa:
cc @gzuber @Francisco-Gamino 
We plan to include supported functions extension version for each stack.

PR #1904, opened 2020-02-19
Make runtime logic consistent when creating function app in Azure

EricJizbaMSFT:
1. Per https://github.com/Azure/azure-functions-host/issues/5628, dedicated and premium plans no longer need to specify `linuxFxVersion` as a Docker image - we can just use `runtime|version` like we already do for consumption
2. In the quick pick, I made sure each pick has the version listed and I got rid of the `.x` (I tried to find a conclusive pattern in other docs related `.x`, but couldn't find one. I just think it's ugly).

||v1|v2|v3|
|---|---|---|---|
|Before|v1before|v2before|v3before|
|After|v1after|v2after|v3after|

3. Added Python 3.8 - related to https://github.com/microsoft/vscode-azurefunctions/issues/1854


EricJizbaMSFT:
> Anywhere to get the list of Runtimes dynamically?

Not possible at the moment. Tracked with https://github.com/microsoft/vscode-azurefunctions/issues/1752, see there for more info

PR #1906, opened 2020-02-20
Add local project support for python 3.8

EricJizbaMSFT:
I still need to update one line in this PR with the exact version of the func cli that adds support for Python 3.8 (which was supposed to happen yesterday), but otherwise this is good to go.

In combination with https://github.com/microsoft/vscode-azurefunctions/pull/1904, fixes https://github.com/microsoft/vscode-azurefunctions/issues/1854

### DIFF
--- a/src/commands/createNewProject/pythonSteps/pythonVersion.ts
+++ b/src/commands/createNewProject/pythonSteps/pythonVersion.ts
@@ -6,6 +6,7 @@
 import * as semver from 'semver';
 import { ext } from '../../../extensionVariables';
 import { getLocalFuncCoreToolsVersion } from '../../../funcCoreTools/getLocalFuncCoreToolsVersion';
+import { FuncVersion, getMajorVersion } from '../../../FuncVersion';
 import { localize } from "../../../localize";
 import { cpUtils } from "../../../utils/cpUtils";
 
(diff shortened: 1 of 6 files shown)


src/commands/createNewProject/pythonSteps/pythonVersion.ts
```diff
@@ -23,22 +24,43 @@ export async function getPythonVersion(pyAlias: string): Promise<string> {
...
-    const minFuncVersion: string = '2.7.1846';
-    try {
-        const currentVersion: string | null = await getLocalFuncCoreToolsVersion();
-        if (currentVersion && semver.lt(currentVersion, minFuncVersion)) {
-            ext.outputChannel.appendLine(localize('outOfDateWarning', 'WARNING: Some Python versions are not supported with version "{0}" of the func CLI. Update to at least "{1}".', currentVersion, minFuncVersion));
-            return versions;
+    const result: string[] = [];
+
+    const versionInfo: [string, string][] = [
+        ['2.0.0', '3.6'],
```
nturinski: Why is this a 2d array now?

EricJizbaMSFT: The first entry is minimum func cli version and the second entry is the python version it supports. It's more clear a few lines down in the for loop. The reason I need a 2d array now is just because it gets more and more complicated when I have to add more python versions

PR #1953, opened 2020-03-03
Add "View Properties" to right click menu of function

EricJizbaMSFT:
I removed it from the right click menu in https://github.com/microsoft/vscode-azurefunctions/pull/1875 in favor of the left click, but I think there's value in having it in both places - same thing we did for "View Deployment Logs" in https://github.com/microsoft/vscode-azurefunctions/pull/1935.

### DIFF
--- a/package.json
+++ b/package.json
@@ -552,6 +552,11 @@
                     "group": "3@2"
                 },
                 {
+                    "command": "azureFunctions.viewProperties",
+                    "when": "view == azFuncTree && viewItem =~ /Remote;.*;Function;/i",
+                    "group": "4@1"
+                },
+                {
                     "command": "azureFunctions.appSettings.add",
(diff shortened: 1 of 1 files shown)


EricJizbaMSFT:
I would flip your logic around and say the other commands deserve their own groups. I think "Copy Function Url" deserves it's own group since that's one of the most useful things we provide. I also think start/stop streaming logs naturally goes together

nturinski:
Yeah, it's just the more groups there are, the more cluttered it looks, and the harder it is to discover the more widely used commands.  I figured it could go with like, refresh or open in portal.

EricJizbaMSFT:
This is at the function level, not the function _app_ level. There's not that many commands:
Screen Shot 2020-02-05 at 6 38 59 PM

PR #1973, opened 2020-03-13
Easier debugging in a webpacked world 🕸📦🌎

EricJizbaMSFT:
This should fix the following error that I'm sure we've all seen when debugging:

> Activating extension 'ms-azuretools.vscode-azurefunctions' failed: Cannot find module './dist/extension.bundle' 

The problem is environment variables don't carry over when you open a new folder. My solution will instead modify "main.js" before every webpack. That means you'll have to be mindful not to commit the changes to "main.js" if you test webpack locally, but worst case you'll only break the F5 experience of other devs on your team (aka not the vsix we ship). I almost never debug with webpack and would gladly deal with that minor annoyance over the existing experience.

cc @StephenWeatherford @philliphoff @bwateratmsft since this could help y'all on ARM/Docker, too

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -17,7 +17,6 @@
             ],
             "preLaunchTask": "npm: compile",
             "env": {
-                "AZCODE_FUNCTIONS_IGNORE_BUNDLE": "1",
                 "DEBUGTELEMETRY": "v",
                 "NODE_DEBUG": ""
             }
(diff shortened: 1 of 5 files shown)


nturinski:
This goes down as the best emoji title in our repo.

EricJizbaMSFT:
Discussed offline

PR #2003, opened 2020-04-01
Allow skipping staging template source in CI ⏭

EricJizbaMSFT:
I like having test failures to kind of "notify" me when the staging templates have changes that affect us, but I don't want to update the tests until the staging templates move to production, which can take a little while. Hopefully adding this environment variable will be a nice middle ground where I can actually get successful builds in the meantime


EricJizbaMSFT:
I'll set the variable when I want to :) I'm not sure - could just be a few one-off builds. The beauty of this is that I can dial it up/down in Azure Pipelines without having to submit PRs to master

PR #2010, opened 2020-04-04
Ask if venv should be created when initiating an existing Python project

NVolcz:
Users are currently asked if a virtual environment (venv) should be created when creating a new Python project, this also triggers creation of tasks for manage builds within the venv. However, when importing/initiating an existing project (e.g. when git cloned), the user is not prompted in the same manner if a venv is to be created. This makes it harder to import an existing project.

I have copied the logic from the create flow to the init flow to provide the user with the same options.


src/commands/initProjectForVSCode/pythonSteps/addPythonInitVSCodeSteps.ts
```diff
@@ -34,6 +38,9 @@ export async function addPythonInitVSCodeSteps(
         } else {
             promptSteps.push(new PythonVenvListStep(venvs));
         }
+    } else if (createPythonVenv) {
```
EricJizbaMSFT, line 41: Looks like there's a few test failures. You can ignore the tests titled "Template Count - Staging Python", but the other two are relevant to this PR.

- Create New Python Project skip venv

Looks like this one actually found a bug. If you do "Create New Project" and select "skip venv", you'll get the same prompt twice. Since "init for vs code" always happens for "create new project", I think you can remove these steps from [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts#L23)

- Create New Python Project skip venv

Just need to update the test to reflect the behavior change. Similar to [this test](https://github.com/microsoft/vscode-azurefunctions/blob/master/test/project/createNewPythonProject.test.ts#L16), you can specify `inputs: [/skip/i]` [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/test/project/initProjectForVSCode.test.ts#L57)

NVolcz: Thanks for the clarifications about the tests. Was not sure how to interpret the result when master were failing as well. Will try to have a look at the bug this weekend or next week.

NVolcz: I believe my latest push fixed the open issues.

src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts
```diff
@@ -3,24 +3,21 @@
...
     context: IProjectWizardContext & IPythonVenvWizardContext,
-    promptSteps: AzureWizardPromptStep<IProjectWizardContext>[],
     executeSteps: AzureWizardExecuteStep<IProjectWizardContext>[]): void {
 
     const createPythonVenv: boolean = !!getWorkspaceSetting<boolean>('createPythonVenv', context.workspacePath);
     context.telemetry.properties.createPythonVenv = String(createPythonVenv);
 
     if (createPythonVenv) {
-        promptSteps.push(new PythonAliasListStep());
         executeSteps.push(new PythonVenvCreateStep());
```
EricJizbaMSFT, line 21: We don't need to add this step either since it's added in `addPythonInitVSCodeSteps.ts` now. In fact, there's only one line from this file that's actually needed, so I would delete this file and just add this line directly to `NewProjectLanguageStep.ts`:

```
executeSteps.push(new PythonProjectCreateStep());
```

src/commands/initProjectForVSCode/pythonSteps/addPythonInitVSCodeSteps.ts
```diff
@@ -16,6 +19,7 @@ export async function addPythonInitVSCodeSteps(
     promptSteps: AzureWizardPromptStep<IProjectWizardContext>[],
     executeSteps: AzureWizardExecuteStep<IProjectWizardContext>[]): Promise<void> {
 
+    const createPythonVenv: boolean = !!getWorkspaceSetting<boolean>('createPythonVenv', context.workspacePath);
     const venvs: string[] = [];
```
EricJizbaMSFT, line 23: Please copy this line over from the other file, so that we can track how often people create a virtual environment
```
context.telemetry.properties.createPythonVenv = String(createPythonVenv);
```

src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts
NVolcz: How about this line? Isn't it needed for the telemetry?
```context.telemetry.properties.createPythonVenv = String(createPythonVenv);```

Comment (NVolcz) [reply src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts:21]: How about this line? Isn't it needed for the telemetry?
context.telemetry.properties.createPythonVenv = String(createPythonVenv);

Comment (EricJizbaMSFT) [reply src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts:21]: Not if you move it to the other file like my other comment suggests

Comment (NVolcz) [reply src/commands/createNewProject/pythonSteps/addPythonCreateProjectSteps.ts]: Sorry. Missed your other comment. I have now moved the code from `addPythonInitVSCodeSteps.ts` into `NewProjectLanguageStep.ts` as suggested.

File: src/commands/createNewProject/NewProjectLanguageStep.ts [hunk 4]
```diff
@@ -84,7 +85,10 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
                 executeSteps.push(await DotnetProjectCreateStep.createStep(context));
                 break;
             case ProjectLanguage.Python:
-                addPythonCreateProjectSteps(context, promptSteps, executeSteps);
+                const createPythonVenv: boolean = !!getWorkspaceSetting<boolean>('createPythonVenv', context.workspacePath);
+                context.telemetry.properties.createPythonVenv = String(createPythonVenv);
```
EricJizbaMSFT, line 89: Haha so close 😂I meant put the telemetry line in `addPythonInitVSCodeSteps.ts` where we're already calling `getWorkspaceSetting`

NVolcz: Haha! That makes more sense ;-). Let's give it another try!

PR #2094, opened 2020-05-13
Clean up output when creating function app/slot

### LINKED ISSUE #2084: There is no "Created new function app..." output log when creating it during deployment
**OS:** Win10
**Build Version:** [20200513.5](https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=19233&view=results)

**Repro Steps:**
1. Open a function project in VS Code.
2. Click 'Deploy to Function App...' button->create a new function app.
3. Check the output after deployment.

**Expect:**
There is "Created new function app..." output log.

**Actual:**
There is no "Created new function app..." output log.
image

**More info:**
This issue does not reproduce when only create a Function App.

EricJizbaMSFT:
We weren't displaying to the output when it successfully creating during a deploy. Fixes #2084

Also I moved the `enableFileLogging` code into the wizard so that the progress notification would still display while it's executing.

### DIFF
--- a/src/commands/createFunctionApp/FunctionAppCreateStep.ts
+++ b/src/commands/createFunctionApp/FunctionAppCreateStep.ts
@@ -5,8 +5,8 @@
 
 import { WebSiteManagementClient, WebSiteManagementModels as SiteModels } from 'azure-arm-website';
 import { Progress } from 'vscode';
-import { WebsiteOS } from 'vscode-azureappservice';
-import { AzureWizardExecuteStep, createAzureClient } from 'vscode-azureextensionui';
+import { SiteClient, WebsiteOS } from 'vscode-azureappservice';
+import { AzureWizardExecuteStep, createAzureClient, parseError } from 'vscode-azureextensionui';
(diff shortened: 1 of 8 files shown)


src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -43,6 +45,18 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             siteConfig: await this.getNewSiteConfig(context),
             reserved: context.newSiteOS === WebsiteOS.linux  // The secret property - must be set to true to make it a Linux plan. Confirmed by the team who owns this API.
         });
+
+        context.siteClient = new SiteClient(context.site, context);
+        if (!context.siteClient.isLinux) { // not supported on linux
+            try {
+                await enableFileLogging(context.siteClient);
```
nturinski, line 52: I think we talked about enabling this by default for slots as well.

EricJizbaMSFT: Could be - I'm gonna let future Eric deal with that

PR #2112, opened 2020-05-18
Add tool to update backup templates

EricJizbaMSFT:
Two main problems I'm trying to fix here:
1. Make the process of updating backup templates more automatic
1. Make the backup templates more accurately reflect the latest templates (For example, right now backup templates have no concept of bundles)

To do that, I created a tool. It's a bit weird TBH, but it works haha. You run the test suite with an environment variable set and it updates all backup templates from the latest template sources (instead of actually running tests). Since it's running _in_ the extension, I get access to all the existing code around templates and it was easy to make backup templates like _the exact same thing_ as latest templates.

The result of this tool is here: https://github.com/microsoft/vscode-azurefunctions/pull/2113

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -57,7 +57,8 @@
                 "DEBUGTELEMETRY": "v",
                 "NODE_DEBUG": "",
                 "ENABLE_LONG_RUNNING_TESTS": "",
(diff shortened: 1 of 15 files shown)


EricJizbaMSFT:
ping @nturinski

PR #2138, opened 2020-05-29
Fix breaking changes

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/725


src/commands/deployments/disconnectRepo.ts
```diff
@@ -3,13 +3,21 @@
...
     if (!node) {
         node = await ext.tree.showTreeItemPicker<DeploymentsTreeItem>(DeploymentsTreeItem.contextValueConnected, context);
     }
-    await node.disconnectRepo(context);
+
+    if (node.parent instanceof SlotTreeItemBase) {
+        await disconnectRepository(context, node.parent.client, node.parent.root);
+        await node.refresh();
+    } else {
+        throw Error(localize('actionNotSupported', 'Action not supported'));
```
alexweininger, line 21: Is this an appropriate error message?

src/commands/deployments/disconnectRepo.ts
```diff
@@ -3,13 +3,21 @@
...
 
 export async function disconnectRepo(context: IActionContext, node?: DeploymentsTreeItem): Promise<void> {
     if (!node) {
         node = await ext.tree.showTreeItemPicker<DeploymentsTreeItem>(DeploymentsTreeItem.contextValueConnected, context);
     }
-    await node.disconnectRepo(context);
+
+    if (node.parent instanceof SlotTreeItemBase) {
+        await disconnectRepository(context, node.parent.client, node.parent.root);
+        await node.refresh();
```
alexweininger, line 19: I didn't see any, but are there any checks I need to do before refreshing?

src/commands/deployments/disconnectRepo.ts
EricJizbaMSFT: Should this ever be possible? If not, make sure to add "Internal error" like [here](https://github.com/microsoft/vscode-azurefunctions/blob/master/src/utils/nonNull.ts#L24)

We have a telemetry query setup specifically for those errors. Since there's nothing the user could do, it's more important that we can find this in telemetry than for them to understand it

src/commands/deployments/disconnectRepo.ts
EricJizbaMSFT: Don't think so

Issue #2134, opened 2020-05-28
Project created several folders down is not automatically detected


SeaDude:
After creating a new Python project using the extension, the project is not listed under "Functions/MySubscription". The files were all created, but the project is not listed.

I'm using the latest VS Code for Linux. Please advise. Thank you

**Example**:
![noProject](https://user-images.githubusercontent.com/8636973/83155107-c140de00-a0b5-11ea-93bd-ecae94992858.gif)

EricJizbaMSFT:
Looks like a bug when you create a project several levels below your opened workspace. I'll look into fixing that, but in the meantime you have a few ways you can workaround it:

1. Open the exact folder you want to use in VS Code _before_ you create the project
1. If you want to keep your current folder open instead of the exact project folder, you can specify "azureFunctions.projectSubpath" in your VS Code settings (`.vscode/settings.json`). It would look something like this:

Settings:
<img width="461" alt="Screen Shot 2020-05-28 at 10 14 57 AM" src="https://user-images.githubusercontent.com/11282622/83172266-30c1c800-a0cc-11ea-8c9d-4f5686224c33.png">

Folder view:
<img width="223" alt="Screen Shot 2020-05-28 at 10 15 16 AM" src="https://user-images.githubusercontent.com/11282622/83172268-328b8b80-a0cc-11ea-8100-b7b6435000af.png">

Azure view:
<img width="223" alt="Screen Shot 2020-05-28 at 10 15 06 AM" src="https://user-images.githubusercontent.com/11282622/83172267-31f2f500-a0cc-11ea-9540-584928b9af90.png">

This will add a "Local Project" view in the Azure Functions tree. For it to show up under your subscription, you would need to deploy using the blue up arrow.

SeaDude:
#1 above worked fine. Thank you for acknowledging the bug (so I know I'm not insane!) and providing unblockers for me to continue dev. 
I appreciate that.

EricJizbaMSFT:
Glad that worked! Re-opening the bug for our own tracking purposes

PR #2182, opened 2020-06-25
Add "projectSubpath" setting if project is several levels down

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2134

### DIFF
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
@@ -7,7 +7,7 @@ import * as fse from 'fs-extra';
 import * as path from 'path';
 import { DebugConfiguration, TaskDefinition, WorkspaceFolder } from 'vscode';
 import { AzureWizardExecuteStep } from 'vscode-azureextensionui';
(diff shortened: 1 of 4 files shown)


src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
```diff
@@ -189,21 +189,29 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
...
+    private async writeSettingsJson(context: IProjectWizardContext, vscodePath: string, language: string, version: FuncVersion): Promise<void> {
         const settings: ISettingToAdd[] = this.settings.concat(
             { key: projectLanguageSetting, value: language },
             { key: funcVersionSetting, value: version },
             // We want the terminal to be open after F5, not the debug console (Since http triggers are printed in the terminal)
             { prefix: 'debug', key: 'internalConsoleOptions', value: 'neverOpen' }
         );
 
+        // Add "projectSubpath" setting if project is far enough down that we won't auto-detect it
+        if (path.posix.relative(context.projectPath, context.workspacePath).startsWith('../..')) {
```
nturinski, line 201: So since a folder that's 5 subfolders deep from the root technically starts '../...', would this code detect those cases too?  Is there any concern about performance?

EricJizbaMSFT: Discussed offline

PR #2192, opened 2020-06-29
Add AppInsights sampling excluded types to host.json

anthonychu:
Update host.json to exclude requests from sampling by default. Matches changes made to [Core Tools](https://github.com/Azure/azure-functions-core-tools/blob/5b0d1c25fa10adcc9a235579d159d7c3e4cf98ff/src/Azure.Functions.Cli/StaticResources/host.json).

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
@@ -79,7 +79,15 @@ local.settings.json`));
             return {};
         } else {
             const hostJson: IHostJsonV2 = {
-                version: '2.0'
+                version: '2.0',
+                logging: {
+                    applicationInsights: {
+                        samplingSettings: {
+                            isEnabled: true,
+                            excludedTypes: 'Request'
+                        }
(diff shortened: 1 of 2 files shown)


src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -79,7 +79,15 @@ local.settings.json`));
             return {};
         } else {
             const hostJson: IHostJsonV2 = {
-                version: '2.0'
+                version: '2.0',
```
EricJizbaMSFT, line 82: This feels like a pretty obscure setting to me. I assume there's a reason this needs to be defaulted in the tooling instead of in the platform? If so - can you add a link to the original discussion in a comment (a TypeScript comment in the file itself, not a PR comment on GitHub)

anthonychu: Found the [discussion](https://github.com/Azure/azure-functions-templates/issues/906). Reason was it's a breaking change if we default. Will add a comment.

Issue #1390, opened 2019-07-18
Add option to execute function with .http file


EricJizbaMSFT:
We already let users execute a timer function (https://github.com/microsoft/vscode-azurefunctions/issues/402). However, we could also let them execute any arbitrary function by pre-populating an `http` file and leveraging the REST client extension. The portal has a similar experience:
![screen shot 2018-05-31 at 9 31 25 am](https://user-images.githubusercontent.com/11282622/40794946-7282960c-64b5-11e8-91e9-29921503aa1f.png)

_as suggested by @fiveisprime offline_

PR #2195, opened 2020-06-29
Support "Execute Function" for all binding types

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1390

We originally discussed leveraging ".http" files, but that gets a bit dicey in terms of authentication - I don't want to put the key directly in the file as plain text and other methods (i.e. environment variables) were non-trivial. Instead, I followed the exact same pattern already established for timer triggers, just with a prompt for the request body. The default value comes from the "sample.dat" file included with each template. The biggest downside is the input box itself (since it can't display much content), but I think it's OK for the time being. People can always copy/paste the value into a file to edit more easily.

Here's a few examples with the default value from "sample.dat":

Http Trigger:
Screen Shot 2020-06-29 at 2 45 03 PM
Screen Shot 2020-06-29 at 2 46 02 PM

Blob trigger (the path to the blob):
Screen Shot 2020-06-29 at 2 44 38 PM

Cosmos DB Trigger (an array of documents):
Screen Shot 2020-06-29 at 2 44 46 PM

Event grid trigger (as mentioned - not the best for an input box):
Screen Shot 2020-06-29 at 2 44 57 PM

### DIFF
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -22,14 +22,16 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
     }
 
     const client: SiteClient | undefined = node instanceof RemoteFunctionTreeItem ? node.parent.parent.root.client : undefined;
+    const triggerBindingType: string | undefined = node.config.triggerBinding?.type;
+    context.telemetry.properties.triggerBindingType = triggerBindingType;
 
     let functionInput: string | {} = '';
     if (!node.config.isTimerTrigger) {
         const prompt: string = localize('enterRequestBody', 'Enter request body');
         let value: string | undefined;
-        if (node.config.triggerBinding?.type) {
+        if (triggerBindingType) {
             const version: FuncVersion = await node.parent.parent.getVersion();
-            value = await ext.templateProvider.tryGetSampleData(context, version, node.config.triggerBinding?.type);
+            value = await ext.templateProvider.tryGetSampleData(context, version, triggerBindingType);
             if (value) {
                 // Clean up the whitespace to make it more friendly for a one-line input box
                 value = value.replace(/[\r\n\t]/g, ' ');
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -55,6 +55,7 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         method = 'GET';
         body = functionInput;
     } else {
+        // https://docs.microsoft.com/azure/azure-functions/functions-manually-run-non-http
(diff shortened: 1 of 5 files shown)


src/commands/executeFunction.ts
```diff
@@ -3,35 +3,75 @@
...
+    if (!node.config.isTimerTrigger) {
+        const prompt: string = localize('enterRequestBody', 'Enter request body');
+        let value: string | undefined;
+        if (triggerBindingType) {
+            const version: FuncVersion = await node.parent.parent.getVersion();
+            value = await ext.templateProvider.tryGetSampleData(context, version, triggerBindingType);
+            if (value) {
+                // Clean up the whitespace to make it more friendly for a one-line input box
+                value = value.replace(/[\r\n\t]/g, ' ');
+                value = value.replace(/ +/g, ' ');
```
nturinski, line 38: Does this turn multiple spaces to one?

src/commands/executeFunction.ts
```diff
@@ -3,35 +3,75 @@
...
+        body = functionInput;
+    } else {
+        url = `${node.parent.parent.hostUrl}/admin/functions/${node.name}`;
+        method = 'POST';
+        body = { input: functionInput };
+    }
+
+    let response: string | undefined;
     await node.runWithTemporaryDescription(localize('executing', 'Executing...'), async () => {
         // https://docs.microsoft.com/azure/azure-functions/functions-manually-run-non-http
```
nturinski, line 65: Seems like it makes sense to move this where you set url for http triggers now.

src/commands/executeFunction.ts
```diff
@@ -42,5 +82,6 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         }
     });
 
-    window.showInformationMessage(localize('executed', 'Executed function "{0}"', name));
+    const message: string = response ? localize('executedWithResponse', 'Executed function "{0}". Response: "{1}"', node.name, response) : localize('executed', 'Executed function "{0}"', node.name);
```
nturinski, line 85: Do the timer triggers have responses?  I noticed that you didn't output it before.

src/commands/executeFunction.ts
EricJizbaMSFT: Yep - "sample.dat" file uses multiple spaces for tabs, which looks weird in the input box

src/commands/executeFunction.ts
EricJizbaMSFT: They don't for the default template. Technically I think users can change the code and add a response for any function type

Issue #1465, opened 2019-08-26
WEBSITE_RUN_FROM_PACKAGE recommendation


mhoeger:
A user reported an issue in https://github.com/Azure/Azure-Functions/issues/1306 where the root cause was that the published function app was not using WEBSITE_RUN_FROM_PACKAGE = 1.

Although VS Code defaults to this setting, if the user creates the function app from the Portal and then deploys through VS Code, the setting is not added. Is it possible to log some sort of warning or just informational text that recommends that they set WEBSITE_RUN_FROM_PACKAGE to improve performance and predictability if it's not detected in the deployed site, especially with Node.js function apps? 

cc: @ColbyTresness

ColbyTresness:
Potentially. @ehamai can you comment on the feasibility of the above? And potentially help us with the data ask?

EricJizbaMSFT:
I still think it's super annoying that `WEBSITE_RUN_FROM_PACKAGE` isn't the default everywhere since it's so clearly better, but as long as we agree it's very low risk to modify `WEBSITE_RUN_FROM_PACKAGE` to `1` I think I'm coming around to this

Do we prompt the user to confirm this change? I hope not - since I think it'll just confuse people. But if we don't prompt, that means all users are forced to have `WEBSITE_RUN_FROM_PACKAGE` if they want to deploy from VS Code. Are there valid use cases where people don't want that setting and do we need to provide a workaround so they can still deploy from VS Code?

EricJizbaMSFT:
Sorry - I forgot this issue was just about a warning/recommendation. I thought it was about changing `WEBSITE_RUN_FROM_PACKAGE` automatically. What exactly do we want?

mhoeger:
@EricJizbaMSFT - I think there's a separate thread going on with Paul, but for Windows apps, my preference is adding `WEBSITE_RUN_FROM_PACKAGE` to `1` automatically if it's not already there (note this has to happen before doing zip deployment). The perf gains are pretty massive (ex: I have a function deployed that only references the durable-functions package and the cold start is almost a minute which is just nuts). I've also seen apps have issues caused by our file change watcher that we enable if editing is enabled.

`WEBSITE_RUN_FROM_PACKAGE` does disallow editing but if people really want to edit it deployed code anyways, they can delete WEBSITE_RUN_FROM_PACKAGE and edit without issues.

PR #2207, opened 2020-07-02
Add WEBSITE_RUN_FROM_PACKAGE to windows apps by default

EricJizbaMSFT:
This setting has big performance benefits per the platform team, but the portal doesn't set it by default (by design). I think a prompt about this would be confusing, so I just do it automatically and log a message in the output channel. If a user has the app setting already (regardless of value) - we won't change it.

Also I moved `verifyAppSettings` after the deploy confirmation prompt because it felt weird to update settings if they're just gonna cancel anyways.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1465


src/tree/SlotTreeItemBase.ts
```diff
@@ -212,7 +214,7 @@ export abstract class SlotTreeItemBase extends AzureParentTreeItem<ISiteTreeRoot
 
     public async isReadOnly(): Promise<boolean> {
         const appSettings: WebSiteManagementModels.StringDictionary = await this.root.client.listApplicationSettings();
-        return !!appSettings.properties && !!(appSettings.properties.WEBSITE_RUN_FROM_PACKAGE || appSettings.properties.WEBSITE_RUN_FROM_ZIP);
+        return [runFromPackageKey, 'WEBSITE_RUN_FROM_ZIP'].some(key => appSettings.properties && envUtils.isEnvironmentVariableSet(appSettings.properties[key]));
```
wwlorey, line 217: This line kind of scared me at first but it makes sense

PR #2277, opened 2020-08-05
Update a few nightly tests based on new azure sdks

EricJizbaMSFT:


test/nightly/createProjectAndDeploy.test.ts
```diff
@@ -118,7 +118,7 @@ async function validateFunctionUrl(appName: string, functionName: string, routeP
     assert.ok(functionUrl.includes(routePrefix), 'Function url did not include routePrefix.');
 
     const client: ServiceClient = createGenericClient();
-    const response: HttpOperationResponse = await client.sendRequest({ method: 'GET', url: functionUrl, body: { name: "World" } });
+    const response: HttpOperationResponse = await client.sendRequest({ method: 'POST', url: functionUrl, body: { name: "World" } });
```
nturinski, line 121: This is unrelated to the SDK change, right?

EricJizbaMSFT: It is related. 'GET' requests aren't supposed to have a body according to some official http spec, and the new sdk actually enforces that. So I had to switch to 'POST' instead

PR #2290, opened 2020-08-10
Clean up logic for getting feeds

EricJizbaMSFT:
There was a problem with the staging bundle feed causing tests to fail. Turns out the functions team will fix it on their end, but I did some investigating in the mean time and came up with a little bit of code cleanup on our side, mainly consolidating around feedUtils for all feeds.

### DIFF
--- a/src/templates/script/ScriptBundleTemplateProvider.ts
+++ b/src/templates/script/ScriptBundleTemplateProvider.ts
@@ -3,7 +3,6 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { HttpOperationResponse } from '@azure/ms-rest-js';
 import * as fse from 'fs-extra';
 import * as path from 'path';
 import { IActionContext, parseError } from 'vscode-azureextensionui';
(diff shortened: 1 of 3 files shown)


src/utils/requestUtils.ts
```diff
@@ -26,7 +26,7 @@ export namespace requestUtils {
 
         try {
             const client: ServiceClient = createGenericClient();
-            return await client.sendRequest(options);
+            return await client.sendRequest(request);
```
nturinski, line 29: 👀

src/templates/script/ScriptBundleTemplateProvider.ts
```diff
@@ -41,9 +38,7 @@ export class ScriptBundleTemplateProvider extends ScriptTemplateProvider {
         const resourcesUrl: string = release.resources.replace('{locale}', language);
 
         const urls: string[] = [release.bindings, resourcesUrl, release.functions];
-        const responses: HttpOperationResponse[] = await Promise.all(urls.map(url => requestUtils.sendRequestWithTimeout({ method: 'GET', url })));
-        // NOTE: r.parsedBody doesn't work because these feeds sometimes return with a BOM char or incorrect content-type
-        [this._rawBindings, this._rawResources, this._rawTemplates] = <[object, object, object[]]>responses.map(r => parseJson(nonNullProp(r, 'bodyAsText')));
+        [this._rawBindings, this._rawResources, this._rawTemplates] = <[object, object, object[]]>await Promise.all(urls.map(url => feedUtils.getJsonFeed(url)));
```
nturinski, line 41: Do you not have to specific the generic object T when you call `getJsonFeed`? Like `getJsonFeed<[object, object, object[]]>`?

EricJizbaMSFT: Nope, not necessary. If anything, It would actually be `getJsonFeed<object>` (remember, it's inside a `map` function`). Typescript is smart enough to infer stuff like that.

Issue #2307, opened 2020-08-18
Deployment slots for Linux Consumption


dariagrigoriu:
Enable deployment slots for Linux Consumption

EricJizbaMSFT:
@dariagrigoriu is this fully supported? I seem to be getting a ServiceUnavailable error consistently when I attempt to sync triggers on a slot. It works perfectly if I deploy the same thing to production. We're using this ARM api:

subscriptions/{subscriptionId}/resourceGroups/{resourceGroupName}/providers/Microsoft.Web/sites/{name}/slots/{slot}/syncfunctiontriggers

PR #2315, opened 2020-08-19
Allow slots on Linux Consumption

EricJizbaMSFT:
I seem to be getting more "Service Unavailable" errors with slots, but I did get it to work at least once 🤷‍♂️

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2307


EricJizbaMSFT:
Seems to be working for me now

Issue #1866, opened 2020-02-05
Enable support for PowerShell 7 functions


eamonoreilly:
We need to create a setting FUNCTIONS_WORKER_RUNTIME_VERSION in the local.settings.json file with a value of ~7 to indicate to the functions host to run PowerShell 7 during function app creation.

EricJizbaMSFT:
Can you give us more info on this? All the docs I see still say PowerShell 6. Also how does this translate when deploying to Azure? Can users target 6 or 7?

eamonoreilly:
Hi Eric,
PowerShell 7 support is rolling out but won't be available until some time in March. The customer can choose the language version of PowerShell (6 or 7) during app creation in Azure although we will default to 7. 
When deploying to Azure, we should check the site config of the function app and look for a property under siteProperties called "PowerShellVersion". If this value exists, we need to make sure FUNCTIONS_WORKER_RUNTIME_VERSION in the local settings matches this version. If this property does not exist (older function apps), then we need to make sure FUNCTIONS_WORKER_RUNTIME_VERSION is set to 6 or else error out since the worker will be set to 6 in the cloud.

I don't think we will need for this work to be out until mid / late March. Please let me know if there is additional information I can provide.
Thanks,
Eamon

sebastianzolg:
Is there any timeline for this?

EricJizbaMSFT:
I investigated this a little while ago and unfortunately the sdk we're using (azure-arm-website) doesn't recognize the "PowerShellVersion" property on siteConfig and was filtering it out. We're finally unblocked on using the new sdk (@azure/arm-appservice - see https://github.com/microsoft/vscode-azuretools/issues/510), so I decided to wait until that's done (should be this milestone) instead of adding a temporary workaround

EricJizbaMSFT:
New sdk doesn't seem to allow powerShellVersion either. Filed an issue here: https://github.com/Azure/azure-sdk-for-js/issues/10552

PR #2317, opened 2020-08-19
Allow PowerShell 7 when creating function app in Azure

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/1866

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -5,6 +5,7 @@
 
 'use strict';
 
+import { WebSiteManagementMappers } from '@azure/arm-appservice';
 import * as vscode from 'vscode';
 import { registerAppServiceExtensionVariables } from 'vscode-azureappservice';
 import { AzExtTreeDataProvider, AzureUserInput, callWithTelemetryAndErrorHandling, createApiProvider, createAzExtOutputChannel, IActionContext, registerEvent, registerUIExtensionVariables } from 'vscode-azureextensionui';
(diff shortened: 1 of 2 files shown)


src/extension.ts
```diff
@@ -79,6 +80,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('java', javaDebugProvider));
         context.subscriptions.push(vscode.debug.registerDebugConfigurationProvider('PowerShell', powershellDebugProvider));
         context.subscriptions.push(vscode.workspace.registerTaskProvider(func, new FuncTaskProvider(nodeDebugProvider, pythonDebugProvider, javaDebugProvider, powershellDebugProvider)));
+
+        // Temporary workaround so that "powerShellVersion" is an allowed property on SiteConfig
+        // https://github.com/Azure/azure-sdk-for-js/issues/10552
+        // tslint:disable-next-line: no-non-null-assertion
+        WebSiteManagementMappers.SiteConfig.type.modelProperties!.powerShellVersion = { serializedName: 'powerShellVersion', type: { name: 'String' } };
```
wwlorey, line 87: I was trying to find documentation for `WebSiteManagementMappers` to understand it but was struggling to find some. So adding `powerShellVersion` here makes that property available without throwing build errors?

EricJizbaMSFT: At the end of the day it's just a temporary workaround that should go away once they release the next version of "arm-appservice". But I'll try to explain how mappers work with a mock example (I didn't find docs - I just went digging on the sdk). Let's say this is the mapper:
```
export const AddressMapper: msRest.CompositeMapper = {
  serializedName: "Address",
  type: {
    name: "Composite",
    className: "Addresss",
    modelProperties: {
      street: {
        serializedName: "street",
        type: {
          name: "String"
        }
      },
      zipCode: {
        serializedName: "zipCode",
        type: {
          name: "String"
        }
      }
    }
};
```

And I make an api call like this:
```
const address: Address = { street: "105th Ave", zipCode: "55555", badProperty: "oops" };
await client.addAddressToContacts(address);
```

The sdk uses the mapper inside "addAddressToContacts" to convert the object into a string request body. The mapper is essentially a schema used to do some extra validation and cleanup (aka removing "badProperty") before doing a JSON.stringify on the object and sending the request.

My problem is that the sdk thinks "powerShellVersion" is a bad property, so it removes it from the object (even though I _did_ set it) before sending the request. This has nothing to do with build errors. It has to do with the actual functionality of sending the request.

PR #2351, opened 2020-09-02
Convert all master branches to main in repos we contribute to

nturinski:


test/createFunction/createFunction.CSharp.test.ts
```diff
@@ -137,7 +137,7 @@ function addSuite(version: FuncVersion): void {
             );
```
EricJizbaMSFT, line 137: Seems like you're missing some changes you've done in other repos, like in package.json and .azure-pipelines

nturinski: Damn, I know I did them... I think I forgot to save the changes.  Need to turn auto-save back on lol

EricJizbaMSFT (approved):

PR #2355, opened 2020-09-02
Check for CSharp rather than C#. Also dont limit to id-lang format

nturinski:
Since not all template triggers follow the {triggerId}-{language} format, I thought it made sense to just see if the actualId included both words we are looking for.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2378


src/commands/createFunction/FunctionListStep.ts
```diff
@@ -50,7 +50,8 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
                     const actualId: string = t.id.toLowerCase();
                     const expectedId: string = options.templateId.toLowerCase();
                     // Check for a match with or without the language in the id
-                    return actualId === expectedId || actualId === `${expectedId}-${language.toLowerCase()}`;
+                    // The id for C# triggers include "CSharp" rather than "C#", so we'll have to search for that instead
+                    return actualId === expectedId || actualId.includes(expectedId) && actualId.includes((language === ProjectLanguage.CSharp ? 'CSharp' : language).toLowerCase());
```
EricJizbaMSFT, line 54: The "includes" logic is too simple. For example, there's an "HttpTrigger" template and an "HttpTriggerWithParameters" template that would cause problems.

Also - we don't have to check for the language in particular. It's already filtered by language in the call to "getFunctionTemplates" a few lines up. The only reason I used language before was because I was doing `===` instead of `includes`.

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -47,10 +47,9 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
             const templates: IFunctionTemplate[] = await ext.templateProvider.getFunctionTemplates(context, context.projectPath, language, version, TemplateFilter.All);
             const foundTemplate: IFunctionTemplate | undefined = templates.find((t: IFunctionTemplate) => {
                 if (options.templateId) {
-                    const actualId: string = t.id.toLowerCase();
-                    const expectedId: string = options.templateId.toLowerCase();
-                    // Check for a match with or without the language in the id
-                    return actualId === expectedId || actualId === `${expectedId}-${language.toLowerCase()}`;
+                    const actualName: string = t.name.toLowerCase().replace(/\s/, '');
```
EricJizbaMSFT, line 50: We can't use the name 😬. It's a display name that could change or could be localized. We've had a bug in the past related to this, especially because the templates are returned from the feed and could change without us testing. The id is much more reliable and shouldn't change. Also, it only really comes in two formats that you need to handle:
- HttpTrigger-JavaScript
- Azure.Function.CSharp.HttpTrigger.3.x

You should be able to figure out some conditions or regex magic to handle those two cases. Your "includes" logic before was close, but it just needs to be more careful about finding the whole word without extra stuff at the end. There's probably some regex magic you could do

nturinski: Oh yeah, I remember when that broke us a while ago... 

Anyway, I think using the boundary characters should work to match _only_ whole word cases without any extra alphanumeric characters.

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -49,8 +49,8 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
                 if (options.templateId) {
                     const actualId: string = t.id.toLowerCase();
                     const expectedId: string = options.templateId.toLowerCase();
-                    // Check for a match with or without the language in the id
-                    return actualId === expectedId || actualId === `${expectedId}-${language.toLowerCase()}`;
+                    // check for only whole word matches and escape the "\b" since it's in a string, not a RegExp expression
+                    return new RegExp(`\\b${expectedId}\\b`, 'gmi').test(actualId);
```
EricJizbaMSFT, line 53: The RegExp looks good to me, but please keep `actualId === expectedId` in the check. You're passing in "HttpTrigger", but I also want this to continue working if people pass in the exact id (aka "HttpTrigger-JavaScript" or "Azure.Function.CSharp.HttpTrigger.3.x") instead of the shortened/generic id

EricJizbaMSFT, line 53: Just to be safe, you should use the "escape-string-regexp" npm package to escape expectedId. Same as https://github.com/microsoft/vscode-azuretools/blob/5999c2ad4423e86f22d2c648027242d8816a50e4/ui/src/callWithMaskHandling.ts#L24

nturinski: Trying to use escape, but it's causing problems with escaping `\b` characters.

image
image

EricJizbaMSFT: Well you only have to escape `expectedId`, not your \b characters :P

nturinski: LOL...right 😔

package-lock.json
```diff
@@ -2352,7 +2352,7 @@
         },
         "commander": {
             "version": "2.8.1",
-            "resolved": "https://registry.npmjs.org/commander/-/commander-2.8.1.tgz",
```
EricJizbaMSFT, line 2355: Oof that https to http thing popped back up for ya. You gotta clear your npm cache or whatever. Let me know if you need help finding the fix

EricJizbaMSFT (approved):

PR #2372, opened 2020-09-07
Update profile.ps1 template to set Disable-AzContextAutosave scope to process instead of current user

Francisco-Gamino:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2371

By default, `Disable-AzContextAutosave` cmdlet call in the profile.ps1 defaults to scope user. We need to set this option for the process, so adding `-Scope Process`. Also, adding Out-Null to omit the cmdlet output. 

cc: @anirudhgarg @AnatoliB @stefanushinardi


Francisco-Gamino:
Hello @EricJizbaMSFT -- could you please take a look? Thanks.

EricJizbaMSFT:
A bit confused because the issue you referenced just lists out the exact changes already in the PR 😝. Would like more details on what the change does and why

Francisco-Gamino:
Hello @EricJizbaMSFT-- I've added more details for the code change. Could you please take another look? Thanks.

EricJizbaMSFT (approved):

PR #2395, opened 2020-09-17
Adding support for Stencil

vrdmr:
We want to allow developers to create Azure Function apps from OpenAPI specification file and this commit enables the use of Autorest.Azure-Functions* plugins to be integrated with VSCode Extension.

/fyi @fiveisprime @EricJizbaMSFT @skylerah @anganti


jeffhollan:
Looks great! One thing I didn't see but may be there is some prompt for folks who may not have autorest installed (which I imagine may be a large portion of folks) - just so they know why it's not working and what they can do or where to go to install it

vrdmr:
@jeffhollan - Thanks. Yes, I am adding a check similar to the one done in APIM to check if [autorest exists on the box](https://github.com/microsoft/vscode-apimanagement/pull/121/files#diff-0a195892fd8b16b28a2e378bddd686a8R80). I might need to do one more check if npm exists on the box and install autorest for the developers using it (after taking their permission)

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,64 @@
...
+        } else {
+            throw new Error(localize("notSupported", "Not a supported language. We currently support C#, Java, Python, and Typescript"));
+        }
+
+        args.push('--generate-metadata:false');
+        args.push(`--output-folder:${wizardContext.projectPath}`);
+
+        ext.outputChannel.show();
+        await window.withProgress({ location: ProgressLocation.Notification, title: localize('generatingFunctions', 'Generating Http trigger functions from OpenAPI...') }, async () => {
+            await cpUtils.executeCommand(ext.outputChannel, undefined, 'autorest', ...args);
```
EricJizbaMSFT, line 41: What's the plan for users who don't have autorest installed? And when do we expect that work to be done?

EricJizbaMSFT (changes requested):

src/commands/createNewProject/IProjectWizardContext.ts
```diff
@@ -17,6 +17,8 @@ export interface IProjectWizardContext extends IActionContext {
     languageFilter?: RegExp;
     version: FuncVersion;
     openBehavior?: OpenBehavior;
+
+    generataFromOpenAPI?: boolean;
```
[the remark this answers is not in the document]
vrdmr: Thanks for catching this.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export class OpenAPICreateStep extends AzureWizardExecuteStep<IFunctionWizardContext> {
+    public priority: number;
+
+    public async execute(wizardContext: IFunctionWizardContext): Promise<void> {
+        const uris: Uri[] = await this.askDocument();
+        const uri: Uri = uris[0];
+
+        const args: string[] = [];
+        args.push(`--input-file:${uri.fsPath} --version:3.0.6314`);
```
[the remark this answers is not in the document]
vrdmr: Currently, this is hard-coded as the new version of autorest/core hasn't been deployed to NpmJS and when passing in the version, we force autorest to download this version from its own repo. Once the autorest/core gets released, I'll remove the `--version` argument.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,64 @@
+import { OpenDialogOptions, ProgressLocation, Uri, window, workspace } from "vscode";
+import { AzureWizardExecuteStep } from "vscode-azureextensionui";
+import { ProjectLanguage } from "../../../constants";
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { cpUtils } from "../../../utils/cpUtils";
+import { IFunctionWizardContext } from "../../createFunction/IFunctionWizardContext";
+
```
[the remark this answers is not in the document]
vrdmr: This creates a new function app project and not a single function - which is why this was in the `createNewProject` folder.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,64 @@
...
+    public shouldExecute(): boolean {
+        return true;
+    }
+
+    public async askDocument(): Promise<Uri[]> {
+        const openDialogOptions: OpenDialogOptions = {
+            canSelectFiles: true,
+            canSelectFolders: false,
+            canSelectMany: false,
+            openLabel: 'Select OpenAPI Specification (V2 or V3) File',
```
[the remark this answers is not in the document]
vrdmr: Here are the samples - there wasn't a way to describe people to choose a file.

image

image

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
vrdmr: We want users to install `autorest` rather than installing it for them - in future, we can install it after asking permission from the user.

Currently, I am going to throw with [Autorest's README](https://aka.ms/autorest ) added to the error.

@JeffHollan had similar comment above - I've added that in the next iteration.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,64 @@
...
+        const uri: Uri = uris[0];
+
+        const args: string[] = [];
+        args.push(`--input-file:${uri.fsPath} --version:3.0.6314`);
+
+        if (wizardContext.language === ProjectLanguage.TypeScript) {
+            args.push('--azure-functions-typescript');
+            args.push('--no-namespace-folders:True');
+        } else if (wizardContext.language === ProjectLanguage.CSharp) {
+            args.push('--namespace:Microsoft.Azure.Stencil');
```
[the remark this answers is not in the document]
vrdmr: I'll need your help with this - I am trying to integrate it but the promp to ask for the NameSpace isn't coming up.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Well FunctionListStep is also used during "Create Function". If this only applies for a new project, you'll have to make sure it doesn't show up for "Create Function"

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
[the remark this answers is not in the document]
EricJizbaMSFT: You'll have to give more details or just push up what you have so far

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah that's pretty bad... Usually this is a one or two word button. I think you'll need to find a different way to describe to users what they're selecting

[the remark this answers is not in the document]
vrdmr: I have set it to `Specification file` for the time being. That is sufficient for now.

src/commands/createNewProject/openAPISteps/OpenAPICreateStep.ts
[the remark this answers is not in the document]
vrdmr: Circumvented the issue with a different approach. You can check the `askNamespace` method here.

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -164,6 +176,14 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
             });
         }
 
+        if (language === ProjectLanguage.CSharp || language === ProjectLanguage.Java || language === ProjectLanguage.Python || language === ProjectLanguage.TypeScript) {
```
EricJizbaMSFT, line 179: FunctionListStep is used for 'Create Project' _and_ 'Create Function'. Seems like you were saying this is only for 'Create Project', in which case you should check `this._isProjectWizard` in this conditional

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,116 @@
...
+        const errorMessage: string = 'Please make sure you have Autorest installed. Visit https://aka.ms/autorest to find more information on Autorest and installation steps.';
+
+        if (!await isNPMInstalled()) {
+            throw new Error(localize("notSupported", `NPM is not installed. ${errorMessage}`));
+        } else if (!await isAutoRestInstalled()) {
+            throw new Error(localize("notSupported", `Autorest could not be detected on the system. ${errorMessage}`));
+        }
+
+        let namespace: string = '';
+        if (wizardContext.language === ProjectLanguage.CSharp || wizardContext.language === ProjectLanguage.Java) {
```
EricJizbaMSFT, line 31: All prompts should be a part of AzureWizardPromptSteps, not AzureWizardExecuteSteps (for various reasons, including the step count and back button).  For `askDocument`, you'll have to create your own prompt step. For `askNamespace`, you should be able to use `DotnetNamespaceStep` for C# and  `JavaPackageNameStep` for Java. Those steps should be added inside `FunctionListStep.getSubWizard` around [here](https://github.com/microsoft/vscode-azurefunctions/pull/2395/files#diff-66c8dd99a66b6c46145ef571643118e6R121), similar to [here](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/FunctionListStep.ts#L79)

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,116 @@
...
+            namespace = await askNamespace(wizardContext.language);
+        }
+
+        args.push(`--input-file:${cpUtils.wrapArgInQuotes(uri.fsPath)} --version:3.0.6314`);
+
+        if (wizardContext.language === ProjectLanguage.TypeScript) {
```
EricJizbaMSFT, line 37: nit: I would use a 'switch case' instead of all these 'else ifs'

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,116 @@
...
+            args.push('--no-namespace-folders:True');
+            args.push('--no-async');
+        } else {
+            throw new Error(localize('notSupported', 'Not a supported language. We currently support C#, Java, Python, and Typescript'));
+        }
+
+        args.push('--generate-metadata:false');
+        args.push(`--output-folder:${wizardContext.projectPath}`);
+
+        ext.outputChannel.show();
```
EricJizbaMSFT, line 57: We try to avoid popping up the output channel because it can take a large chunk of screen real-estate and get annoying. If you really think the output is important, you can include a link in the message like [this](https://github.com/microsoft/vscode-azuretools/blob/5999c2ad4423e86f22d2c648027242d8816a50e4/appservice/src/deploy/deploy.ts#L70) to pop it up

src/commands/createFunction/FunctionListStep.ts
vrdmr: I started investigating after looking at your comment. I feel this is useful in both the cases - and your hunch was correct. As people can _Skip for now_, the access to creating HttpTriggers from OpenAPI goes away, and that is not the ideal. Which is why I moved this out of the `this._isProjectWizard` conditional.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: Thanks. noob mistake. Coming from Python doesn't help. :)

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: I've shortened it and removed extra log lines from it. I felt the pop-up is important as the OpenAPI generation can take some time and users may want to know what is happening in the background. The idea of adding the output_window link is great. I did that - but when I click the output window link now, it doesn't open the window.

I read [this](https://github.com/microsoft/vscode-azuretools/blob/5999c2ad4423e86f22d2c648027242d8816a50e4/appservice/src/tree/DeploymentTreeItem.ts#L30) comment - Do I need to register my command somewhere to allow the output_window link to be clickable?
```typescript
/**
 * NOTE: This leverages two commands prefixed with `ext.prefix` that should be registered by each extension: "showOutputChannel" and "viewDeploymentLogs"
 */
```

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: Thanks a lot for the pointers - I was able to use them and simplify my code a lot. :)

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
EricJizbaMSFT: You should use this command, which is already registered: https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/registerCommands.ts#L109

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -113,6 +114,25 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
...
+                case ProjectLanguage.CSharp:
+                    promptSteps.push(new DotnetNamespaceStep());
+                    break;
+                default:
+                    break;
+            }
+
+            executeSteps.push(await OpenAPICreateStep.createStep(context));
+
+            const title: string = localize('createFunction', 'Create new {0}', '');
```
EricJizbaMSFT, line 134: Create new ... what? Right now it doesn't say

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -113,6 +114,25 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
 
             const title: string = localize('createFunction', 'Create new {0}', template.name);
             return { promptSteps, executeSteps, title };
+        } else if (context.generateFromOpenAPI) {
+            const promptSteps: AzureWizardPromptStep<IFunctionWizardContext>[] = [];
+            const executeSteps: AzureWizardExecuteStep<IFunctionWizardContext>[] = [];
+
+            switch (context.language) {
+                case ProjectLanguage.Java:
+                    promptSteps.push(new JavaFunctionNameStep());
```
EricJizbaMSFT, line 123: JavaFunctionNameStep -> JavaPackageNameStep

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,107 @@
...
+        await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
+            await cpUtils.tryExecuteCommand(ext.outputChannel, undefined, 'autorest', ...args);
+        });
+    }
+
+    public shouldExecute(): boolean {
+        return true;
+    }
+
+    public async askDocument(): Promise<Uri[]> {
```
EricJizbaMSFT, line 71: You need to create an `AzureWizardPromptStep` for this. There shouldn't be any prompts as a part of an "execute" step

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,114 @@
...
+        return new OpenAPICreateStep();
+    }
+
+    public async execute(wizardContext: IFunctionWizardContext & IJavaProjectWizardContext & IDotnetFunctionWizardContext): Promise<void> {
+        // const uris: Uri[] = await this.askDocument();
+        const uris: Uri[] = nonNullProp(wizardContext, 'openApiSpecificationFile');
+        const uri: Uri = uris[0];
+        const args: string[] = [];
+
+        args.push(`--input-file:${cpUtils.wrapArgInQuotes(uri.fsPath)} --version:3.0.6320`);
```
EricJizbaMSFT, line 32: I'm still confused about hard-coding this version. When can we remove it - before this PR is merged? And what happens if the users doesn't have the latest version installed on their machine - will it fail?

src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
```diff
@@ -21,7 +21,8 @@ export class TypeScriptProjectCreateStep extends JavaScriptProjectCreateStep {
 
     protected packageJsonDevDeps: { [key: string]: string } = {
         '@azure/functions': '^1.0.2-beta2',
-        typescript: '^3.3.3'
+        typescript: '^3.3.3',
+        '@azure/core-http': '^1.1.4'
```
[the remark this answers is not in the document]
vrdmr: Is there a selective way to add this dependency? for the OpenAPI generated projects? We need this `@azure/core-http` dependency for the apps we create.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: The `--version` argument was mainly as 6320 hadn't been deployed yet. But with [@autorest/core@3.0.6320](https://www.npmjs.com/package/@autorest/core) now deployed. It is not needed anymore now. Removing it now.

src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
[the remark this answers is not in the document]
EricJizbaMSFT: Yeah you should be able to find a way to add it for just your case, probably during `OpenAPICreateStep.execute`. Most projects will not need `@azure/core-http`, so we don't want to add it by default.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        args.push('--generate-metadata:false');
+        args.push(`--output-folder:${cpUtils.wrapArgInQuotes(wizardContext.projectPath)}`);
+
+        ext.outputChannel.appendLog(localize('statutoryWarning', 'Using the plugin could overwrite your custom changes to the functions.'));
+        const title: string = localize('generatingFunctions', 'Generating from OpenAPI...Check [output window](command:{0}) for status.', ext.prefix + '.showOutputChannel');
+
+        await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
+            await cpUtils.tryExecuteCommand(ext.outputChannel, undefined, 'autorest', ...args);
```
EricJizbaMSFT, line 64: This should use `executeCommand` instead of `tryExecuteCommand`, otherwise the user won't know if an error occurred

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,89 @@
...
+                args.push(`--namespace:${nonNullProp(wizardContext, 'javaPackageName')}`);
+                args.push('--azure-functions-java');
+                break;
+            case ProjectLanguage.Python:
+                args.push('--azure-functions-python');
+                args.push('--no-namespace-folders:True');
+                args.push('--no-async');
+                break;
+            default:
+                throw new Error(localize('notSupported', 'Not a supported language. We currently support C#, Java, Python, and Typescript'));
```
EricJizbaMSFT, line 53: You don't support JavaScript?

vrdmr: Not yet. Once we do JS and PS, we'll be adding them as well.

src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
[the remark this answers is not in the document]
vrdmr: Done.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -0,0 +1,130 @@
...
+                }
+            });
+            context.errorHandling.suppressDisplay = true;
+        }
+
+        throw new Error(message);
+    }
+}
+
+async function addAutorestSpecificTypescriptDependencies(context: IFunctionWizardContext): Promise<void> {
```
EricJizbaMSFT, line 99: A few problems with this code:
1. It uses `coreHttp` as the package name instead of `@azure/core-http` in package.json
1. It asks to overwrite `package.json` when creating a project in an empty folder
1. It always overwrites the entire file, instead of editing the existing file

Here are some better options without those problems.

### Option 1:
```typescript
await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'npm', 'install', '@azure/core-http', '--save-dev');
```

### Option 2:
```typescript
const coreHttp: string = '@azure/core-http';
const coreHttpVersion: string = '^1.1.4';
const packagePath: string = path.join(context.projectPath, 'package.json');

await confirmEditJsonFile(packagePath, (data: { devDependencies?: { [key: string]: string } }): {} => {
    // tslint:disable-next-line: strict-boolean-expressions
    data.devDependencies = data.devDependencies || {};
    if (!data.devDependencies[coreHttp]) {
        data.devDependencies[coreHttp] = coreHttpVersion;
    }
    return data;
});
```

vrdmr: @EricJizbaMSFT Thanks. I've used the 2nd option to do this work.

EricJizbaMSFT (approved):

PR #2401, opened 2020-09-18
App Settings API

nturinski:
Tested this with Static Web Apps, seems like this is all I have to do to leverage upload/download app settings from the Function extension.

Bumping the version as well so I wasn't creating false telemetry events.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azurefunctions",
-    "version": "0.24.1",
+    "version": "0.24.2-alpha",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
--- a/src/commands/api/uploadAppSettingsFromApi.ts
+++ b/src/commands/api/uploadAppSettingsFromApi.ts
@@ -2,12 +2,12 @@
 *  Copyright (c) Microsoft Corporation. All rights reserved.
 *  Licensed under the MIT License. See License.txt in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
-import { OutputChannel } from "vscode";
 import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
-import { ISimpleAppSettingsClient, uploadAppSettingsInternal } from "../appSettings/uploadAppSettings";
+import { ISimpleAppSettingsClient } from "../../vscode-azurefunctions.api";
+import { uploadAppSettingsInternal } from "../appSettings/uploadAppSettings";
 
-export async function uploadAppSettingsFromApi(client: ISimpleAppSettingsClient, outputChannel: OutputChannel, workspacePath?: string): Promise<void> {
+export async function uploadAppSettingsFromApi(client: ISimpleAppSettingsClient, workspacePath?: string): Promise<void> {
     return await callWithTelemetryAndErrorHandling('api.uploadAppSettings', async (context: IActionContext) => {
-        await uploadAppSettingsInternal(context, client, outputChannel, workspacePath);
+        await uploadAppSettingsInternal(context, client, workspacePath);
     });
 }
--- a/src/commands/api/uploadAppSettingsFromApi.ts
+++ b/src/commands/api/uploadAppSettingsFromApi.ts
@@ -1,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { AppSettingsTreeItem } from "vscode-azureappservice";
+import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
+import { uploadAppSettings } from "../appSettings/uploadAppSettings";
+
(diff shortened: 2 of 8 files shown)


src/vscode-azurefunctions.api.d.ts
```diff
@@ -3,12 +3,16 @@
...
 
+import { AppSettingsTreeItem } from "vscode-azureappservice";
+
 export interface AzureFunctionsExtensionApi {
     apiVersion: string;
 
     revealTreeItem(resourceId: string): Promise<void>;
 
     createFunction(options: ICreateFunctionOptions): Promise<void>;
+    downloadAppSettings(node?: AppSettingsTreeItem): Promise<void>;
```
EricJizbaMSFT, line 14: You can't guarantee this `AppSettingsTreeItem` class will be the same across extensions, so we can't use it. You can pass something like the function app id and go from there

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "vscode-azurefunctions",
-    "version": "0.24.1",
+    "version": "0.24.2-alpha",
```
EricJizbaMSFT, line 3: Please do alpha bump in a separate PR. They're unrelated and the app settings stuff may take longer to review/merge while the alpha bump should be super quick

src/vscode-azurefunctions.api.d.ts
nturinski: An issue I'm running with that approach is that the tree view is not shared across extensions.  Therefore, even if I pass the application tree id itself, it won't be able to find the node.  As far as I know, the platform team doesn't have plans to expose the Function App associated to the SWA to the Functions API endpoint, so I'm not sure there'll be a way to workaround this issue.

nturinski: A workaround to this is to pass the AzExtTreeDataProvider.  I feel like this makes the API pretty rigid, but considering we're dealing with an Azure resource anyway, I think that's okay.  Let me know if you have any thoughts on it, but that's what I'll implement for now.

nturinski:
This file https://github.com/microsoft/vscode-azurefunctions/blob/43b4b12ac36f7f9b864f25d5701822a098a8ade9/src/commands/appSettings/uploadAppSettings.ts pretty much has all of the preliminary stuff I wanted to ask about.

What I'm doing to use the outputChannel is terrible to me.  I'm wondering if there's a better way, or would it be better to just use `appendLine` in this case so that it doesn't matter if it is a vscode.OutputChannel.  I feel like having the output be in the wrong channel is definitely something that's going to be filed by CTI...

The rest of what I'm doing I'm basically okay with-- wanted to see what you thought before proceeding with downloadAppSettings.

src/vscode-azurefunctions.api.d.ts
```diff
@@ -9,12 +9,24 @@ export interface AzureFunctionsExtensionApi {
...
     createFunction(options: ICreateFunctionOptions): Promise<void>;
+    downloadAppSettings(client: ISimpleAppSettingsClient): Promise<void>;
+    uploadAppSettings(client: ISimpleAppSettingsClient): Promise<void>;
 }
 
 export type ProjectLanguage = 'JavaScript' | 'TypeScript' | 'C#' | 'Python' | 'PowerShell' | 'Java';
-
 export type ProjectVersion = '~1' | '~2' | '~3';
 
+export interface ISimpleAppSettingsClient {
```
EricJizbaMSFT, line 19: Let's rename `ISimpleAppSettingsClient` to `IAppSettingsClient`

I feel like the only reason you didn't do that already is because of `IAppSettingsClient` in vscode-azureappservice, but I don't think that's a good enough reason to pick a worse name here. They both have a legitimate reason to be called `IAppSettingsClient` and we can work around any problems arising from that.

src/commands/api/uploadAppSettingsFromApi.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azureextensionui";
+import { ISimpleAppSettingsClient } from "../../vscode-azurefunctions.api";
+import { uploadAppSettingsInternal } from "../appSettings/uploadAppSettings";
+
+export async function uploadAppSettingsFromApi(client: ISimpleAppSettingsClient, workspacePath?: string): Promise<void> {
```
EricJizbaMSFT, line 9: Let's remove `workspacePath`. I don't see it used in SWA or even listed in "vscode-azurefunctions.api.d.ts"

nturinski: Yeah, I meant to remove that since Functions has logic regarding retrieving the workspacePath if not provided 👍

src/vscode-azurefunctions.api.d.ts
nturinski: Yeah, that is pretty much the reason why.  That and this isn't really a "complete" `AppSettingsClient` since you wouldn't be able to use it with an `AppSettingsTreeNode`.  But I'm okay with changing it, just didn't want to cause confusion.

EricJizbaMSFT (approved):

Issue #2326, opened 2020-08-21
When deploying 64 bit project, Azure function app should be updated to 64 bit mode


prabh-62:
## System 
- Visual Studio Code - `1.48.1`
- [Azure Functions extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azurefunctions) - `0.23.0`

Create a 64-bit Azure Functions project and `Deploy to Functions App` using extension
<img width="1299" alt="64_bit_vscode" src="https://user-images.githubusercontent.com/13006107/90841891-e39c6d80-e32b-11ea-8446-e69ee5e0ea79.png">

VSCode extension doesn't update project setting in Azure portal if previously app was in 32-bit mode.
<img width="1257" alt="32_bit_func_portal" src="https://user-images.githubusercontent.com/13006107/90842007-2bbb9000-e32c-11ea-8c66-ae59856eee28.png">

After successfully deploying 64 bit, if we hit any functions, API returns `HTTP 503: Service Unavailable`
<img width="601" alt="64_bit_not_running" src="https://user-images.githubusercontent.com/13006107/90842039-4261e700-e32c-11ea-8665-171ee61e539f.png">

### Expected Outcome
Extension should update Azure Functions app to 64-bit in Azure Portal as well

prabh-62:
@EricJizbaMSFT I saw the `help-wanted` label and created a Pull request for this issue. Looking for feedback on the PR https://github.com/microsoft/vscode-azurefunctions/pull/2429

EricJizbaMSFT:
Fixed in https://github.com/microsoft/vscode-azurefunctions/pull/2429

PR #2429, opened 2020-10-01
Update workerProcess to 64bit if platformTarget is x64

prabh-62:
Issue: https://github.com/microsoft/vscode-azurefunctions/issues/2326

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -98,11 +98,13 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
     const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
     if (functionProject === undefined) {
-        throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
+        return;
+        // throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
     }
     const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
     if (projectFiles.length === 0) {
-        throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+        return;
+        // throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
     }
     const mainProject: string = path.join(functionProject, projectFiles[0]);
     const platformTarget: string | undefined = await dotnetUtils.tryGetPlatformTarget(mainProject);
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -20,6 +20,7 @@ import { dotnetUtils } from '../../utils/dotnetUtils';
 import { isPathEqual } from '../../utils/fs';
 import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
 import { verifyInitForVSCode } from '../../vsCodeConfig/verifyInitForVSCode';
+import { tryGetFunctionProjectRoot } from '../createNewProject/verifyIsProject';
 import { notifyDeployComplete } from './notifyDeployComplete';
 import { runPreDeployTask } from './runPreDeployTask';
 import { validateRemoteBuild } from './validateRemoteBuild';
@@ -95,8 +96,15 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 }
 
 async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
-    const projFiles: string[] = await dotnetUtils.getProjFiles(language, context.workspaceFolder.uri.fsPath);
-    const mainProject: string = path.join(context.workspaceFolder.uri.fsPath, projFiles[0]);
+    const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
+    if (functionProject === undefined) {
+        throw new Error(localize('failedToFindFuncHost', 'Unable not locate Azure Functions project.'));
+    }
+    const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
+    if (projectFiles.length === 0) {
+        throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+    }
+    const mainProject: string = path.join(functionProject, projectFiles[0]);
     const platformTarget: string | undefined = await dotnetUtils.getPlatformTarget(mainProject);
     if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
         const config: WebSiteManagementModels.SiteConfigResource = {
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -99,7 +99,7 @@ async function tryToUpdateWorkerProcessTo64BitIfRequired(context: IDeployContext
     try {
         const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
         if (functionProject === undefined) {
-            throw new Error(localize('failedToFindFuncHost', 'Unable not locate Azure Functions project.'));
+            throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
         }
         const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
         if (projectFiles.length === 0) {
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -66,16 +66,7 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     }
 
     if (language === ProjectLanguage.CSharp) {
-        const projFiles: string[] = await dotnetUtils.getProjFiles(language, context.workspaceFolder.uri.fsPath);
-        const mainProject: string = path.join(context.workspaceFolder.uri.fsPath, projFiles[0]);
-        const platformTarget: string | undefined = await dotnetUtils.getPlatformTarget(mainProject);
-        if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
-            const config: WebSiteManagementModels.SiteConfigResource = {
-                use32BitWorkerProcess: false
-            };
-            context.stopAppBeforeDeploy = true;
-            await node.root.client.updateConfiguration(config);
-        }
+        await updateWorkerProcessTo64BitIfRequired(context, siteConfig, node, language);
     }
 
     await verifyAppSettings(context, node, version, language);
@@ -101,6 +92,19 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     );
 
     await notifyDeployComplete(context, node, context.workspaceFolder.uri.fsPath);
+}
+
+async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
+    const projFiles: string[] = await dotnetUtils.getProjFiles(language, context.workspaceFolder.uri.fsPath);
+    const mainProject: string = path.join(context.workspaceFolder.uri.fsPath, projFiles[0]);
+    const platformTarget: string | undefined = await dotnetUtils.getPlatformTarget(mainProject);
+    if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+        const config: WebSiteManagementModels.SiteConfigResource = {
+            use32BitWorkerProcess: false
+        };
+        context.stopAppBeforeDeploy = true;
+        await node.root.client.updateConfiguration(config);
+    }
 }
 
 async function validateGlobSettings(context: IActionContext, fsPath: string): Promise<void> {
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { WebSiteManagementModels } from '@azure/arm-appservice';
+import * as path from 'path';
 import * as vscode from 'vscode';
 import { deploy as innerDeploy, getDeployFsPath, getDeployNode, IDeployContext, IDeployPaths, showDeployConfirmation } from 'vscode-azureappservice';
 import { IActionContext } from 'vscode-azureextensionui';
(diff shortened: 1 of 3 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -63,6 +65,19 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         validateGlobSettings(context, context.effectiveDeployFsPath);
     }
 
+    if (language === ProjectLanguage.CSharp) {
+        const projFiles: string[] = await dotnetUtils.getProjFiles(language, context.workspaceFolder.uri.fsPath);
+        const mainProject: string = path.join(context.workspaceFolder.uri.fsPath, projFiles[0]);
```
prabh-62, line 70: I feel like this approach might not work where functions apps are nested inside directory. How do I get access to right `*.csproj` file ?

src/commands/initProjectForVSCode/InitVSCodeStep/DotnetInitVSCodeStep.ts
```diff
@@ -63,11 +63,11 @@ export class DotnetInitVSCodeStep extends InitVSCodeStepBase {
 
         const projFilePath: string = path.join(projectPath, projFileName);
 
-        const versionInProjFile: string = await dotnetUtils.getPropertyInProjFile(projFilePath, 'AzureFunctionsVersion');
+        const versionInProjFile: FuncVersion | undefined = await dotnetUtils.getFuncVersion(projFilePath);
```
EricJizbaMSFT, line 66: I want this to be the exact version in the proj file before `tryParseFuncVersion` is called. We collect this in telemetry in the next line, and it's helpful to know what "unexpected" versions people are using

src/utils/dotnetUtils.ts
```diff
@@ -26,7 +27,20 @@ export namespace dotnetUtils {
         return await getPropertyInProjFile(projFilePath, 'TargetFramework');
     }
 
-    export async function getPropertyInProjFile(projFilePath: string, prop: string): Promise<string> {
+    export async function getFuncVersion(projFilePath: string): Promise<FuncVersion | undefined> {
+        const version: string = await getPropertyInProjFile(projFilePath, 'AzureFunctionsVersion');
+        return tryParseFuncVersion(version);
+    }
+
+    export async function getPlatformTarget(projFilePath: string): Promise<string | undefined> {
```
EricJizbaMSFT, line 35: If a function returns `undefined` instead of throwing an error, it should have "try" at the beginning of the name

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,26 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+        throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+    }
+    const mainProject: string = path.join(functionProject, projectFiles[0]);
+    const platformTarget: string | undefined = await dotnetUtils.getPlatformTarget(mainProject);
+    if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+        const config: WebSiteManagementModels.SiteConfigResource = {
+            use32BitWorkerProcess: false
+        };
+        context.stopAppBeforeDeploy = true;
+        await node.root.client.updateConfiguration(config);
```
EricJizbaMSFT, line 114: We try to avoid silent changes to a user's function app. Please add a warning similar to this: https://github.com/microsoft/vscode-azurefunctions/blob/d1318cad6809dbe65029dfe2012c76a1b17d0df5/src/commands/deploy/verifyAppSettings.ts#L86

For example, maybe they didn't actually want 64-bit and they will cancel this deploy and change their local project instead of the app in Azure.

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,26 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+    }
+    const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
+    if (projectFiles.length === 0) {
+        throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+    }
+    const mainProject: string = path.join(functionProject, projectFiles[0]);
+    const platformTarget: string | undefined = await dotnetUtils.getPlatformTarget(mainProject);
+    if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+        const config: WebSiteManagementModels.SiteConfigResource = {
+            use32BitWorkerProcess: false
```
EricJizbaMSFT, line 111: Does this preserve all existing properties on the site config? Or does it change them back to the default? I think it would be safer if you edited the existing `siteConfig` object and sent that in `updateConfiguration`

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,26 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+        };
+        context.stopAppBeforeDeploy = true;
```
EricJizbaMSFT, line 113: Did you add this because you have to restart a function app every time you update the bitness? If so, that's not quite how this property was meant to be used. If a restart is necessary, I'd rather just do the full restart here, aka before the deploy

src/commands/deploy/deploy.ts
```diff
@@ -63,6 +66,10 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         validateGlobSettings(context, context.effectiveDeployFsPath);
     }
 
+    if (language === ProjectLanguage.CSharp) {
```
EricJizbaMSFT, line 69: I believe the bitness can only be changed on Windows function apps, right? If so, you should also check
`&& !node.root.client.isLinux`

EricJizbaMSFT (changes requested):

src/commands/deploy/deploy.ts
prabh-62: My intent here is to only update app to 64 bit and affect nothing else. If I  use existing `siteConfig` and somebody mutates `siteConfig` object, then I will unintentionally update other settings as well.

I can use existing `siteConfig` object if you believe it is safe to do so.

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,26 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
 }
 
+async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
+    const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
+    if (functionProject === undefined) {
+        throw new Error(localize('failedToFindFuncHost', 'Unable not locate Azure Functions project.'));
+    }
+    const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
+    if (projectFiles.length === 0) {
+        throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
```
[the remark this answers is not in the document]
prabh-62: Added a `try catch` around the new code so it doesn't stop the deployment

src/commands/initProjectForVSCode/InitVSCodeStep/DotnetInitVSCodeStep.ts
prabh-62: reverted the change

src/utils/dotnetUtils.ts
prabh-62: Acknowledged

src/commands/deploy/deploy.ts
prabh-62: Acknowledged

src/commands/deploy/deploy.ts
prabh-62: if we don't restart the app, then we get an error `Function host is not running`

src/commands/deploy/deploy.ts
prabh-62: Acknowledged

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,48 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+            await node.root.client.updateConfiguration(config);
+        }
+    } catch (error) {
+        // swallow cancellations
+        const parsedError: IParsedError = parseError(error);
+        if (!parsedError.isUserCancelledError) {
+            return;
+        }
+        // We try out best to update app to 64 bit
+        // If we don't succeed, we shouldn't stop the deployment
```
EricJizbaMSFT, line 131: This isn't quite what I meant. If the user explicitly chooses to update the bitness, it's okay if an error stops deployment. They're not blocked, because they can just deploy again and select "Deploy Anyway" the next time.

I suggest you get rid of the entire catch block and instead change the above lines where you `throw new Error` to just `return;`. Those cases _would_ block a user from deploying, which is what we're trying to avoid.

src/commands/deploy/deploy.ts
EricJizbaMSFT: In that case, I would before if you called `restartFunctionApp` right after you update the site config, instead of setting `context.stopAppBeforeDeploy`

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,48 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+        if (projectFiles.length === 0) {
+            throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+        }
+        const mainProject: string = path.join(functionProject, projectFiles[0]);
+        const platformTarget: string | undefined = await dotnetUtils.tryGetPlatformTarget(mainProject);
+        if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+            const message: string = localize('overwriteSetting', 'Detected 64 bit project. Update setting in Azure Portal?');
+            const deployAnyway: vscode.MessageItem = { title: localize('deployAnyway', 'Deploy Anyway'), isCloseAffordance: true };
+            const dialogResult: vscode.MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, DialogResponses.yes, deployAnyway);
+            if (dialogResult.isCloseAffordance === true) {
```
EricJizbaMSFT, line 114: No need to do this `isCloseAffordance` stuff. Just check `if (dialogResult === deployAnyway) {`

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,48 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+            throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
+        }
+        const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
+        if (projectFiles.length === 0) {
+            throw new Error(localize('unableToFindProj', 'Unable to detect project file.'));
+        }
+        const mainProject: string = path.join(functionProject, projectFiles[0]);
+        const platformTarget: string | undefined = await dotnetUtils.tryGetPlatformTarget(mainProject);
+        if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+            const message: string = localize('overwriteSetting', 'Detected 64 bit project. Update setting in Azure Portal?');
```
EricJizbaMSFT, line 111: How about this:

> The remote app targets "{0}", but your local project targets "{1}". Update remote app to "{1}"? This will restart the app.

We are updating the setting for them, so they don't have to use the portal like you said. Also, I want them to know we will restart the app.

src/commands/deploy/deploy.ts
prabh-62: I tested few times by manually switching app to 32 bit and making local project 64 bit. In all cases, function app had no issues. So, I am removing the code to restart app.

src/commands/deploy/deploy.ts
prabh-62: Removed code which threw error

src/commands/deploy/deploy.ts
prabh-62: Changed

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,33 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
 }
 
+async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
+    const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
+    if (functionProject === undefined) {
+        return;
+        // throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
+    }
+    const projectFiles: string[] = await dotnetUtils.getProjFiles(language, functionProject);
+    if (projectFiles.length === 0) {
```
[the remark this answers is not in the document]
prabh-62: Done

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,33 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     await notifyDeployComplete(context, node, context.workspaceFolder.uri.fsPath);
 }
 
+async function updateWorkerProcessTo64BitIfRequired(context: IDeployContext, siteConfig: WebSiteManagementModels.SiteConfigResource, node: SlotTreeItemBase, language: ProjectLanguage): Promise<void> {
+    const functionProject: string | undefined = await tryGetFunctionProjectRoot(context.workspaceFolder.uri.fsPath);
+    if (functionProject === undefined) {
+        return;
+        // throw new Error(localize('failedToFindFuncHost', 'Unable to locate Azure Functions project.'));
```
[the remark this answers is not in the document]
prabh-62: Removed

src/commands/deploy/deploy.ts
```diff
@@ -88,6 +95,33 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+    }
+    const mainProject: string = path.join(functionProject, projectFiles[0]);
+    const platformTarget: string | undefined = await dotnetUtils.tryGetPlatformTarget(mainProject);
+    if (platformTarget === 'x64' && siteConfig.use32BitWorkerProcess === true) {
+        const message: string = localize('overwriteSetting', 'The remote app targets "{0}", but your local project targets "{1}". Update remote app to "{1}"?', '32 bit', '64 bit');
+        const no64Bit: vscode.MessageItem = { title: localize('noTo64Bit', 'No'), isCloseAffordance: true };
+        const dialogResult: vscode.MessageItem = await ext.ui.showWarningMessage(message, { modal: true }, DialogResponses.yes, no64Bit);
+        if (dialogResult === no64Bit) {
+            return;
+        }
```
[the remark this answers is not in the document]
prabh-62: My apologies, I didn't see that this was a suggestion (not a comment). I committed similar code to the branch

EricJizbaMSFT (approved):
Thanks!

prabh-62:
Thank you for reviewing the pull request! Eric

src/commands/deploy/deploy.ts
lovettchris: I guess this fix is not live yet on https://ms.portal.azure.com/ ? I'm still seeing this problem ...

lovettchris: Also, how does one change a function app to 64bit using an "az command"  - for example I want to create my "az functionapp create" my app as 64bit to start with...

Issue #2359, opened 2020-09-03
Support comments in "local.settings.json"


wdy0719:
Cannot build/create azure function when local.settings.json contains comment lines (which works in vs2019).

Action: azureFunctions.createFunction
Error type: Error
Error Message: Failed to parse "local.settings.json": Unexpected token / in JSON at position 287.

Version: 0.24.0
OS: win32
OS Release: 10.0.19041
Product: Visual Studio Code
Product Version: 1.48.2
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:33:862999extension.bundle.js:33:862999
fulfilled extension.bundle.js:33:862293
```

</details>

![image](https://user-images.githubusercontent.com/11709227/92162968-d41b2b00-ede7-11ea-94ae-eafd96bcd0ec.png)

martinhenrysson:
I would also like support for comments. It is sometimes useful when switching between environments and using code that was checked-in from Visual Studio 2019.

acds:
I'd suggest just changing value name (like append an `X`) such that it different than the code expects so the parameter is config is not found via the json, but defers Environment others other configured source.

JSON does not support comments as standard. if you want to actually add comments just use a element labels:
```JSON
{
    "_comment":  "Comments with // or /**/ are NEVER used in a JSON file!!!",
    "regulardata": "stuff",
    "mySettingX": "my setting with X appended so code does not pick it up from json config"
}
```

PR #2430, opened 2020-10-01
Don't throw error if there are comments in local.settings.json

EricJizbaMSFT:
The only downside of the jsonc parser that I'm using (which is actually maintained by a VS Code dev) is I have to handle errors myself. I feel like the new message is at least as good as the old message, which is all that I care about.

Old:
Screen Shot 2020-09-29 at 5 43 39 PM

New:
Screen Shot 2020-09-29 at 5 41 33 PM

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2359

### DIFF
--- a/src/utils/parseJson.ts
+++ b/src/utils/parseJson.ts
@@ -3,12 +3,40 @@
  *  Licensed under the MIT License. See License.md in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as jsonc from 'jsonc-parser';
+import { localize } from '../localize';
+
 /**
- * Has extra logic to remove a BOM character if it exists
+ * Has extra logic to remove a BOM character if it exists and handle comments
  */
 export function parseJson<T extends object>(data: string): T {
     if (data.charCodeAt(0) === 0xFEFF) {
         data = data.slice(1);
     }
-    return <T>JSON.parse(data);
+
+    const errors: jsonc.ParseError[] = [];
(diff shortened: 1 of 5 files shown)


src/utils/parseJson.ts
```diff
@@ -3,12 +3,40 @@
...
     if (data.charCodeAt(0) === 0xFEFF) {
         data = data.slice(1);
     }
-    return <T>JSON.parse(data);
+
+    const errors: jsonc.ParseError[] = [];
+    const result: T = <T>jsonc.parse(data, errors, { allowTrailingComma: true });
+    if (errors.length > 0) {
+        const [line, column]: [number, number] = getLineAndColumnFromOffset(data, errors[0].offset);
+        throw new Error(localize('jsonParseError', '{0} near line "{1}", column "{2}"', jsonc.printParseErrorCode(errors[0].error), line, column));
```
nturinski, line 21: Are there ever multiple errors?  Would we ever want to output all of the errors and direct them to the console?

nturinski (approved):

src/utils/parseJson.ts
EricJizbaMSFT: Yeah there can be multiple errors, but displaying only the first error is consistent with the previous behavior. Plus, VS Code will handle displaying all errors in the "Problems" tab when you open up the file, like this:
Screen Shot 2020-10-05 at 3 40 12 PM

Issue #2176, opened 2020-06-22
Support initializing custom handler apps


anthonychu:
Related #2175 

Custom handler apps have:

* `FUNCTIONS_WORKER_RUNTIME` set to `Custom`
* a `customHandler` section in host.json
* should launch the same way as any other runtime stack (`func start`).
    - some modifications to the startup command may be needed but that's up to the customer to do that ([example](https://github.com/anthonychu/azure-functions-deno-template/blob/66adfe7a17c675396821367b9867a0e12c02f3a2/.vscode/tasks.json))

See https://github.com/Azure/azure-functions-core-tools/issues/2052 for host.json

* Add `Custom Handler` to *Select a language* prompt
* We're adding templates to bundles https://github.com/Azure/azure-functions-templates/issues/955
    - templates will only include a *function.json* and no code file

anthonychu:
@pragnagopa @yojagad

anthonychu:
@pragnagopa @EricJizbaMSFT Please review details.

pragnagopa:
cc @soninaren
ACK. Will update the issue once we have bundles ready to test.

PR #2440, opened 2020-10-06
Support creating custom handler projects

EricJizbaMSFT:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2176

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/CustomProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/CustomProjectCreateStep.ts
@@ -1,0 +1,21 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { IHostJsonV2 } from '../../../funcConfig/host';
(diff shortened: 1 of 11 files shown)


src/commands/createNewProject/ProjectCreateStep/CustomProjectCreateStep.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export class CustomProjectCreateStep extends ScriptProjectCreateStep {
+    protected async getHostContent(): Promise<IHostJsonV2> {
+        const hostJson: IHostJsonV2 = await super.getHostContent();
+        hostJson.customHandler = {
+            description: {
+                defaultExecutablePath: '',
+                workingDirectory: '',
+                arguments: []
+            }
```
wwlorey, line 17: Just verifying that `"enableForwardingHttpRequest": false` shouldn't be included here. I noticed that the example from https://github.com/Azure/azure-functions-core-tools/issues/2052#issue-642663971 includes it but looks like it may have been removed during PR review

wwlorey (approved):

src/commands/createNewProject/ProjectCreateStep/CustomProjectCreateStep.ts
EricJizbaMSFT: Yeah it was taken out in PR. Sounds like it already defaults to false if not specified - they were just discussing if it should be there for the sake of discoverability

Issue #2460, opened 2020-10-12
Disconnect the debugger automatically when debugging the PowerShell project


v-xinda:
**OS:** Windows 10
**Build:** 20201011.1

**Repro Steps:**
1.  Create a PowerShell project and open in VS Code.
2.  Set a breakpoint and debug it.
3. Check the result.

**Expect:**
The debugger doesn't disconnect automatically.

**Actual:**
Disconnect the debugger automatically.

**More Info:**
This issue doesn't reproduce for JavaScript/TypeScript/C# project.

EricJizbaMSFT:
Related to https://github.com/Azure/azure-functions-powershell-worker/issues/538

PR #2465, opened 2020-10-14
Add runtime-specific "$func-watch" and fix PowerShell attach

EricJizbaMSFT:
PowerShell doesn't work with the new message "Worker process started and initialized" (added in https://github.com/microsoft/vscode-azurefunctions/pull/2200). We've had other issues in the past where only one language is affected, so I decided to add runtime-specific versions of "func-watch". Since this gets hard-coded in the user's "tasks.json", I want to add one for each runtime now, even though PowerShell is the only one that needs it

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2460


wwlorey (approved):

package.json
```diff
@@ -793,6 +793,36 @@
...
+                "label": "%azureFunctions.problemMatchers.funcNodeWatch%",
+                "base": "$func-watch"
+            },
+            {
+                "name": "func-powershell-watch",
+                "label": "%azureFunctions.problemMatchers.funcPowerShellWatch%",
+                "base": "$func-watch",
+                "background": {
+                    "activeOnStart": true,
+                    "beginsPattern": "^.*(Job host stopped|signaling restart).*$",
```
nturinski, line 818: I didn't see it in a quick search, but Func Core Tools aren't localized, right?

nturinski (approved):

package.json
EricJizbaMSFT: Nah

PR #2478, opened 2020-10-19
Add EventGrid template to C# verified category

EricJizbaMSFT:


test/templateCount.test.ts
```diff
@@ -21,7 +21,7 @@ function addSuite(source: TemplateSource | undefined): void {
             [ProjectLanguage.JavaScript, FuncVersion.v3, 14],
             [ProjectLanguage.CSharp, FuncVersion.v1, 12],
```
nturinski, line 22: Would you not have to change the count of v1 triggers too?

nturinski (approved):

test/templateCount.test.ts
EricJizbaMSFT: It was already in there for v1

PR #2491, opened 2020-10-26
Adding Stencil Python requirements update and better error reporting

vrdmr:
Following changes in the PR:

- Fixed Python's requirements for Stencil generated apps - They require `msrest` requirements for the models generated. That dependency will be added to the `requirments.txt` file for the Python HTTP Trigger functions generated by Stencil.
- Added relevant URLs in the logs and error pop-ups such that users can easily find links to the documentation or for reporting errors when either `autorest` or a specific plugin fails to generate functions.

/cc: @skylerah, @EricJizbaMSFT, @fiveisprime

### DIFF
--- a/src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
+++ b/src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
@@ -69,8 +69,7 @@ export class OpenAPICreateStep extends AzureWizardExecuteStep<IFunctionWizardCon
             try {
                 await cpUtils.executeCommand(ext.outputChannel, undefined, 'autorest', ...args);
             } catch (error) {
-                const message: string = localize('autorestNotRunSuccessfully',
-                    `Failed to run "autorest.${pluginRunPrefix}-${pluginRun}" | Click "Report an Issue" to report the error.`);
+                const message: string = localize('autorestNotRunSuccessfully', `Failed to run "autorest.${pluginRunPrefix}-${pluginRun}" | Click "Report an Issue" to report the error.`);
                 if (!wizardContext.errorHandling.suppressDisplay) {
                     window.showErrorMessage(message, DialogResponses.reportAnIssue).then(async result => {
                         if (result === DialogResponses.reportAnIssue) {
--- a/src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
+++ b/src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
@@ -3,14 +3,15 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as fse from 'fs-extra';
 import * as path from 'path';
 import { ProgressLocation, Uri, window } from "vscode";
 import { AzureWizardExecuteStep, DialogResponses, IActionContext } from 'vscode-azureextensionui';
-import { ProjectLanguage } from "../../../constants";
+import { ProjectLanguage, requirementsFileName } from "../../../constants";
 import { ext } from "../../../extensionVariables";
 import { localize } from "../../../localize";
 import { cpUtils } from "../../../utils/cpUtils";
-import { confirmEditJsonFile } from '../../../utils/fs';
+import { confirmEditJsonFile, confirmOverwriteFile } from '../../../utils/fs';
 import { nonNullProp } from '../../../utils/nonNull';
 import { openUrl } from '../../../utils/openUrl';
 import { IJavaProjectWizardContext } from '../../createNewProject/javaSteps/IJavaProjectWizardContext';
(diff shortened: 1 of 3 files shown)


src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -29,46 +30,68 @@ export class OpenAPICreateStep extends AzureWizardExecuteStep<IFunctionWizardCon
...
-        const title: string = localize('generatingFunctions', 'Generating from OpenAPI...Check [output window](command:{0}) for status.', ext.prefix + '.showOutputChannel');
+        ext.outputChannel.appendLog(localize('urlsForIssues', 'If there are any issues, please visit https://aka.ms/stencil for usage information or report issues at https://aka.ms/stencil/issues.'));
+        const title: string = localize('generatingFunctions', 'Generating from OpenAPI Specification...Check [output window](command:{0}) for status.', ext.prefix + '.showOutputChannel');
 
         await window.withProgress({ location: ProgressLocation.Notification, title }, async () => {
-            await cpUtils.executeCommand(ext.outputChannel, undefined, 'autorest', ...args);
+            try {
+                await cpUtils.executeCommand(ext.outputChannel, undefined, 'autorest', ...args);
+            } catch (error) {
+                const message: string = localize('autorestNotRunSuccessfully', `Failed to run "autorest.${pluginRunPrefix}-${pluginRun}" | Click "Report an Issue" to report the error.`);
```
EricJizbaMSFT, line 72: Please remove `| Click "Report an Issue" to report the error.`. We try to keep messages short, and it's pretty redundant with the button itself.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -109,3 +132,33 @@ async function addAutorestSpecificTypescriptDependencies(context: IFunctionWizar
...
+
+    // Some custom requirements might not have newline character added already.
+    const dependenciesToAdd: string = `\n${dependenciesToAddArray.join('\n')}\n`;
+
+    if (await fse.pathExists(requirementsPath)) {
+        try {
+            oldRequirements = (await fse.readFile(requirementsPath)).toString();
+            newRequirements = oldRequirements.concat(dependenciesToAdd);
+        } catch (error) {
+            // If we failed to parse or edit the existing file, just ask to overwrite the file completely
```
EricJizbaMSFT, line 153: I think you copied this from `confirmEditJsonFile`, but it doesn't really make sense in this case. We're not parsing or editing JSON, so there's way fewer chances for problems. The only real error could happen from reading the file, but if that fails we probably wouldn't be able to write to it either. I would get rid of the whole catch block.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -109,3 +132,33 @@ async function addAutorestSpecificTypescriptDependencies(context: IFunctionWizar
...
+    // Current list of dependencies to be added to support stencil generated projects.
+    const dependenciesToAddArray: string[] = ['msrest'];
+
+    // Some custom requirements might not have newline character added already.
+    const dependenciesToAdd: string = `\n${dependenciesToAddArray.join('\n')}\n`;
+
+    if (await fse.pathExists(requirementsPath)) {
+        try {
+            oldRequirements = (await fse.readFile(requirementsPath)).toString();
+            newRequirements = oldRequirements.concat(dependenciesToAdd);
```
EricJizbaMSFT, line 151: This should check if `oldRequirements` already has 'msrest' before trying to add it

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -109,3 +132,33 @@ async function addAutorestSpecificTypescriptDependencies(context: IFunctionWizar
...
+
+    let oldRequirements: string = '';
+    let newRequirements: string = '';
+    const requirementsPath: string = path.join(context.projectPath, requirementsFileName);
+
+    // Current list of dependencies to be added to support stencil generated projects.
+    const dependenciesToAddArray: string[] = ['msrest'];
+
+    // Some custom requirements might not have newline character added already.
+    const dependenciesToAdd: string = `\n${dependenciesToAddArray.join('\n')}\n`;
```
EricJizbaMSFT, line 146: Use `os.EOL` instead of `\n`

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: Done. Made it better for future usage as well, to make I also take care of a deleted requirements.txt - making this self-sufficient.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
vrdmr: Done. Just added a warning log to make sure the user is aware of the issue and takes action.

src/commands/createFunction/openAPISteps/OpenAPICreateStep.ts
```diff
@@ -3,109 +3,111 @@
...
+            default:
+                break;
         }
     }
 
     public shouldExecute(): boolean {
         return true;
     }
 }
-
```
EricJizbaMSFT, line 79: Seems like you did some decently large refactoring and feature changes unrelated to my comments. In the future, please refrain from changes like that while a PR is in progress. Multiple smaller PRs will get reviewed/approved/merged much faster than one larger/evolving PR.

ejizba:
Closing due to lack of activity

Issue #2494, opened 2020-10-28
[CustomHandler]Remove preview flag 


pragnagopa:
cc @anthonychu

Stacks updates here: pragnagopa/azure-functions-supported-runtime-stacks#21

Issue #2482, opened 2020-10-19
Add Python 3.9 preview


stefanushinardi:
Python 3.9 preview's ETA is Mid November 2020

Issue #2385, opened 2020-09-15
Add Node 14


anthonychu:
* Allow creating an app in Azure with Node 14
* Publish an app to Azure with Node 14

Please use the latest version of Core Tools that should unblock Node 14: https://github.com/Azure/azure-functions-core-tools/releases/tag/3.0.2996

 /cc @mhoeger

fiveisprime:
This should be available on the platform soon and we'll have a plan within the next week on when this should show up as a preview language in the extension.

anthonychu:
@fiveisprime Can we enable this during the week of Nov 2?

anthonychu:
Available in Windows and Linux now. This is unblocked and ready to enable. Is it possible to get this out by Nov 9? @fiveisprime

anthonychu:
@fiveisprime @EricJizbaMSFT Updated initial comment with more details. Thanks!

PR #2513, opened 2020-11-03
Use new api to display runtime stacks

EricJizbaMSFT:
The main point of this PR is to replace the hard-coded stacks in `functionStacks.ts` with a dynamic API call. The schema is a bit different, but the functionality is pretty much the exact same.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2325
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2385
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2482
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2494


src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -0,0 +1,39 @@
...
+export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWizardContext> {
+    public async prompt(context: IFunctionAppWizardContext): Promise<void> {
+        const placeHolder: string = localize('selectRuntimeStack', 'Select a runtime stack.');
+        context.newSiteStack = (await ext.ui.showQuickPick(getStackPicks(context), { placeHolder })).data;
+
+        if (!context.newSiteStack.minorVersion.stackSettings.linuxRuntimeSettings) {
+            context.newSiteOS = WebsiteOS.windows;
+        } else if (!context.newSiteStack.minorVersion.stackSettings.windowsRuntimeSettings) {
+            context.newSiteOS = WebsiteOS.linux;
+        } else if (!context.advancedCreation) {
```
nturinski, line 22: Shouldn't you default to the `preferredOs` first and if it's advanced creation, then check the `minorVersion.stackSettings`?

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    public shouldPrompt(context: IFunctionAppWizardContext): boolean {
+        return !context.newSiteStack;
+    }
+
+    public async getSubWizard(context: IFunctionAppWizardContext): Promise<IWizardOptions<IFunctionAppWizardContext> | undefined> {
+        if (context.newSiteOS === undefined) {
+            return { promptSteps: [new SiteOSStep()] };
+        } else {
+            await setLocationsTask(context);
```
nturinski, line 35: This seems like a behavioral change.  Before `setLocationsTask` was always called in SubscriptionTreeItem if  `!advancedCreation` but `newSiteOS` can be set in either scenario, right?

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
EricJizbaMSFT: Realistically the order here won't make a difference. The point of the above checks is to see if that stack even supports both OS's - if it only supports one OS, we just use that OS. It should be impossible for the "preferredOS" to be an OS that the stack doesn't support

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
EricJizbaMSFT: It's also set in `SiteOSStep` [here](https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/createAppService/SiteOSStep.ts#L21). In the past we _always_ prompted for OS during advanced create, but now we don't. Thus I have to `setLocationsTask` here to maintain existing behavior.

EricJizbaMSFT: Err actually that was true for App Service - Functions is a bit different. It'll overwrite the locationsTask after the "HostPlan" step [here](https://github.com/microsoft/vscode-azurefunctions/blob/b6d5cd3a140338de5351fdeefafd33ef65af1213/src/commands/createFunctionApp/FunctionAppHostingPlanStep.ts#L21) (which happens after we prompt for runtime/OS).

So setting it here isn't technically necessary, but it's consistent with App Service and doesn't really hurt anything.

nturinski (approved):

PR #2530, opened 2020-11-10
Prep for release

EricJizbaMSFT:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 1.0.1 - 2020-11-11
+
+### Added
+- Dynamically retrieve the runtime stacks available in Azure so that the extension supports them as soon as possible, including upcoming preview support for Node 14
```
nturinski, line 6: This kind of feels like an engineering/implementation detail that users may not necessarily care about.  Can we not just say added preview support for Node 14?

nturinski (approved):

CHANGELOG.md
EricJizbaMSFT: Mmm I agree this kind of stuff is lower priority, but I wouldn't say it's worthless to users. It's low enough priority that I didn't put it in the App Service changelog, but worth it in this case because it's the whole point of the Functions release. They might not care how we get stacks, but I think they care that new stacks show up faster

> Can we not just say added preview support for Node 14?

I don't know exactly when Node 14 will happen and I don't want to try to time our release with that

PR #2542, opened 2020-11-20
Leverage "Report Issue" command

ejizba:
And suppress "Report an Issue" button

Depends on https://github.com/microsoft/vscode-azuretools/pull/819

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,7 +8,7 @@
 import { WebSiteManagementMappers } from '@azure/arm-appservice';
 import * as vscode from 'vscode';
 import { registerAppServiceExtensionVariables } from 'vscode-azureappservice';
-import { AzExtTreeDataProvider, AzureUserInput, callWithTelemetryAndErrorHandling, createApiProvider, createAzExtOutputChannel, IActionContext, registerEvent, registerUIExtensionVariables } from 'vscode-azureextensionui';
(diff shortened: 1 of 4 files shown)


src/extension.ts
```diff
@@ -67,6 +67,10 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
         ext.templateProvider = new CentralTemplateProvider();
 
+        // Suppress "Report an Issue" button for all errors in favor of the command
+        registerErrorHandler(c => c.errorHandling.suppressReportIssue = true);
```
nturinski, line 71: I think I know the answer, but wanna consider maybe doing A/B testing here to see the number of errors reported as well as errors with filled out repro steps?

nturinski (approved):

src/extension.ts
ejizba: Sure we can do A/B testing - I'll do "A" and you do "B" in Static Web Apps!

For real, though, as the person that [originally](https://github.com/microsoft/vscode-azuretools/issues/129) [made](https://github.com/microsoft/vscode-azuretools/pull/136) the button, I think I always viewed it as a temporary measure while we had low usage and while the extension was in a rough/early state. Errors happen a lot in software, and our button incorrectly implies that the error needs to be reported and thus may not be solve-able by the user themselves. Show me some other products that pop up a big ol' button for every error and I may reconsider, but I don't think those products exist.

PR #2567, opened 2020-12-04
VS Code Functions Extension changes for one click local development experience

shimedh:
This adds the ability to open VS Code, download and unzip the associated function app project including its dev container, and the .vscode metadata for the function app project. For this to happen, the VS Code request uri handler has to be activated. To activate this, you would need to use the vscode protocol from the browser with a URL such as this - vscode://ms-azuretools.vscode-azurefunctions/?resourceId=<appResourceId>&defaultHostName=<appHostName>&devcontainer=<devContainerName>&language=<appLanguage>&downloadAppContent=<true/false>

If the vscode request uri handler is not activated, the extension will continue to operate as is.


shimedh:
Sending a new PR after fixing conflicts

PR #2568, opened 2020-12-04
VS Code Functions Extension changes for one click local development experience

shimedh:
This adds the ability to open VS Code, download and unzip the associated function app project including its dev container, and the .vscode metadata for the function app project. For this to happen, the VS Code request uri handler has to be activated. To activate this, you would need to use the vscode protocol from the browser with a URL such as this - vscode://ms-azuretools.vscode-azurefunctions/?resourceId=&defaultHostName=&devcontainer=&language=&downloadAppContent=<true/false>

If the vscode request uri handler is not activated, the extension will continue to operate as is.


shimedh:
@ejizba Could you please help review these changes? Also could you help me understand the test failure in my PR build. They seem unrelated to my changes. Also when I go to the build pipeline's Test task I see the below message:
image

ejizba:
You can ignore the test failures - they are unrelated to your changes.

shimedh:
> You can ignore the test failures - they are unrelated to your changes.

@ejizba Thanks for the note. The PR is ready for review. Could you please help review when you get some time?

src/utils/requestUtils.ts
```diff
@@ -48,4 +50,43 @@ export namespace requestUtils {
             stream.pipe(fse.createWriteStream(filePath).on('finish', resolve).on('error', reject));
         });
     }
+
+    // tslint:disable-next-line:no-any
+    export async function getFunctionAppMasterKey(azureSession: AzureSession, resourceId: string, accessToken: any): Promise<HttpOperationResponse> {
+        const subscriptionId: string = getSubscriptionFromId(resourceId, false);
+        const subscriptionContext: ISubscriptionContext = {
```
ejizba, line 57: You're doing a lot of stuff manually here that doesn't need to be done that way. For example, use `ext.tree.findTreeItem` to get a `SlotTreeItemBase` matching `resourceId ` and then use `SlotTreeItemBase.client.listHostKeys`

shimedh: @ejizba I see. In our case we are not in a local project yet, but setting it up. Does the ext.tree.findTreeItem work even before the local project is setup?

src/extension.ts
```diff
@@ -107,3 +144,133 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 // tslint:disable-next-line:no-empty
 export function deactivateInternal(): void {
 }
+
```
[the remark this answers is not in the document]
shimedh: @ejizba  Sure, are you suggesting moving all the code for our feature into a separate utility file and calling it from here?

src/extension.ts
```diff
@@ -107,3 +144,133 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+    }
+}
+
+// tslint:disable-next-line:no-any
+async function setupAzureAccount(account: AzureAccount): Promise<any> {
+    try {
+        await vscode.commands.executeCommand('azure-account.login');
+        return await account.sessions[0].credentials2.getToken();
+    } catch (err) {
+        vscode.window.showErrorMessage(localize('failedAzureAccSetupErrorMessage', 'Failed to setup Azure account. Please try again.'));
```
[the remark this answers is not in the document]
shimedh: @ejizba I am keeping this and the other comment open for now as I have clarifying questions. Since I am not able to use ext.tree.findTreeItem  as mentioned below, I need the account and token information for my current setup.

shimedh:
@ejizba I have updated the PR by resolving most of the comments. I have follow up questions on two comments which I have kept unresolved. Could you please take a look?

src/utils/requestUtils.ts
ejizba: We basically center our entire extension around the tree, so yes you should be able to get `findTreeItem` to work. In this case you're trying to get their function app node, which is an Azure resource that should be there as long as they're signed in, regardless of anything related to the local project. Could be a problem with your resourceId 🤷‍♂️  I would suggest using the debugger or looking at other examples in the codebase.

src/extension.ts
```diff
@@ -59,7 +78,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         const validateEventId: string = 'azureFunctions.validateFunctionProjects';
         // tslint:disable-next-line:no-floating-promises
         callWithTelemetryAndErrorHandling(validateEventId, async (actionContext: IActionContext) => {
-            await verifyVSCodeConfigOnActivate(actionContext, vscode.workspace.workspaceFolders);
+            if (ext.context.globalState.get(GlobalStates.initProjectWithoutConfigVerification) === true) {
```
ejizba, line 81: Just call `initProjectForVSCode` before you reload the window and you shouldn't have to mess with the globalState

src/utils/localDevelopmentUtils.ts
```diff
@@ -0,0 +1,152 @@
...
+
+interface ILocalDevelopmentRequiredInputs {
+    resourceId: string | null;
+    defaultHostName: string | null;
+    devContainerName: string | null;
+    language: string | null;
+    downloadAppContent: string | null;
+}
+
+export namespace localDevelopmentUtils {
```
ejizba, line 22: "localDevelopment" could describe over half the features in our extension and "utils" is a last resort folder for small utility functions that can't be placed anywhere better because they're used by multiple features. I would appreciate if you took some time familiarizing yourself with our file structure and made an attempt at a better location/name. If you're still struggling to find a place after you've addressed the other comments, I'll come up with a suggestion at that time

src/utils/requestUtils.ts
shimedh: @ejizba In our case, we are passing the correct resourceId and the context as below based on other examples I saw in the code:
const slotTreeItem: SlotTreeItemBase | undefined = await ext.tree.findTreeItem(resourceId, { ...actionContext, loadAll: true });

However, stepping through and debugging the findTreeItem function seems to get only subscription level cachedChildren on line 206 of F:\repos\vscode-azurefunctions\node_modules\vscode-azureextensionui\out\src\tree\AzExtTreeDataProvider.js
It does not get any resourceId specific cachedChildren and hence always returns undefined.

It almost seems like the tree is not fully loaded in our specific case.

ejizba: Just push up what you have so far and I can try it out

src/extension.ts
shimedh: verifyVSCodeConfigOnActivate is existing code for all other scenarios. Even if I change to always call initProjectForVSCode , we would still need the global state to determine whether to pass in the filePath and language so that users don't need to do the selection again for our scenario as we already know the language from the uri.

If there is any confusion here, it would be better to discuss it over a call to expedite this process.

src/utils/localDevelopmentUtils.ts
shimedh: I have moved this around. If there is still a concern, it would be better to discuss it over a call to expedite this process.

src/utils/requestUtils.ts
shimedh: I am adding the code to findTreeItem for you to try it out. However, I am keeping my existing code as well for now so it does not break our entire scenario. I can clean it up once findTreeItem is working. Also another thing to note is that this section of code is executed before VSCode reloads and initProjectForVSCode gets called.

In order to trigger our scenario you will need to use the below:
vscode://ms-azuretools.vscode-azurefunctions/?resourceId=<appResourceId>&defaultHostName=<appHostName>&devcontainer=<devContainerName>&language=<appLanguage>&downloadAppContent=<true/false>

ejizba: As suspected, it is a problem with the resourceId. [This line](https://github.com/microsoft/vscode-azurefunctions/pull/2568/files#diff-bec5299b320724002801862064e523e95e27d75971447e824195446c0a658b11R108) is converting it to lowercase, but our tree uses normal casing as given to us by Azure APIs and as seen in the portal (i.e. `resourceGroups` instead of `resourcegroups`). If I keep the casing correct, `slotTreeItem` is defined for me.

shimedh: I can update the code to pass in the resourceId before we convert it. However, based on ARM's standards all resourceId's are treated case insensitive as resourceId's are unique irrespective of casing - https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules.
It is always good engineering practice to make unique ids case insensitive.

Currently maybe there is no scenario that gets affected by this in VSCode, but, in future if there is a scenario of passing in resourceId from an external client and the resourceId casings are changed on the client side, then VSCode side will break.

I have created an issue for tracking this - https://github.com/microsoft/vscode-azurefunctions/issues/2617

src/extension.ts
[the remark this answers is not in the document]
shimedh: Tried using ext.azureAccountTreeItem.getIsLoggedIn() but it only tells if I am logged in or not. How can I automatically prompt the login without using the logic around AzureAccount extension in case user is not logged in?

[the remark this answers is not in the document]
shimedh: Updated

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,122 @@
...
+            const projectFilePath: string = projectFilePathUri.fsPath;
+            const devContainerFolderPathUri: vscode.Uri = vscode.Uri.joinPath(projectFilePathUri, '.devcontainer');
+            ext.context.globalState.update(GlobalStates.projectFilePath, projectFilePathUri.fsPath);
+            await callWithTelemetryAndErrorHandling('azureFunctions.extractContentAndDownloadDevContainerFiles', async (_actionContext: IActionContext) => {
+                if (downloadAppContent === 'true') {
+                    // tslint:disable-next-line: no-unsafe-any
+                    await extract(downloadFilePath, { dir: projectFilePath });
+                }
+                await requestUtils.downloadFile(
+                    `https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/containers/${devContainerName}/.devcontainer/devcontainer.json`,
```
ejizba, line 61: I don't think we can rely on this URL to work going forward. For example, maybe they change the default branch from "master" to "main" (a common thing happening lately) or maybe they change the file structure.

An easy way to fix that would be to just get rid of dev containers stuff. @fiveisprime what do you think - is dev containers worth it for this? A few reasons against it IMO:
1. If the point of using the dev container is to reduce the friction of installing dev tools on my local machine, I'm not sure this does that. For example, if I have a JavaScript or C# project, I probably already have Node.js or .NET on my machine, but it seems less likely that I have Docker. The only remaining tool to install is the func cli, but we already have some logic to help users with that (and we want to keep improving that logic) and it's only necessary if you want to debug 
1. I don't think our users typically use dev containers for local development (might be worth checking telemetry). I'm also pretty sure the dev containers extension is the least used Remote extension (compared to SSH and WSL)

src/utils/requestUtils.ts
```diff
@@ -48,4 +50,27 @@ export namespace requestUtils {
             stream.pipe(fse.createWriteStream(filePath).on('finish', resolve).on('error', reject));
         });
     }
+
+    // tslint:disable-next-line:no-any
+    export async function getFunctionAppMasterKey(resourceId: string, actionContext: IActionContext): Promise<WebSiteManagementModels.HostKeys | undefined> {
```
ejizba, line 55: This function isn't really related to `requestUtils` anymore. Just use the two lines directly instead of creating a function here

src/utils/requestUtils.ts
```diff
@@ -48,4 +50,27 @@ export namespace requestUtils {
...
+        const slotTreeItem: SlotTreeItemBase | undefined = await ext.tree.findTreeItem(resourceId, { ...actionContext, loadAll: true });
+        return await slotTreeItem?.client.listHostKeys();
+    }
+
+    export async function downloadFunctionAppContent(defaultHostName: string, filePath: string, masterKey: string): Promise<void> {
```
ejizba, line 60: No need to copy a bunch of code here. I'd suggest you tweak `downloadFile` like this and then you can just pass in the request options:
```typescript
    export async function downloadFile(requestOptionsOrUrl: string | RequestPrepareOptions, filePath: string): Promise<void> {
        await fse.ensureDir(path.dirname(filePath));
        const request: WebResource = new WebResource();
        request.prepare(typeof requestOptionsOrUrl === 'string' ? { method: 'GET', url: requestOptionsOrUrl } : requestOptionsOrUrl);
```

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,122 @@
...
+    defaultHostName: string | null;
+    devContainerName: string | null;
+    language: string | null;
+    downloadAppContent: string | null;
+}
+
+export async function setupProjectFolder(uri: vscode.Uri, filePath: string): Promise<void> {
+    const requiredInputs: ILocalDevelopmentRequiredInputs = getAllRequiredInputs(uri.query);
+    const resourceId: string | null = requiredInputs.resourceId;
+    const defaultHostName: string | null = requiredInputs.defaultHostName;
```
ejizba, line 24: You shouldn't need `defaultHostName` as long as we have `resourceId`. You can get it from the node.
```
slotTreeItem.client.defaultHostUrl
```

src/downloadAzureProject/setupProjectFolder.ts
shimedh: @ejizba 
We want to use devContainers for this feature so that customers don't need to install each language specific libraries on their machine.

Also, this feature is a two step process and the first step would be an msi which will install all the requirements like Docker, etc on the customer's machine before they hit this flow. This was mentioned in our initial demo meeting with Matt and you.

For the URL, we are working with Functions team to get an API which would return the devContainer but, till then we would be using this method to unblock the feature.

We can discuss this over a call if you like.

src/extension.ts
```diff
@@ -40,6 +49,13 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
 
     registerUIExtensionVariables(ext);
     registerAppServiceExtensionVariables(ext);
+    // NOTE: Example call for opening vscode with query parameters -
+    // vscode://ms-azuretools.vscode-azurefunctions/?resourceId=<appResourceId>&defaultHostName=<appHostName>&devcontainer=<devContainerName>&language=<appLanguage>&downloadAppContent=<true/false>
+    vscode.window.registerUriHandler({
+        async handleUri(uri: vscode.Uri): Promise<void> {
```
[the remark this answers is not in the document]
shimedh: I prefer to keep the name to be more verbose. I will change it to the below instead:
vscode.window.registerUriHandler({
        handleUri: downloadAzureProjectFromUri
    });

src/utils/azure.ts
```diff
@@ -11,7 +11,8 @@ import { createStorageClient } from './azureClients';
 import { nonNullProp, nonNullValue } from './nonNull';
 
 function parseResourceId(id: string): RegExpMatchArray {
```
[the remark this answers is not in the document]
shimedh: Sure will revert these changes. I have also created the below issue for tracking the case insensitive check with azure resourceId's - https://github.com/microsoft/vscode-azurefunctions/issues/2617

src/extension.ts
shimedh: Not sure I quite follow this one. When I revert this code from extension.ts and call initProjectForVSCode at the end of 
setupProjectFolder try block, I am getting the below message when I run the code.

image


With my current code with the global states I see the following which is initializing it fine:
image

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,115 @@
...
+    const downloadAppContent: string = getRequiredQueryParameter(parsedQuery, 'downloadAppContent');
+
+    ext.context.globalState.update(GlobalStates.projectLanguage, getProjectLanguageForLanguage(language));
+    const toBeDeletedFolderPathUri: vscode.Uri = vscode.Uri.joinPath(vsCodeFilePathUri, 'temp');
+
+    try {
+        const functionAppName: string = getNameFromId(resourceId);
+        const downloadFilePath: string = vscode.Uri.joinPath(toBeDeletedFolderPathUri, `${functionAppName}.zip`).fsPath;
+
+        vscode.window.showInformationMessage(localize('settingUpFunctionAppLocalProjInfoMessage', `Setting up project for function app '${functionAppName}' with language '${language}'.`));
```
ejizba, line 34: This would work better than `showInformationMessage`:

`vscode.window.withProgress({ location: vscode.ProgressLocation.Notification`

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,115 @@
...
+                `https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/containers/${devContainerName}/.devcontainer/devcontainer.json`,
+                vscode.Uri.joinPath(devContainerFolderPathUri, 'devcontainer.json').fsPath
+            );
+            await requestUtils.downloadFile(
+                `https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/containers/${devContainerName}/.devcontainer/Dockerfile`,
+                vscode.Uri.joinPath(devContainerFolderPathUri, 'Dockerfile').fsPath
+            );
+        });
+        vscode.window.showInformationMessage(localize('restartingVsCodeInfoMessage', 'Restarting VS Code with your function app project'));
+        vscode.commands.executeCommand('vscode.openFolder', vscode.Uri.file(projectFilePath));
```
ejizba, line 73: Feels like opening a new window would be better than the existing window, since I don't think the user gets to pick which window handles the uri. Just pass `true` for the third parameter (`forceNewWindow`)

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,115 @@
...
+                    await requestUtils.downloadFile(requestOptions, downloadFilePath);
+                }
+            });
+        }
+
+        const projectFilePathUri: vscode.Uri = vscode.Uri.joinPath(vsCodeFilePathUri, `${functionAppName}`);
+        const projectFilePath: string = projectFilePathUri.fsPath;
+        const devContainerFolderPathUri: vscode.Uri = vscode.Uri.joinPath(projectFilePathUri, '.devcontainer');
+        ext.context.globalState.update(GlobalStates.projectFilePath, projectFilePathUri.fsPath);
+        await callWithTelemetryAndErrorHandling('azureFunctions.extractContentAndDownloadDevContainerFiles', async (_actionContext: IActionContext) => {
```
ejizba, line 58: > We want to use devContainers for this feature so that customers don't need to install each language specific libraries on their machine.

I'm just skeptical people want Docker on their machine, but I'm pretty confident a JavaScript developer would want Node.js.

> Also, this feature is a two step process and the first step would be an msi which will install all the requirements like Docker, etc on the customer's machine before they hit this flow. 

Does the msi already exist? What about mac/linux? 

> This was mentioned in our initial demo meeting with Matt and you.

Fair enough, but a quick demo meeting does not mean an experience has been finalized.

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,115 @@
...
+    const toBeDeletedFolderPathUri: vscode.Uri = vscode.Uri.joinPath(vsCodeFilePathUri, 'temp');
+
+    try {
+        const functionAppName: string = getNameFromId(resourceId);
+        const downloadFilePath: string = vscode.Uri.joinPath(toBeDeletedFolderPathUri, `${functionAppName}.zip`).fsPath;
+
+        vscode.window.showInformationMessage(localize('settingUpFunctionAppLocalProjInfoMessage', `Setting up project for function app '${functionAppName}' with language '${language}'.`));
+
+        if (downloadAppContent === 'true') {
+            // NOTE: We don't want to download app content for compiled languages.
```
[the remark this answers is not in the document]
shimedh: When customers create a compiled language functionApp from Azure Portal, we don't allow then to create or manage any functions from the Portal and point them to using VSCode instead. This will help those customers to setup their local projects with all necessary packages and devcontainer and they can easily add functions and then publish to azure from VSCode. We will have all this documented on our end on the Azure Portal as well as any documentations needed.

src/extension.ts
```diff
@@ -93,6 +108,10 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         WebSiteManagementMappers.SiteConfig.type.modelProperties!.powerShellVersion = { serializedName: 'powerShellVersion', type: { name: 'String' } };
         // tslint:disable-next-line: no-non-null-assertion
         WebSiteManagementMappers.SiteConfigResource.type.modelProperties!.powerShellVersion = { serializedName: 'properties.powerShellVersion', type: { name: 'String' } };
+
+        context.subscriptions.push(vscode.window.registerUriHandler({
+            handleUri: downloadAzureProjectFromUri
```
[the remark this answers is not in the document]
shimedh: In that case I can call it handleUri as you previously suggested. I can use the action parameter in the query with either of the two values for our scenario: downloadContentAndSetupProject OR setupProject. I will also get rid of "downloadAppContent".

We need the two values based on my other comment regarding "downloadAppContent" parameter.

src/downloadAzureProject/setupProjectFolder.ts
shimedh: We will be working on the msi on our end and will not enable this feature from our end till we have that ready. We will also have a corresponding file needed for mac/linux as this is our complete experience that we need to ship.

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Updating

src/extension.ts
shimedh: Seems like updating to force it to open in a new window based on your other comment fixed this as well. I will update the code and remove the unneeded global states

src/downloadAzureProject/setupProjectFolder.ts
shimedh: @ejizba  and @fiveisprime Do you want me to setup a call to talk about this and some other questions related to our feature that came up in this PR?

src/downloadAzureProject/handleUri.ts
```diff
@@ -0,0 +1,45 @@
...
+// vscode://ms-azuretools.vscode-azurefunctions/?resourceId=<appResourceId>&defaultHostName=<appHostName>&devcontainer=<devContainerName>&language=<appLanguage>&action=<'downloadContentAndSetupProject' OR 'setupProject'>
+export async function handleUri(uri: vscode.Uri): Promise<void> {
+    await callWithTelemetryAndErrorHandling('azureFunctions.handleUri', async (context: IActionContext) => {
+        const isLoggedIn: boolean = await ext.azureAccountTreeItem.getIsLoggedIn();
+        if (!isLoggedIn) {
+            await vscode.commands.executeCommand('azure-account.login');
+        }
+
+        const parsedQuery: querystring.ParsedUrlQuery = querystring.parse(uri.query);
+        const action: string = getRequiredQueryParameter(parsedQuery, 'action');
```
ejizba, line 28: This should be fast, so check this before you do `getIsLoggedIn` and `login` above. No need to make the user login if the uri isn't gonna work anyways

src/downloadAzureProject/handleUri.ts
```diff
@@ -0,0 +1,45 @@
...
+
+        if (action === HandleUriActions.downloadContentAndSetupProject || action === HandleUriActions.setupProject) {
+            const filePathUri: vscode.Uri[] = await ext.ui.showOpenDialog({ canSelectFolders: true, canSelectFiles: false, canSelectMany: false });
+            await setupProjectFolder(uri, filePathUri[0], context);
+        }
```
ejizba, line 33: Throw an error here like `throw new RangeError(localize('invalidAction', 'Invalid action "{0}".', action));` in an else case

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,101 @@
...
+                const defaultHostName: string | undefined = slotTreeItem?.client.defaultHostName;
+
+                if (!!hostKeys && hostKeys.masterKey && defaultHostName) {
+                    const requestOptions: RequestPrepareOptions = {
+                        method: 'GET',
+                        url: `https://${defaultHostName}/admin/functions/download?includeCsproj=true&includeAppSettings=true`,
+                        headers: { 'x-functions-key':  hostKeys.masterKey }
+                    };
+                    await requestUtils.downloadFile(requestOptions, downloadFilePath);
+                }
```
ejizba, line 49: This should throw an error if we're failing to get the tree item or host keys rather than silently continuing

src/downloadAzureProject/handleUri.ts
```diff
@@ -0,0 +1,45 @@
...
+    });
+}
+
+export function getRequiredQueryParameter(parsedQuery: querystring.ParsedUrlQuery, key: string): string {
+    const value: string | string[] | undefined = parsedQuery[key];
+    if (value && typeof value === 'string') {
+        return value;
+    } else {
+        vscode.window.showErrorMessage(localize('invalidInputsErrorMessage', 'Invalid inputs. Please try again.'));
```
ejizba, line 42: As mentioned, get rid of `showErrorMessage` or it'll show double messages. Also, I like the "missing query" message better.

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,101 @@
...
+            await requestUtils.downloadFile(
+                `https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/containers/${devContainerName}/.devcontainer/Dockerfile`,
+                vscode.Uri.joinPath(devContainerFolderPathUri, 'Dockerfile').fsPath
+            );
+            await initProjectForVSCode(context, projectFilePath, getProjectLanguageForLanguage(language));
+            vscode.window.showInformationMessage(localize('restartingVsCodeInfoMessage', 'Restarting VS Code with your function app project'));
+            vscode.commands.executeCommand('vscode.openFolder', vscode.Uri.file(projectFilePath), true);
+        });
+    } catch (err) {
+        vscode.window.showErrorMessage(localize('failedLocalProjSetupErrorMessage', 'Failed to set up your local project. Please try again.'));
```
ejizba, line 72: I feel like this generic message isn't very helpful, so I'd prefer to get rid of the catch block and let the original error bubble up. If you want to keep the generic message, please do something [like this](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/templates/script/ScriptBundleTemplateProvider.ts#L72) where you include the original error in the message and use `throw` instead of `showErrorMessage`

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            const projectFilePathUri: vscode.Uri = vscode.Uri.joinPath(vsCodeFilePathUri, `${functionAppName}`);
+            const projectFilePath: string = projectFilePathUri.fsPath;
+            const devContainerFolderPathUri: vscode.Uri = vscode.Uri.joinPath(projectFilePathUri, '.devcontainer');
+            if (downloadAppContent) {
+                // tslint:disable-next-line: no-unsafe-any
+                await extract(downloadFilePath, { dir: projectFilePath });
+            }
+            await requestUtils.downloadFile(
```
ejizba, line 59: Sure you can set up a meeting to discuss the dev container/docker stuff. 

> When customers create a compiled language functionApp from Azure Portal, we don't allow then to create or manage any functions from the Portal and point them to using VSCode instead. This will help those customers to setup their local projects with all necessary packages and devcontainer and they can easily add functions and then publish to azure from VSCode. We will have all this documented on our end on the Azure Portal as well as any documentations needed.

Other than the reasons I've mentioned so far, another reason to avoid dev containers is to keep portal docs consistent with our [getting started tutorial](https://docs.microsoft.com/azure/azure-functions/create-first-function-vs-code-csharp) which has never used them.

src/downloadAzureProject/setupProjectFolder.ts
shimedh: I am keeping the try catch and going with the second option of throwing the error

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Throwing an error in else

src/downloadAzureProject/setupProjectFolder.ts
shimedh: I have setup a meeting for 1/25 to discuss these open docker and dev container related comments.

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Discussed in the meeting

src/downloadAzureProject/setupProjectFolder.ts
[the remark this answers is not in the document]
shimedh: Discussed in the meeting

src/downloadAzureProject/setupProjectFolder.ts
shimedh: As discussed in the meeting, we are not going to enable this feature till we have the msi/pkg piece in.

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Discussed on the meeting

src/downloadAzureProject/setupProjectFolder.ts
shimedh: @ejizba Could you provide pointers on how to add feature flag in our handleUri entry point? As that seems like the last thing remaining to get this in

ejizba: It's just a VS Code setting. Define it in package.json:

https://github.com/microsoft/vscode-azureappservice/blob/57ba1a52dc44a351ace10a079ad548770ea66b03/package.json#L850

And reference it inside handleUri like this:

https://github.com/microsoft/vscode-azureappservice/blob/b5a0bdfd6fc4d0bef8dcb6c8167318355f7f88a7/src/commands/remoteDebug/getRemoteDebugLanguage.ts#L20

src/downloadAzureProject/setupProjectFolder.ts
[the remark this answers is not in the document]
ejizba: It sounded like C#Script was the priority for .NET in the portal, in which case you need to update your logic because we have a specific C#Script [ProjectLanguage](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/constants.ts#L20). Also in that case, you _can_ download the app content. Other than that, I'm a little fuzzy on what we said for C# (aka _not_ script). What are we doing in that case? Sounded like y'all didn't care about that case, so I'd rather not deal with it in VS Code

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Thanks updated!

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,103 @@
...
+                `https://raw.githubusercontent.com/microsoft/vscode-dev-containers/master/containers/${devContainerName}/.devcontainer/devcontainer.json`,
+                vscode.Uri.joinPath(devContainerFolderPathUri, 'devcontainer.json').fsPath
+            );
+            await requestUtils.downloadFile(
```
ejizba, line 65: Just to make sure we're on the same page - I'm still not satisfied with the dev container implementation for reasons we've already discussed. My concerns no longer block this PR because of the feature flag, but will need to be addressed before we can remove the feature flag in the future. Putting aside the implementation, I'm also hoping we can get some user feedback on the design of the dev container piece before the flag is removed to help determine if we need it at all.

package.json
```diff
@@ -1050,7 +1051,12 @@
                         "type": "boolean",
                         "description": "%azureFunctions.showReloadTemplates%",
                         "default": false
-                    }
+                    },
+		      "azureFunctions.enableDownloadContentAndSetupProject": {
```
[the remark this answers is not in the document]
shimedh: sure. will update

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const toBeDeletedFolderPathUri: vscode.Uri = vscode.Uri.joinPath(vsCodeFilePathUri, 'temp');
+
+    try {
+        const functionAppName: string = getNameFromId(resourceId);
+        const downloadFilePath: string = vscode.Uri.joinPath(toBeDeletedFolderPathUri, `${functionAppName}.zip`).fsPath;
+
+        await vscode.window.withProgress({ location: vscode.ProgressLocation.Notification, title: localize('settingUpFunctionAppLocalProjInfoMessage', `Setting up project for function app '${functionAppName}' with language '${language}'.`) }, async () => {
+            if (downloadAppContent) {
+                // NOTE: We don't want to download app content for compiled languages.
```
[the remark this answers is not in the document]
shimedh: Yes, I will update to the C#Script ProjectLanguage. And with that we would be downloading content as well in our flow.
Yup, for C# (aka not script) we don't need to worry about in VS Code for our scenario

src/downloadAzureProject/setupProjectFolder.ts
shimedh: Yes as we discussed in the meeting as well, we are working on getting the API changes to make the devcontainer part better. Also having this deployed behind a feature flag will help us test this end to end once we have our msi/pkg part done and do some user testing as well to get feedback. We will share the feedback with Matt and you.

src/downloadAzureProject/setupProjectFolder.ts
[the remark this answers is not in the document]
ejizba: So doesn't that mean we only need to support "downloadContentAndSetupProject", not "setupProject"?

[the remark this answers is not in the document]
shimedh: We need to support both as we need setupProject for Java as Java is a compiled language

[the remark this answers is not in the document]
ejizba: For our extension, C# is used 8 times more than Java. I would prefer we don't support _any_ compiled language because I don't think it's useful in the context of this feature, but if we're going to do a compiled language it just feels like we should do C# before Java.

Fwiw, our C#Script usage is basically non-existent (C# is used 100 times more than C#Script), so that's debate-ably not worth supporting, either.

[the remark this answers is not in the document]
shimedh: This is all behind a feature flag. We can pivot on ideas as we get feedback on usage from customers internally in the team.

src/downloadAzureProject/setupProjectFolder.ts
```diff
@@ -0,0 +1,103 @@
...
+        );
+    }
+}
+
+function getProjectLanguageForLanguage(language: string): ProjectLanguage {
+    switch (language) {
+        case 'powershell':
+            return ProjectLanguage.PowerShell;
+        case 'node':
+            return ProjectLanguage.TypeScript;
```
ejizba, line 85: Shouldn't this be JavaScript? I didn't think the portal supported TypeScript...

shimedh: Sure will update

ejizba (approved):
Just gotta merge with master and fix the lint issues

shimedh:
> Just gotta merge with master and fix the lint issues

@ejizba I have fixed the linter errors. The build seems to fail on the test step and does not seem related to my changes. Can you help merge this in?

Issue #2550, opened 2020-11-27
.NET 5 - add debugging


anthonychu:
Add the ability to debug.

* Select the correct process when attaching
* Automatically scaffold launch config if .NET 5 worker app is detected (maybe based on having the .NET 5 worker package in csproj)

PR #2579, opened 2020-12-09
Debugging with .NET 5

ejizba:
.NET 5 support for Azure Functions will be in a separate worker process and I ran into two main problems with our "pickProcess" command when trying to get it to work:
1. We would return a process id before the child worker process had even started. My fix here was to replace some existing "waiting" logic with a loop that pings "admin/host/status" to see if the app is running
1. The innermost child process with .NET 5 seemed to be something like "conhost.exe" on Windows, which is not what we want to attach to. Now, I filter to only "dotnet" or "func" processes

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2550

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -3,9 +3,10 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { ServiceClient, WebResource } from '@azure/ms-rest-js';
 import * as unixPsTree from 'ps-tree';
 import * as vscode from 'vscode';
-import { IActionContext, UserCancelledError } from 'vscode-azureextensionui';
+import { createGenericClient, IActionContext, UserCancelledError } from 'vscode-azureextensionui';
 import { hostStartTaskName } from '../constants';
 import { IPreDebugValidateResult, preDebugValidate } from '../debug/validatePreDebug';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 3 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -73,69 +73,87 @@ async function startFuncTask(context: IActionContext, workspaceFolder: vscode.Wo
...
             }
 
             const taskInfo: IRunningFuncTask | undefined = runningFuncTaskMap.get(workspaceFolder);
-            // Ensure func task runs for at least 1 second in case it fails on startup and thus `taskError` will be thrown instead
-            if (taskInfo && Date.now() > taskInfo.startTime + 1000) {
-                return taskInfo.processId.toString();
+            if (taskInfo) {
+                try {
+                    // wait for status url to indicate functions host is running
+                    await client.sendRequest(statusRequest);
```
nturinski, line 89: https://github.com/Azure/azure-functions-core-tools/issues/185#issuecomment-315484240

Looks like you could check the state === 'Running' as well.  Not really sure, but I imagine you could have instances where it returns 200, but is still initializing?

src/commands/pickFuncProcess.ts
```diff
@@ -73,69 +73,87 @@ async function startFuncTask(context: IActionContext, workspaceFolder: vscode.Wo
...
+ * 1. On Windows, the rootPid is almost always the parent PowerShell process
+ * 2. On Unix, the rootPid may be a wrapper around the main func exe if installed with npm
+ * 3. Starting with the .NET 5 worker, Windows sometimes has an inner process we _don't_ want like 'conhost.exe'
+ * The only processes we should want to attach to are the "func" process itself or a "dotnet" process running a dll, so we will pick the innermost one of those
  */
-async function getInnermostUnixPid(pid: string): Promise<string> {
-    return await new Promise<string>((resolve: (pid: string) => void, reject: (e: Error) => void): void => {
-        unixPsTree(parseInt(pid), (error: Error | undefined, children: unixPsTree.PS[]) => {
+async function pickChildProcess(rootPid: string): Promise<string> {
+    const children: OSAgnosticProcess[] = process.platform === 'win32' ? await getWindowsChildren(rootPid) : await getUnixChildren(rootPid);
```
nturinski, line 130: Seems like you were polling the children before with the while loop/delay, but I don't see that behavior anymore.  Is it not necessary?

ejizba: It was replaced by the polling of "admin/host/status" instead

nturinski (approved):

Issue #1401, opened 2019-07-23
VS Code instance freezes after terminating vscode debugging session on javascript HttpTrigger function


SDP190:
I have latest vscode version running on windows 10.
I am able to run Azure Functions locally on VS code using AzureFunctions ext. for VS Code, however when I run the debugger (F5) on my JavaScript Azure Function V2 (HttpTrigger), then when I stop it, very often it causes VS Code instance to crash and freeze.

Any help on this?

Its a standard project created with VS Code.

**VS Code Info (Copied from About window):**
Version: 1.36.1 (system setup)
Commit: 2213894ea0415ee8c85c5eea0d0ff81ecc191529
Date: 2019-07-08T22:59:35.033Z
Electron: 4.2.5
Chrome: 69.0.3497.128
Node.js: 10.11.0
V8: 6.9.427.31-electron.0
OS: Windows_NT x64 10.0.18362

**Node.js Version:**
`node -v  --> v10.15.3`

EricJizbaMSFT:
@SDP190 What language is your project? Does it show any errors?

SDP190:
My language is JavaScript node.js
It does not show any errors!

EricJizbaMSFT:
Is your project a standard project created from VS Code? If not - is it open source and can you share a link?

Also just to confirm - can you copy your exact VS Code version and your Node.js version

SDP190:
@EricJizbaMSFT See original comment updated

askpatrickw:
I experience this intermittently as well with Python.

standard project created with vscode

> VS Code Version: 1.36.1 (user setup)
> Commit: 2213894ea0415ee8c85c5eea0d0ff81ecc191529
> Date: 2019-07-08T22:59:35.033Z
> Electron: 4.2.5
> Chrome: 69.0.3497.128
> Node.js: 10.11.0
> V8: 6.9.427.31-electron.0
> OS: Windows_NT x64 10.0.18362

> (.env) PS: github\project> python --version
> Python 3.6.8

EricJizbaMSFT:
@SDP190 @askpatrickw If you can get concrete repro steps that would be great. I'm having a hard time repro-ing. That being said - I actually submitted a PR a little bit ago that might affect this experience: https://github.com/microsoft/vscode-azurefunctions/pull/1389 (The PR title says C#, but it should affect other languages as well)

[...]

PR #2580, opened 2020-12-09
Change how we end func process after debugging

ejizba:
One of the biggest problems with `funcExecution.terminate` is that it closes the terminal pane, meaning users can't see any errors if "func host start" failed. `process.kill`, on the other hand, will keep it open

I'm also hoping this will help https://github.com/microsoft/vscode-azurefunctions/issues/1401. I briefly got the freeze to repro on my Windows machine, but not consistently enough to be sure I can close the issue. Anyways, users have reported that doing "ctrl+c" in the terminal or hitting the trash can button never froze VS Code, which is what makes me think this might work.


nturinski (approved):
🗑️

PR #2583, opened 2020-12-09
Fix timing issues with pickProcess

ejizba:
Fixed two problems that can happen if you mash the "stop debugging" and "start debugging" buttons:
1. It never stops the previous func task and you get an error like "Failed to stop previous running Functions host"
    1. Solved this by adding `stopFuncTaskIfRunning` to `waitForPrevFuncTaskToStop`. Previously it would only stop it if the "onDidTerminateDebugSession" event was fired
1. It starts a second func task and you get a port conflict error
    1. Solved this by using `executeIfNotActive` (taskUtils copied from [here](https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/utils/taskUtils.ts))

### DIFF
--- a/src/utils/taskUtils.ts
+++ b/src/utils/taskUtils.ts
@@ -3,8 +3,8 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { Task, tasks as codeTasks, TaskScope, WorkspaceFolder } from "vscode";
-import { isPathEqual, isSubpath } from "./fs";
+import { Task, tasks as codeTasks, WorkspaceFolder } from "vscode";
+import { isPathEqual } from "./fs";
 
 export namespace taskUtils {
     export function getFsPathFromTask(task: Task): string | undefined {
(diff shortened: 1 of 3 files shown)


nturinski (approved):

src/utils/taskUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+            return workspaceFolder.uri?.fsPath;
+        } else {
+            return undefined;
+        }
+    }
+
+    /**
+     * Returns true if the task's scope is a workspace folder matching the given path or if the task is not associated with a path
+     */
+    export function isTaskInScopeOfPath(task: Task, fsPath: string): boolean {
```
nturinski, line 22: Is it worth removing functions in util files that aren't currently being used?

ejizba: Mmm could go either way. In general, I think it's okay to give "utils" files a bit of lee-way with our normal rules. For example, we allow a folder "utils" that doesn't have a very descriptive name and we frequently just copy code from other repos

Issue #2547, opened 2020-11-25
Templates do not update after changing bundles version


anthonychu:
1. Create new JS function app, skip function creation.
1. Update host.json to use Bundles 2.x
1. Create new function - Durable entities templates are not available
1. Reload window
1. Create new function - Durable entities templates are there

ejizba:
So far it hasn't been worth listening for changes to host.json or anything else. We only update when a user re-loads VS Code. Not entirely sure it's worth listening, given how rarely a user would change major bundle versions. Another options is to add some sort of "Reload Templates" command, which could be accessed from the command palette or at the bottom of the template list (if we want it extra-discoverable)

nturinski:
I think with the `Reload templates` option, we can go ahead and close this issue.

PR #2587, opened 2020-12-10
Add option to reload templates and clear cache

ejizba:
We've run into a few bugs with the templates where it would be nice to give users an option to clear the cache, for example https://github.com/Azure/azure-functions-core-tools/issues/2338

I added an option to the end of the "Create Function" quick pick because it's easiest to clear the cache if I know what language/project/version to clear (as opposed to like a separate command palette command). But I also thought it was ugly and didn't want it to show it all the time, so I hid it behind a setting
Screen Shot 2020-12-09 at 6 00 14 PM

Related to https://github.com/microsoft/vscode-azurefunctions/issues/2547

### DIFF
--- a/package.json
+++ b/package.json
@@ -1045,6 +1045,11 @@
                         "type": "boolean",
                         "description": "%azureFunctions.showHiddenStacks%",
                         "default": false
+                    },
+                    "azureFunctions.showReloadTemplates": {
+                        "type": "boolean",
+                        "description": "%azureFunctions.showReloadTemplates%",
+                        "default": false
                     }
                 }
             }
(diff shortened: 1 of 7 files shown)


BorCob:
Speaking of issues with templates, could you have a look at this: https://github.com/microsoft/vscode-azurefunctions/issues/2588

bgoldstein14:
I'm hoping that this "feature" is actually a fix for an issue we're currently seeing in Visual Studio Code. The only option for Azure Function creation now is one to use OpenAPI to create an HTTP Trigger which is not very helpful.

nturinski (approved):

ejizba:
@bgoldstein14 that was actually fixed with https://github.com/microsoft/vscode-azurefunctions/pull/2584, but yes this PR is related. See here for more information and a workaround: https://github.com/microsoft/vscode-azurefunctions/issues/2588#issuecomment-742698428

PR #2628, opened 2021-01-20
Finalize enabling bots

ejizba:
They're working! The reason needs-more-info failed is because "pingDays" is apparently a required parameter. It's used to ping the assignee (if there is one) after 80 days if we still haven't done anything.

To finalize the actions, I removed the "readonly" flag and changed to reference a specific release tag instead of "main". I'll create the tag once https://github.com/microsoft/vscode-azuretools/pull/838 is merged.


nturinski (approved):

nturinski:
🎉🎉🥳🎊


...🤖🤖🤖

Issue #2654, opened 2021-02-05
Unable to run func tasks after updating to VS Code 1.53.0


justinneff:
After installing the VS Code 1.53.0 update, I am now unable to run func related tasks. The error message received is different depending on how the task is run.

When run from the command palette, I get this message in an error toast:

```
There is no task provider registered for tasks of type "func".
```

When task used as `preLaunchTask` this message is in the Tasks Output:

```
Error: The func task detection didn't contribute a task for the following configuration:
{
    "type": "func",
    "label": "runApiFunctionsHost",
    "command": "host start",
    "problemMatcher": "$func-node-watch",
    "isBackground": true,
    "dependsOn": "api build",
    "options": {
        "cwd": "${workspaceFolder}/services/api"
    }
}
The task will be ignored.
```

I have tried uninstalling and reinstalling the Azure Functions extension, but that did not help. This configuration was working prior to the VS Code update so I am assuming that is somehow related. For reference, I have included the relevant portions of the tasks and launch configurations.

**tasks.json**

```
  {
    "type": "func",
    "command": "host start",
    "problemMatcher": "$func-node-watch",
    "isBackground": true,
    "dependsOn": "api build",
    "options": { "cwd": "${workspaceFolder}/services/api" }
  },
```

**launch.json**

```
    {
      "name": "Attach to Node Functions",
      "type": "node",
      "request": "attach",
      "port": 9229,
      "preLaunchTask": "runApiFunctionsHost",
      "cwd": "${workspaceFolder}/services/api"
    }
```

ejizba:
Yeah looks like this is [related to](https://github.com/microsoft/vscode-azurefunctions/issues/2611) how our extension is activated. As a workaround, you can downgrade VS Code to 1.52 or just make sure you open the Azure view each time you open VS Code, which will activate our extension. This is only a problem when your project is several folders down in the workspace - projects in the root or one level down should still activate us automatically.

We'll look into the proper fix

justinneff:
Thanks for the workaround!

ejizba:
We released v1.2.1, which includes a fix for this. @justinneff thanks for filing!

PR #2657, opened 2021-02-08
Search entire workspace for "host.json" in activation event

ejizba:
We initially chose to only search the root and one level down for performance reasons. However, VS Code had a bug where they would actually search the whole workspace even though we didn't specify it. They fixed that bug in the latest release and it seems like some users have come to rely on that behavior.

I looked into other activation events, but didn't find any that would work. I feel like we should make this change to get us back to the previous behavior, but we could always do more investigating in the future to help with startup performance.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2654

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,8 +79,7 @@
         "onCommand:azureFunctions.addBinding",
         "onCommand:azureFunctions.setAzureWebJobsStorage",
         "onCommand:azureFunctions.reportIssue",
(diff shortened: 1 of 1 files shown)


nturinski (approved):

nturinski:
Should we revisit this due to performance though?  Isn't a recursive search kind of expensive?

PR #2666, opened 2021-02-11
Switch to ESLint

wwlorey:


src/vsCodeConfig/launch.ts
```diff
@@ -11,20 +11,19 @@ const versionKey: string = 'version';
 export const launchVersion: string = '0.2.0';
 
 export function getDebugConfigs(folder: WorkspaceFolder): DebugConfiguration[] {
-    // tslint:disable-next-line: strict-boolean-expressions
     return getLaunchConfig(folder).get<DebugConfiguration[]>(configurationsKey) || [];
 }
 
 export function updateDebugConfigs(folder: WorkspaceFolder, configs: DebugConfiguration[]): void {
-    getLaunchConfig(folder).update(configurationsKey, configs);
+    void getLaunchConfig(folder).update(configurationsKey, configs);
```
ejizba, line 18: await

src/vsCodeConfig/tasks.ts
```diff
@@ -10,20 +10,19 @@ const versionKey: string = 'version';
 export const tasksVersion: string = '2.0.0';
 
 export function getTasks(folder: WorkspaceFolder): ITask[] {
-    // tslint:disable-next-line: strict-boolean-expressions
     return getTasksConfig(folder).get<ITask[]>(tasksKey) || [];
 }
 
 export function updateTasks(folder: WorkspaceFolder, tasks: ITask[]): void {
-    getTasksConfig(folder).update(tasksKey, tasks);
+    void getTasksConfig(folder).update(tasksKey, tasks);
```
ejizba, line 17: await

ejizba (commented):
It seems like eslint is better at applying rules to "Thenable" and not just "Promise". That's a good thing, and it means we may have uncovered bugs related to not awaiting things (probably very minor, edge-case bugs, but still). I feel like I did a pretty thorough search, so if you update at least the code in the 4 files I commented on we should be good

ejizba (approved):

Issue #781, opened 2018-11-04
Host auto-restart on save closes debug session


jeffhollan:
Feature Request:

The azure functions host will automatically detect file changes and reload. This is nice for javascript where you can make a quick edit in code and have host pick it up right away. However in VS Code the debug session terminates somewhere in this process. Not a huge deal but would be nice if the file watch and host restart could maintain a debug session.  May include an ask on restart behavior for azure/azure-functions-core-tools

Extension version: 0.12.0
VS Code version: Code - Insiders 1.29.0-insider (d43f6cb9947c57a3254b556e69aeacc2b2025184, 2018-11-02T15:50:03.156Z)
OS version: Darwin x64 18.2.0

EricJizbaMSFT:
This is being tracked in the functions host: https://github.com/Azure/azure-functions-host/issues/3543

FISHMANPET:
It's not super clear to me from reading the bugs in azure-functions-host, but I think there is a difference in behavior in vscode somehow.

I'm working on Python functions, and as this bug and the attached external bug says, when I change one of my function files the session ends.

However I can run the same command in a terminal (in this case bash in WSL): (`. .venv/bin/activate && func host start`)

When I do that, if I update a function file, the func command auto restarts correctly.  So as far as I can tell the function tools are operating correctly (at least for me, with Python) but the extension isn't behaving correctly when the host restarts.

Another thing about the current behavior.  It looks like what's happening when a debug session is intered is pip install is run, and then in a new terminal the func command is run.  In addition to this method not handling the restart correctly, the new terminal goes away when the func command restarts (and this may be the root of the issue).

I discovered this because my function was writing files into the function directory (not knowing this would restart the function host) but to me it just looked like my code had made the function host crash.  it wasn't until I ran it outside of vscode that I saw that my file changes were causing the restart.

EricJizbaMSFT:
@FISHMANPET there's definitely differences in behavior for VS Code. Here's what we do:
1. We set an environment variable telling the func cli what arguments to use to make the python process debug-able (If you search the logs for `ptvsd_launcher.py`, that's what I'm talking about) 
1. We stop the func task after the debug session ends (which happens automatically when the job host stops). I've added a setting to control this behavior in #1565

If I run `func host start` in the terminal, it successfully restarts after a file save as you described. However, if I leverage the setting I just added, I can now see the full logs after a file save _while in debug mode_ and it fails to restart the host with this error:
[...]

PR #2695, opened 2021-02-19
Add restart to the launch config for nodejs projects

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/781
Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/573

Tracking for Python: https://github.com/microsoft/vscode-python/issues/15431

We discussed offline if we should retroactively ask users to alter their launch configurations for them, but we decided it was probably going to annoy most current users rather than help them.  It's pretty easy for a user to add the restart property to their own `launch.json`.

### DIFF
--- a/src/debug/NodeDebugProvider.ts
+++ b/src/debug/NodeDebugProvider.ts
@@ -15,7 +15,8 @@ export const nodeDebugConfig: DebugConfiguration = {
     type: 'node',
     request: 'attach',
     port: defaultNodeDebugPort,
-    preLaunchTask: hostStartTaskName
+    preLaunchTask: hostStartTaskName,
+    restart: true
 };
 
 export class NodeDebugProvider extends FuncDebugProviderBase {


src/debug/NodeDebugProvider.ts
```diff
@@ -15,7 +15,8 @@ export const nodeDebugConfig: DebugConfiguration = {
     type: 'node',
     request: 'attach',
     port: defaultNodeDebugPort,
-    preLaunchTask: hostStartTaskName
+    preLaunchTask: hostStartTaskName,
+    restart: true
```
ejizba, line 19: As discussed offline, "restart" doesn't seem to work on Mac. We'll need to figure out why that is before we can merge

stefanushinardi:
@nturinski ping to check if there is any progress for this PR

nturinski:
Using the latest func CLI tools (3.0.3331) and VS Code (1.58.0) I can confirm that this still doesn't work on MacOS.  I need to investigate why this isn't working

alexweininger:
On Mac I'm only able to get "restart" to work when I run func start with the following arguments:

`func start --javascript --language-worker -- "--inspect=9229"`

nturinski:
Confirmed that the workaround works on Windows as well though the command is `func start --javascript --language-worker="--inspect=9229"`

With this PR, we should change the `func host start` task to include those parameters so that "restart" will work on both Windows and Mac.

I'll have to do some more investigation, but we should probably omit the port number to the inspect flag.  The port number can be whatever the user defines it to be, and I believe that there is logic in the extension to retrieve the port number dynamically.  

I've confirmed that it works without it on Windows and this [workaround](https://github.com/Azure/azure-functions-core-tools/issues/1738#issuecomment-622553961) didn't include it.

nturinski:
Closing due to lack of activity.

PR #2713, opened 2021-03-02
Only include GA versions for .NET "getFramework"

ejizba:
I got an error using the JsonCLI tool because it tried to use .NET 5, but I only had preview versions installed on that machine. I think that's a fair error, so I'll exclude preview versions from our "getFramework" logic. My new regex basically just looks for a GA version followed by either EOL or whitespace (which would exclude a hyphen)

Typical values for "versions" look like this:
```
5.0.103
```
```
2.1.807 [/usr/local/share/dotnet/sdk]
3.1.403 [/usr/local/share/dotnet/sdk]
5.0.100-rc.2.20479.15 [/usr/local/share/dotnet/sdk]
5.0.103 [/usr/local/share/dotnet/sdk]
```

### DIFF
--- a/src/templates/dotnet/executeDotnetTemplateCommand.ts
+++ b/src/templates/dotnet/executeDotnetTemplateCommand.ts
@@ -58,19 +58,37 @@ async function getFramework(context: IActionContext, workingDirectory: string | 
             // ignore
         }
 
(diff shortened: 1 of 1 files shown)


nturinski (approved):
Discussed offline.  Having a preview version will straight up block all .NET project creation which is a terrible experience.

PR #2716, opened 2021-03-03
Support listing functions and "copy function url" for .NET 5

ejizba:
.NET 5 doesn't use the typical "function.json" files, which breaks two scenarios:

 ### List functions locally

There's a url I can ping (`http://localhost:7071/admin/functions`) to get them. It's not the best because the project has to be running, but that's when you probably want to "Copy Function Url" or "Execute Functions" anyways. I also have a placeholder tree item like this:
Screen Shot 2021-03-02 at 6 37 11 PM

### Copy function url for remote functions

In this case, the `config` property on `FunctionEnvelope` usually corresponds to the contents of the "functions.json" file, but it's an empty object for .NET 5. We'll use `invokeUrlTemplate` instead - it doesn't have _all_ the information we might want, but it gets http triggers working which is the most important part.

### DIFF
--- a/src/tree/AzureAccountTreeItemWithProjects.ts
+++ b/src/tree/AzureAccountTreeItemWithProjects.ts
@@ -78,15 +78,17 @@ export class AzureAccountTreeItemWithProjects extends AzureAccountTreeItemBase {
                     } else {
                         let preCompiledProjectPath: string | undefined;
                         let effectiveProjectPath: string;
-                        const compiledProjectPath: string | undefined = await getCompiledProjectPath(projectPath, language);
-                        if (compiledProjectPath) {
+                        let isIsolated: boolean | undefined;
+                        const compiledProjectInfo: CompiledProjectInfo | undefined = await getCompiledProjectInfo(projectPath, language);
+                        if (compiledProjectInfo) {
                             preCompiledProjectPath = projectPath;
-                            effectiveProjectPath = compiledProjectPath;
+                            effectiveProjectPath = compiledProjectInfo.compiledProjectPath;
+                            isIsolated = compiledProjectInfo.isIsolated;
                         } else {
                             effectiveProjectPath = projectPath;
                         }
 
-                        const treeItem: LocalProjectTreeItem = new LocalProjectTreeItem(this, { effectiveProjectPath, folder, language, version, preCompiledProjectPath });
+                        const treeItem: LocalProjectTreeItem = new LocalProjectTreeItem(this, { effectiveProjectPath, folder, language, version, preCompiledProjectPath, isIsolated });
                         this._projectDisposables.push(treeItem);
                         children.push(treeItem);
                     }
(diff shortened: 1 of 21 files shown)


src/tree/remoteProject/RemoteFunctionTreeItem.ts
```diff
@@ -59,18 +59,28 @@ export class RemoteFunctionTreeItem extends FunctionTreeItemBase {
...
+        return await this.parent.parent.client.getFunction(this.name);
+    }
+
     public async getKey(): Promise<string | undefined> {
-        switch (this.config.authLevel) {
-            case HttpAuthLevel.admin:
-                const hostKeys: WebSiteManagementModels.HostKeys = await this.client.listHostKeys();
-                return nonNullProp(hostKeys, 'masterKey');
-            case HttpAuthLevel.function:
+        if (this._config.authLevel === HttpAuthLevel.anonymous) {
```
nturinski, line 67: Could just use `isAnonymous` here

src/tree/AzureAccountTreeItemWithProjects.ts
```diff
@@ -142,12 +144,15 @@ export class AzureAccountTreeItemWithProjects extends AzureAccountTreeItemBase {
     }
 }
 
-async function getCompiledProjectPath(projectPath: string, projectLanguage: ProjectLanguage): Promise<string | undefined> {
+type CompiledProjectInfo = { compiledProjectPath: string; isIsolated: boolean };
+
+async function getCompiledProjectInfo(projectPath: string, projectLanguage: ProjectLanguage): Promise<CompiledProjectInfo | undefined> {
```
nturinski, line 149: Is the idea to make this future proof for other languages? Otherwise I feel like it makes more sense to only call this _if_ we're dealing with a C#/F# project rather than checking it in here.

src/tree/FunctionTreeItemBase.ts
```diff
@@ -77,20 +101,32 @@ export abstract class FunctionTreeItemBase extends AzExtTreeItem {
...
 
-        const route: string = (this.config.triggerBinding && this.config.triggerBinding.route) || this.name;
-        const hostJson: IParsedHostJson = await this.parent.parent.getHostJson();
-        triggerUrl.pathname = `${hostJson.routePrefix}/${route}`;
+    private async refreshTriggerUrl(): Promise<void> {
+        const hostUrl = new url.URL(this.parent.parent.hostUrl);
+        let triggerUrl: url.URL;
+        if (this._func?.invokeUrlTemplate) {
+            triggerUrl = new url.URL(this._func?.invokeUrlTemplate);
+            triggerUrl.protocol = hostUrl.protocol; // invokeUrlTemplate seems to use the wrong protocol sometimes. Make sure it matches the hostUrl
```
nturinski, line 123: Is there any information about this?

src/tree/localProject/LocalFunctionTreeItem.ts
```diff
@@ -3,26 +3,42 @@
...
+    public async refreshFuncEnvelope(): Promise<WebSiteManagementModels.FunctionEnvelope | undefined> {
+        try {
+            const response = await requestUtils.sendRequestWithExtTimeout({
+                url: `${this.parent.parent.hostUrl}/admin/functions/${this.name}`,
+                method: 'GET'
+            });
+            return requestUtils.convertToAzureSdkObject(response.parsedBody);
+        } catch {
+            // project isn't running
+            return undefined;
```
nturinski, line 38: There aren't any other errors that we should be catching?

src/tree/localProject/LocalFunctionsTreeItem.ts
```diff
@@ -64,6 +72,37 @@ export class LocalFunctionsTreeItem extends FunctionsTreeItemBase {
...
+                method: 'GET'
+            });
+            return await this.createTreeItemsWithErrorHandling(
+                <WebSiteManagementModels.FunctionEnvelope[]>functions.parsedBody,
+                'azFuncInvalidLocalFunction',
+                async func => {
+                    func = requestUtils.convertToAzureSdkObject(func);
+                    return LocalFunctionTreeItem.create(this, nonNullProp(func, 'name'), func.config, undefined, func, context);
+                },
+                func => func.name
```
nturinski, line 92: Would it be worth to mention to start debugging to view this?  I imagine that this would only throw an error is func.name were to be `undefined`.  Could the user potentially kill their func task without the event handler catching it?  Like if they closed it via their terminal or process manager?

nturinski (approved):

src/tree/AzureAccountTreeItemWithProjects.ts
ejizba: We already use this for Java, too

ejizba: Ya gotta uncollapse some lines of code on Github to see it

src/tree/FunctionTreeItemBase.ts
ejizba: Nah just from my own testing. Pretty harmless thing to do even if they fix the problem

PR #2719, opened 2021-03-03
Fix bug where .NET templates are refreshed too often

ejizba:
When we get the projectTemplateKey from the VS Code [setting](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/createFunction.ts#L61), the default value is an empty string. Well, we don't want that to affect the sessionProjKey, otherwise it'll refresh the templates all the time. Sure enough, the unit tests were having a lot of timeout problems because of this.

I also had to re-work the function tests a tiny bit. Turns out they were benefiting from the fact that we constantly refreshed the templates. Fix was to make sure .NET 5 and .NET 3.1 tests use a unique project path


ejizba:
Gotta rework this a little bit to handle some more cases. Will submit a new PR when ready.

PR #2720, opened 2021-03-04
Fix bug where .NET templates are refreshed too often

ejizba:
Tests were timing out on .NET quite a bit - I cleaned some stuff up in both the source code and tests to fix it. Two main fixes in the source code:
1. Don't clear out the sessionProjKey if the value from the VS Code setting is an empty string (that's actually the default)
2. Reduce the number of times we read the csproj to get the projectTemplateKey

### DIFF
--- a/src/templates/CentralTemplateProvider.ts
+++ b/src/templates/CentralTemplateProvider.ts
@@ -9,6 +9,7 @@ import { ProjectLanguage, projectTemplateKeySetting, TemplateFilter } from '../c
 import { ext, TemplateSource } from '../extensionVariables';
 import { FuncVersion } from '../FuncVersion';
 import { localize } from '../localize';
+import { nonNullValue } from '../utils/nonNull';
 import { DotnetTemplateProvider } from './dotnet/DotnetTemplateProvider';
 import { getDotnetVerifiedTemplateIds } from './dotnet/getDotnetVerifiedTemplateIds';
 import { IBindingTemplate } from './IBindingTemplate';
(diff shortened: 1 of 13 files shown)


src/templates/CentralTemplateProvider.ts
```diff
@@ -106,6 +107,12 @@ export class CentralTemplateProvider implements Disposable {
         }
     }
 
+    public async getProjectTemplateKey(projectPath: string | undefined, language: ProjectLanguage, version: FuncVersion, projectTemplateKey: string | undefined): Promise<string> {
+        const cachedProviders = await this.getCachedProviders(projectPath, language, version, projectTemplateKey);
+        const provider = nonNullValue(cachedProviders.providers[0], 'firstProvider');
```
nturinski, line 112: Why do we only care about the firstProvider?

src/templates/TemplateProviderBase.ts
```diff
@@ -156,32 +157,37 @@ export abstract class TemplateProviderBase implements Disposable {
...
         this._projKeyMayHaveChanged = true;
     }
 
-    public async hasProjKeyChanged(projKey: string | undefined): Promise<boolean> {
+    /**
+     * Returns true if the key changed
+     */
+    public async updateProjKeyIfChanged(projKey: string | undefined): Promise<boolean> {
+        let hasChanged: boolean;
         if (!this.refreshProjKey) {
```
nturinski, line 176: Do you need this conditional still?  If there is no refreshProjKey, I'd imagine the other conditions would all just be false and you'd hit the else statement.

nturinski (approved):

src/templates/TemplateProviderBase.ts
ejizba: I like being explicit. Plus TypeScript would complain if I didn't have it
Screen Shot 2021-03-03 at 7 04 17 PM

src/templates/CentralTemplateProvider.ts
ejizba: Add a comment in the code. .NET only has one provider and I'm not sure how we'll handle this once we get to other languages using project keys

Issue #2723, opened 2021-03-04
The port in the copied url from "Local Project tree view" is different from the one copied in terminal


v-ruizh:
**OS:** All
**Build Version:** 20210304.14

**Repro Steps:**
1. Create a JavaScript project.
2. Set a breakpoint and debug this project.
3. Copy and check the function url from "Local Project tree view" and Terminal during debugging.

**Expect:**
The copied url from "Local Project tree view" is the same as the one copied from Terminal.

**Actual:**
The port in the copied url from "Local Project tree view" is different from the one copied in Terminal.

- Url in "Local Project tree view": http://localhost:**9229**/api/HttpTrigger1
- Url in Terminal: http://localhost:**7071**/api/HttpTrigger1

**More Info:**
1. This issue does not reproduce for C# project.
2. Past the url which copied from "Local Project tree view" in browser and the breakpoint will not be hit.
3. You can refresh the "Local Project tree view" before copying the url.

PR #2726, opened 2021-03-04
Fix debugPort vs. funcPort

ejizba:
I got the port used for debugging (i.e. 9229 for node.js) mixed up with the port the func host runs on (usually 7071)

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2723

### DIFF
--- a/src/funcCoreTools/funcHostTask.ts
+++ b/src/funcCoreTools/funcHostTask.ts
@@ -17,7 +17,7 @@ const funcTaskStartedEmitter = new vscode.EventEmitter<vscode.WorkspaceFolder | 
 export const onFuncTaskStarted = funcTaskStartedEmitter.event;
 
 export const runningFuncPortMap = new Map<vscode.WorkspaceFolder | vscode.TaskScope | undefined, string>();
-export const defaultFuncPort: string = '7071';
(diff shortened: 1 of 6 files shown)


src/funcCoreTools/funcHostTask.ts
```diff
@@ -13,6 +13,12 @@ export interface IRunningFuncTask {
 
 export const runningFuncTaskMap: Map<vscode.WorkspaceFolder | vscode.TaskScope, IRunningFuncTask> = new Map<vscode.WorkspaceFolder | vscode.TaskScope, IRunningFuncTask>();
 
+const funcTaskStartedEmitter = new vscode.EventEmitter<vscode.WorkspaceFolder | vscode.TaskScope | undefined>();
+export const onFuncTaskStarted = funcTaskStartedEmitter.event;
+
+export const runningFuncPortMap = new Map<vscode.WorkspaceFolder | vscode.TaskScope | undefined, string>();
+export const defaultFuncPort: string = '7071';
```
nturinski, line 20: Is there any point in exporting this?  I don't see it used anywhere else.

nturinski (approved):

PR #2727, opened 2021-03-04
Use an older mac agent for pipelines

ejizba:
Mac builds started failing for several repos. Seems like it's related to the agent being updated recently to 10.15 because 10.14 works for me

### DIFF
--- a/.azure-pipelines/main.yml
+++ b/.azure-pipelines/main.yml
@@ -24,7 +24,7 @@ jobs:
 
 - job: macOS
   pool:
-    vmImage: macOS-latest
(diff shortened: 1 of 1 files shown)


nturinski:
Err, shouldn't this have passed if that were the case?

ejizba:
:P Did you look at the build? There's unrelated timeouts in the tests

It passed here: https://dev.azure.com/ms-azuretools/AzCode/_build/results?buildId=31648&view=logs&j=a5e52b91-c83f-5429-4a68-c246fc63a4f7

nturinski (approved):

Issue #2656, opened 2021-02-08
C# debugging fails in v1.54 of VS Code


howardginsburg:
With the most recent update to the Insiders version, I cannot debug an Azure Functions project.

Steps to reproduce:
1. Run -> Start Debugging
2. The project will build and the Azure Function runtime will start.
3. Error window in right corner of VS Code says "Cannot read property 'map' of undefined".
4. Debug stops.
5. Azure Function runtime must be killed through task manager or my shutting down VS Code.

ejizba:
What type of project is this - C#, JavaScript, etc.? And what version of the func cli do you have?

xsoheilalizadeh:
@ejizba 

I have the same issue with the latest version v1.2.1, the project is c#.

```
                  %%%%%%
                 %%%%%%
            @   %%%%%%    @
          @@   %%%%%%      @@
       @@@    %%%%%%%%%%%    @@@
     @@      %%%%%%%%%%        @@
       @@         %%%%       @@
         @@      %%%       @@
           @@    %%      @@
                %%
                %


Azure Functions Core Tools
Core Tools Version:       3.0.3233 Commit hash: d1772f733802122a326fa696dd4c086292ec0171
Function Runtime Version: 3.0.15193.0

Version: 1.54.0-insider (user setup)
Commit: cd97629c4458c05b16a1fd444d21c61295ba2529
Date: 2021-02-09T08:37:01.449Z
Electron: 11.2.3
Chrome: 87.0.4280.141
Node.js: 12.18.3
V8: 8.7.220.31-electron.0
OS: Windows_NT x64 10.0.19042
```

howardginsburg:
C# project
CLI: 3.0.3233

ejizba:
Hey are you guys still seeing this behavior? I saw some problems for a few days but it kind of went away - maybe VS Code fixed the bug on their end...

howardginsburg:
Yes, the problem still exists for me even after updating to the latest VSCode Insider build and Functions extension.

ejizba:
What OS are you on? I still haven't seen the particular error "Cannot read property 'map' of undefined"...

howardginsburg:
Edition	Windows 10 Enterprise
Version	20H2
Installed on	‎10/‎13/‎2020
OS build	19042.804
Serial number	037373502857
Experience	Windows Feature Experience Pack 120.2212.551.0

runtimeware:
[...]

PR #2736, opened 2021-03-05
Workaround incorrect processId bug for C# debugging

ejizba:
Temporary fix for https://github.com/microsoft/vscode-azurefunctions/issues/2656

Basically - if we detect the processId is no longer running, we'll just use the activeTerminal's processId. The active terminal _should_ be the `func host start` task, but I'm not entirely sure it will always be (hence _temporary_ workaround). When I test on older versions of VS Code (aka without the bug), `isRunning` always returns true and `taskInfo.processId` already matches `activeTerminal.processId`. 

Obviously if VS Code fixes their bug we won't need this, but I figured best to have a fix available on our side if we want it.


nturinski:
Also a temporary fix for https://github.com/microsoft/vscode-azurefunctions/issues/2737#issuecomment-794562983

nturinski (approved):

PR #2751, opened 2021-03-18
Support C# "bool" parameter during "Create Function"

ejizba:
Before:
Screen Shot 2021-03-18 at 2 15 22 PM

After:
Screen Shot 2021-03-18 at 2 14 34 PM

### DIFF
--- a/src/commands/addBinding/settingSteps/BooleanPromptStep.ts
+++ b/src/commands/addBinding/settingSteps/BooleanPromptStep.ts
@@ -3,21 +3,21 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { QuickPickItem } from "vscode";
+import { IAzureQuickPickItem } from "vscode-azureextensionui";
+import { BindingSettingValue } from "../../../funcConfig/function";
+import { envUtils } from "../../../utils/envUtils";
 import { IBindingWizardContext } from "../IBindingWizardContext";
 import { BindingSettingStepBase } from "./BindingSettingStepBase";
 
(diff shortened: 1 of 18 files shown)


src/commands/addBinding/settingSteps/BooleanPromptStep.ts
```diff
@@ -9,10 +9,15 @@ import { BindingSettingStepBase } from "./BindingSettingStepBase";
...
-        const picks: QuickPickItem[] = [
+        let picks: QuickPickItem[] = [
             { label: 'true', description: '' },
             { label: 'false', description: '' }
         ];
-        return (await context.ui.showQuickPick(picks, { placeHolder: this._setting.label })).label;
+
+        if (this._setting.defaultValue?.toLowerCase() === 'false') {
+            picks = picks.reverse();
+        }
```
wwlorey, line 19: Could you elaborate on why you reverse the picks here? Also why add `this._setting.description` as a potential placeholder below?

ejizba: > Could you elaborate on why you reverse the picks here?

If the default is "false", I just want "false" to be first.

> Also why add this._setting.description as a potential placeholder below?

So "label" is generally the name of the setting and "description" is a description of it. For input boxes I can use both ("label" for "placeHolder" and "description" for "prompt"), but for quick picks I only have "placeHolder". We hardly ever use BooleanPromptStep, and it just happened to look better in this case which is why I added it. The only risk is that placeHolder can get cut off if it's too long (a description will generally be longer than a label), but you can still hover over it to see the tooltip, which has the whole thing.

wwlorey (approved):

PR #2770, opened 2021-04-13
npm update

ejizba:
Ran into this problem with xregexp, which is why I hard-coded the version to 4.3.0: https://github.com/slevithan/xregexp/issues/305


wwlorey (approved):
Do you think it's worth filing a new issue with xregexp since the issue you linked is closed and the problem persists?

ejizba:
> Do you think it's worth filing a new issue with xregexp since the issue you linked is closed and the problem persists?

When they closed the issue, they specifically fixed it for anyone targeting "webworker" or "web", but we target "node". They mentioned a few options for people targeting "node", including marking "xregexp" as an external, waiting for Node.js 14, or trying to import the package in other ways. Aka it's possible the issue is "by design" for node, and also possible the issue is in our webpack config. I'm a bit skeptical the problem is on us because all of our other packages work, but that issue is pretty dense - I'd have to spend a decent amount of time investigating before I'd feel comfortable filing a new issue, which doesn't seem worth it. I tried a few of the options mentioned and this fix was the easiest one that actually worked.

Issue #2760, opened 2021-03-29
How to bypass  .NET3 | .NET5 in showQuickPick UI


dmitriklem:
I am following code in [this example](https://github.com/microsoft/vscode-azurestaticwebapps/blob/d2816e900a75841abff6a8dcd171de877ca9d9ab/src/commands/createHttpFunction.ts#L27).  Is there any way to provide .NET3 | .NET5 as the parameters/functional settings, so they are not shown in the UI (by showQuickPick) during "funcApi.createFunction" call?

nturinski:
Related: https://github.com/microsoft/vscode-azurestaticwebapps/issues/323

PR #2773, opened 2021-04-15
Add targetFramework as a parameter for createFunction

nturinski:
I couldn't find an `id` property that I could use on the runtimeWorkers.  My main concern with using targetFramework is that would `netcoreapp3.1` be replaced by `netcoreapp3.2` when they start supporting .NET 3.2?  Any suggestion on what I should pass in for a more stable id?

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2760

### DIFF
--- a/src/commands/createNewProject/IProjectWizardContext.ts
+++ b/src/commands/createNewProject/IProjectWizardContext.ts
@@ -23,6 +23,8 @@ export interface IProjectWizardContext extends IActionContext {
 
     generateFromOpenAPI?: boolean;
     openApiSpecificationFile?: Uri[];
+
+    targetNetwork?: string;
 }
 
 export type OpenBehavior = 'AddToWorkspace' | 'OpenInNewWindow' | 'OpenInCurrentWindow' | 'AlreadyOpen' | 'DontOpen';
--- a/src/commands/createNewProject/IProjectWizardContext.ts
+++ b/src/commands/createNewProject/IProjectWizardContext.ts
@@ -24,7 +24,7 @@ export interface IProjectWizardContext extends IActionContext {
     generateFromOpenAPI?: boolean;
     openApiSpecificationFile?: Uri[];
 
-    targetNetwork?: string;
+    targetFramework?: string;
 }
 
 export type OpenBehavior = 'AddToWorkspace' | 'OpenInNewWindow' | 'OpenInCurrentWindow' | 'AlreadyOpen' | 'DontOpen';
--- a/src/commands/createNewProject/createNewProject.ts
+++ b/src/commands/createNewProject/createNewProject.ts
@@ -49,7 +49,6 @@ export async function createNewProjectInternal(context: IActionContext, options:
     const version: string = options.version || getGlobalSetting(funcVersionSetting) || await tryGetLocalFuncVersion() || latestGAVersion;
     const projectTemplateKey: string | undefined = getGlobalSetting(projectTemplateKeySetting);
     const wizardContext: Partial<IFunctionWizardContext> & IActionContext = Object.assign(context, options, { language, version: tryParseFuncVersion(version), projectTemplateKey });
-    wizardContext.targetNetwork = options.targetNetwork;
 
     if (options.folderPath) {
         FolderListStep.setProjectPath(wizardContext, options.folderPath);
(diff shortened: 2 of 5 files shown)


src/commands/createNewProject/IProjectWizardContext.ts
```diff
@@ -23,6 +23,8 @@ export interface IProjectWizardContext extends IActionContext {
 
     generateFromOpenAPI?: boolean;
     openApiSpecificationFile?: Uri[];
+
+    targetNetwork?: string;
```
ejizba, line 27: Where did you get the word `targetNetwork`? Should be `targetFramework`

src/commands/createNewProject/createNewProject.ts
```diff
@@ -49,6 +49,7 @@ export async function createNewProjectInternal(context: IActionContext, options:
     const version: string = options.version || getGlobalSetting(funcVersionSetting) || await tryGetLocalFuncVersion() || latestGAVersion;
     const projectTemplateKey: string | undefined = getGlobalSetting(projectTemplateKeySetting);
     const wizardContext: Partial<IFunctionWizardContext> & IActionContext = Object.assign(context, options, { language, version: tryParseFuncVersion(version), projectTemplateKey });
+    wizardContext.targetNetwork = options.targetNetwork;
```
ejizba, line 52: This should not be necessary because of the Object.assign just above

src/vscode-azurefunctions.api.d.ts
```diff
@@ -78,4 +79,9 @@ export interface ICreateFunctionOptions {
      * If set to true, it will not try to open the folder after create finishes. Defaults to false
      */
     suppressOpenFolder?: boolean;
+
+    /**
+     * If set, it will automatically select the worker runtime for .NET with the matching targetNetwork
+     */
+    targetNetwork?: TargetFramework;
```
ejizba, line 86: I don't think I want to start hard-coding target framework versions in the api. Can we just leave this as `string`?

ejizba (commented):
Target framework is pretty stable - I would trust it. Also I don't think there's ever going to be `netcoreapp3.2` because `net5.0` _is_ the next release. You can see more about the versioning plans here: https://dotnet.microsoft.com/platform/support/policy/dotnet-core

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
```diff
@@ -28,6 +28,12 @@ export class DotnetRuntimeStep extends AzureWizardPromptStep<IProjectWizardConte
         } else if (runtimes.length === 1) {
             // No need to prompt if it only supports one
             setWorkerRuntime(context, runtimes[0]);
+        } else if (context.targetNetwork) {
+            // if a targetNetwork was provided from createNewProject
+            const workerRuntime = runtimes.find(runtime => runtime.targetFramework === context.targetNetwork);
+            if (workerRuntime) {
+                setWorkerRuntime(context, workerRuntime);
+            }
```
ejizba, line 36: I feel like this should throw an error instead of falling through. Something like `Unrecognized target framework "{0}". Available frameworks: "{1}".`

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
```diff
@@ -28,6 +28,14 @@ export class DotnetRuntimeStep extends AzureWizardPromptStep<IProjectWizardConte
         } else if (runtimes.length === 1) {
             // No need to prompt if it only supports one
             setWorkerRuntime(context, runtimes[0]);
+        } else if (context.targetFramework) {
```
ejizba, line 31: I think this case should be before the `runtimes.length === 1` case. If `runtimes` is `[netcoreapp2.1]` and `context.targetFramework` is `netcoreapp3.1`, I want it to throw your "Unrecognized target framework" error instead of using `2.1`

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
```diff
@@ -28,6 +28,14 @@ export class DotnetRuntimeStep extends AzureWizardPromptStep<IProjectWizardConte
...
+            // if a targetFramework was provided from createNewProject
+            const workerRuntime = runtimes.find(runtime => runtime.targetFramework === context.targetFramework);
+            if (!workerRuntime) {
+                throw new Error(localize('unknownFramework', 'Unrecognized target framework "{0}". Available frameworks: "{1}".', context.targetFramework, runtimes.join(', ')));
```
ejizba, line 35: Does `runtimes.join` work? I thought they were objects, not strings

ejizba (approved):

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
nturinski: Yeah, you're right, looks like I'll have to use a .map to get the frameworks.

nturinski: image

PR #2784, opened 2021-04-28
Take app setting validation from shared package

ejizba:
This logic is unique to Functions and I wanted to consolidate all app settings verification into this repo.

Depends on https://github.com/microsoft/vscode-azuretools/pull/893

### DIFF
--- a/src/commands/deploy/verifyAppSettings.ts
+++ b/src/commands/deploy/verifyAppSettings.ts
@@ -14,17 +14,24 @@ import { localize } from '../../localize';
 import { SlotTreeItemBase } from '../../tree/SlotTreeItemBase';
 import { getFunctionsWorkerRuntime, isKnownWorkerRuntime } from '../../vsCodeConfig/settings';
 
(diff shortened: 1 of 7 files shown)


ejizba:
ping @nturinski @wwlorey @neelip

src/commands/deploy/verifyAppSettings.ts
```diff
@@ -91,3 +98,45 @@ async function verifyWebContentSettings(context: IActionContext, remotePropertie
     context.telemetry.properties.webContentSettingsRemoved = String(shouldRemoveSettings);
     return shouldRemoveSettings;
 }
+
+function verifyLinuxRemoteBuildSettings(context: IActionContext, remoteProperties: { [propertyName: string]: string }, bools: VerifyAppSettingBooleans): boolean {
```
nturinski, line 102: Where did all of these extra settings come from? I don't think this was in the appservice package.

nturinski: Ah never mind, found the deleted file.  Thanks Eric 🥴

nturinski (approved):

Issue #2744, opened 2021-03-11
C# Debugging fails when setting LocalHttpPort in local.settings.json


lwoollett-smudge:
Hi guys, with #2738 not fixing my issues with the debugger I kept looking for solutions.

Turns out it was inside my local.settings.json
This local.settings will fail to attach the debugger:
```
{
  "IsEncrypted": false,
  "Values": {
    "AzureWebJobsStorage": "",
    "FUNCTIONS_WORKER_RUNTIME": "dotnet",
  },
  "Host": {
    "LocalHttpPort": 9003,
    "CORS": "*",
    "CORSCredentials": false
  }
}
```
Whereas this will succeed:
```
{
  "IsEncrypted": false,
  "Values": {
    "AzureWebJobsStorage": "",
    "FUNCTIONS_WORKER_RUNTIME": "dotnet",
  },
  "Host": {
    "CORS": "*",
    "CORSCredentials": false
  }
}
```

`dotnet --info` output:
```
.NET Core SDK (reflecting any global.json):
 Version:   3.1.406
 Commit:    f42e6b201a

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.16
 OS Platform: Darwin
 RID:         osx.10.16-x64
 Base Path:   /Users/luke/Development/Runtimes/dotnet-sdk-3.1.406-osx-x64/sdk/3.1.406/

Host (useful for support):
  Version: 3.1.12
  Commit:  0267ad09c6

.NET Core SDKs installed:
  3.1.406 [/Users/luke/Development/Runtimes/dotnet-sdk-3.1.406-osx-x64/sdk]
  ```

VSCode version 1.54.1

Asked @danielwhyte-smudge to replicate and he was also able to get this behaviour

ejizba:
Yeah we do not take `LocalHttpPort` into account. We'll look into fixing.

In the meantime, any particular reason you need to use "9003"? It should work if you use the default, "7071". As a workaround, I think you can replicate the same behavior if you search for all instances of `host start` in your `.vscode` folder and replace it with `host start --port 9003`

lwoollett-smudge:
Yeah have been switching over to `host start --port x` while we've been waiting for a fix, the bonus is it gets committed into version control as well so postman scripts are easier to share 🍾 

For the current project i'm working on we're using 90xx as a port for when we're running multiple functions locally.

PR #2786, opened 2021-04-28
Take "LocalHttpPort" into account when determining func port

ejizba:
Users can control the func port from their "local.settings.json" and it causes problems when we ignore that

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2744

### DIFF
--- a/src/funcCoreTools/funcHostTask.ts
+++ b/src/funcCoreTools/funcHostTask.ts
@@ -3,8 +3,12 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import * as path from 'path';
 import * as vscode from 'vscode';
 import { IActionContext, registerEvent } from 'vscode-azureextensionui';
+import { tryGetFunctionProjectRoot } from '../commands/createNewProject/verifyIsProject';
+import { localSettingsFileName } from '../constants';
+import { getLocalSettingsJson } from '../funcConfig/local.settings';
 import { getWorkspaceSetting } from '../vsCodeConfig/settings';
 
 export interface IRunningFuncTask {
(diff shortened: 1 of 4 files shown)


src/funcCoreTools/funcHostTask.ts
```diff
@@ -62,14 +66,33 @@ export function stopFuncTaskIfRunning(workspaceFolder: vscode.WorkspaceFolder):
     }
 }
 
-export function getFuncPortFromTask(funcTask: vscode.Task | undefined): string {
-    let port: string = defaultFuncPort;
+export async function getFuncPortFromTask(funcTask: vscode.Task | undefined, projectPathOrTaskScope: string | vscode.WorkspaceFolder | vscode.TaskScope): Promise<string> {
```
neelip, line 69: I think method name needs a change to make it more general.

nturinski, line 69: Do you have to add `TaskScope` as a type here?  

It's kind of confusing since `vscode.TaskScope` is an enum of 1 or 2, but the scope property in task includes WorkspaceFolder.  I know that this is technically correct, but you're only using this parameter as a projectPath, so it seems like it'd make more sense to just ensure that it's a WorkspaceFolder [here](https://github.com/microsoft/vscode-azurefunctions/pull/2786/files?diff=split&w=1#diff-cef073404d3356451ab0401754e928cd82a926e6ec3fd4cab28f22d2d1adb8ceR37)

nturinski: `getFuncPortFromTaskOrProject`?

Personally, I think it's kind of a long-winded name but it is technically doing two things now.

I guess you could also write another function called `getFuncPortFromLocalSettings` that just does the new part that you added and call it in here as well. 🤷

nturinski (approved):

src/funcCoreTools/funcHostTask.ts
ejizba: > It's kind of confusing since vscode.TaskScope is an enum of 1 or 2, but the scope property in task includes WorkspaceFolder. 

I don't pick the types, VS Code does: https://github.com/microsoft/vscode/blob/69373b783347d93e598caef5be7a80099ba73983/src/vs/vscode.d.ts#L6641

> so it seems like it'd make more sense to just ensure that it's a WorkspaceFolder here

I don't really see the difference. Feels like 6 of one, half dozen of another

ejizba: > I think function name needs a change to make it more general.

My preference when you review and don't like a name is that you make an alternative suggestion in your comment to "prove" there's a better name that still makes sense. As Nathan mentioned, even his better name is kinda long-winded and some functions are just hard to pick a name for.

> getFuncPortFromTaskOrProject

Sure

ejizba: > But I already approved it so if you feel like they're equivalent, then feel free to ignore.

I'd explain more of my reasoning if this was worth our time to debate. Seems like we both agree it's not worth our time, so yeah I'll go ahead and merge as-is

PR #2787, opened 2021-04-29
Add setting for template source

ejizba:


src/commands/createFunction/FunctionListStep.ts
```diff
@@ -144,9 +144,14 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
         let templateFilter: TemplateFilter = getWorkspaceSetting<TemplateFilter>(templateFilterSetting, context.projectPath) || TemplateFilter.Verified;
 
         while (!context.functionTemplate) {
-            const placeHolder: string = this._isProjectWizard ?
+            let placeHolder: string = this._isProjectWizard ?
                 localize('selectFirstFuncTemplate', "Select a template for your project's first function") :
                 localize('selectFuncTemplate', 'Select a template for your function');
+
+            if (ext.templateProvider.templateSource) {
```
nturinski, line 151: I'm understanding correctly that this will equal `""` if the default is being used?

nturinski (approved):

src/commands/createFunction/FunctionListStep.ts
ejizba: Uh well not sure if it's `undefined` or an empty string, but either way should work the same

PR #2794, opened 2021-05-10
Use app service specific location step

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/906


src/tree/SubscriptionTreeItem.ts
```diff
@@ -104,12 +103,15 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
         }
 
         if (!context.advancedCreation) {
+            LocationListStep.addStep(wizardContext, promptSteps);
```
nturinski, line 106: Since custom locations are only supported for advanced scenarios, would it be worth it to leave location list step at the end for basic create? That way people using the more common scenario won't notice a change.

ejizba: It's still at the end

PR #2799, opened 2021-05-13
Wait for plan to exist before saying create is done

ejizba:
We have the occasional weird issue because the plan comes back as `undefined` immediately after creating a function app. It can cause nightly build failures, which was my main motivation to fix. Although we've baked some best guesses into the code if we see `undefined`, I think it would be best to just wait here until it exists.

In my testing, it takes about 30 extra seconds - feels reasonable to me in the context of all the other resources being created, but open to other's thoughts. I figure I can check telemetry after we release and reassess based on how long it's taking for real users.


nturinski (approved):

ejizba:
Closing in favor of https://github.com/microsoft/vscode-azurefunctions/pull/2801

PR #2828, opened 2021-05-26
Let users debug even if local.settings.json can't be parsed

ejizba:
I discovered parse errors in telemetry for a very small number of users and figured we could handle this better. For C# we currently block debugging, but I tried it out and the func cli will still work (for the most part).


nturinski (approved):
In the case that the func cli doesn't work, what kind of error is thrown?  Is it just seen in the terminal?

ejizba:
This parse error is the first and only "unexpected" error I've seen from preDebugValidate, but yes theoretically if the func cli doesn't work the error will be seen in the terminal

PR #2838, opened 2021-06-03
Give target framework priority when getting C# templates

ejizba:
I'm seeing this error pop up in telemetry:

> Failed to get latest templates: Failed to find templates for .NET worker "netcoreapp3.1".

It's because users have v2 listed as their func version in their VS Code settings, which does not support .net core 3.1 as listed in their csproj. In this case, we will print a warning to the output channel and just use the templates from v3.

### DIFF
--- a/src/templates/dotnet/DotnetTemplateProvider.ts
+++ b/src/templates/dotnet/DotnetTemplateProvider.ts
@@ -8,10 +8,12 @@ import * as path from 'path';
 import { RelativePattern, workspace } from 'vscode';
 import { IActionContext } from 'vscode-azureextensionui';
 import { ProjectLanguage } from '../../constants';
-import { FuncVersion } from '../../FuncVersion';
+import { ext } from '../../extensionVariables';
+import { FuncVersion, getMajorVersion } from '../../FuncVersion';
 import { localize } from '../../localize';
 import { cliFeedUtils } from '../../utils/cliFeedUtils';
 import { dotnetUtils } from '../../utils/dotnetUtils';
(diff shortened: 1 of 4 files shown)


src/templates/dotnet/DotnetTemplateProvider.ts
```diff
@@ -62,24 +64,40 @@ export class DotnetTemplateProvider extends TemplateProviderBase {
...
+                        context.telemetry.properties.effectiveProjectRuntime = newVersion;
+                        const oldMajorVersion = getMajorVersion(this.version);
+                        const newMajorVersion = getMajorVersion(newVersion);
+                        const warning = localize('mismatchProjKeyWarning', 'WARNING: "{0}" is not supported on Azure Functions v{1}. Using templates from v{2} instead.', projKey, oldMajorVersion, newMajorVersion);
+                        ext.outputChannel.appendLog(warning);
+                        return templateVersion;
+                    }
+                }
+            }
+            throw new Error(localize('projKeyError', '"{0}" is not supported for Azure Functions version "{1}".', projKey, this.version));
```
nturinski, line 89: Would it make sense to not throw an error and just let the debug happen?  They'll get an error whenever they start debugging, right?  I'm wondering if there are any configurations where it might work (like a -preview version or something) that the extension might not recognize

nturinski (approved):

src/templates/dotnet/DotnetTemplateProvider.ts
ejizba: This is for creating a function, not debugging

PR #2854, opened 2021-06-16
Add action-specific templateProvider for parallel tests

ejizba:
I'm splitting up my 'parallel tests' work into smaller PRs. The focus of this PR is to add a new `ActionVariable` class in `extensionVariables.ts` because I need a way to have a unique `templateProvider` for each action when running in parallel. _Most_ of this PR is monotonous changes as a consequence of that. The interesting files are `extensionVariables.ts` as mentioned, and `global.test.ts` which handles `templateProvider` on the test side of things.

One consequence of this work is that I can't use `executeCommand` from the tests as often, but I think this is fine because it actually more accurately reports errors as test failures when you call the function directly. The only downside is we lose a bit of telemetry logging, but I added that back in `runWithTestActionContext`.

### DIFF
--- a/extension.bundle.ts
+++ b/extension.bundle.ts
@@ -16,10 +16,13 @@
 // At runtime the tests live in dist/tests and will therefore pick up the main webpack bundle at dist/extension.bundle.js.
 export * from 'vscode-azureappservice';
 export * from 'vscode-azureextensionui';
+export * from './src/commands/copyFunctionUrl';
 export * from './src/commands/createFunction/createFunction';
 export * from './src/commands/createFunction/dotnetSteps/DotnetNamespaceStep';
+export * from './src/commands/createFunctionApp/createFunctionApp';
 export * from './src/commands/createNewProject/createNewProject';
 export * from './src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep';
+export * from './src/commands/deploy/deploy';
 export * from './src/commands/deploy/verifyAppSettings';
 export * from './src/commands/initProjectForVSCode/initProjectForVSCode';
 export * from './src/constants';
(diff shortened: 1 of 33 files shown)


nturinski (approved):
I got a question though:

Is there really an benefit to attaching the templateProvider to the extentionVariables now?  Wouldn't it be the same to just have a static function that gets/creates a templateProvider?

ejizba:
I'm not entirely sure what you mean by "benefit" - I don't think either option mentioned would have a functional difference. Like... `ext` is just used to organize the code and I feel like `templateProvider` still fits. I also don't think this pattern will be unique to the `templateProvider` long-term, so I don't want to put it in a templateProvider-specific file.

nturinski:
Yeah, but I guess it's confusing to me since we used to use extensionVariables for objects/singletons that should be accessed globally.  But now it kind of feels like it's being used as like a global factory to make a bunch of local copies of templateProvider.

For example, the output channel should always have just the one per instance of the extension.

ejizba:
> But now it kind of feels like it's being used as like a global factory to make a bunch of local copies of templateProvider.

The local copies are only used during tests, though. When we actually run the extension, there's just one templateProvider being used. So in that sense, I think it's still primarily an extension-scoped variable.

> For example, the output channel should always have just the one per instance of the extension.

Actually, one problem with parallel tests is that the logs get all jumbled between tests. A solution could be to create an action-specific output channel, perhaps one that caches the logs and then prints them all at once at the end of the action. So the output channel is actually an example of why I think this is a generic pattern we might use for several extension variables.

ejizba:
Since you already approved, I'll go ahead and merge to unblock the rest of my work on the parallel tests. I'm open to continuing this discussion, though, because whatever we decide on here will likely become a common pattern if/when more and more tests are run in parallel.

Issue #2659, opened 2021-02-08
Improve startup performance related to workspaceContains activation event


ejizba:
Related to https://github.com/microsoft/vscode-azurefunctions/issues/2654, we have to search the whole workspace or debugging could fail. As the VS Code team [put it](https://github.com/microsoft/vscode/issues/110510#issuecomment-770133700):
> Using workspaceContains:**/... will start a search across the user's full workspace which may use a lot of CPU. And, if this search takes longer than 5s, the search will be canceled and your extension will be activated anyway.

Ideally we find a better way to activate. Some new activation events that could help from VS Code:
- An event based on a VS Code setting (probably "azureFunctions.projectSubpath") - already tracked in https://github.com/microsoft/vscode/issues/55299
- An event based on the task provider, in our case "func"

heaths:
Alternatively, could the depth be limited? We use this technique for VSIX extensions in Visual Studio for manifest discovery. IIRC, the max depth is 3 which made a huge difference. Perhaps the max depth could be specified by the extension, with a user-specified/default max.

ejizba:
@heaths I think maybe you meant to comment on a _VS Code_ issue? This repo is for an _extension_ for VS Code, not VS Code itself. Our issues and the Rust issue are definitely related, but we can't fix it for y'all haha

Fwiw, you might be able to replicate a depth-limit yourself by listing multiple activation events like:
```
workspaceContains:*/example.txt
workspaceContains:*/*/example.txt
workspaceContains:*/*/*/example.txt
```

heaths:
You're right. 🤦‍♂️ I saw microsoft/vscode and failed to read the rest. But, yes, that might be a good workaround.

Issue #2578, opened 2020-12-08
Does not detect functions in nested folders


stewartadam:
When opening up a large project that consists of several C# projects (including some Azure Functions), the extension will never detect/offer to produce settings. One must open the specific folder containing the Azure Function .csproj file for it to offer to build launch configuration, meaning users need to either (1) do so and then merge that configuration, adjusting paths, into their parent workspace; or (2) keep two editor windows open - one for editing the broader solution, and a second for launching the Azure Function.

Asks:

1. The extension should find and offer to produce missing launch configurations for any Azure Function csproj files in the workspace, not just the top-level opened folder.
2. It would be nice if the Function detection could be built-in to the existing VS Code launch configuration detection instead of using its own separate configuration detection.

ejizba:
We do have a setting "azureFunctions.projectSubpath" which is supposed to help for this situation. Basically, set that to the subpath of the functions project, run the "Initialize Project for Use with VS Code" command (we prompt users to do this in a few places, or you can access it from the command palette), and things should just start working

> The extension should find and offer to produce missing launch configurations for any Azure Function csproj files in the workspace, not just the top-level opened folder.

We automatically look for projects at the root of the workspace or 1 folder level down. This detection logic happens when users first open a VS Code window and we didn't want to search the entire workspace because that could negatively affect startup performance.

> It would be nice if the Function detection could be built-in to the existing VS Code launch configuration detection instead of using its own separate configuration detection.

This is worth looking into. I think I tried it a while ago and the VS Code API wasn't quite good enough. For example I think it _only_ let us specify content to go into the "launch.json" file and we would need to edit the "tasks.json" and "settings.json" files as well.

stewartadam:
> We do have a setting "azureFunctions.projectSubpath"

Am I correct in understanding that this would only work/assume you have a single Azure Function in the workspace? Although detecting Functions a few folders level deep is part of the request, the other is that projects often have several functions so it'd be nice to have the configs detected and setup for more than just one hardcoded in `azureFunctions.projectSubpath`.

ejizba:
There's a bit of confusion here in terminology between "function" and "project". That setting should point to the root of your project, where the "host.json" file is, but that project can contain multiple functions.

If you want to debug multiple _projects_ at once, we have this doc on how to do it: https://github.com/microsoft/vscode-azurefunctions/wiki/Multiple-function-projects

stewartadam:
[...]

PR #2858, opened 2021-06-17
Search at all levels when looking for a project

ejizba:
We've had various issues reported in the past about projects not being found, and we've generally resisted for perf reasons, but I feel like it's finally time to just do it. A few notes:
1. I still give priority to the root and one level down
2. I switched to use VS Code's `findFiles` api, which is simpler and hopefully better perf than searching ourselves with `readdir`
3. As of https://github.com/microsoft/vscode-azurefunctions/pull/2657, we're activating based on the whole workspace anyways, so I feel like we might as well support any project that caused us to activate

I also changed how `suppressPrompt` works because I was annoyed how often I had to change from the default. Now 'silent' (the most common case) is the default. I also split the other case into a modal or non-modal option to better fit each scenario.

Related to https://github.com/microsoft/vscode-azurefunctions/issues/2578 and https://github.com/microsoft/vscode-azurefunctions/issues/2659

### DIFF
--- a/src/commands/createNewProject/verifyIsProject.ts
+++ b/src/commands/createNewProject/verifyIsProject.ts
@@ -24,7 +24,7 @@ export async function isFunctionProject(folderPath: string): Promise<boolean> {
 export type MultiProjectPromptBehavior = 'silent' | 'prompt' | 'modalPrompt';
 
 /**
- * Checks root folder and subFolders one level down
+ * Checks root folder and one level down first, then all levels of tree
  * If a single function project is found, returns that path.
  * If multiple projects are found, prompt to pick the project.
  */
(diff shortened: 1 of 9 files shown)


src/commands/createNewProject/verifyIsProject.ts
```diff
@@ -18,50 +18,56 @@ export async function isFunctionProject(folderPath: string): Promise<boolean> {
     return await fse.pathExists(path.join(folderPath, hostFileName));
 }
 
+/**
+ * Describes what to do if multiple projects are found
+ */
+export type MultiProjectPromptBehavior = 'silent' | 'prompt' | 'modalPrompt';
+
 /**
  * Checks root folder and subFolders one level down
```
nturinski, line 27: Description is out of date.

nturinski (approved):

PR #2864, opened 2021-06-22
Fix venvUtil tests

ejizba:
The old terminal profile settings were deprecated and don't work the same, so I have to update the tests
https://code.visualstudio.com/updates/v1_56#_profile-improvements

The production code doesn't reference the settings at all since it uses `vscode.env.shell`, so that code can stay the same.


neelip (commented):
The build seems to be failing. Is it expected to fail?

neelip (approved):
Never mind. I noticed build for windows has been failing for a while.

PR #2865, opened 2021-06-22
Fix Add binding test

ejizba:
After https://github.com/microsoft/vscode-azurefunctions/pull/2856, the Adding Binding test started timing out. Basically it's the first test and "preLoadTemplates" wasn't quite pre-loading correctly. Fix was to pass `testWorkspaceFolders[0]` (the folder used for all non-parallel tests) as the projectPath

When I was investigating the issue, I also worked on improving the perf by re-using `templatesTask` more often. We used to have a separate templatesTask per projectPath, but .NET is the only one where the projectPath actually matters

### DIFF
--- a/src/templates/TemplateProviderBase.ts
+++ b/src/templates/TemplateProviderBase.ts
@@ -144,6 +144,10 @@ export abstract class TemplateProviderBase implements Disposable {
         return key;
     }
 
+    public supportsProjKey(): boolean {
+        return !!this.refreshProjKey;
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


src/templates/TemplateProviderBase.ts
```diff
@@ -144,6 +144,10 @@ export abstract class TemplateProviderBase implements Disposable {
         return key;
     }
 
+    public supportsProjKey(): boolean {
```
neelip, line 147: nit: comment on it's use

neelip (approved):

PR #2873, opened 2021-07-12
Use `lastStep` to improve funnel telemetry

ejizba:
Depends on https://github.com/microsoft/vscode-azuretools/pull/956 and https://github.com/microsoft/vscode-azuretools/pull/957


nturinski (commented):
Did you intentionally not add stepNames to any ui calls that were in wizard steps?

src/commands/remoteDebugJava/remoteDebugJavaFunctionApp.ts
```diff
@@ -40,7 +40,7 @@ export async function remoteDebugJavaFunctionApp(context: IActionContext, node?:
                 const appSettings: WebSiteManagementModels.StringDictionary = await client.listApplicationSettings();
                 if (needUpdateSiteConfig(siteConfig) || (appSettings.properties && needUpdateAppSettings(appSettings.properties))) {
                     const confirmMsg: string = localize('confirmRemoteDebug', 'The configurations of the selected app will be changed before debugging. Would you like to continue?');
-                    const result: vscode.MessageItem = await context.ui.showWarningMessage(confirmMsg, { modal: true }, DialogResponses.yes, DialogResponses.learnMore, DialogResponses.cancel);
+                    const result: vscode.MessageItem = await context.ui.showWarningMessage(confirmMsg, { modal: true }, DialogResponses.yes, DialogResponses.learnMore);
```
neelip, line 43: missing 'stepName'?

neelip (approved):

ejizba:
> Did you intentionally not add stepNames to any ui calls that were in wizard steps?

Yes because the wizard step names are already pretty accurate

src/commands/remoteDebugJava/remoteDebugJavaFunctionApp.ts
ejizba: 👀  Yeah I got a bit lazy for scenarios that I don't really care about... and this is one of those scenarios. there's just so many prompts

nturinski (approved):
There were a couple of spots where a `stepName` is missing.  I assume it's probably because it's the one and only step in a command (these ui dialogs aren't part of the wizard steps), in which case we already know where they canceled and it's probably not that useful to keep track of.

My only concern is would it be easier to write telemetry queries for cancelled events if every cancelled action had a `stepName` defined rather than having to infer it from the command id.  Would it make any sense to just add the command callbackId as the stepName if it is left undefined and isn't part of a wizard?

ejizba:
> I assume it's probably because it's the one and only step in a command 

Nah it's more that it's too annoying to specify a stepName for every single possible warning. And fwiw when there's no step name the fallback behavior is to use a shortened version of the message (see [here](https://github.com/microsoft/vscode-azuretools/blob/main/ui/src/userInput/AzExtUserInput.ts#L74)), so the property _is_ still defined in telemetry, just maybe not as succinctly or accurately as using a stepName.

> Would it make any sense to just add the command callbackId as the stepName if it is left undefined and isn't part of a wizard?

As stated, the property should never be left undefined.

Issue #2551, opened 2020-11-27
.NET 5 - Add deployment


anthonychu:
Add ability to deploy to an already created function app. Perhaps ensuring app settings / site configs are properly set.

ejizba:
Deployment was unblocked in https://github.com/microsoft/vscode-azurefunctions/pull/2586. Still haven't done work to verify/add the app settings & site config

PR #2876, opened 2021-07-13
Add validation of dotnet vs. dotnet-isolated runtime

ejizba:
I'm talking about this error:
Screen Shot 2021-07-12 at 5 18 21 PM

Previously it only checked "node" vs "dotnet", etc.

Also, during preDebugValidate, we will now write the correct worker runtime if it's missing, rather than always "dotnet".

Last piece needed to fix https://github.com/microsoft/vscode-azurefunctions/issues/2551


ejizba:
Merging to get a build, but feel free to review tomorrow

PR #2885, opened 2021-07-14
Upgrade tools package and typescript

nturinski:
Relies on https://github.com/microsoft/vscode-azuretools/pull/958


ejizba (approved):

nturinski:
I assume that the timeout is specific to Windows since it's slow and I didn't suddenly break bindings.  But let me know if that is not the case 😬

PR #2889, opened 2021-07-16
Use `confirmOverwriteSettings` from appservice package

ejizba:


src/commands/appSettings/confirmOverwriteSettings.ts
```diff
@@ -1,84 +0,0 @@
-/*---------------------------------------------------------------------------------------------
```
nturinski, line 1: 🥳

nturinski (approved):

PR #2902, opened 2021-08-03
Updating Extract to Newest Version

ChloeQuijano:
Updated the extract which previously had a bug. Included changes to the `src/templates/script/ScriptTemplateProvider.ts` file to adapt to the updated extract, specifically it no longer allows callbacks and forces to promises instead.


package-lock.json
```diff
@@ -1,13126 +1,8 @@
 {
     "name": "vscode-azurefunctions",
     "version": "1.4.2-alpha",
-    "lockfileVersion": 2,
+    "lockfileVersion": 1,
```
ejizba, line 4: Looks like your npm version is out of date. Please update (I'm on v7.19.1) and make sure there are minimal changes to this file

src/templates/script/ScriptTemplateProvider.ts
```diff
@@ -58,15 +58,24 @@ export class ScriptTemplateProvider extends TemplateProviderBase {
...
+                            resolve();
+                        },
+                        (err) => {
+                            if (err) {
+                                reject(err);
+                            }
+                        }
+                    )
+                }
+            )
```
ejizba, line 78: I think we can switch from a callback to a promise now. All of this would change to one line:
```
await extract(filePath, { dir: templatesPath });
```

package-lock.json
refortie: Pushed a new version with updated npm

ejizba (approved):

Issue #2896, opened 2021-07-29
Cannot have two projects with the same name


paulaustin-automutatio:
Does this occur consistently? Yes
Repro steps:

1. Create 2 Azure static web site apps with the functions in the sub folder api
2. Go to the Azure Functions list and refresh and it doesn't like it as they both have the same project name (api), even though they are in different folders

Action: AzureTreeDataProvider.getChildren
Error type: Error
Error Message: An element with the following id already exists: /localProjectapi

Version: 1.4.1
OS: win32
OS Release: 10.0.19043
Product: Visual Studio Code
Product Version: 1.58.2
Language: en

<details>
<summary>Call Stack</summary>

```
extension.bundle.js:2:4327421extension.bundle.js:2:4327421
AzExtTreeDataProvider.<anonymous> extension.bundle.js:2:4327281
fulfilled extension.bundle.js:2:4325085
```

</details>

PR #2911, opened 2021-08-09
Fix having two local projects with same name

alexweininger:
This is a fix for #2896.

Here is what it now looks like: 

image

### DIFF
--- a/src/tree/localProject/LocalProjectTreeItemBase.ts
+++ b/src/tree/localProject/LocalProjectTreeItemBase.ts
@@ -4,21 +4,21 @@
  *--------------------------------------------------------------------------------------------*/
 
 import * as path from 'path';
+import { WorkspaceFolder } from 'vscode';
 import { AzExtParentTreeItem, TreeItemIconPath } from 'vscode-azureextensionui';
 import { localize } from '../../localize';
 import { treeUtils } from '../../utils/treeUtils';
-import { isMultiRootWorkspace } from '../../utils/workspace';
 
 export abstract class LocalProjectTreeItemBase extends AzExtParentTreeItem {
     public readonly label: string = localize('localProject', 'Local Project');
     public suppressMaskLabel: boolean = true;
-    private readonly _projectName: string;
+    private readonly _projectSubpath: string;
     private readonly _id: string;
 
-    public constructor(parent: AzExtParentTreeItem, projectPath: string) {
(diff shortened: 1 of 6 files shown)


src/tree/localProject/LocalProjectTreeItemBase.ts
```diff
@@ -7,23 +7,26 @@ import * as path from 'path';
...
 export abstract class LocalProjectTreeItemBase extends AzExtParentTreeItem {
     public readonly label: string = localize('localProject', 'Local Project');
     public suppressMaskLabel: boolean = true;
     private readonly _projectName: string;
+    private readonly _id: string;
 
     public constructor(parent: AzExtParentTreeItem, projectPath: string) {
         super(parent);
-        this._projectName = path.basename(projectPath);
+        this._projectName = isMultiRootWorkspace() ? `${path.basename(path.dirname(projectPath))}/${path.basename(projectPath)}` : path.basename(projectPath);
```
ejizba, line 20: We know the containing WorkspaceFolder when creating the project tree items, so I think it'll simplify things if you pass that into this constructor. For example, I think this would be simpler and more robust:
```typescript
this._projectName = path.relative(path.dirname(folder.uri.fsPath), projectPath);
```

src/tree/localProject/LocalProjectTreeItemBase.ts
```diff
@@ -7,23 +7,26 @@ import * as path from 'path';
 import { AzExtParentTreeItem, TreeItemIconPath } from 'vscode-azureextensionui';
 import { localize } from '../../localize';
 import { treeUtils } from '../../utils/treeUtils';
+import { isMultiRootWorkspace } from '../../utils/workspace';
 
 export abstract class LocalProjectTreeItemBase extends AzExtParentTreeItem {
     public readonly label: string = localize('localProject', 'Local Project');
     public suppressMaskLabel: boolean = true;
     private readonly _projectName: string;
```
ejizba, line 15: nit: Rename to `_projectSubpath` or something similar now that it's not just the name

ejizba (approved):

ejizba:
@alexweininger in the future use a keyword in the PR description so that it'll close the issue automatically

https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #2916, opened 2021-08-12
Switch to shared ESLint config

wwlorey:


test/api.test.ts
```diff
@@ -8,6 +8,7 @@ import { Extension, extensions } from "vscode";
 import { runWithInputs } from 'vscode-azureextensiondev';
 import { AzureExtensionApiProvider } from "vscode-azureextensionui/api";
 import { nonNullValue, ProjectLanguage, registerOnActionStartHandler } from '../extension.bundle';
+// eslint-disable-next-line no-restricted-imports
 import { AzureFunctionsExtensionApi } from '../src/vscode-azurefunctions.api';
```
alexweininger, line 12: Is it worth trying to add a pattern to [this rule](https://github.com/microsoft/vscode-azuretools/blob/main/eslint-config-azuretools/test.js#L14) so we can remove these disable comments?

I tested out this glob and it worked: `**/src/!(*.api)`

wwlorey: I dig it. Let me get automatic publishing for the config package set up then I'll revisit this PR

wwlorey: After looking into this more - it doesn't look like ESLint plays nice with glob exclude patterns. [This pattern](https://globster.xyz/?q=**%2Fsrc%2F**%2F!(*.api)&f=dir%2Fsrc%2Futils%2FvscodeUtils%2Cdir%2Fsrc%2Fvscode-azurefunctions.api%2Csrc%2Fblah.api%2Csrc%2Fblah%2Cthis%2Fpath%2Fshould%2Fbe%2Ffine) works outside of ESLint. But when it's used with an ESLint rule, normal `src` imports aren't matched. And using multiple patterns didn't do the trick either. So I vote we just special case these

alexweininger: Ok, at least we tried 😄

alexweininger (approved):

PR #2919, opened 2021-08-16
Remove dependency of functions maven plugin for java templates

Flanker32:
### Issues to resolve
Currently we use functions maven plugin to get java function templates and generate new function, which has a hard dependency for maven plugin and could not work for java projects with other build tools like gradle. 
Besides, create new function with maven may have poor performance as maven goal may take much time to initialization.

### Solution
- Create Java Function class by ts directly instead of invoke `mvn functions:add`
- Refactor java template provider to get latest template from remote resource

### Next steps
In this PR I still use the version of functions maven plugin as the template version to keep compatibility with history data, we may need set the version for template in the future.

### DIFF
--- a/src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
+++ b/src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
@@ -2,15 +2,14 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
-
+import * as fse from 'fs-extra';
 import { IActionContext } from 'vscode-azureextensionui';
-import { ext } from '../../../extensionVariables';
-import { IFunctionTemplate } from '../../../templates/IFunctionTemplate';
 import { mavenUtils } from "../../../utils/mavenUtils";
 import { nonNullProp } from '../../../utils/nonNull';
 import { getJavaFunctionFilePath, IJavaProjectWizardContext } from '../../createNewProject/javaSteps/IJavaProjectWizardContext';
 import { FunctionCreateStepBase } from '../FunctionCreateStepBase';
 import { getBindingSetting, IFunctionWizardContext } from '../IFunctionWizardContext';
+import { IJavaFunctionTemplate, IJavaFunctionWizardContext } from './IJavaFunctionWizardContext';
 
 export class JavaFunctionCreateStep extends FunctionCreateStepBase<IFunctionWizardContext & IJavaProjectWizardContext> {
     private constructor() {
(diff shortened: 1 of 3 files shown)


src/templates/java/JavaTemplateProvider.ts
```diff
@@ -22,42 +22,41 @@ interface IRawJavaTemplates {
     templates: object[];
 }
 
-/**
- * Java templates largely follow the same formatting as script templates, but they come from maven
- */
+const repositoryUrl: string = "https://repo1.maven.org/maven2/com/microsoft/azure/azure-functions-maven-plugin/maven-metadata.xml";
+const templatesUrl: string = "https://aka.ms/java-function-templates";
```
ejizba, line 26: Please work with the Functions team to get these templates added to the main templates repo and templates feed

- Repo: https://github.com/Azure/azure-functions-templates
- Feed: https://functionscdn.azureedge.net/public/cli-feed-v4.json

I don't think using a GitHub url directly is reliable enough for use in the released version of our extension.

Flanker32: @ejizba Thanks for your feedback. We are now syncing with Function team to add Java to templates repo, but there seems to be some history issue there, I'll update here once there are updates.

Besides, now `JavaTemplateProvider` depends on functions maven plugin, which only works for maven project. Get templates from git repo could help users get the latest java templates for all projects, besides, we could fall back to backup template shipped with extension once there are issues with Github connection.  Or we could keep origin java template provider but just update the java function creation step, which will keep current behavior for maven and use shipped template for other java projects(like gradle)?

ejizba: I don't like using a link to GitHub. The cli-feed has an established process for moving templates to a "staging" environment first (that's what the "prerelease" tags are for) and we run unit tests against that environment to make sure everything looks good before it moves to "production". A direct link to GitHub does not have any process like that, so it's much more possible a change can break the released version of our extension.

I'm okay with using the backup templates however you want to do that. That option is much less prone to bugs since everything is just shipped with the extension and there's no "external" points of integration over the internet. Obviously the main downside is that you have to wait for the next release of the extension to update the templates

Flanker32: Yes, I agree that GitHub link is not a stable solution for function template, I've revert the changes for Java template provider but just update the FunctionCreationStep, will update the template provider once we add java templates to template repository.

src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
```diff
@@ -22,36 +21,34 @@ export class JavaFunctionCreateStep extends FunctionCreateStepBase<IFunctionWiza
...
+        const path: string = getJavaFunctionFilePath(context.projectPath, packageName, functionName);
+        await fse.writeFile(path, content)
         return getJavaFunctionFilePath(context.projectPath, packageName, functionName);
     }
 }
 
-function removeLanguageFromId(id: string): string {
-    return id.split('-')[0];
+function substituteParametersInTemplate(template: IJavaFunctionTemplate, args: Map<string, string>): string {
+    let javaTemplate = template.templateFiles["function.java"];
```
ejizba, line 49: Rather than hard-code the file name (which may change) - how about we just use the first file like `Object.values(template.templateFiles)[0]`

src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
```diff
@@ -2,15 +2,14 @@
  *  Copyright (c) Microsoft Corporation. All rights reserved.
```
ejizba, line 2: > Yes, I agree that GitHub link is not a stable solution for function template, I've revert the changes for Java template provider but just update the FunctionCreationStep, will update the template provider once we add java templates to template repository.

Just to clarify - do you want this merged with just the `FunctionCreationStep` stuff or do you want to wait until you can update the template provider? We're about to do a release of the extension so anything merged in the next day or two will be included in that release

src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
```diff
@@ -22,36 +21,34 @@ export class JavaFunctionCreateStep extends FunctionCreateStepBase<IFunctionWiza
         return new JavaFunctionCreateStep();
```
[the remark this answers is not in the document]
Flanker32: Yes, I've removed the static create step

src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
Flanker32: We may add other files (like `sample.dat`) to java templates, and using the first file may also meet the same issue

src/commands/createFunction/javaSteps/JavaFunctionCreateStep.ts
Flanker32: I think we could release this PR first as a perf enhancement for Java, and I'll update the template provider later.

ejizba (approved):

PR #2922, opened 2021-08-18
Add net6.0 to JsonCli

ejizba:
The NuGet and MSBuild tasks didn't play well with .NET 6 (I think because it's still in preview), so I switched to DotNetCoreCLI tasks and that seemed to fix it.

### DIFF
--- a/tools/JsonCli/.azure-pipelines/main.yml
+++ b/tools/JsonCli/.azure-pipelines/main.yml
@@ -31,18 +31,25 @@ jobs:
     inputs:
       version: 5.0.x
 
-  - task: NuGetCommand@2
-    displayName: 'restore'
+  - task: UseDotNet@2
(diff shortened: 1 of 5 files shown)


nturinski (approved):

nturinski:
I was curious though, do our test machines have the .NET SDK installed by default?  I was just looking at this post about the differences between the two: https://stackoverflow.com/questions/56166335/difference-between-netcore-visualstudio-build-and-msbuild-task-in-azure-pipeli and it seems like it'd be better if we could just use MSBuild.

ejizba:
I have tasks to make sure all the right .NET Core sdks are installed: https://github.com/microsoft/vscode-azurefunctions/blob/main/tools/JsonCli/.azure-pipelines/main.yml#L19

I'm not following why MSBuild is better

PR #2923, opened 2021-08-18
Add .NET 6 version of JsonCli

ejizba:
So that people can do v4 stuff with just .NET 6 installed


resources/dotnetJsonCli/net5.0/Microsoft.TemplateEngine.JsonCli.runtimeconfig.json
```diff
@@ -4,6 +4,9 @@
     "framework": {
       "name": "Microsoft.NETCore.App",
       "version": "5.0.0"
+    },
+    "configProperties": {
+      "System.Reflection.Metadata.MetadataUpdater.IsSupported": false
```
neelip, line 9: I saw that this property applies only to 6. Why is this specifically set?

neelip (approved):

resources/dotnetJsonCli/net5.0/Microsoft.TemplateEngine.JsonCli.runtimeconfig.json
ejizba: These files are auto-generated as a part of the build and it doesn't seem to hurt so I'd rather leave it in. Even though this file isn't targeting 6, the build machine has 6 installed so that could be part of why it's added

Issue #2887, opened 2021-07-15
Running Function Tools with HTTPS Fails to Pick Process for Debug


nerddtvg:
Sometimes we need to run the function apps with HTTPS locally to test OAuth which requires an encrypted URL scheme. However, the extension is hard-coded to look for the status page with the url `http://localhost:${funcPort}/admin/host/status` which will fail. Since the function core tools do not support both HTTP and HTTPS simultaneously, this causes issues. It would be best if the extension either (a) detected HTTPs via the configs/command line and used it or (b) tried both. This also means supporting self-signed certificates with no chain of trust.

A workaround is to use `pickProcess` instead of `azureFunctions.pickProcess` in `launch.json`, manually restart the function app on changes, and choose the PID from the provided list.

https://github.com/microsoft/vscode-azurefunctions/blob/4e7b9a52b465046e53ab1fa886bf8bf349a6452d/src/commands/pickFuncProcess.ts#L87

Error received when trying with HTTPS:

```
Failed to detect running Functions host within "60" seconds. You may want to adjust the "azureFunctions.pickProcessTimeout" setting.
```

ejizba:
Thanks for the detailed bug report, we'll look into it. In the meantime, the workaround you described is the best one I'm aware of

PR #2930, opened 2021-08-24
Support local projects using https

ejizba:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2887

Depends on https://github.com/microsoft/vscode-azuretools/pull/995

### DIFF
--- a/src/tree/localProject/LocalProjectTreeItem.ts
+++ b/src/tree/localProject/LocalProjectTreeItem.ts
@@ -13,7 +13,8 @@ import { IParsedHostJson, parseHostJson } from '../../funcConfig/host';
 import { getLocalSettingsJson, ILocalSettingsJson, MismatchBehavior, setLocalAppSetting } from '../../funcConfig/local.settings';
 import { getFuncPortFromTaskOrProject, isFuncHostTask, onFuncTaskStarted, runningFuncPortMap } from '../../funcCoreTools/funcHostTask';
 import { FuncVersion } from '../../FuncVersion';
-import { ApplicationSettings, IProjectTreeItem } from '../IProjectTreeItem';
+import { requestUtils } from '../../utils/requestUtils';
+import { ApplicationSettings, FuncHostRequest, IProjectTreeItem } from '../IProjectTreeItem';
 import { isLocalProjectCV, matchesAnyPart, ProjectResource, ProjectSource } from '../projectContextValues';
 import { createRefreshFileWatcher } from './createRefreshFileWatcher';
 import { LocalFunctionsTreeItem } from './LocalFunctionsTreeItem';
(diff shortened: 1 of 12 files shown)


src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -62,14 +63,27 @@ export class LocalProjectTreeItem extends LocalProjectTreeItemBase implements Di
...
 
-        return `http://localhost:${port}`;
+        const url = `http://localhost:${port}`;
+        try {
+            await requestUtils.sendRequestWithExtTimeout(context, { url: url, method: 'GET' });
+        } catch {
+            try {
+                const httpsUrl = url.replace('http', 'https');
+                await requestUtils.sendRequestWithExtTimeout(context, { url: httpsUrl, method: 'GET', rejectUnauthorized: false });
+                return { url: httpsUrl, rejectUnauthorized: false };
```
neelip, line 80: I think adding a comment mentioning to update this later would be a good idea if this is not an optimal approach in terms of security.

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -62,14 +63,27 @@ export class LocalProjectTreeItem extends LocalProjectTreeItemBase implements Di
...
         let port = runningFuncPortMap.get(this.workspaceFolder);
         if (!port) {
             const funcTask: Task | undefined = (await tasks.fetchTasks()).find(t => t.scope === this.workspaceFolder && isFuncHostTask(t));
             port = await getFuncPortFromTaskOrProject(context, funcTask, this.effectiveProjectPath);
         }
 
-        return `http://localhost:${port}`;
+        const url = `http://localhost:${port}`;
+        try {
+            await requestUtils.sendRequestWithExtTimeout(context, { url: url, method: 'GET' });
```
nturinski, line 75: If the http localhost doesn't exist, this just throws an error immediately, right?

ejizba: Yeah it errors out quickly in my experience

src/tree/localProject/LocalProjectTreeItem.ts
ejizba: I don't plan to update this later. The reason I'm okay with it is because it's specifically for the `localhost` case, meaning it's only reaching a url on the user's machine. This is similar to how the Databases extension sets it to false for the emulator, which is also using localhost: https://github.com/microsoft/vscode-cosmosdb/blob/cb70ebec689a3217709546b258497c9e2ad1afa7/src/docdb/getCosmosClient.ts#L17

neelip (approved):

nerddtvg:
Thank you very much for adding this functionality!

PR #2969, opened 2021-09-17
Add sort pre-commit hook

alexweininger:
Depends on https://github.com/microsoft/vscode-azuretools/pull/1002

Uses husky to call the sort script added to vscode-azureextensiondev in a pre-commit hook.


package.json
```diff
@@ -393,281 +393,281 @@
             ],
             "view/item/context": [
                 {
-                    "command": "azureFunctions.selectSubscriptions",
-                    "when": "view == azFuncTree && viewItem == azureextensionui.azureSubscription",
-                    "group": "inline"
+                    "command": "azureFunctions.addBinding",
```
nturinski, line 396: I'm not sure if we want to actually sort these alphabetically since they're currently (loosely) sorted by group and `viewItem`.

alexweininger: Makes sense. I'll exclude this from the sort

nturinski (approved):

PR #2972, opened 2021-09-22
Ignore Azurite Files in Function App

lechnerc77:
When using the Azurite Extension in Visual Studio Code (<https://marketplace.visualstudio.com/items?itemName=Azurite.azurite>) several files get created by this extension in the Azure Functions project directory that are only relevant for _local development_. In analogy to e.g.  `local.settings.json` they should be ignored by git by default and therefore be part of the standard `.gitignore` template of the extension. 

This pull request contains the required enhancement of the creation step of the `.gitignore` file for scripting languages.

As far as I saw the .NET language fetches the `.gitignore template from the repository <https://github.com/Azure/azure-functions-templates>, so the enhancement must be done there.


ejizba (approved):
Thanks!

PR #3016, opened 2021-12-03
Drop unsupported .NET versions from JsonCli tool

bwateratmsft:
.NET Core 2.* and 3.0 are EOL; it'd be best to stop building and stop shipping them. I removed the 2.0 entirely and bumped the 3.0 to 3.1, which is supported until 3 Dec 2022.

### DIFF
--- a/tools/JsonCli/.azure-pipelines/main.yml
+++ b/tools/JsonCli/.azure-pipelines/main.yml
@@ -17,14 +17,9 @@ jobs:
       TeamName: 'AzureTools'
 
   - task: UseDotNet@2
-    displayName: 'Use .NET Core sdk 2.0.x'
+    displayName: 'Use .NET Core sdk 3.1.x'
     inputs:
-      version: 2.0.x
-
-  - task: UseDotNet@2
-    displayName: 'Use .NET Core sdk 3.0.x'
-    inputs:
-      version: 3.0.x
+      version: 3.1.x
 
   - task: UseDotNet@2
     displayName: 'Use .NET sdk 5.0.x'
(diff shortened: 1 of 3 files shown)


tools/JsonCli/src/Microsoft.TemplateEngine.JsonCli.csproj
```diff
@@ -2,7 +2,7 @@
 
     <PropertyGroup>
         <OutputType>Exe</OutputType>
-        <TargetFrameworks>netcoreapp2.0;netcoreapp3.0;net5.0;net6.0</TargetFrameworks>
+        <TargetFrameworks>netcoreapp3.1;net5.0;net6.0</TargetFrameworks>
```
ejizba, line 5: There's several other places you'd need to update. I think for this PR, this file might be the only place:
https://github.com/microsoft/vscode-azurefunctions/blob/main/tools/JsonCli/.azure-pipelines/main.yml

After all that gets merged/built you'll need to update the bits here:
https://github.com/microsoft/vscode-azurefunctions/tree/main/resources/dotnetJsonCli

And the logic here:
https://github.com/microsoft/vscode-azurefunctions/blob/main/src/templates/dotnet/executeDotnetTemplateCommand.ts#L47

bwateratmsft: Good to know, thanks @ejizba!

tools/JsonCli/.azure-pipelines/main.yml
```diff
@@ -17,14 +17,9 @@ jobs:
...
     inputs:
-      version: 2.0.x
-
-  - task: UseDotNet@2
-    displayName: 'Use .NET Core sdk 3.0.x'
-    inputs:
-      version: 3.0.x
+      version: 3.1.x
 
   - task: UseDotNet@2
```
ejizba, line 24: Not entirely related, but you should be able to remove `includePreviewVersions: true` from the .NET 6 task since its no longer preview

ejizba (approved):

tools/JsonCli/src/Microsoft.TemplateEngine.JsonCli.csproj
bwateratmsft: Updated the signing project too.

bwateratmsft: Followup PR is #3018.

PR #3018, opened 2021-12-06
Drop unsupported .NET Core versions and nearly-unsupported Python version

bwateratmsft:
Followup from #3016. Removes the unsupported .NET Core 2.0 and 3.0 versions, replacing 3.0 with 3.1. Updates the version detection logic correspondingly.

Since [6.0 is now LTS](https://dotnet.microsoft.com/platform/support/policy/dotnet-core), I reordered the prioritization of versions to 6.0, 5.0, 3.1.

**NOTE**: Git is trying really hard to describe these file changes as renames in a weird and confusing way; what I did was completely delete the contents of `resources/dotnetJsonCli`, recreate the folders within it, and then copy/paste in the files from the published drop.

**NOTE**: I didn't initially intend to include Python changes as part of this, but changing the pool to `macOS-latest` got the .NET tests working...at the expense of Python 3.6. However, Python 3.6 is out-of-support in just a couple of weeks, so thematically it seemed to fit this PR too. I did not remove some references to 3.6 that appeared to be part of a supportability matrix with Functions.


src/templates/dotnet/executeDotnetTemplateCommand.ts
```diff
@@ -60,25 +61,26 @@ async function getFramework(context: IActionContext, workingDirectory: string |
         }
 
         // Prioritize "LTS", then "Current", then "Preview"
```
ejizba, line 63: Looks good to me other than the test failures. The Windows timeouts are probably unrelated, but I think Mac was working just fine before this change. Unfortunately nothing jumps out at me as to _why_ they're failing

ejizba (approved):

src/templates/dotnet/executeDotnetTemplateCommand.ts
bwateratmsft: Sounds good. I'll figure out what's up.

bwateratmsft: The Windows failure is probably transient; one of the builds (for `bmw/netVersions2`) succeeded, one (for the PR) failed. The Mac one is definitely suspicious; I'm looking into it now.

bwateratmsft: Still need to work out why, but despite the "Use .NET SDK..." tasks, the .NET version info visible in the test is as follows:
```
2.1.507
2.1.302 [/Users/runner/.dotnet/sdk]
2.1.403 [/Users/runner/.dotnet/sdk]
2.1.507 [/Users/runner/.dotnet/sdk]
```

bwateratmsft: In contrast, on Linux, the versions and paths above match. So it looks like something is up with `PATH`...on Mac...imagine that.

alexweininger (approved):

PR #3032, opened 2022-01-07
Add option to exclude app settings from upload based on name

alexweininger:
Needed to fix https://github.com/microsoft/vscode-azurestaticwebapps/issues/509

Added an `exclude: (string | RegExp)[]` parameter to `uploadAppSettings` and exposed it in the Functions API.


src/commands/appSettings/uploadAppSettings.ts
```diff
@@ -50,10 +50,25 @@ export async function uploadAppSettingsInternal(context: IActionContext, client:
             remoteSettings.properties = {};
         }
 
+        const excludedAppSettings: string[] = [];
+        if (exclude) {
+            Object.keys(localSettings.Values).forEach((settingName) => {
+                if (exclude.some((exclusion) => typeof exclusion === 'string' ? settingName === exclusion : settingName.match(exclusion))) {
```
nturinski, line 56: Settings are case-insensitive so you should probably change the settingName to all lower case along with the exclusion when you do the comparisons.

nturinski (approved):

src/commands/appSettings/uploadAppSettings.ts
alexweininger: Good catch, changed it to `settingName.match(new RegExp(exclusion, 'i'))`. This creates a new RegExp that is case-insensitive.

src/commands/api/uploadAppSettingsFromApi.ts
```diff
@@ -6,8 +6,8 @@ import { callWithTelemetryAndErrorHandling, IActionContext } from "vscode-azuree
 import { IAppSettingsClient } from "../../vscode-azurefunctions.api";
 import { uploadAppSettingsInternal } from "../appSettings/uploadAppSettings";
 
-export async function uploadAppSettingsFromApi(client: IAppSettingsClient): Promise<void> {
+export async function uploadAppSettingsFromApi(client: IAppSettingsClient, exclude?: (RegExp | string)[]): Promise<void> {
     return await callWithTelemetryAndErrorHandling('api.uploadAppSettings', async (context: IActionContext) => {
-        await uploadAppSettingsInternal(context, client);
+        await uploadAppSettingsInternal(context, client, undefined, exclude);
```
bwateratmsft, line 11: One thing I like to do is put the called function variable name next to `undefined` when it is passed, so it's understandable what _would_ go there if it wasn't `undefined` without needing to look at the called function.
i.e.
```suggestion
        await uploadAppSettingsInternal(context, client, /* workspaceFolder */ undefined, exclude);
```

bwateratmsft (approved):

src/commands/api/uploadAppSettingsFromApi.ts
alexweininger: Good suggestion. 

Also, not sure if you know about this but VS Code recently added a setting to provide inlay hints for parameter names:

image

image

bwateratmsft: I did not know about this! That is awesome and I'm turning it on now.

PR #3046, opened 2022-02-14
package migration for @microsoft packages and T2

nturinski:
Like Brandon, I'll try to point out anything that isn't just a changed import or API call.


src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -83,46 +84,22 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
...
-                            name: "String"
-                        }
-                    }
-                }
-            }
-        };
-
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
-        (<any>site).extendedLocation = { name: customLocation.id, type: 'customLocation' };
+        site.extendedLocation = { name: customLocation.id, type: 'customLocation' };
```
nturinski, line 89: Similar to Brandon's PR with the appservice package, this mapper was used as a workaround to allow some newer properties without actually having to upgrade.  

https://github.com/microsoft/vscode-azuretools/pull/1061/files#r802981607

src/tree/SlotsTreeItem.ts
```diff
@@ -45,27 +46,23 @@ export class SlotsTreeItem extends AzExtParentTreeItem {
         }
 
         const client: WebSiteManagementClient = await createWebSiteClient([context, this]);
-        const webAppCollection: WebSiteManagementModels.WebAppCollection = this._nextLink ?
-            await client.webApps.listSlotsNext(this._nextLink) :
-            await client.webApps.listSlots(this.parent.site.resourceGroup, this.parent.site.siteName);
-
-        this._nextLink = webAppCollection.nextLink;
+        const webAppCollection: Site[] = await uiUtils.listAllIterator(client.webApps.listSlots(this.parent.site.resourceGroup, this.parent.site.siteName));
```
nturinski, line 49: As Brandon mentioned, the paging doesn't actually work so we can't implement loadMore as of now.  I guess we should track this, though I'm not entirely sure where

src/tree/localProject/LocalFunctionsTreeItem.ts
```diff
@@ -87,11 +87,11 @@ export class LocalFunctionsTreeItem extends FunctionsTreeItemBase {
                 rejectUnauthorized: hostRequest.rejectUnauthorized
             });
             return await this.createTreeItemsWithErrorHandling(
-                <WebSiteManagementModels.FunctionEnvelope[]>functions.parsedBody,
+                <FunctionEnvelope[]>functions.parsedBody,
                 'azFuncInvalidLocalFunction',
                 async func => {
                     func = requestUtils.convertToAzureSdkObject(func);
-                    return LocalFunctionTreeItem.create(context, this, nonNullProp(func, 'name'), func.config, undefined, func);
+                    return LocalFunctionTreeItem.create(context, this, nonNullProp(func, 'name'), new ParsedFunctionJson(func.config), undefined, func);
```
nturinski, line 94: So the typing here changed to  Record<string, unknown> which is basically just a Set which is basically represented as an object.  Passing in that data type to ParsedFunctionJson() seems to create the JSON object that we need for local tree items, which is why I ended up just creating it rather than changing the tree items to accept a Record<string, unknown> typing.

src/tree/remoteProject/RemoteFunctionsTreeItem.ts
```diff
@@ -43,13 +43,11 @@ export class RemoteFunctionsTreeItem extends FunctionsTreeItemBase {
...
         const client = await this.parent.site.createClient(context);
-        let funcs: WebSiteManagementModels.FunctionEnvelopeCollection;
+        let funcs: FunctionEnvelope[];
         const maxTime = Date.now() + 2 * 60 * 1000;
         let attempt = 1;
         while (true) {
-            funcs = this._nextLink ?
-                await client.listFunctionsNext(this._nextLink) :
-                await client.listFunctions();
+            funcs = await client.listFunctions();
```
nturinski, line 50: This doens't use the `uiUtils` because it's implemented in the `siteClient`

src/tree/SlotsTreeItem.ts
alexweininger: We should create an issue on the SDK repository for this. Brandon mentioned creating an issue, not sure if he got to it.

alexweininger (approved):

bwateratmsft (approved):

src/tree/SlotsTreeItem.ts
nturinski: I'll include the issue anywhere in code where we need to implement load more

nturinski: Yeah, I anticipate the following tree items could be potential issues:

- Resource groups
- Storage blobs/file shares
- Database documents

At least in the context of extensions that we already have

PR #3080, opened 2022-03-15
Update bots to use common VSCode ones

bwateratmsft:
The `locker` and `needs-more-info-closer` bots have been working smoothly for the Docker extension. We don't use the `stale-closer` bot, so the only good place to test it is here.

Once we've verified that all three bots are working as expected here, I'll open PRs to add them to the rest of the AzCode repos.

### DIFF
--- a/.github/workflows/locker.yml
+++ b/.github/workflows/locker.yml
@@ -8,16 +8,17 @@ jobs:
   main:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


nturinski (approved):

bwateratmsft:
I'm going to keep an eye on what the bot does; if necessary we can undo any changes it makes.

PR #3087, opened 2022-03-24
Add and register resolver and remove treeProvider

nturinski:
Followed the steps laid out by Alex in https://github.com/microsoft/vscode-azureresourcegroups/issues/135.  I'm sure there'll be some things we have to go back and change, but seems like the commands play nice cross extensions (for the most part) and tsc seems happy enough.


package.json
```diff
@@ -85,6 +85,13 @@
     ],
```
bwateratmsft, line 85: Above (can't add the comment there) there's an activation event for `onView:azFuncTree` that can be removed.

package.json
```diff
@@ -85,6 +85,13 @@
...
     "main": "./main.js",
     "contributes": {
+        "x-azResources": {
+            "activation": {
+                "onFetch": [
```
bwateratmsft, line 90: Have we considered `onResolve`?

package.json
```diff
@@ -388,262 +395,262 @@
...
+                    "when": "view == azureResourceGroups && viewItem == logFiles",
                     "group": "1@1"
                 },
                 {
                     "command": "azureFunctions.refresh",
-                    "when": "view == azFuncTree && viewItem == folder",
+                    "when": "view == azureResourceGroups && viewItem == folder",
                     "group": "1@1"
                 }
             ],
```
bwateratmsft, line 656: Below (can't add the comment there) the view registration for `azFuncTree` can be removed.

src/tree/SlotTreeItemBase.ts
```diff
@@ -3,215 +3,103 @@
...
+import { IParsedHostJson } from '../funcConfig/host';
+import { FuncVersion } from '../FuncVersion';
 import { ApplicationSettings, FuncHostRequest, IProjectTreeItem } from './IProjectTreeItem';
-import { matchesAnyPart, ProjectResource, ProjectSource } from './projectContextValues';
-import { RemoteFunctionsTreeItem } from './remoteProject/RemoteFunctionsTreeItem';
+import { ProjectSource } from './projectContextValues';
+import { ResolvedFunctionAppResource } from './ResolvedFunctionAppResource';
 
-export abstract class SlotTreeItemBase extends AzExtParentTreeItem implements IProjectTreeItem {
+export class SlotTreeItemBase extends AzExtParentTreeItem implements IProjectTreeItem {
```
bwateratmsft, line 14: Since this is no longer abstract shouldn't we just rename it to `SlotTreeItem`?

nturinski: Yeah, it's a little bit confusing since it is technically the production (top level) app _and_ the `SlotTreeItem`.  I'm not sure keeping it as `base` really helps distinguish that though 😅

package.json
nturinski: I'm not totally savvy to how this works, but I figured that the fetch would always happen before the resolve.  But I'm totally okay to add it to the manifest.

bwateratmsft: I'd suggest using `onResolve` anytime possible--since it's the latest and most accurate activation--and only using `onFetch` (or even earlier, `onView`) if necessary.

alexweininger:
Is this PR meant to be against the main branch?

nturinski:
We never started a "unified" branch on a per extension level, but we probably should, huh?  Once this is approved, I'll just make it the unified branch.  Sound good?

alexweininger:
> We never started a "unified" branch on a per extension level, but we probably should, huh? Once this is approved, I'll just make it the unified branch. Sound good?

👍  I've been doing that on the other extensions too

package.json
bwateratmsft: Having both `onFetch` and `onResolve` is equivalent to `onFetch` only, since it will happen first.

PR #3109, opened 2022-04-14
Implement unified workspace API

alexweininger:
Want to get this in for the bug bash!

image


package.json
```diff
@@ -369,6 +371,16 @@
                     "when": "view == azureResourceGroups && viewItem == azureextensionui.azureSubscription",
                     "group": "inline"
                 },
+                {
+                    "command": "azureFunctions.createFunction",
+                    "when": "view == azFuncTree && viewItem == azFuncLocalProject",
```
nturinski, line 376: Should the view be `azFuncTree`? I pretty much removed all traces of it-- not azureWorkspace?

src/LocalResourceProvider.ts
```diff
@@ -0,0 +1,97 @@
...
+import { dotnetUtils } from "./utils/dotnetUtils";
+import { getWorkspaceSetting } from "./vsCodeConfig/settings";
+
+export class FunctionsLocalResourceProvider implements LocalResourceProvider {
+
+    public disposables: Disposable[] = [];
+
+    public async provideResources(parent: AzExtParentTreeItem): Promise<AzExtTreeItem[] | null | undefined> {
+
+        return await callWithTelemetryAndErrorHandling('AzureAccountTreeItemWithProjects.provideResources', async (context: IActionContext) => {
```
nturinski, line 26: Are you able to remove `AzureAccountTreeItemWithProjects` now?  That's where this implementation is from, right?

alexweininger: We can, but lets do it later since the tests rely on it... 😭

nturinski (approved):

PR #3112, opened 2022-04-14
Remove `portfinder`

bwateratmsft:


bwateratmsft:
After merging I think dependabot will update #3111, which we can then merge.

alexweininger (approved):

Issue #3115, opened 2022-04-15
`createFunction` API loses options if no folder is given


Charles-Gagnon:
1. Call `createFunction` API and pass in some options, such as language (but NOT `folderPath`) in an empty workspace (no project open)
2. When prompted, choose to `Create New Project`

Expected : Create project flow keeps options passed in

Actual : The options are lost, so for example if you specified a language it will still prompt to select a language later on in the flow

The issue seems to be here : https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/createFunction.ts#L80

The `azureFunctions.createNewProject` command is called to restart the flow but the options are never passed in (and the current execution is ended due to throwing an error farther down.

PR #3116, opened 2022-04-15
Fix "Create Project" flow losing options

Charles-Gagnon:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3115

Pass in the options to the create project flow so they're passed on correctly there


nturinski (approved):
Thanks a lot for your contribution!

Issue #3139, opened 2022-04-24
There is a duplicate "Refresh" action in "Files/Logs/Deployments" context menu after installing App Service extension unified build


v-xinda:
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32 (unified)
**Azure App Service Extension Build Version:** 20220421.6 (unified)
**Azure Function Extension build Version:** 20220421.8 (unified)
**Regression:** Not a regression

**Repro Steps:**
1. Expand a Function App node in Azure Resources view.
2. Right click the "Files" node to open its context menu.
3. Check whether there is no duplicate action in the context menu.

**Expect:**
There is no duplicate action in the context menu.
![image](https://user-images.githubusercontent.com/57662131/164960369-6faa690b-c672-43c5-833c-47857168d8c2.png)

**Actual:**
There is a duplicate "Refresh" action in the context menu.
![image](https://user-images.githubusercontent.com/57662131/164960377-481aee22-faf2-4730-b05d-9474ba37554a.png)

**More Info:**
This issue also reproduces for "Logs" and "Deployments" nodes.

alexweininger:
Closed by #3141

PR #3141, opened 2022-04-25
Fix context menus

### LINKED ISSUE #3140: There are two extra actions in "Deployments" context menu after installing App Service extension unified build
**OS:** All
**Azure App Service Extension Build Version:** 20220421.6 (unified)

**Repro Steps:**
1. Expand a Function App node in Azure Resources view.
2. Configure deployment source is None.
3. Right click the "Deployments" node to open its context menu.
4. Check whether there is no extra action in the context menu.

**Expect:**
There is no extra action in the context menu.
image

**Actual:**
There are two extra actions in the context menu.
image

alexweininger:
Fixes #3140 #3139

### DIFF
--- a/src/commands/openInPortal.ts
+++ b/src/commands/openInPortal.ts
@@ -8,7 +8,7 @@ import { AzExtTreeItem, IActionContext, nonNullProp } from '@microsoft/vscode-az
 import { ext } from '../extensionVariables';
 import { ResolvedFunctionAppResource } from '../tree/ResolvedFunctionAppResource';
 
-export async function openInPortal(context: IActionContext, node?: AzExtTreeItem): Promise<void> {
+export async function openDeploymentInPortal(context: IActionContext, node?: AzExtTreeItem): Promise<void> {
     if (!node) {
         node = await ext.rgApi.tree.showTreeItemPicker<AzExtTreeItem>(new RegExp(ResolvedFunctionAppResource.productionContextValue), context);
     }
--- a/src/commands/openInPortal.ts
+++ b/src/commands/openInPortal.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { openInPortal as uiOpenInPortal } from '@microsoft/vscode-azext-azureutils';
-import { AzExtTreeItem, IActionContext } from '@microsoft/vscode-azext-utils';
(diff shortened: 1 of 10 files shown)


src/tree/ResolvedFunctionAppResource.ts
```diff
@@ -190,12 +190,24 @@ export class ResolvedFunctionAppResource implements ResolvedAppResourceBase {
         const sourceControl: SiteSourceControl = await client.getSourceControl();
         const proxyTree: SlotTreeItem = this as unknown as SlotTreeItem;
 
-        this.deploymentsNode = new DeploymentsTreeItem(proxyTree, this.site, siteConfig, sourceControl);
+        this.deploymentsNode = new DeploymentsTreeItem(proxyTree, {
+            site: this.site,
+            siteConfig,
+            sourceControl,
+            contextValuesToAdd: ['azureFunctions']
```
nturinski, line 197: I think it makes sense to either use `azureFunctions` everywhere in this repo or `azFunc`.  Mixing and matching could get confusing.

alexweininger: Ok I switched them all to `azFunc`. I'd prefer `azureFunctions`, but there is already a lot of usages of `azFunc` and I'd rather not introduce (more) new bugs.

src/commands/openInPortal.ts
```diff
@@ -13,5 +13,5 @@ export async function openInPortal(context: IActionContext, node?: AzExtTreeItem
         node = await ext.rgApi.tree.showTreeItemPicker<AzExtTreeItem>(new RegExp(ResolvedFunctionAppResource.productionContextValue), context);
     }
 
-    await uiOpenInPortal(node, node.fullId);
+    await uiOpenInPortal(node, `${nonNullProp(node, 'parent').parent?.id}/Deployments/${nonNullProp(node, 'id')}`);
```
nturinski, line 16: What's up with this? Doesn't this only work for very specific tree nodes?

alexweininger: The Functions open in portal command is now only used for deployment tree items, we could rename the command to make that more clear.

nturinski: Could you add a comment to that effect?  This would be super confusing.  Or maybe just change `openInPortal` to `openDeploymentInPortal`

nturinski (approved):

Issue #3127, opened 2022-04-22
An error occurs when executing "Start Streaming Logs" action for a Function App


v-ruizh:
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32 (unified)
**Azure Function Extension build Version:** 20220421.8 (unified)
**Regression:** Not a Regression

**Repro Steps:**
1. Create a Function App.
2. Right click this Function App and execute "Start Streaming Logs" action.
3. Check whether succeeds to execute "Start Streaming Logs" action or not.

**Expect:**
Succeed to execute "Start Streaming Logs" action.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/45614674/164607435-904909de-410c-44d8-ae48-05c4bf9a8ecc.png)

PR #3142, opened 2022-04-27
Fix slot context menus

### LINKED ISSUE #3134: Commands miss from the slot context menu
**OS:** All

**Repro Steps:**
1. Right click a slot -> Check the context menu.
2. Check whether no command misses from the context menu.

**Expect:**
No command misses from the context menu.

**Actual:**
Commands miss from the context menu.
image

alexweininger:
Fixes #3134, #3127

### DIFF
--- a/src/tree/SlotTreeItem.ts
+++ b/src/tree/SlotTreeItem.ts
@@ -10,7 +10,12 @@ import { FuncVersion } from '../FuncVersion';
 import { treeUtils } from '../utils/treeUtils';
 import { ApplicationSettings, FuncHostRequest, IProjectTreeItem } from './IProjectTreeItem';
 import { ProjectSource } from './projectContextValues';
+import { RemoteFunctionTreeItem } from './remoteProject/RemoteFunctionTreeItem';
 import { ResolvedFunctionAppResource } from './ResolvedFunctionAppResource';
+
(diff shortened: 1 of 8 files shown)


src/tree/SlotTreeItem.ts
```diff
@@ -86,7 +91,7 @@ export class SlotTreeItem extends AzExtParentTreeItem implements IProjectTreeIte
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        return await this.resolved.loadMoreChildrenImpl(_clearCache, context);
+        return await this.resolved.loadMoreChildrenImpl.call(this, _clearCache, context) as AzExtTreeItem[];
```
nturinski, line 94: Do you think we should just go ahead and use `.call` on all the Impl methods?

alexweininger: Since `loadMoreChildrenImpl` is the only method where we need to cast `this as unknown as SlotTreeItem`, I'm pretty confident it's the only place we need to use `.call`. The typings here actually work in our favor and will warn us in the future (we will be forced to cast `this as unknown as SlotTreeItem`).

Issue #3132, opened 2022-04-22
An error occurs when executing "Delete Setting..." command from command palette


v-ruizh:
**OS:** All
**Azure Resource Extension Build Version:** 20220421.32 (unified)
**Azure Function Extension build Version:** 20220421.8 (unified)
**Regression:** Previous Build

**Repro Steps:**
1. Create a Function App.
2. F1 -> Azure Functions: Delete Setting... -> Select a resource group -> Select a Function App.
3. Check whether the "Select App Setting" list shows or not.

**Expect:**
The "Select App Setting" option shows.
![image](https://user-images.githubusercontent.com/45614674/164619465-273c752e-6690-4a79-a25d-b8aa332303f9.png)

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/45614674/164619342-55c18fe0-f2f4-47c5-9f63-8ed4f3943699.png)

**More Info:**
This issue does not reproduce on the previous Azure Function unified build 20220421.1.

v-mengwe:
This issue also reproduces for the command ""Azure Function:Edit Settings..." from the command palette.

PR #3149, opened 2022-05-03
Replace showTreeItemPicker with pickAppResource

### LINKED ISSUE #3135: The new created function app is not focused after clicking "Click to view resource" from the "AZURE: ACTIVITYLOG" window
**OS:** All

**Repro Steps:**
1. Create a new function app.
2. Navigate to the "AZURE: ACTIVITYLOG" window -> Expand the log -> Click "Click to view resource".
3. Check whether the new created function app is focused or not.

**Expect:**
The new created function app is focused.

**Actual:**
The new created function app is not focused.

alexweininger:
Fixes #3135 and #3132

### DIFF
--- a/src/commands/addBinding/addBinding.ts
+++ b/src/commands/addBinding/addBinding.ts
@@ -36,7 +36,13 @@ export async function addBinding(context: IActionContext, data: Uri | LocalFunct
     } else {
         if (!data) {
             const noItemFoundErrorMessage: string = localize('noLocalProject', 'No matching functions found. C# and Java projects do not support this operation.');
-            data = await ext.rgApi.tree.showTreeItemPicker<LocalFunctionTreeItem>(/Local;ReadWrite;Function;/i, { ...context, noItemFoundErrorMessage });
+            data = await ext.rgApi.pickAppResource<LocalFunctionTreeItem>({ ...context, noItemFoundErrorMessage }, {
+                filter: {
+                    type: 'microsoft.web/sites',
+                    kind: 'functionapp'
+                },
+                expectedChildContextValue: /Local;ReadWrite;Function;/i
+            });
         }
 
         if (!data.functionJsonPath) {
--- a/src/commands/addBinding/addBinding.ts
+++ b/src/commands/addBinding/addBinding.ts
@@ -5,7 +5,7 @@
 
 import { AzureWizard, IActionContext } from "@microsoft/vscode-azext-utils";
 import { Uri, WorkspaceFolder } from "vscode";
-import { ProjectLanguage, projectTemplateKeySetting } from "../../constants";
+import { functionFilter, ProjectLanguage, projectTemplateKeySetting } from "../../constants";
 import { ext } from "../../extensionVariables";
 import { FuncVersion } from "../../FuncVersion";
 import { localize } from "../../localize";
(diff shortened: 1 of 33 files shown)


src/commands/addBinding/addBinding.ts
```diff
@@ -36,7 +36,10 @@ export async function addBinding(context: IActionContext, data: Uri | LocalFunct
     } else {
         if (!data) {
             const noItemFoundErrorMessage: string = localize('noLocalProject', 'No matching functions found. C# and Java projects do not support this operation.');
-            data = await ext.rgApi.tree.showTreeItemPicker<LocalFunctionTreeItem>(/Local;ReadWrite;Function;/i, { ...context, noItemFoundErrorMessage });
+            data = await ext.rgApi.pickAppResource<LocalFunctionTreeItem>({ ...context, noItemFoundErrorMessage }, {
```
nturinski, line 39: This is going to be in the workspaceProvider, so don't think that this will work.  Since we don't change the context value in the workspace, should be able to just do `ext.rgApi.workspaceTree.showTreeItemPicker`

src/commands/browseWebsite.ts
```diff
@@ -4,14 +4,18 @@
...
 import { ResolvedFunctionAppResource } from '../tree/ResolvedFunctionAppResource';
 import { SlotTreeItem } from '../tree/SlotTreeItem';
 import { openUrl } from '../utils/openUrl';
 
 export async function browseWebsite(context: IActionContext, node?: SlotTreeItem): Promise<void> {
     if (!node) {
-        node = await ext.rgApi.tree.showTreeItemPicker<SlotTreeItem>(new RegExp(ResolvedFunctionAppResource.productionContextValue), context);
+        node = await ext.rgApi.pickAppResource<SlotTreeItem>(context, {
+            filter: functionFilter,
+            expectedChildContextValue: new RegExp(ResolvedFunctionAppResource.productionContextValue)
```
nturinski, line 17: Do you need the `expectedChildContextValue`?  Isn't the `productionContextValue` the `AppResource` itself?

src/commands/viewProperties.ts
```diff
@@ -14,7 +15,10 @@ import { SlotTreeItem } from '../tree/SlotTreeItem';
 
 export async function viewProperties(context: IActionContext, node?: SlotTreeItem | RemoteFunctionTreeItem | LocalFunctionTreeItem): Promise<void> {
     if (!node) {
-        node = await ext.rgApi.tree.showTreeItemPicker<SlotTreeItem>(new RegExp(ResolvedFunctionAppResource.productionContextValue), context);
+        node = await ext.rgApi.pickAppResource<SlotTreeItem>(context, {
```
nturinski, line 18: We're getting rid of this command in favor of RG, right?

src/commands/swapSlot.ts
```diff
@@ -5,13 +5,17 @@
...
 import { ext } from '../extensionVariables';
 import { ResolvedFunctionAppResource } from '../tree/ResolvedFunctionAppResource';
 import { SlotTreeItem } from '../tree/SlotTreeItem';
 
 export async function swapSlot(context: IActionContext, sourceSlotNode?: SlotTreeItem): Promise<void> {
     if (!sourceSlotNode) {
-        sourceSlotNode = await ext.rgApi.tree.showTreeItemPicker<SlotTreeItem>(ResolvedFunctionAppResource.pickSlotContextValue, { ...context, suppressCreatePick: true });
+        sourceSlotNode = await ext.rgApi.pickAppResource<SlotTreeItem>({ ...context, suppressCreatePick: true }, {
+            filter: functionFilter,
+            expectedChildContextValue: new RegExp(ResolvedFunctionAppResource.productionContextValue)
```
nturinski, line 17: I think that this should be `ResolvedFunctionAppResource.slotContextValue`

src/commands/viewProperties.ts
alexweininger: Yeah I was going to do that afterwards

src/commands/browseWebsite.ts
alexweininger: This unintentionally forces it to get resolved. And it doesn't behave weirdly. If we leave it out then we have to manually resolve it after calling `pickAppResource`.

src/commands/viewProperties.ts
alexweininger: So actually, the view properties command needs to still exist, but just not on the app resource. So it should never show up in the tree.

PR #3153, opened 2022-05-04
Hide openInPortal command from palette

alexweininger:
This command now only handles opening deployments in the portal, so I think it's ok to remove from the palette.

Alternatively, we could rebrand this command as "Open Deployment in Portal", but we don't have any other commands like that so I felt it wasn't necessary.


bwateratmsft (approved):

alexweininger:
Note: pushed an extra commit up after Brandon approved, reverted to the correct commit (8cd963d)

PR #3163, opened 2022-05-11
Check for sql binding triggers and skip storage prompt if so

nturinski:
SqlInputBinding triggers and SqlOutputBinding triggers both don't actually need a storage account for debugging, but we do a blanket check of it not being an httpTrigger, then prompt for an Azure storage account.

A partner team is using the Functions API to create a project for their users and having that prompt is a bit of a roadblocker for their UI.  This should hopefully fix that.

### DIFF
--- a/src/templates/IFunctionTemplate.ts
+++ b/src/templates/IFunctionTemplate.ts
@@ -19,6 +19,7 @@ export interface IFunctionTemplate {
     language: string;
     isHttpTrigger: boolean;
     isTimerTrigger: boolean;
+    isSqlBindingTrigger: boolean;
     userPromptedSettings: IBindingSetting[];
     categories: TemplateCategory[];
 }
--- a/src/templates/IFunctionTemplate.ts
+++ b/src/templates/IFunctionTemplate.ts
@@ -19,7 +19,7 @@ export interface IFunctionTemplate {
     language: string;
     isHttpTrigger: boolean;
     isTimerTrigger: boolean;
-    isSqlBindingTrigger: boolean;
+    isSqlBindingTemplate: boolean;
     userPromptedSettings: IBindingSetting[];
     categories: TemplateCategory[];
 }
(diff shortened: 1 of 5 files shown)


src/templates/IFunctionTemplate.ts
```diff
@@ -19,6 +19,7 @@ export interface IFunctionTemplate {
     language: string;
     isHttpTrigger: boolean;
     isTimerTrigger: boolean;
+    isSqlBindingTrigger: boolean;
```
Charles-Gagnon, line 22: I'd suggest `isSqlBindingTemplate` - we don't (currently) have trigger support so this is slightly misleading as it is.

Charles-Gagnon, line 22: You also may want to consider refactoring this slightly to just be `templateType` or something - and then instead of having separate values like `isHttpTrigger` and `isTimerTrigger` you just have the type be `httpTrigger`, `timerTrigger`, `sqlBinding`, etc.

bwateratmsft (dismissed):

src/templates/dotnet/parseDotnetTemplates.ts
```diff
@@ -74,6 +74,7 @@ function parseDotnetTemplate(rawTemplate: IRawTemplate): IFunctionTemplate {
     return {
         isHttpTrigger: /^http/i.test(rawTemplate.Name) || /webhook$/i.test(rawTemplate.Name),
         isTimerTrigger: /^timer/i.test(rawTemplate.Name),
+        isSqlBindingTrigger: /Sql.*Binding/i.test(rawTemplate.Name),
```
bwateratmsft, line 77: `/Sql.*Binding/i` should maybe be a constant since it's used here and in `parseScriptTemplates.ts`.

src/templates/IFunctionTemplate.ts
nturinski: I agree long-term, but since we're just trying to get this in soon with minimal impact, this seemed like the easiest solution.

nturinski:
Moving PR into unified branch since it was merged from main already

PR #3183, opened 2022-05-19
Cherry-picked commits for 1.6.3 release

nturinski:


src/extension.ts
```diff
@@ -99,7 +99,7 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
         createFunction: createFunctionFromApi,
         downloadAppSettings: downloadAppSettingsFromApi,
         uploadAppSettings: uploadAppSettingsFromApi,
-        apiVersion: '1.7.0'
+        apiVersion: '1.8.0'
```
bwateratmsft, line 102: I think it does _not_ need to be changed, but does this need to be bumped to 1.9.0 in `main`?

src/vscode-azurefunctions.api.d.ts
```diff
@@ -83,4 +85,10 @@ export interface ICreateFunctionOptions {
      * If set, it will automatically select the worker runtime for .NET with the matching targetFramework
      */
     targetFramework?: string | string[];
+
+    /**
+     * If set, it will include a step that will be executed prior to OpenFolderStep determined by the priority of the step
+     * OpenFolder priority is 250 (https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createNewProject/OpenFolderStep.ts#L11)
```
bwateratmsft, line 91: We should make this comment explicit about whether >250 means first, or <250 means first

bwateratmsft (approved):

bwateratmsft:
I think your merge branches are wrong; we should make a new `rel/1.6.3` branch based on `rel/1.6.2`, and then create the PR from a third branch against `rel/1.6.3`

VasuBhog:
Discussed with @nturinski that as long as the planned release for next Tuesday happens we would not be needing this. However, if it does not then the latest we would need these specific changes would be May 25th AM morning due to Build.

src/extension.ts
nturinski: I think the API would remain the same because there'd be no changes to the API between 1.6.3 and whatever we release in main.

bwateratmsft: I thought so but wasn't 100%.

PR #3196, opened 2022-05-26
Set account tree item

### LINKED ISSUE #3191: Error when picking Create new local app setting from local.settings.json when creating blob trigger (Error getSubscriptionPromptStep)
Repro steps:
1. Create new Function from the workspace
2. Create new project when prompted
3. Select JavaScript (happens on any language)
4. Select Azure Blob Storage trigger
5. Select "BlobTrigger1" as function name
6. On "Select setting from "local.settings.json" ", choose +Create new local app setting

Expected:
New value created and written to local.settings.json upon creation

Actual:
Error appears with "Cannot read properties of undefined (reading 'getSubscriptionPromptStep')

alexweininger:
Going to wait to merge this until after releasing v1.7.1

Fixes #3191 

Is this too hacky? We could properly expose this in the host api, but I don't think it belongs there.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -6,7 +6,7 @@
 'use strict';
 
 import { registerAppServiceExtensionVariables } from '@microsoft/vscode-azext-azureappservice';
-import { registerAzureUtilsExtensionVariables } from '@microsoft/vscode-azext-azureutils';
+import { AzureAccountTreeItemBase, registerAzureUtilsExtensionVariables } from '@microsoft/vscode-azext-azureutils';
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):
> Is this too hacky? We could properly expose this in the host api.

Hacky yes, but I don't think it's something we should put on the API.

alexweininger:
> > Is this too hacky? We could properly expose this in the host api.
> 
> Hacky yes, but I don't think it's something we should put on the API.

We're on the same page then 😄

Issue #3199, opened 2022-05-27
An error occurs when debugging a TypeScript project


v-ruizh:
**OS:** All
**Build Version:** 20220526.13

**Repro Steps:**
1. Create a TypeScript project with HttpTrigger function.
2. Debug it.
3. Check whether succeeds to debug the project or not.

**Expect:**
Succeed to debug a TypeScript project.

**Actual:**
An error occurs when debugging a TypeScript project.
![image](https://user-images.githubusercontent.com/45614674/170675176-80da179c-924d-4f32-bff5-7990eb42ae33.png)

**More Info:**
This issue does not reproduce after executing command "npm i --save-dev @types/node".

alexweininger:
This reproduces on v1.7.1 also

ejizba:
This was due to a recent release of the "@azure/functions" types package which references "Buffer" for the first time. Filed this issue to investigate if we can fix on the types side, but I'm not optimistic: https://github.com/Azure/azure-functions-nodejs-worker/issues/587

The "template" for TypeScript projects is owned by the VS Code extension, so it might be on the extension to address. We went back and forth on whether "@types/node" should be included in TypeScript projects by default in the past (for example [here](https://github.com/microsoft/vscode-azurefunctions/pull/1745)), but that was when `@azure/functions` didn't reference any node types. The main problem is knowing which version to pick.

alexweininger:
A [postinstall script](https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts) could check the installed node version and install the corresponding `@types/node` package.

ejizba:
Do you have examples of other types packages doing that? I don't really want to re-invent the wheel here as I think it's a very common thing for types packages to reference node types. I'd rather just follow what other popular packages are doing

alexweininger:
> Do you have examples of other types packages doing that?

Yeah I don't have any examples. 😄

I just tested out enabling the `skipLibCheck` option in the `tsconfig.json` and it makes the error go away. Maybe the template could enable that by default?

I know that's not optimal, but [TypeScript is exploring better options for skipLibCheck](https://github.com/microsoft/TypeScript/issues/30511#issuecomment-1103259948).

ejizba:
[...]

PR #3212, opened 2022-06-06
Add node types to TS project template

ejizba:
I used x range (`1.x`) instead of caret (`^1.0.0`) to match ` yo code`, although I don't really care. Decided not to check local node version, just the func version. Didn't seem like Brandon cared about being super exact either.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3199

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
     "name": "vscode-azurefunctions",
     "displayName": "Azure Functions",
     "description": "%azureFunctions.description%",
-    "version": "1.7.4-alpha.0",
+    "version": "1.7.4-alpha.1",
     "publisher": "ms-azuretools",
     "icon": "resources/azure-functions.png",
     "aiKey": "AIF-d9b70cd4-b9f9-4d70-929b-a071c400b217",
--- a/src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
@@ -42,24 +42,31 @@ export class TypeScriptProjectCreateStep extends JavaScriptProjectCreateStep {
     protected getPackageJsonDevDeps(context: IProjectWizardContext): { [key: string]: string } {
         // NOTE: Types package matches node worker version, not func host version
         // See version matrix here: https://www.npmjs.com/package/@azure/functions
-        let nodeWorkerVersion: string;
+        let funcTypesVersion: string;
+        // For the node types package, we'll use the latest LTS version possible
+        // See version matrix here: https://docs.microsoft.com/azure/azure-functions/functions-versions?pivots=programming-language-javascript#languages
+        let nodeTypesVersion: string;
         switch (context.version) {
             case FuncVersion.v4:
-                nodeWorkerVersion = '3';
+                funcTypesVersion = '3';
+                nodeTypesVersion = '16';
                 break;
             case FuncVersion.v3:
-                nodeWorkerVersion = '2';
+                funcTypesVersion = '2';
+                nodeTypesVersion = '14';
                 break;
             case FuncVersion.v2:
-                nodeWorkerVersion = '1';
+                funcTypesVersion = '1';
+                nodeTypesVersion = '10';
                 break;
             default:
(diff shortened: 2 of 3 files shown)


src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
```diff
@@ -42,24 +42,31 @@ export class TypeScriptProjectCreateStep extends JavaScriptProjectCreateStep {
...
                 break;
             case FuncVersion.v3:
-                nodeWorkerVersion = '2';
+                funcTypesVersion = '2';
+                nodeTypesVersion = '14';
                 break;
             case FuncVersion.v2:
-                nodeWorkerVersion = '1';
+                funcTypesVersion = '1';
+                nodeTypesVersion = '10';
```
bwateratmsft, line 60: 😨

What's the support state for Functions v2?

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurefunctions",
     "displayName": "Azure Functions",
     "description": "%azureFunctions.description%",
-    "version": "1.7.4-alpha.0",
+    "version": "1.7.4-alpha.1",
```
bwateratmsft, line 5: @nturinski this would be needed in package-lock.json as well.

src/commands/createNewProject/ProjectCreateStep/TypeScriptProjectCreateStep.ts
ejizba: https://docs.microsoft.com/en-us/azure/azure-functions/language-support-policy#retirement-policy-exceptions

30 September 2022 is the listed date to end support for Node 10 (the latest version for v2). Users will still be able to run Node 10, but this is the main difference:
> If you're running functions apps using an unsupported language version, you'll be required to upgrade before receiving support for the function apps.

Aka my team doesn't have to respond to incidents until user's upgrade

package.json
nturinski: Damn it, I didn't want to actually pull this down and do it for him lol.

ejizba: Lol I can do it if you tell me what it's about

ejizba: Interesting - for the record I think you could automate that in the dev ops pipeline. I assume the purpose is to differentiate prerelease builds? If you chop off the ".0" from the version in git and just add it every build from main I think you'd get the same thing with less manual steps and less git churn. For example add something like this to your azure pipelines:

```
CURR_VER=$(node -p "require('./package.json').version")
npm version "$CURR_VER.$TODO"
```

Where TODO is either a build id or [counter](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#counter) depending on your preferences

alexweininger: The reason we started bumping the versions like this is if you've installed an extension from a VSIX, VS Code won't let you install (update) the same extension from a VSIX unless the extension version is different. 

I think your solution would do the trick as well. I didn't know about counter, that's cool.

ejizba: Anyways I bumped the lock file for this PR. I see tests are failing, but looks like tests have been failing for a while on main so hard to tell if my PR caused that. Let me know if you want me to investigate

alexweininger (approved):
The failing tests aren't related to these changes.

package.json
bwateratmsft: > 

I could check again but I thought I hadn't experienced that issue. I thought the problem was that the publicly-released version of the extension had a version number higher than what we were using internally for the unified work, and VSCode kept auto-updating it.

alexweininger: > > The reason we started bumping the versions like this is if you've installed an extension from a VSIX, VS Code won't let you install (update) the same extension from a VSIX unless the extension version is different.
> > I think your solution would do the trick as well. I didn't know about counter, that's cool.
> 
> I could check again but I thought I hadn't experienced that issue. I thought the problem was that the publicly-released version of the extension had a version number higher than what we were using internally for the unified work, and VSCode kept auto-updating it.

I think you're right. I can't reproduce the issue I described either. For some reason I thought for sure that was an issue we had. 

If the issue you described is the only reason we need these version bump checks, then we can probably remove them...

PR #3235, opened 2022-07-07
Support for new Python language model

philliphoff:
Adds support for the new Python language model, where bindings are attributes within the application code rather than in separate JSON files.


package.json
```diff
@@ -833,6 +833,11 @@
                             ""
                         ]
                     },
+                    "azureFunctions.projectLanguageModel": {
+                        "scope": "resource",
+                        "type": "number",
```
bwateratmsft, line 838: Do we need any constraints on the values allowed? e.g. `maximum` or `minimum`?

src/commands/initProjectForVSCode/InitVSCodeLanguageStep.ts
```diff
@@ -22,21 +22,24 @@ export class InitVSCodeLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
-            { label: ProjectLanguage.CSharpScript },
-            { label: ProjectLanguage.FSharp },
-            { label: ProjectLanguage.FSharpScript },
-            { label: ProjectLanguage.Java },
-            { label: ProjectLanguage.JavaScript },
-            { label: ProjectLanguage.PowerShell },
-            { label: ProjectLanguage.Python },
-            { label: ProjectLanguage.TypeScript },
-            { label: ProjectLanguage.Custom }
+        const languagePicks: IAzureQuickPickItem<{ language: ProjectLanguage, model?: number }>[] = [
```
bwateratmsft, line 25: This `{ language: ProjectLanguage, model?: number }` "type" is used in a couple places; it might benefit from a proper type definition.

src/commands/initProjectForVSCode/InitVSCodeLanguageStep.ts
```diff
@@ -22,21 +22,24 @@ export class InitVSCodeLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
+            { label: ProjectLanguage.CSharp, data: { language: ProjectLanguage.CSharp } },
```
bwateratmsft, line 26: These values (or at least, similar) are used in multiple places; we should make a constant for them. That might eliminate the need for the type definition just above if it only remains in one place...

src/commands/initProjectForVSCode/initProjectForVSCode.ts
```diff
@@ -43,10 +43,11 @@ export async function initProjectForVSCode(context: IActionContext, fsPath?: str
     }
 
     language = language || getGlobalSetting(projectLanguageSetting) || await detectProjectLanguage(context, projectPath);
+    const languageModel: number | undefined = getGlobalSetting(projectLanguageModelSetting) || await detectProjectLanguageModel(language, projectPath);
```
bwateratmsft, line 46: Global setting refers specifically to user settings, i.e. workspace/workspace folder settings are ignored. Does it make sense for a language model setting to be a user setting?

src/tree/localProject/LocalFunctionsTreeItem.ts
```diff
@@ -34,7 +34,9 @@ export class LocalFunctionsTreeItem extends FunctionsTreeItemBase {
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        if (this.parent.isIsolated) {
+        const isParentPythonV2Plus = this.parent.langauge === ProjectLanguage.Python && this.parent.languageModel && this.parent.languageModel > 1;
```
bwateratmsft, line 37:
```suggestion
        const isParentPythonV2Plus = this.parent.language === ProjectLanguage.Python && this.parent.languageModel && this.parent.languageModel > 1;
```

bwateratmsft, line 37: Can we reuse the `isPythonV2Plus` utility?

src/debug/validatePreDebug.ts
```diff
@@ -120,7 +155,10 @@ async function validateAzureWebJobsStorage(context: IActionContext, projectLangu
                 return new ParsedFunctionJson(await fse.readJSON(functionJsonPath));
             }));
 
-            if (functions.some(f => !f.isHttpTrigger)) {
+            // NOTE: Currently, Python V2+ requires storage to be configured, even for HTTP triggers.
+            const isPythonV2Plus = projectLanguage === 'Python' && projectLanguageModel && projectLanguageModel > 1;
```
bwateratmsft, line 159: Can we reuse the `isPythonV2Plus` utility?

src/debug/validatePreDebug.ts
```diff
@@ -99,6 +111,29 @@ function getMatchingWorkspace(debugConfig: vscode.DebugConfiguration): vscode.Wo
...
+    const validateTools = getWorkspaceSetting<boolean>('validateFuncCoreTools', workspacePath) !== false;
+
+    if (validateTools && isPythonV2Plus(projectLanguage, projectLanguageModel)) {
+        const version = await getLocalFuncCoreToolsVersion(context, workspacePath);
+
+        // TODO: Update range with official version.
+        const expectedVersionRange = '>=4.0.4515';
+
+        if (version && !semver.satisfies(version, expectedVersionRange)) {
+            const message: string = localize('invalidFunctionVersion', 'The version of installed Functions tools "{0}" is not sufficient for this project type ("{1}")?', version, expectedVersionRange);
```
bwateratmsft, line 127: The message ends in a question mark but isn't actually a question?

(teehee)

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
```diff
@@ -30,10 +32,34 @@ const defaultRequirements: string = `# DO NOT include azure-functions-worker in
...
+
+const defaultFunctionAppSource = `import azure.functions as func
+
+app = func.FunctionApp(auth_level=func.AuthLevel.ANONYMOUS)
+
+# Uncomment to create an HTTP trigger-based function.
+# @app.function_name(name="HttpTrigger1")
+# @app.route(route="hello")
+# def test_function(req: func.HttpRequest) -> func.HttpResponse:
+#    return func.HttpResponse("HttpTrigger1 function processed a request!!!")
```
bwateratmsft, line 45:
```suggestion
#    return func.HttpResponse("HttpTrigger1 function processed a request!!!!!!!!!")
```

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -106,13 +108,17 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
 
         const wizardOptions: IWizardOptions<IProjectWizardContext> = { promptSteps, executeSteps };
 
-        // All languages except Java support creating a function after creating a project
+        const isPythonV2Plus = context.language === ProjectLanguage.Python && context.languageModel && context.languageModel > 1;
```
bwateratmsft, line 111: Can we reuse the `isPythonV2Plus` utility?

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
```diff
@@ -0,0 +1,48 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
```
bwateratmsft, line 6: The Azure extensions use https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/utils/AzExtFsExtra.ts instead of `fs-extra`.

package-lock.json
```diff
@@ -74,7 +74,7 @@
                 "webpack-cli": "^4.6.0"
             },
             "engines": {
-                "vscode": "^1.58.0"
+                "vscode": "^1.66.0"
```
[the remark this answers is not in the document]
philliphoff: Yes; it was already `1.66.0` in the `package.json`.

package.json
philliphoff: There's no theoretical maximum, but I'll set a `minimum` of 1.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
philliphoff: I'm letting the Functions team decide how excited they want to be about a successfully processed request.

src/commands/initProjectForVSCode/InitVSCodeLanguageStep.ts
philliphoff: Agree there seems to be some repetition/overlap in areas of the codebase, though there are some subtle differences.  For this PR, still being new to the codebase, I was trying not to do too much refactoring in hopes of not introducing major regressions.  There is a similar new model for Node.js coming down the road that will help to show where some of that commonality lies and be a good opportunity for further refactoring.

src/commands/initProjectForVSCode/initProjectForVSCode.ts
philliphoff: The purpose of this logic is to "VS Code"-ify an existing project, so the assumption is that there likely wouldn't be a local VS Code settings file.  Hence, most of the settings are inferred from non-VS Code content (like for the project language) if they're not set at the global level.

src/debug/validatePreDebug.ts
philliphoff: It was supposed to be rhetorical.  :-) Updated to a boring, old period.

src/tree/localProject/LocalFunctionsTreeItem.ts
philliphoff: It's remarkable that spelling checks still aren't a common thing in linters.  Updated.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
```diff
@@ -48,6 +74,32 @@ export class PythonProjectCreateStep extends ScriptProjectCreateStep {
...
+
+        // Python V2+ model currently does not work when extension bundles are specified.
+        if (this.isModelV2Plus) {
+            hostJson.extensionBundle = undefined;
+        }
+
+        return hostJson;
+    }
+
+    private get isModelV2Plus(): boolean {
```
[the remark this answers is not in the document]
philliphoff: Consolidated.

src/commands/createNewProject/NewProjectLanguageStep.ts
philliphoff: Consolidated

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -92,6 +95,12 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
 
             addBindingSettingSteps(template.userPromptedSettings, promptSteps);
 
+            const isV2PythonModel = context.language === ProjectLanguage.Python && context.languageModel && context.languageModel > 1;
```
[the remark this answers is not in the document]
philliphoff: Consolidated.

src/tree/localProject/LocalFunctionsTreeItem.ts
philliphoff: Consolidated.

src/debug/validatePreDebug.ts
philliphoff: Consolidated.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
```diff
@@ -30,10 +33,34 @@ const defaultRequirements: string = `# DO NOT include azure-functions-worker in
 azure-functions
 `;
 
+const azureWebJobsFeatureFlagsKey = 'AzureWebJobsFeatureFlags';
+
+const defaultFunctionAppSource = `import azure.functions as func
```
nturinski, line 38: Is this just a placeholder until the Python language model templates are available or are we going to have to hardcode the source code like this?

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
```diff
@@ -30,10 +33,34 @@ const defaultRequirements: string = `# DO NOT include azure-functions-worker in
...
+# @app.function_name(name="HttpTrigger1")
+# @app.route(route="hello")
+# def test_function(req: func.HttpRequest) -> func.HttpResponse:
+#    return func.HttpResponse("HttpTrigger1 function processed a request!!!")
+`;
+
 export class PythonProjectCreateStep extends ScriptProjectCreateStep {
     protected gitignore: string = pythonGitignore;
 
+    constructor(private readonly model?: number) {
```
nturinski, line 52: JavaScript is getting the same language model treatment. Should we just add the optional parameter to the `ScriptProjectCreateStep`?

src/constants.ts
```diff
@@ -33,6 +34,8 @@ export enum ProjectLanguage {
     Custom = 'Custom'
 }
 
+export const previewPythonModel: number = 2;
```
nturinski, line 37: Could you explain what these numbers mean? I'm not really sure what the significances of the numbers are for the models, but you set a minimum of 1 for the setting.

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
philliphoff: Updated.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
philliphoff: For this preview version, we're now bundling as "backup templates" so they're in the standard `templates.json` format in the `resources` folder.

src/commands/createFunction/scriptSteps/PythonFunctionCreateStep.ts
```diff
@@ -0,0 +1,49 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import * as fse from 'fs-extra';
```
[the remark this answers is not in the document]
philliphoff: Updated.

src/debug/validatePreDebug.ts
```diff
@@ -99,6 +108,29 @@ function getMatchingWorkspace(debugConfig: vscode.DebugConfiguration): vscode.Wo
...
+
+    if (validateTools && isPythonV2Plus(projectLanguage, projectLanguageModel)) {
+        const version = await getLocalFuncCoreToolsVersion(context, workspacePath);
+
+        // TODO: Update range with official version.
+        const expectedVersionRange = '>=4.0.4515';
+
+        if (version && !semver.satisfies(version, expectedVersionRange)) {
+            const message: string = localize('invalidFunctionVersion', 'The version of installed Functions tools "{0}" is not sufficient for this project type ("{1}").', version, expectedVersionRange);
+            const debugAnyway: vscode.MessageItem = { title: localize('debugWithInvalidFunctionVersionAnyway', 'Debug anyway') };
```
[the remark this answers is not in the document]
philliphoff: Agree, though I'd say it's out of scope for the purposes of this initial preview.  (E.g. perhaps it'd make sense as the new model goes GA)?

src/constants.ts
philliphoff: The number itself is arbitrary and just corresponds to a specific iteration of the language model.  `1` is the (current) model (and assumed if omitted).  `2` is the "new" model (with binding attributes).  If there are any significant changes made to the model in the future (such that tooling would need to distinguish between it and previous iterations), that model would be assigned an incremented number.

src/commands/createNewProject/ProjectCreateStep/PythonProjectCreateStep.ts
philliphoff: I'm reluctant to make changes to accommodate Node.js until we see the extent to which there is real overlap.  I suspect that would make sense, but tried my best not to do significant refactoring until we actually get to the point of having more than a single, concrete example of the needed changes.

src/commands/initProjectForVSCode/InitVSCodeLanguageStep.ts
philliphoff: How about, if I see it in one more place, I promise to do so. :-)

alexweininger:
I like the auto open markdown feature, really cool (and tricky to implement IIRC) that it grabs the markdown file from the template and displays it rendered.

I wish we could do something similar with walkthroughs, ex: spawn a walkthrough from a markdown file, or maybe a special walkthrough.json file. Could be really useful for templates, samples, etc.

alexweininger:
When I append a new function to a file, it's not smart about making sure the new function name is unique. Sounds like it's out of scope for the time being. But something we might revisit later.

philliphoff:
> When I append a new function to a file, it's not smart about making sure the new function name is unique. Sounds like it's out of scope for the time being. But something we might revisit later.

Yes, it is exactly things of that sort that require a robust language service in order to solve, but which take time we didn't think we had for this initial preview.  Even then it's likely to result in a long tail of possible code structure to try to cover.

philliphoff:
> I like the auto open markdown feature, really cool (and tricky to implement IIRC) that it grabs the markdown file from the template and displays it rendered.

It's admittedly a hack (in particular, to support VS Code's need to reopen the document on a new file) to avoid having to manage a local disk cache), but I think it will work well enough for most cases.

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { AzureWizardPromptStep, IAzureQuickPickItem, IAzureQuickPickOptions } from '@microsoft/vscode-azext-utils';
+import { FunctionLocation, IPythonFunctionWizardContext } from './IPythonFunctionWizardContext';
+import { localize } from '../../../localize';
+
+export class PythonScriptStep extends AzureWizardPromptStep<IPythonFunctionWizardContext> {
+    public async prompt(wizardContext: IPythonFunctionWizardContext): Promise<void> {
+        const entries = await vscode.workspace.fs.readDirectory(vscode.Uri.file(wizardContext.projectPath));
```
nturinski, line 14: `AzExtFsExtra` does have a `readDirectory()` command though the return types are a little bit different.

resources/backupTemplates/pystein/templates/templates.json
```diff
@@ -0,0 +1,4968 @@
...
+        "host.json": "{\r\n    \"version\": \"2.0\",\r\n    \"logging\": {\r\n      \"applicationInsights\": {\r\n        \"samplingSettings\": {\r\n          \"isEnabled\": true,\r\n          \"excludedTypes\": \"Request\"\r\n        }\r\n      }\r\n    },\r\n    \"extensionBundle\": {\r\n      \"id\": \"Microsoft.Azure.Functions.ExtensionBundle\",\r\n      \"version\": \"[3.*, 4.0.0)\"\r\n    }\r\n  }",
```
nturinski, line 1847: Alex mentioned that the `host.json` can't contain the `extensionBundle` property in order to debug (and you have denoted as such in the code.) I noticed that all the templates have host.jsons that include it by default, so whenever you create a new trigger, it automatically updates the `host.json` to include the property again.

nturinski (commented):
There is no way to pass an argument to change the default trigger names right now since everything is static content in the backupTemplates, right?

I was surprised by the lack of prompt steps, but I'm assuming when the template feed is actually serving the templates, that there will be a way to pass in parameters for starting names and other properties similar to the other templates?

NOTE: I left a comment in the `templates.json` but it might be annoying to view due to the size of that file so I'll leave it here as well:

> Alex mentioned that the `host.json` can't contain the `extensionBundle` property in order to debug (and you have denoted as such in the code.) I noticed that all the templates have host.jsons that include it by default, so whenever you create a new trigger, it automatically updates the `host.json` to include the property again.

resources/backupTemplates/pystein/templates/templates.json
philliphoff: Ah, that's where the bundle was sneaking back in; we saw that yesterday but I hadn't yet pinpointed when it happened.  Agree that it could be removed from the (one) PyStein project template; I'll note that to the Functions team.  That said, the resurrected bundle property is added from a "default" that's independent of any template (.  I'll need to disable that for PyStein projects.

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
philliphoff: Noted, though I'd prefer to use the VS Code APIs when available (and reasonably convenient).  If anything, it makes it easier for the extension to support their new remote host/filesystem scenarios.  (Yes, there are lots of other things that *wouldn't* work for Functions in those scenarios, but still...)

src/tree/localProject/LocalFunctionsTreeItem.ts
```diff
@@ -33,7 +34,9 @@ export class LocalFunctionsTreeItem extends FunctionsTreeItemBase {
     }
 
     public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
-        if (this.parent.isIsolated) {
+        const isParentPythonV2Plus = isPythonV2Plus(this.parent.language, this.parent.languageModel);
+
+        if (this.parent.isIsolated || isParentPythonV2Plus) {
```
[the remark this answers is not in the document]
philliphoff: Reasonable suggestion; done (changed to `getChildrenForHostedProjects()` and updated comments).

src/templates/script/parseScriptTemplates.ts
```diff
@@ -228,7 +229,8 @@ export function parseScriptTemplate(rawTemplate: IRawTemplate, resources: IResou
         language,
         userPromptedSettings,
         templateFiles: rawTemplate.files || {},
-        categories: rawTemplate.metadata.category || []
+        categories: rawTemplate.metadata.category || [],
+        categoryStyle: rawTemplate.metadata.categoryStyle
```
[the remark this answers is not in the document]
philliphoff: Yes, the Functions team is using the `categoryStyle` metadata property to indicate their (currently one and only one PyStein project template).

src/templates/script/PysteinTemplateProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+
+        // Find the first Python Preview project root template...
+        return templates.functionTemplates.find(template =>
+            template.language === ProjectLanguage.Python
+            && template.id.endsWith('-Python-Preview')
+            && template.categoryStyle === 'projectroot') as IScriptFunctionTemplate;
+    }
+
+    public async updateBackupTemplates(): Promise<void> {
+        await Promise.resolve();
```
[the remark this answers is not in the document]
philliphoff: Yes, added comment to that effect.

src/templates/script/PysteinTemplateProvider.ts
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
```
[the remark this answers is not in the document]
philliphoff: Removed.  Wish there was a linter for unneeded lint pragmas.

src/debug/validatePreDebug.ts
```diff
@@ -98,6 +107,29 @@ function getMatchingWorkspace(debugConfig: vscode.DebugConfiguration): vscode.Wo
...
+/**
+ * Ensure that that Python V2+ projects have an appropriate version of Functions tools installed.
+ */
+async function validateFunctionVersion(context: IActionContext, projectLanguage: string | undefined, projectLanguageModel: number | undefined, workspacePath: string): Promise<boolean> {
+    const validateTools = getWorkspaceSetting<boolean>('validateFuncCoreTools', workspacePath) !== false;
+
+    if (validateTools && isPythonV2Plus(projectLanguage, projectLanguageModel)) {
+        const version = await getLocalFuncCoreToolsVersion(context, workspacePath);
+
+        // TODO: Update range with official version.
```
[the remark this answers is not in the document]
philliphoff: I've updated the comments to be more clear (and the version number).  The Functions team are not sure what the version number for the official "preview" build with Pystein will be, so the currently available build version number will serve as the minimum bar (for now).

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+
+        const functionAppPath: string = path.join(context.projectPath, pythonFunctionAppFileName);
+
+        await window.showTextDocument(await workspace.openTextDocument(Uri.file(functionAppPath)));
+
+        const gettingStartedPath = path.join(context.projectPath, gettingStartedFileName);
+
+        await showMarkdownPreviewFile(Uri.file(gettingStartedPath));
```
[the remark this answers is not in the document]
philliphoff: So, it turns out there is (the not quite documented `markdown.showPreviewToSide` command, closely related to the "standard" `markdown.showPreview` command)!  I've updated the utility method to allow either, and updated the function/project creation steps to open side-by-side, which is quite nice.

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -36,31 +38,33 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
-            { label: ProjectLanguage.Python, data: ProjectLanguage.Python },
-            { label: ProjectLanguage.Java, data: ProjectLanguage.Java },
-            { label: ProjectLanguage.PowerShell, data: ProjectLanguage.PowerShell },
-            { label: localize('customHandler', 'Custom Handler'), data: ProjectLanguage.Custom }
+        let languagePicks: IAzureQuickPickItem<{ language: ProjectLanguage, model?: number } | undefined>[] = [
+            { label: ProjectLanguage.JavaScript, data: { language: ProjectLanguage.JavaScript } },
+            { label: ProjectLanguage.TypeScript, data: { language: ProjectLanguage.TypeScript } },
+            { label: ProjectLanguage.CSharp, data: { language: ProjectLanguage.CSharp } },
+            { label: ProjectLanguage.Python, data: { language: ProjectLanguage.Python } },
+            { label: pythonNewModelPreview, data: { language: ProjectLanguage.Python, model: previewPythonModel } },
```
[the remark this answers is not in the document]
philliphoff: We went around and around about how best to expose the PyStein model (and with what wording) and settled on a separate "language" entry.  I think this makes sense as, eventually, this new model will become the default and the old will be phased out, and we'll be back at a single Python entry in the list.

With respect to a v1/v2 prompt, considered it, but we erred on the side of limiting the number of prompts shown to users when creating a new project, the idea being that each prompt tends to lose (in particular, new) users.

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
```diff
@@ -0,0 +1,224 @@
...
+            this.localSettingsJson = JSON.parse(localSettingsContent) as ILocalSettingsJson;
+        } else {
+            // Python V2+ model currently requires the EnableWorkerIndexing feature flag be enabled.
+            this.localSettingsJson.Values ??= {};
+            this.localSettingsJson.Values['AzureWebJobsFeatureFlags'] = 'EnableWorkerIndexing';
+        }
+
+        await super.executeCore(context, progress);
+
+        const files = [
```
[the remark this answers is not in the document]
philliphoff: I'm a bit conflicted about this as the underlying script base type explicitly manages certain files, in useful ways, so blindly pulling template files could unintentionally override that behavior.  (E.g. the PyStein step explicitly pulls certain files out of the template to get them into the form the underlying type requires.) We *could* instead do what you suggest but exclude the files we know are handled differently, but that could still present a problem if the underlying script base type is updated to explicitly handle more files.  In that case, I don't see it as much different than what we're doing now.

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
nturinski: Well, under the hood, AzExtFsExtra is using all VS Code APIs which was the whole point of this [PR](https://github.com/microsoft/vscode-azurefunctions/pull/3268).

https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/utils/AzExtFsExtra.ts#L94-L100

It just offers some QOL, like not having to wrap projectPath with `vscode.Uri.file()`, but definitely not a ship blocker or anything.

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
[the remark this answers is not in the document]
nturinski: Yeah, I guess my thought was that we would just trust the templates feed team to serve the correct files with each template. I think that's what we're doing with other Script templates. 

But since we're going to have to change the template provider whenever the templates are being served dynamically anyway, we can always revisit the list of template files then? A static list of files makes sense for a static hard-coded template to me.

nturinski (approved):
I'm good to approve this. I kind of want to merge this sooner so we can start to get CTI doing test passes on it, but I'd understand wanting to wait for Brandon to rise from the dead to give it a thorough review. 🧟

src/templates/script/PysteinTemplateProvider.ts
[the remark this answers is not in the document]
nturinski: @motm32 Sounds like a pretty good extension idea... 🤔💭

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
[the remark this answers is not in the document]
philliphoff: Ok, you've won me over, I've switch to an "exlusion list" instead of an "inclusion list".

src/commands/createFunction/scriptSteps/PythonScriptStep.ts
philliphoff: Ah, ok, I see, then forget everything I said!  For some reason I thought `AzExtFsExtra` was still using `fs` under the hood.  :-)  Eh, in this particular case I find direct use of the VS Code API easier to read than using the wrapper.  I could go either way.

philliphoff:
> There is no way to pass an argument to change the default trigger names right now since everything is static content in the backupTemplates, right?
>
> I was surprised by the lack of prompt steps, but I'm assuming when the template feed is actually serving the templates, that there will be a way to pass in parameters for starting names and other properties similar to the other templates?

@nturinski Agreed that this initial preview is limited in scope with respect to customization of the template (i.e. there effectively is none).  To do it properly requires more than just token replacement in the added function source, as there may be other changes needed elsewhere (e.g. new imports, validation against existing function names, etc., and, in particular, when they introduce PyStein's "blueprints").  We really need a robust language service that allows source manipulation a la Roslyn, and I'm not sure we have that now (and our time frame didn't really allow it in any case).  This is one reason we're trying to lean on the "inline" documentation in the function templates, to help users understand what needs to change in order to add new functions.

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
```diff
@@ -19,6 +20,16 @@ import { showMarkdownPreviewFile } from '../../../utils/textUtils';
 
 const gettingStartedFileName = 'getting_started.md';
 
+async function fileExists(uri: vscode.Uri): Promise<boolean> {
```
nturinski, line 23: You could use [this](https://github.com/microsoft/vscode-azuretools/blob/main/utils/src/utils/AzExtFsExtra.ts#L17-L21) to save some lines, though I will go back and add a `fileExists` utility function later.

nturinski: You would also not have to wrap the file paths with `Uri.file()` as well.

nturinski (approved):

src/commands/createNewProject/ProjectCreateStep/PysteinProjectCreateStep.ts
philliphoff: Doesn't `stat()` throw if the file *doesn't* exist?

philliphoff:
> I'm good to approve this. I kind of want to merge this sooner so we can start to get CTI doing test passes on it, but I'd understand wanting to wait for Brandon to rise from the dead to give it a thorough review. 🧟

Agree that I'd like CTI taking a look and making sure I didn't horribly break the existing scenarios.  @bwateratmsft did an early review of the draft PR, but I'm happy for him to tear it apart further when he's back next week (and I'm out).  I can then clean up when I get back.

bwateratmsft:
I took a brief look at the changes since my review and didn't see anything that concerned me. Agree, getting it in sooner than later is a good idea so CTI can take a look. Thanks!

Issue #3165, opened 2022-05-13
[Suggestion] Don't show Logic Apps in Functions Extension


anthonywhite:
Or if you think you must, show them with a Logic App icon.

After all, there are Logic Apps extensions...

nturinski:
The payload includes `functionapp` in its kind.  We should filter `workflowapp` out of what we display

![image](https://user-images.githubusercontent.com/5290572/175430527-8b0f555f-4a14-42d6-a023-bf8cb56afdee.png)

alexweininger:
Fixed by https://github.com/microsoft/vscode-azureresourcegroups/pull/332

PR #3238, opened 2022-07-12
Don't resolve Logic Apps

alexweininger:
Related to https://github.com/microsoft/vscode-azurefunctions/issues/3165 and https://github.com/microsoft/vscode-azureresourcegroups/issues/246

### DIFF
--- a/src/FunctionAppResolver.ts
+++ b/src/FunctionAppResolver.ts
@@ -20,6 +20,6 @@ export class FunctionAppResolver implements AppResourceResolver {
     }
 
     public matchesResource(resource: AppResource): boolean {
-        return resource.type.toLowerCase() === 'microsoft.web/sites' && !!resource.kind?.includes('functionapp');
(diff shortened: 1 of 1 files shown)


src/FunctionAppResolver.ts
```diff
@@ -20,6 +20,6 @@ export class FunctionAppResolver implements AppResourceResolver {
     }
 
     public matchesResource(resource: AppResource): boolean {
-        return resource.type.toLowerCase() === 'microsoft.web/sites' && !!resource.kind?.includes('functionapp');
+        return resource.type.toLowerCase() === 'microsoft.web/sites' && !!resource.kind?.includes('functionapp') && !resource.kind?.includes('workflowapp');
```
bwateratmsft, line 23: A comment on that part would be good. Maybe also break up the expression into three lines for ease of reading.

alexweininger: Will do. Also been thinking about if we should put the "is this an X app" util functions into azure utils.

bwateratmsft (approved):

PR #3260, opened 2022-07-28
Update backup templates

nturinski:
Can't run any tests without them because of the new stacks. Plus we should be updating these with major releases anyway.


alexweininger (approved):
[Instructions for how to update the backup templates for future reference](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/templates/README.md#updating-backup-templates)

alexweininger:
> Plus we should be updating these with major releases anyway.

We could automate this if we wanted

nturinski:
> > Plus we should be updating these with major releases anyway.
> 
> We could automate this if we wanted

Maybe, but you need to install Java and then Maven in order to grab the new templates. I feel like it'd be more trouble than it's worth to get that running on a build machine

Issue #3266, opened 2022-07-29
Turn workspace related unit tests back on


nturinski:
Until the testing suite actually installs the Azure Resources extension and gets activated, these tests will not work as intended.

PR #3267, opened 2022-07-29
Fix tests to use v4 Core Tools

nturinski:
Remove the preview labels, some other misc. fixes. Also removed the local binding tests, issue being tracked here #3266


nturinski:
note: Windows test failed only due a timeout

bwateratmsft (approved):

PR #3268, opened 2022-07-29
Replace fse with AzExtFsUtils where possible

nturinski:
I couldn't get rid of `fse` entirely unfortunately...

- There's a `createWriteStream` command in `requestUtils` that there's no equivalent in `workspace.fs`
- The gulpfile doesn't actually have access to the utils package because the gulpfile doesn't run with the extension as far as I know. Whenever one is resolved, we should be able to move `fse-extra` to a devDependency or replace with `fs` calls
- A couple of calls to copy had to have the "overwrite" property added to them. It was causing issues in the tests when the backup templates were being updated.


bwateratmsft:
> I couldn't get rid of `fse` entirely unfortunately...
> 
> * There's a `createWriteStream` command in `requestUtils` that there's no equivalent in `workspace.fs`

Can't we make our own leveraging `fs.writeFile` along with an in-memory stream buffer?

bwateratmsft (approved):

nturinski:
> > I couldn't get rid of `fse` entirely unfortunately...
> > 
> > * There's a `createWriteStream` command in `requestUtils` that there's no equivalent in `workspace.fs`
> 
> Can't we make our own leveraging `fs.writeFile` along with an in-memory stream buffer?

Yeah, I think we can but I wanted to do that separately just since it's a bit more of a complex change than just changing out some fs calls with VS Code's implementationm

PR #3273, opened 2022-08-03
Push Framework argument for .net7.0 

nturinski:
@ejizba Just wanted to check with you on this-- so I should only have to pass `--arg: Framework net7.0` on project creation since that is when the `.csproj` is generated and that's where `Framework` is actually replacing `TargetFrameworkValue`.

On a `FunctionCreate`, assumedly the project is already created with the appropriate `.csproj` file.

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/DotnetProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/DotnetProjectCreateStep.ts
@@ -42,9 +42,8 @@ export class DotnetProjectCreateStep extends ProjectCreateStepBase {
         const functionsVersion: string = 'v' + majorVersion;
         const projTemplateKey = nonNullProp(context, 'projectTemplateKey');
         const args = ['--identity', identity, '--arg:name', cpUtils.wrapArgInQuotes(projectName), '--arg:AzureFunctionsVersion', functionsVersion];
-        if (/net7.[0-9]/.test(projTemplateKey)) {
-            args.push('--arg:Framework', cpUtils.wrapArgInQuotes('net7.0'));
-        }
+        // defaults to net6.0 if there is no targetFramework
+        args.push('--arg:Framework', cpUtils.wrapArgInQuotes(context.workerRuntime?.targetFramework));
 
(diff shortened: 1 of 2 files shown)


alexweininger (commented):
Is this not a CLI bug? I feel like it should do this for us.

src/commands/createNewProject/ProjectCreateStep/DotnetProjectCreateStep.ts
```diff
@@ -41,7 +41,12 @@ export class DotnetProjectCreateStep extends ProjectCreateStepBase {
         }
         const functionsVersion: string = 'v' + majorVersion;
```
ejizba, line 42: cc @liliankasem 

For Alex's context, this is a new change by Lilian to the .NET isolated templates. PR here: https://github.com/Azure/azure-functions-templates/pull/1240. Basically, in the past the isolated worker only really supported one .NET version at a time (or at least only one we cared about for the sake of templates). Now, they support net48, net6.0, and net7.0 all at once which is why a new parameter was introduced. If the arg is not passed, it will currently default to net6.0.

Also - it's important to note that the version of the `dotnet` cli being used to create the project does not necessarily match the version of the project. I can have all three versions installed on my machine but I should still be able to pick which one I use for my project.

alexweininger (approved):

nturinski:
For reference, our tests have been creating csproj files this way forever:

```
protected async initializeTestFolder(testFolder: string): Promise<void> {
        await super.initializeTestFolder(testFolder);
        await AzExtFsExtra.writeFile(path.join(testFolder, 'test.csproj'), `<Project Sdk="Microsoft.NET.Sdk">
    <PropertyGroup>
        <TargetFramework>${this._targetFramework}</TargetFramework>
    </PropertyGroup>
    <ItemGroup>
        <PackageReference Include="${this._isIsolated ? 'Microsoft.Azure.Functions.Worker.Sdk' : 'Microsoft.NET.Sdk.Functions'}" Version="1.0.0" />
    </ItemGroup>
</Project>`);
    }
```

So I think adding the target framework for all versions shouldn't break anything. Also tried with v3 and it doesn't seem to find the new argument.

Issue #3264, opened 2022-07-28
View properties doesn't display site config


ejizba:
Does this occur consistently? Yeah
Repro steps:
1. Right click on a function app
2. View properties

### Expected

It'll show a ton of properties from the site config. This used to work (https://github.com/microsoft/vscode-azurefunctions/pull/2164) and I'm guessing it broke with the new unified view

### Actual

This is what it displays:
```
{
    "id": "/subscriptions/...",
    "name": "...",
    "type": "Microsoft.Web/sites",
    "kind": "functionapp",
    "location": "centraluseuap",
    "tags": {}
}
```
Which unfortunately does not have very much information at all. 

Version: 1.7.4
OS: darwin
OS Release: 21.6.0
Product: Visual Studio Code
Product Version: 1.69.2
Language: en

nturinski:
Yep, you are correct.  The data doesn't get set when the func app loads, it got set when `viewProperties` is called in the Functions extension.  But since it's using RG's `viewProperties` command, it just has the old, basic data.

Sorry about that 🙇🏻🦖

PR #3281, opened 2022-08-08
Load siteConfig on resolve

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3264

It's easier to just add the siteConfig on resolve than to try to call `viewProperties` from the Functions extension. It makes the resolve a little bit slower, but it doesn't seem to be that noticeable.

### DIFF
--- a/src/FunctionAppResolver.ts
+++ b/src/FunctionAppResolver.ts
@@ -5,21 +5,15 @@ import { ResolvedFunctionAppResource } from "./tree/ResolvedFunctionAppResource"
 import { createWebSiteClient } from "./utils/azureClients";
 
 export class FunctionAppResolver implements AppResourceResolver {
-    public async resolveResource(subContext: ISubscriptionContext, resource: AppResource): Promise<ResolvedFunctionAppResource | null> {
+    public async resolveResource(subContext: ISubscriptionContext, resource: AppResource): Promise<ResolvedFunctionAppResource | undefined> {
         return await callWithTelemetryAndErrorHandling('resolveResource', async (context: IActionContext) => {
-            try {
-                const client = await createWebSiteClient({ ...context, ...subContext });
-                const rg = getResourceGroupFromId(nonNullProp(resource, 'id'));
-                const name = nonNullProp(resource, 'name');
-                const site = await client.webApps.get(rg, name);
-                site.siteConfig = await client.webApps.getConfiguration(rg, name)
-                return new ResolvedFunctionAppResource(subContext, site);
-
-            } catch (e) {
-                console.error({ ...context, ...subContext });
-                throw e;
-            }
-        }) ?? null;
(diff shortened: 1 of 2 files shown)


src/FunctionAppResolver.ts
```diff
@@ -9,7 +9,10 @@ export class FunctionAppResolver implements AppResourceResolver {
         return await callWithTelemetryAndErrorHandling('resolveResource', async (context: IActionContext) => {
             try {
```
alexweininger, line 10: While we're making changes here, lets remove this try/catch and console.log

src/FunctionAppResolver.ts
```diff
@@ -5,18 +5,15 @@ import { ResolvedFunctionAppResource } from "./tree/ResolvedFunctionAppResource"
...
-
-            } catch (e) {
-                console.error({ ...context, ...subContext });
-                throw e;
-            }
-        }) ?? null;
+            const client = await createWebSiteClient({ ...context, ...subContext });
+            const rg = getResourceGroupFromId(nonNullProp(resource, 'id'));
+            const name = nonNullProp(resource, 'name');
+            const site = await client.webApps.get(rg, name);
```
bwateratmsft, line 13: Doesn't have to be in this PR, but can these be made lazy so that they aren't fetched before they are needed?

bwateratmsft (approved):

alexweininger (approved):

src/FunctionAppResolver.ts
```diff
@@ -10,9 +10,9 @@ export class FunctionAppResolver implements AppResourceResolver {
             const client = await createWebSiteClient({ ...context, ...subContext });
             const rg = getResourceGroupFromId(nonNullProp(resource, 'id'));
             const name = nonNullProp(resource, 'name');
+
             const site = await client.webApps.get(rg, name);
-            site.siteConfig = await client.webApps.getConfiguration(rg, name)
-            return new ResolvedFunctionAppResource(subContext, site);
+            return ResolvedFunctionAppResource.createResolvedFunctionAppResource(context, subContext, site);
```
bwateratmsft, line 15: Not sure I understand the function of the latest commit (where `siteConfig` is loaded async).

nturinski: I wanted to make it lazy, but it would require changes in `viewProperties` in Resource Groups. Unless I can make a get/set data asynchronous, but I don't think that's possible.

Loading the siteConfig in the background is to not slow down the resolve, but still having it likely by the time a user views the properties.

bwateratmsft: Does part of the `siteConfig` end up being displayed?

bwateratmsft: Yeah, I think that's the way to go (change `viewProperties`). But I think it probably should be done in a separate PR.

bwateratmsft (approved):

PR #3303, opened 2022-08-31
Use language model in template provider caching.

### LINKED ISSUE #3300: An error occurs if a Java/Python (New Model Preview) project opens in VS Code previously when creating a new Python project
**OS:** Win 10/ Mac (Inter)
**Build Version:** 20220819.8

**Repro Steps:**
1. Create and open a Java/Python (New Model Preview) project in VS Code.
2. F1 -> Click "Azure Functions: Create New Project..." -> Select a new folder -> Select Python language -> Select an existing Python interpreter -> Select "HTTP trigger" -> Enter the default function name -> Click "Open in current window".
3. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
image

**More Info:**
1. Don't ask to select **Authorization level** when creating a python project if a Python (New Model Preview) project opens in VS Code previously.
image
2. This issue does not reproduce when creating projects with other languages.
3. This issue does not reproduce when no project opens in VS Code.

philliphoff:
Use the language model in the template provider cache key to differentiate providers of templates of different models (e.g. Python from Python V2).

Resolves #3300.

### DIFF
--- a/src/templates/CentralTemplateProvider.ts
+++ b/src/templates/CentralTemplateProvider.ts
@@ -124,8 +124,8 @@ export class CentralTemplateProvider implements Disposable {
     }
 
     private getCachedProvidersKey(language: ProjectLanguage, languageModel: number | undefined, version: FuncVersion): string {
-        // NOTE: VS Code treats lack of a language model project setting as a 0, so treat undefined === 0.
-        return `${language}:${languageModel !== undefined ? languageModel : 0}:${version}`;
+        // NOTE: VS Code treats lack of a language model project setting as a 0, so treat undefined === null === 0.
+        return `${language}:${languageModel ?? 0}:${version}`;
     }
 
     private tryGetCachedProviders(projectPath: string | undefined, language: ProjectLanguage, languageModel: number | undefined, version: FuncVersion): CachedProviders | undefined {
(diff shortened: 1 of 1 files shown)


src/templates/CentralTemplateProvider.ts
```diff
@@ -123,12 +123,13 @@ export class CentralTemplateProvider implements Disposable {
         return await provider.getProjKey(context);
     }
 
-    private getCachedProvidersKey(language: ProjectLanguage, version: FuncVersion): string {
-        return language + version;
+    private getCachedProvidersKey(language: ProjectLanguage, languageModel: number | undefined, version: FuncVersion): string {
+        // NOTE: VS Code treats lack of a language model project setting as a 0, so treat undefined === 0.
+        return `${language}:${languageModel !== undefined ? languageModel : 0}:${version}`;
```
bwateratmsft, line 128:
```suggestion
        return `${language}:${languageModel ?? 0}:${version}`;
```
Nullish coalescing operator has the slight advantage of also catching `null`, in addition to `undefined` (even though in theory it shouldn't be happening...). Plus IMO it's a little cleaner but that's personal preference.

bwateratmsft (approved):

src/templates/CentralTemplateProvider.ts
philliphoff: Ok, I'll buy that.

bwateratmsft (approved):

PR #3316, opened 2022-09-06
All language model to be zero (i.e. undefined).

### LINKED ISSUE #3315: An error occurs for the setting "azureFunctions.projectLanguageModel" with the default value "0"
**OS:** Win 10/Mac (Pro)
**Build Version:** 20220902.7

**Repro Steps:**
1. Install the Azure Function extension.
2. Navigate to the settings page for the Azure Function extension.
3. Check whether the value of the setting "azureFunctions.projectLanguageModel" >=1 or not.

**Expect:**
The value of the setting "azureFunctions.projectLanguageModel" >=1 without any error.

**Actual:**
An error occurs for the setting "azureFunctions.projectLanguageModel" with the default value "0".
image

philliphoff:
Update `package.json` description of the project language model to allow zero as a value (which is treated the same as undefined), as that's what the VS Code API does.

Resolves #3315

### DIFF
--- a/package.json
+++ b/package.json
@@ -838,7 +838,7 @@
                         "scope": "resource",
                         "type": "number",
                         "description": "%azureFunctions.projectLanguageModel%",
-                        "minimum": 1
+                        "minimum": 0
                     },
                     "azureFunctions.projectTemplateKey": {
                         "scope": "resource",


bwateratmsft (approved):

philliphoff:
Grr, that flaky Windows test is annoying.

PR #3320, opened 2022-09-07
Fixes some issues related to the functionality of the "Connect to Github" command

### LINKED ISSUE #3214: It keeps "Connect to a GitHub Repository..." status after connecting to a GitHub repository
**OS:** All
**Azure Function Extension build Version:** 20220607.6

**Repro Steps:**
1. Expand a Function App -> Expand the "Deployment" node -> Click "Connect to a GitHub Repository...".
2. Check whether succeeds to connect the GitHub repository or not.

**Expect:**
Succeed to connect the GitHub repository.
image

**Actual:**
It keeps "Connect to a GitHub Repository..." status.
image

**More Info:**
1. Refresh can fix this issue.
2. This issue also reproduces on the latest released Function extension 1.7.3.

### LINKED ISSUE #3325: An error occurs when executing "Connect to a GitHub Repository..." command
**OS:** All
**Build Version:** 20220906.4

**Repro Steps:**
1. F1 -> Azure Functions: Connect to a GitHub Repository... -> Select a Function App.
2. Check whether the next step is "Choose organization" or not.

**Expect:**
The next step is "Choose organization".

**Actual:**
An error occurs.
image

**More Info:**
This issue does not reproduce when executing the action from Function App Deployments node.

MicroFish91:
Second Edit:  I'm liking this version of the implementation less and less.  I realized the edge-case I was running into had to do with how `showTreeItemPicker` is automatically executing its associated 'Connect To Github' command which is putting two calls to the same command on the stack.  I'm hoping v2 picker will handle this case better so we can just root directly into the `DeploymentsTreeItem` node.. let me convert this to a draft PR until discussed further...

------------------------------------------------------------------

Fixes #3214 
Edit:  Also now fixes #3325 

A lot of these changes ended up being slightly outside the scope of just fixing the issue.  I noticed that the command palette version of this command was always throwing an error and not functioning the way I would expect it to function, so I went ahead and added some support for it so it doesn't do that every time.  

There were a few tricky nuances to implementing due to the nature of how our tree item pickers work and how the nodes we are interested in are nested.  For example, I couldn't just `showTreeItem` or `pickAppResource` on `DeploymentTreeItem.contextValue` because there is a deployment node tucked away under `Slots` that is also a valid option and it would never allow the user to select it this way.... so, begrudgingly I ended up using the context on the `GenericTreeItem` representing `ConnectToGithub` as the anchor point instead.  I tested all combinations and I think it works in all situations now.

I would have also been open to removing it from the command palette entirely but seeing as it was still there, I assumed people wanted to keep it there as well.

Note:  This kind of issue might be a possible good idea/test case for v2 tree picker?

Edit:  Oh also, I am using the `SlotTreeItem` as the parent node to trigger for refresh due to some slightly inconsistent UI refresh behavior I noticed when refreshing the deployment node from under the `SlotsTreeItem` (basically when refreshing close to `SlotsTreeItem` deployment, the commit children won't show up without a slight wait and additional refresh... I assumed this must be related to some sort of timing issue).  The UI inconsistency may not be a huge deal though, so if for performance reasons we want to refresh closer to the node of interest, I'm happy to go in and change it to do so.

### DIFF
--- a/src/commands/deployments/connectToGitHub.ts
+++ b/src/commands/deployments/connectToGitHub.ts
@@ -5,20 +5,29 @@
 
 import { DeploymentsTreeItem, editScmType } from "@microsoft/vscode-azext-azureappservice";
 import { GenericTreeItem, IActionContext } from "@microsoft/vscode-azext-utils";
-import { functionFilter, ScmType } from "../../constants";
+import { connectToGithubContextValue, functionFilter, ScmType } from "../../constants";
 import { ext } from "../../extensionVariables";
+import { ResolvedFunctionAppResource } from "../../tree/ResolvedFunctionAppResource";
 import { isSlotTreeItem, SlotTreeItem } from "../../tree/SlotTreeItem";
+import { treeUtils } from "../../utils/treeUtils";
 
 export async function connectToGitHub(context: IActionContext, target?: GenericTreeItem): Promise<void> {
-    const parentNode: SlotTreeItem = await ext.rgApi.pickAppResource<SlotTreeItem>(context, {
-        filter: functionFilter,
-    });
-
+    let parentNode: SlotTreeItem;
     let connectToNode: GenericTreeItem;
 
     if (!target) {
-        connectToNode = await ext.rgApi.appResourceTree.showTreeItemPicker<GenericTreeItem>('ConnectToGithub', context, parentNode);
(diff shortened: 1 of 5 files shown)


MicroFish91:
My understanding of the scope of this problem has changed, closing this out in favor of https://github.com/microsoft/vscode-azurefunctions/pull/3349

Issue #1234, opened 2019-04-28
deploy error


ywyt738:
Repro steps:1.2.Action: azureFunctions.deployError type: TypeErrorError Message: Cannot read property 'name' of undefinedVersion: 0.16.0OS: win32VS Code Version: 1.33.1<details><summary>Call Stack</summary>```ResourceGroupCreateStep.<anonymous> (dist/extension.bundle.js:165:73094)Generator.next (<anonymous>)module.exports.n (dist/extension.bundle.js:165:72800)new Promise (<anonymous>)module.exports.n (dist/extension.bundle.js:165:72577)ResourceGroupCreateStep.execute (dist/extension.bundle.js:165:73017)AzureWizard.<anonymous> (dist/extension.bundle.js:277:681795)Generator.next (<anonymous>)module.exports.n (dist/extension.bundle.js:277:679901)new Promise (<anonymous>)module.exports.n (dist/extension.bundle.js:277:679678)module.exports.t.AzureWizard.execute.s.window.withProgress.t (dist/extension.bundle.js:277:681434)c._withProgress (d:/Microsoft VS extensionHostProcess.js:441:75)c.withProgress (d:/Microsoft VS extensionHostProcess.js:440:836)Object.withProgress (d:/Microsoft VS extensionHostProcess.js:702:393)AzureWizard.<anonymous> (dist/extension.bundle.js:277:681375)Generator.next (<anonymous>)module.exports.n (dist/extension.bundle.js:277:679901)new Promise (<anonymous>)module.exports.n (dist/extension.bundle.js:277:679678)AzureWizard.execute (dist/extension.bundle.js:277:681327)FunctionAppProvider.<anonymous> (dist/extension.bundle.js:172:156890)Generator.next (<anonymous>)o (dist/extension.bundle.js:172:152743)process._tickCallback (internal/process/next_tick.js:68:7)```</details>

PR #3335, opened 2022-09-14
Update `appservice` package

### LINKED ISSUE #3188: An error "ServiceUnavailable" occurs in Files node and Logs node of a Windows Function App
**OS:** Win10/M1
**Azure Resource Extension Build Version:** 20220523.7
**Azure Function Extension build Version:** 20220523.6

**Repro Steps:**
1. Create a **Windows** Function app.
2. Expand this web app -> Files node and Logs node.
3. Check whether there is no error in Files node and Logs node.

**Expect:**
There is no error in Files node and Logs node.

**Actual:**
An error "ServiceUnavailable" occurs in Files node and Logs node.
image

**More Info:**
This issue does not reproduce for a Linux Function App.

MicroFish91:
Fixes #3188 

Change made here: [AppService #1234](https://github.com/microsoft/vscode-azuretools/pull/1234)

When I was testing again today, it seemed like it was mostly happening with .NET runtimes... not sure why but it has been harder to repro today than previously... however, when running the old and new implementation side-by-side, this change seems to fix it when it does occur while the old implementation will just hang on the service unavailable error...

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
                 "@azure/arm-storage": "^17.0.0",
                 "@azure/ms-rest-js": "^2.2.1",
                 "@azure/storage-blob": "^12.5.0",
-                "@microsoft/vscode-azext-azureappservice": "^0.7.3",
+                "@microsoft/vscode-azext-azureappservice": "^0.7.4",
                 "@microsoft/vscode-azext-azureutils": "^0.3.4",
                 "@microsoft/vscode-azext-utils": "^0.3.12",
                 "escape-string-regexp": "^4.0.0",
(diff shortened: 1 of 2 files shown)


alexweininger:
> When I was testing again today, it seemed like it was mostly happening with .NET runtimes... not sure why but it has been harder to repro today than previously... however, when running the old and new implementation side-by-side, this change seems to fix it when it does occur while the old implementation will just hang on the service unavailable error...

So the error still occurs occasionally?

alexweininger (approved):

bwateratmsft (approved):

MicroFish91:
> So the error still occurs occasionally?

I feel like I may have seen it resurface again one time in all my tests using the new implementation, but it disappeared upon refresh.  With the old implementation the error would surface pretty regularly and refreshing would not fix it.

Side-note:  I also sometimes get `BadGateway` errors, but I was already getting those using the old implementation as well - refreshing fixes those in both scenarios for me.

Issue #3269, opened 2022-07-30
Subscription not found when creating function app with trial account


scook1802:
Does this occur consistently? 
Repro steps:

1.
2.

Action: azureFunctions.createFunctionApp
Error type: SubscriptionNotFound
Error Message: Subscription redacted:id was not found.

Version: 1.7.4
OS: win32
OS Release: 10.0.22000
Product: Visual Studio Code
Product Version: 1.69.2
Language: en

<details>
<summary>Call Stack</summary>

```
handleErrorResponse extension.bundle.js:2:1823131
deserializeResponseBody extension.bundle.js:2:1823862
```

</details>

WallyWally12:
> Does this occur consistently? Repro steps:
> 
> 
> Action: azureFunctions.createFunctionApp Error type: SubscriptionNotFound Error Message: Subscription redacted:id was not found.
> 
> Version: 1.7.4 OS: win32 OS Release: 10.0.22000 Product: Visual Studio Code Product Version: 1.69.2 Language: en
> 
> Call Stack

Same here (new user)!

nturinski:
Are you guys using student accounts by chance?

WallyWally12:
Hi,

No, just normal accounts in trial mode with the 200 dollar allowance. I even used two different accounts, but both fail.
I was only able to upload to an App Service, however I’m now deviating from the Microsoft training, so that not ideal.


Best regards, Met vriendelijke groet,

Ingmar Waller
Solution Architect & Product owner ORTEC Routing & Dispatch

+31 (0)6 1313 7176 | ***@***.******@***.***> | LinkedIn<http://www.linkedin.com/in/ingmarwaller> | ortec.com<https://ortec.com/>

MJSanfelippo:
Hi, I am running into this too. I created a new Azure subscription on my personal account with my $150/month free Azure credits as MSFT employee. I am following [this guide](https://docs.microsoft.com/en-us/learn/modules/develop-azure-functions/5-create-function-visual-studio-code) and running into this same issue.

raff-run:
Having the exact same problem here. Just created a 200 dollar free acount with this my email.

BartoszCki:
Same as described by @raff-run

minaelee:
[...]

PR #3343, opened 2022-09-19
Register providers before checking name availability

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3269

This issue is happening with _NEW_ subscriptions. I don't think the trial/pay-as-you go is actually the issue here. The storage provider isn't registered, but we are calling `isRelatedNameAvailable` for storage accounts before the provider is verified to be registered.

Since it's school season, I think we're seeing an influx of these issues due to the fact that classes are starting up and students are registering trial accounts.

### DIFF
--- a/src/tree/SubscriptionTreeItem.ts
+++ b/src/tree/SubscriptionTreeItem.ts
@@ -5,19 +5,20 @@
 
 import { Site, WebSiteManagementClient } from '@azure/arm-appservice';
 import { AppInsightsCreateStep, AppInsightsListStep, AppKind, AppServicePlanCreateStep, CustomLocationListStep, IAppServiceWizardContext, LogAnalyticsCreateStep, SiteNameStep, WebsiteOS } from '@microsoft/vscode-azext-azureappservice';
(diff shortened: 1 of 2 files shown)


alexweininger:
Do you know if this was a regression?

alexweininger (approved):
Although I don't think we've seen similar issues in other repos, we should check if we need to make this change in other places we use `VerifyProvidersStep`.

edit: also nice work 🚀

nturinski:
> Do you know if this was a regression?

I don't think so. We haven't changed anything with the generating related name code. It could be something changed on the Azure Storage that requires the provider be registered in order to call `checkNameAvailability`? Or it _could_ be something to do with trial accounts?

IHVH:
> Fixes #3269
> 
...
> Since it's school season, I think we're seeing an influx of these issues due to the fact that classes are starting up and students are registering trial accounts.

It's really, this subscription is used for student learning.

alexweininger:
@IHVH 

A fix for this bug has been released in https://github.com/microsoft/vscode-azurefunctions/releases/tag/v1.8.0, please let us know if you're still experiencing this issue.

IHVH:
I'm not sure if #3269 is related to my question #3363
But yes my problem is still there
Thanks @alexweininger !

PR #3357, opened 2022-09-23
Fixed extra dialogue pop up when creating a C# project with Azure Blob/Queue Storage trigger  

### LINKED ISSUE #3023: There is an extra dialog when creating a C# project with "Azure blob Storage trigger" function
**OS:** Win10
**Build Version:** 20211206.2

**Repro Steps:**

1. Open an empty folder in VS Code -> Click "Create New Project..." -> Select the folder -> Select C# language -> Select a .NET runtime -> Select "Azure Blob Storage trigger" -> Press Enter -> Select "Create new local app setting" -> Select a subscription -> **Select an existing storage account** -> Press Enter with default value.
2. Check whether there is no extra dialog.

**Expect:**
There is no extra dialog.

**Actual:**
There is an extra dialog. (we have selected an existing storage account in the step1)
image

**More Info:**

1. This issue doesn't reproduce for other languages.
2. This issue also reproduces when creating a C# project with "Azure Queue Storage trigger" function.

motm32:
Fixed #3023.

Also could potentially remove this [if statement][1] but I was not sure if anything else depended on this.

[1]: https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/createFunction/FunctionSubWizard.ts#:~:text=if%20((,%7D


nturinski (commented):
I don't think we can just remove CSharp from the list of project languages that can validate on debug since the comment below still applies.

Do we know why this is occurring?  I don't think we can remove this check from the if statement, but maybe we can alter the prompt step's criteria?

motm32:
> Do we know why this is occurring? I don't think we can remove this check from the if statement, but maybe we can alter the prompt step's criteria?

It seems like this is occurring because of the pushed prompt step in the if statement linked above. I had thought this prompt step may not be needed anymore (in this case) since we are already being asked to link a storage account in previous steps. Since this is only happening with C# and Azure Storage blob/queue triggers I could put a check to skip over this prompt step if those are chosen.

src/commands/appSettings/IAzureWebJobsStorageWizardContext.ts
```diff
@@ -3,9 +3,11 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { StorageAccount } from "@azure/arm-storage";
 import { IActionContext, ISubscriptionContext } from "@microsoft/vscode-azext-utils";
 
 export interface IAzureWebJobsStorageWizardContext extends IActionContext, Partial<ISubscriptionContext> {
     projectPath: string;
+    storageAccount?: StorageAccount;
```
nturinski, line 11: Does this actually get set anywhere? I'd assume not since you just added it to the context definition 😅

motm32: I was also thinking that when I originally went to write this code I found the storage account is set in StorageAccountListStep.js which I couldn't find a way to access that data, but I randomly went to debug my code it does seem to set the storage account data correctly as shown in the image below. 

image

nturinski: Oh right, in that case, what you probably should do is just extend `IStorageAccountWizardContext` from the `vscode-azext-azureutils` package since that's where it is actually setting the `storageAccount` property.

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -45,7 +45,9 @@ export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardC
     }
 
     public shouldPrompt(context: T): boolean {
-        return !context.azureWebJobsStorageType && context.storageAccount === undefined && context.newStorageAccountName === undefined;
+        //Only want to prompt if no storage account it selected
```
nturinski, line 48: Was this written by afternoon Megan? lmao
```suggestion
        // Only should prompt if no storage account was selected
```

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -45,7 +45,9 @@ export class AzureWebJobsStoragePromptStep<T extends IAzureWebJobsStorageWizardC
...
+        context.azureWebJobsStorageType = (context.storageAccount !== undefined || context.newStorageAccountName !== undefined) ? 'azure' : undefined;
```
nturinski, line 49: You can use `!!` as a shorthand in JS to check to see if a value is truthy or not. Wasn't sure if you were familiar with it. It's a little goofy, but it's using `!` to first get the opposite boolean value, and then the next `!` converts it back. The one caveat is if 0 or "" would be valid inputs, this would give you the incorrect value.

```suggestion
        context.azureWebJobsStorageType = (!!context.storageAccount || !!context.newStorageAccountName) ? 'azure' : undefined;
```

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
motm32: It was but also included distractions from book talk which doesn't help :)

PR #3362, opened 2022-09-28
Automatically `npm install` after running `Create New Project` (JS/TS)

### LINKED ISSUE #3034: Automatically run npm install after initializing TypeScript project
Is it possible to run `npm i` after initializing a TS project (maybe do it for JS too for consistency)?

MicroFish91:
Closes #3034


src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
```diff
@@ -56,8 +57,26 @@ export class ScriptFunctionCreateStep extends FunctionCreateStepBase<IScriptFunc
...
+
+        await this.installDependencies(language, context.projectPath);
         return fileName ? path.join(functionPath, fileName) : functionJsonPath;
     }
 
+    private async installDependencies(language: string, projectPath: string): Promise<void> {
+        switch (language) {
+            case ProjectLanguage.CSharpScript:
+            case ProjectLanguage.FSharpScript:
+            case ProjectLanguage.JavaScript:
```
bwateratmsft, line 69: This `switch` statement isn't implemented correctly; this will essentially do `nodeUtils.installDependencies(projectPath)` for *all* project types.
This is what you're looking for:
```typescript
        switch (language) {
            case ProjectLanguage.JavaScript:
            case ProjectLanguage.TypeScript:
                await nodeUtils.installDependencies(projectPath);
                break;
            case ProjectLanguage.CSharpScript:
            case ProjectLanguage.FSharpScript:
            case ProjectLanguage.PowerShell:
            case ProjectLanguage.Python:
            default:
                break;
        }
```

src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
```diff
@@ -56,8 +57,26 @@ export class ScriptFunctionCreateStep extends FunctionCreateStepBase<IScriptFunc
 
         const language: ProjectLanguage = nonNullProp(context, 'language');
         const fileName: string | undefined = getScriptFileNameFromLanguage(language);
+
+        await this.installDependencies(language, context.projectPath);
         return fileName ? path.join(functionPath, fileName) : functionJsonPath;
     }
 
+    private async installDependencies(language: string, projectPath: string): Promise<void> {
```
bwateratmsft, line 65: Maybe `installDependenciesIfNeeded` instead, since it doesn't always do something?

src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
MicroFish91: Oh right, that makes sense, thank you for catching that!

nturinski: I think I'd rather there be a separate `JavaScriptFunctionCreateStep` that handles this instead of having a switch case for generic `ScriptFunctionCreateStep`.  We already have a `TypeScruptFunctionCreateStep` as it is.

src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
```diff
@@ -36,6 +40,12 @@ export class JavaScriptProjectCreateStep extends ScriptProjectCreateStep {
                 devDependencies: this.getPackageJsonDevDeps(context)
             });
         }
+
```
nturinski, line 43: I liked having it as a function still. I think it'd be nicer to have it as a private function on the class.

Also, bit of a nit, but seeing `PackageManger.npm` as the command is a bit weird to me. We both know that the package manager is `npm` which is also the name of the CLI command, but it doesn't feel very intuitive to me to pass it in as a command arg. But that's definitely a non-issue.

nturinski (dismissed):

nturinski (approved):

PR #3364, opened 2022-10-05
Fixes function deployment error when AzureWebJobsStorage connection string is "UseDevelopmentStorage=true"

### LINKED ISSUE #3298: Cannot release function unless local settings points at a live resource
We're trying to release a queue-triggered function in Azure 

The local.settings.json file needs to have an AzureWebJobStorage connection string which is a live resource (instead of the recommended "UseDevelopmentStorage=true;") for the release to work.

If the AzureWebJobsStorage connection string is "UseDevelopmentStorage=true;", the function deployment attempts to connect to a local resource, even though this value is overridden in the function app's config in the portal. This leads to an error 
```Retry failed after 6 tries. Retry settings can be adjusted in ClientOptions.Retry. (An attempt was made to access a socket in a way forbidden by its access permissions. (127.0.0.1:10001))```

Curiously, if the AzureWebJobStorage is updated to point at a storage account, the release will succeed and the AzureWebJobStorage value from the portal configuration will be used thereafter.

Relates to [this](https://github.com/microsoft/vscode-azurefunctions/issues/1523)

### PR DESCRIPTION
**Pull request by motm32**
Fixes #3298

### DIFF
--- a/src/commands/appSettings/uploadAppSettings.ts
+++ b/src/commands/appSettings/uploadAppSettings.ts
@@ -53,6 +53,11 @@ export async function uploadAppSettingsInternal(context: IActionContext, client:
         }
 
         const excludedAppSettings: string[] = [];
+        if (localSettings.Values["AzureWebJobsStorage"] === 'UseDevelopmentStorage=true') {
+            delete localSettings.Values?.["AzureWebJobsStorage"];
+            excludedAppSettings.push("AzureWebJobsStorage");
+        }
+
         if (exclude) {
             Object.keys(localSettings.Values).forEach((settingName) => {
                 if (exclude.some((exclusion) => typeof exclusion === 'string' ? settingName.toLowerCase() === exclusion.toLowerCase() : settingName.match(new RegExp(exclusion, 'i')))) {
--- a/src/commands/appSettings/uploadAppSettings.ts
+++ b/src/commands/appSettings/uploadAppSettings.ts
@@ -53,6 +53,7 @@ export async function uploadAppSettingsInternal(context: IActionContext, client:
         }
 
         const excludedAppSettings: string[] = [];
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/commands/appSettings/uploadAppSettings.ts [hunk 1]
```diff
@@ -53,6 +53,11 @@ export async function uploadAppSettingsInternal(context: IActionContext, client:
         }
 
         const excludedAppSettings: string[] = [];
+        if (localSettings.Values["AzureWebJobsStorage"] === 'UseDevelopmentStorage=true') {
```
nturinski, line 56: Definitely good to have a clarifying comment here about this one. In this case, it's probably good to also cite the issue since it describes the scenario pretty well.

nturinski (approved):

src/commands/appSettings/uploadAppSettings.ts
```diff
@@ -53,6 +53,12 @@ export async function uploadAppSettingsInternal(context: IActionContext, client:
         }
 
         const excludedAppSettings: string[] = [];
+        // Don't want AzureWebJobsStorage to be overwritten in this case as it should point at a live resource: https://github.com/microsoft/vscode-azurefunctions/issues/3298
+        if (localSettings.Values["AzureWebJobsStorage"] === 'UseDevelopmentStorage=true') {
```
MicroFish91, line 57: Small nit, but I think there's a constant we've been using for `UseDevelopmentStorage=true`

src/commands/appSettings/uploadAppSettings.ts
MicroFish91: Small nit, but I think we already have a constant that could be used here representing `UseDevelopmentStorage=true`

nturinski: He is correct, `localEmulatorConnectionString`.

PR #3365, opened 2022-10-06
Updated templates for Dynamic Concurrency 

### LINKED ISSUE #3248: Update templates for dynamic concurrency
Dynamic concurrency is a performance feature that went GA for functions end of May.

We want to enable the feature by default for new apps, and it requires updating host.json settings for the app.

motm32:
Fixes #3248.

### DIFF
--- a/src/commands/createFunction/FunctionCreateStepBase.ts
+++ b/src/commands/createFunction/FunctionCreateStepBase.ts
@@ -56,14 +56,16 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
         await verifyExtensionBundle(context, template);
 
         const cachedFunc: ICachedFunction = { projectPath: context.projectPath, newFilePath, isHttpTrigger: template.isHttpTrigger };
-        if (context.functionTemplate?.defaultFunctionName === 'ServiceBusQueueTrigger' || context.functionTemplate?.defaultFunctionName === 'BlobTrigger' || context.functionTemplate?.defaultFunctionName === 'QueueTrigger') {
-            const hostFilePath: string = path.join(context.projectPath, hostFileName);
-            const hostJson = await AzExtFsExtra.readJSON<IHostJsonV2>(hostFilePath);
-            hostJson.concurrency = {
-                dynamicConcurrencyEnabled: true,
-                snapshotPersistenceEnabled: true
+        const hostFilePath: string = path.join(context.projectPath, hostFileName);
+        if (await AzExtFsExtra.pathExists(hostFilePath)) {
+            if (context.functionTemplate?.isDynamicConcurrent) {
+                const hostJson = await AzExtFsExtra.readJSON<IHostJsonV2>(hostFilePath);
+                hostJson.concurrency = {
+                    dynamicConcurrencyEnabled: true,
+                    snapshotPersistenceEnabled: true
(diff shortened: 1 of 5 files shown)


src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -5,13 +5,16 @@
...
+import { hostFileName } from '../../constants';
 import { ext } from '../../extensionVariables';
+import { IHostJsonV2 } from '../../funcConfig/host';
 import { localize } from '../../localize';
 import { IFunctionTemplate } from '../../templates/IFunctionTemplate';
 import { nonNullProp } from '../../utils/nonNull';
 import { verifyExtensionBundle } from '../../utils/verifyExtensionBundle';
 import { getContainingWorkspace } from '../../utils/workspace';
 import { IFunctionWizardContext } from './IFunctionWizardContext';
+import path = require('path');
```
bwateratmsft, line 17: Does this not work?
```suggestion
import * as path from 'path';
```
If it does work, ESLint will also want to reorder the imports.

bwateratmsft (approved):

bwateratmsft (dismissed):

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -53,6 +56,15 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
         await verifyExtensionBundle(context, template);
 
         const cachedFunc: ICachedFunction = { projectPath: context.projectPath, newFilePath, isHttpTrigger: template.isHttpTrigger };
+        if (context.functionTemplate?.defaultFunctionName === 'ServiceBusQueueTrigger' || context.functionTemplate?.defaultFunctionName === 'BlobTrigger' || context.functionTemplate?.defaultFunctionName === 'QueueTrigger') {
```
nturinski, line 59: You shouldn't use properties like `defaultFunctionName`, `name`, or `label` when you can help it. There's always a chance that these could change, and sometimes localization can screw these conditionals up.

I think the preferred method would be to check the `id` property instead.

Also, rather than have a conditional like this, I think it'd be better to change the `IFunctionTemplate` to have a property that is `isDynamicConcurrent` (or something like that) which would have this check. It would be a little bit more tidy 😊

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -53,6 +56,15 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
...
+            const hostFilePath: string = path.join(context.projectPath, hostFileName);
+            const hostJson = await AzExtFsExtra.readJSON<IHostJsonV2>(hostFilePath);
```
nturinski, line 61: It's not likely, but you should still ensure the file is there before you try to do a read/write action on it. People could have created a project and deleted their `host.json` during the wizard or something ridiculous like that.

nturinski (changes requested):

nturinski (approved):
Looks great! 😊

alexweininger:
JSON schema is being updated over at https://github.com/SchemaStore/schemastore/pull/2516

nturinski (approved):

Issue #3369, opened 2022-10-10
Cannot read properties of undefined (reading 'isLinux')


mumurug-MSFT:
**Details:**
VS code version: 1.72.0
Azure Functions extension version: v.1.8.1
Python extension version: v.2022.16.0
OS: Windows

**Repro Steps:**
Following https://learn.microsoft.com/en-us/azure/azure-functions/functions-develop-vs-code?tabs=python#republish-project-files to deploy Azure Functions app to Azure caused the below error:

![image](https://user-images.githubusercontent.com/95652148/194883340-72cbfe82-ebe6-4b16-a7e0-48777aa9ab44.png)

Unable to deploy the app via VS code.

karlwirenrocket:
Having the same issue on Version: 1.71.0

AlexFilipovici:
Having the same issue - Cannot read properties of undefined (reading 'createClient')

Following https://learn.microsoft.com/en-us/azure/azure-functions/create-first-function-vs-code-typescript#deploy-the-project-to-azure

VS code version: 1.72.1
Azure Functions extension version: v.1.8.1

nturinski:
Hi all, I've found a workaround to this issue. The tutorial will have you press the "Deploy" cloud icon on the local workspace, which is where the issue is.

The two methods you can use are:
- Open Resources > Function App > <your_function_app>
- Right-click the function app
- Click "Deploy to Function App..."
![image](https://user-images.githubusercontent.com/5290572/195707520-6b83fcef-ffe8-480c-9c4e-cf78c7cdf6b0.png)

The other method is:
- Press F1 to open the Command Palette
- Type `>Azure Functions: Deploy to Function App...`
- Select `Azure Functions: Deploy to Function App...`
- Follow the prompts to select your Function App
![image](https://user-images.githubusercontent.com/5290572/195707789-a7749651-e40a-4430-8a55-5d525d03cf3d.png)

Sorry about the issue! Hopefully these unblock you as we work on resolving this.

AlexFilipovici:
Thanks, @nturinski!

nturinski:
Hi all, just wanted to let you know v1.8.2 has been released and should resolve this issue.

PR #3370, opened 2022-10-13
Check if the passed in node is a LocalProjectTreeItem

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3369

Because of this [change](https://code.visualstudio.com/updates/v1_72#_selected-tree-items-passed-to-viewtitle-actions), the Local Project tree item will get passed in to deploy which is what's causing the error.

I think when we use the new wrapper, since we pass in `node` and `nodes` , it might be easier for us to tell when a node was right-clicked vs. just selected, though I'm not sure how VS Code passes us those args yet.

I assume that once that changes, we can change how we do this check.

### DIFF
--- a/src/utils/treeUtils.ts
+++ b/src/utils/treeUtils.ts
@@ -23,6 +23,7 @@ export namespace treeUtils {
         return ext.context.asAbsolutePath('resources');
     }
 
+    // replace with azext-utils when it's released
     export function isAzExtTreeItem(ti: unknown): ti is AzExtTreeItem {
         return !!ti && (ti as AzExtTreeItem).fullId !== undefined && (ti as AzExtTreeItem).fullId !== null;
     }
--- a/src/utils/treeUtils.ts
+++ b/src/utils/treeUtils.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { TreeItemIconPath } from '@microsoft/vscode-azext-utils';
+import { AzExtTreeItem, TreeItemIconPath } from '@microsoft/vscode-azext-utils';
 import * as path from 'path';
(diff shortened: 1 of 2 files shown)


alexweininger (dismissed):
Wow. I can repro the `Cannot read properties of undefined (reading 'createClient')` error, but not the isLinux one. Do you know how to repro that?

nturinski:
> Wow. I can repro the `Cannot read properties of undefined (reading 'createClient')` error, but not the isLinux one. Do you know how to repro that?

I haven't been able to repro it either. Still kinda poking around, but really not sure how it's happening. What's crazy is once you have selected something, even if you unselect it, it still passes that tree item in. Super annoying. Only way to unselect something is to reload the window.

alexweininger:
> Only way to unselect something is to reload the window.

I noticed the same thing. Weird.

src/utils/treeUtils.ts
```diff
@@ -22,4 +22,8 @@ export namespace treeUtils {
     function getResourcesPath(): string {
         return ext.context.asAbsolutePath('resources');
     }
+
+    export function isAzExtTreeItem(ti: unknown): ti is AzExtTreeItem {
```
bwateratmsft, line 26: We're planning to add this function to the core utils library (https://github.com/microsoft/vscode-azuretools/pull/1229/files#diff-a94c5ee42a49df43d3b3624aea93155299963c5571dcefed509b755d942640ae). Can we either move that function now, so this PR can use it, or otherwise make a note in the code here to start using that function from utils when it is available?

alexweininger (approved):

src/utils/treeUtils.ts
alexweininger: Exists here too https://github.com/microsoft/vscode-azuretools/blob/main/appservice/src/deploy/getDeployNode.ts#L13

PR #3382, opened 2022-10-19
Added End of Life warning button for when a stack has an upcoming or is past its EOL

### LINKED ISSUE #3353: Show warning when a stack is selected that has an upcoming or is past its end of life (EOL)
The Azure Functions Stacks API [has been updated to include End of Life dates](https://github.com/Azure/azure-functions-ux/pull/6862/files) based on our [Azure Functions language runtime support policy table](https://docs.microsoft.com/en-us/azure/azure-functions/language-support-policy#retirement-policy-exceptions).

We now would like to have the Azure Functions extension for VS Code to show warnings when a stack is selected and has an EOL date within **6 months** (or if EOL already passed).

Verbiage for the warning can be based on the Important note on the [Azure Functions runtime versions overview](https://learn.microsoft.com/azure/azure-functions/functions-versions) documentation:
"After the deadline, function apps can be created and deployed, and existing apps continue to run. However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them".

This extension already uses the Stacks API.

motm32:
Fixes #3353. 

I still am not sure if this is the most discoverable since the button isn't the most noticeable but may be the best option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1060,7 +1060,7 @@
     "devDependencies": {
         "@azure/arm-resources": "^3.0.0",
         "@microsoft/eslint-config-azuretools": "^0.1.0",
-        "@microsoft/vscode-azext-dev": "^0.1.3",
+        "@microsoft/vscode-azext-dev": "^0.1.4",
         "@types/fs-extra": "^8.1.0",
         "@types/gulp": "^4.0.6",
         "@types/gulp-filter": "^3.0.33",
--- a/package.json
+++ b/package.json
@@ -1038,6 +1038,11 @@
                         "type": "boolean",
                         "default": false,
                         "description": "%azureFunctions.showPysteinModel%"
+                    },
+                    "azureFunctions.endOfLifeWarning": {
+                        "type": "boolean",
+                        "description": "%azureFunctions.endOfLifeWarning%",
+                        "default": true
                     }
                 }
             }
--- a/package.json
+++ b/package.json
@@ -1070,7 +1070,7 @@
         "@types/ps-tree": "^1.1.0",
         "@types/request": "2.0.7",
         "@types/semver": "^5.5.0",
-        "@types/vscode": "1.57.0",
+        "@types/vscode": "^1.63.0",
         "@types/websocket": "^1.0.0",
         "@types/xml2js": "^0.4.5",
         "@types/xregexp": "4.3.0",
--- a/package.json
+++ b/package.json
@@ -1069,7 +1069,7 @@
         "@types/ps-tree": "^1.1.0",
         "@types/request": "2.0.7",
         "@types/semver": "^5.5.0",
-        "@types/vscode": "^1.63.0",
+        "@types/vscode": "^1.66.0",
         "@types/websocket": "^1.0.0",
         "@types/xml2js": "^0.4.5",
         "@types/xregexp": "4.3.0",
--- a/package.nls.json
+++ b/package.nls.json
@@ -33,7 +33,7 @@
     "azureFunctions.enableJavaRemoteDebugging": "Enable remote debugging for Java Functions Apps running on Windows. (experimental)",
     "azureFunctions.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
     "azureFunctions.enableRemoteDebugging": "Enable remote debugging for Node.js Function Apps running on Linux App Service plans. Consumption plans are not supported. (experimental)",
-    "azureFunctions.endOfLifeWarning": "Enable End Of Life Warning",
+    "azureFunctions.endOfLifeWarning": "Show a warning when creating Function Apps with stacks within 6 months of their end of life.",
     "azureFunctions.executeFunction": "Execute Function Now...",
     "azureFunctions.funcCliPath": "The path to the 'func' executable to use for debug and deploy tasks. For example, set it to 'node_modules/.bin/func' if using the func cli installed as a local npm package.",
     "azureFunctions.initProjectForVSCode": "Initialize Project for Use with VS Code...",
--- a/package.nls.json
+++ b/package.nls.json
@@ -33,6 +33,7 @@
     "azureFunctions.enableJavaRemoteDebugging": "Enable remote debugging for Java Functions Apps running on Windows. (experimental)",
     "azureFunctions.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
     "azureFunctions.enableRemoteDebugging": "Enable remote debugging for Node.js Function Apps running on Linux App Service plans. Consumption plans are not supported. (experimental)",
+    "azureFunctions.endOfLifeWarning": "Enable End Of Life Warning",
     "azureFunctions.executeFunction": "Execute Function Now...",
     "azureFunctions.funcCliPath": "The path to the 'func' executable to use for debug and deploy tasks. For example, set it to 'node_modules/.bin/func' if using the func cli installed as a local npm package.",
     "azureFunctions.initProjectForVSCode": "Initialize Project for Use with VS Code...",
--- a/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
+++ b/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
@@ -13,14 +13,17 @@ export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAp
     public async prompt(context: IFunctionAppWizardContext): Promise<void> {
         const settingKey: string = 'endOfLifeWarning';
         if (getWorkspaceSetting<boolean>(settingKey)) {
-            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up." +
-                "After the deadline, function apps can be created and deployed, and existing apps continue to run." +
-                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
-            const result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
-            if (result === DialogResponses.learnMore) {
+            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
+                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
+                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
+            const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
+            do {
                 await openUrl(funcVersionLink);
-            }
-            else if (result === DialogResponses.dontWarnAgain) {
+                result = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
+            } while (result === DialogResponses.learnMore);
+
+            if (result === DialogResponses.dontWarnAgain) {
                 await updateGlobalSetting(settingKey, false);
             }
         }
--- a/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
+++ b/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
@@ -1,0 +1,31 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+import { AzureWizardPromptStep, DialogResponses, openUrl } from "@microsoft/vscode-azext-utils";
+import { MessageItem } from "vscode";
+import { funcVersionLink } from "../../../FuncVersion";
+import { localize } from "../../../localize";
+import { getWorkspaceSetting, updateGlobalSetting } from "../../../vsCodeConfig/settings";
+import { IFunctionAppWizardContext } from "../IFunctionAppWizardContext";
+
+export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAppWizardContext> {
+    public async prompt(context: IFunctionAppWizardContext): Promise<void> {
+        const settingKey: string = 'endOfLifeWarning';
+        if (getWorkspaceSetting<boolean>(settingKey)) {
+            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up." +
+                "After the deadline, function apps can be created and deployed, and existing apps continue to run." +
+                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
+            const result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
+            if (result === DialogResponses.learnMore) {
+                await openUrl(funcVersionLink);
+            }
+            else if (result === DialogResponses.dontWarnAgain) {
+                await updateGlobalSetting(settingKey, false);
+            }
+        }
+    }
+    public shouldPrompt(context: IFunctionAppWizardContext): boolean {
+        return !!context.newSiteStack;
+    }
+}
--- a/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
+++ b/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
@@ -13,22 +13,27 @@ export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAp
     public async prompt(context: IFunctionAppWizardContext): Promise<void> {
         const settingKey: string = 'endOfLifeWarning';
         if (getWorkspaceSetting<boolean>(settingKey)) {
-            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
-                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
-                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
-            const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
-            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
-            while (result === DialogResponses.learnMore) {
+            let result: Promise<MessageItem> = this.getWarningMessage(context);
+            while (await result === DialogResponses.learnMore) {
                 await openUrl(funcVersionLink);
-                result = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
+                result = this.getWarningMessage(context);
             }
 
-            if (result === DialogResponses.dontWarnAgain) {
+            if (await result === DialogResponses.dontWarnAgain) {
                 await updateGlobalSetting(settingKey, false);
             }
         }
     }
+
     public shouldPrompt(context: IFunctionAppWizardContext): boolean {
         return !!context.newSiteStack;
+    }
+
+    private async getWarningMessage(context: IFunctionAppWizardContext): Promise<MessageItem> {
+        const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
+            "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
+            "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
+        const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+        return await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain);
     }
 }
--- a/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
+++ b/src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
@@ -2,7 +2,7 @@
 *  Copyright (c) Microsoft Corporation. All rights reserved.
 *  Licensed under the MIT License. See License.md in the project root for license information.
 *--------------------------------------------------------------------------------------------*/
-import { AzureWizardPromptStep, DialogResponses, openUrl } from "@microsoft/vscode-azext-utils";
+import { AzureWizardPromptStep, DialogResponses, IActionContext, openUrl } from "@microsoft/vscode-azext-utils";
 import { MessageItem } from "vscode";
 import { funcVersionLink } from "../../../FuncVersion";
 import { localize } from "../../../localize";
@@ -13,13 +13,13 @@ export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAp
     public async prompt(context: IFunctionAppWizardContext): Promise<void> {
         const settingKey: string = 'endOfLifeWarning';
         if (getWorkspaceSetting<boolean>(settingKey)) {
-            let result: Promise<MessageItem> = this.getWarningMessage(context);
-            while (await result === DialogResponses.learnMore) {
+            let result: MessageItem = await this.showEOLWarningMessage(context);
+            while (result === DialogResponses.learnMore) {
                 await openUrl(funcVersionLink);
-                result = this.getWarningMessage(context);
+                result = await this.showEOLWarningMessage(context);
             }
 
-            if (await result === DialogResponses.dontWarnAgain) {
+            if (result === DialogResponses.dontWarnAgain) {
                 await updateGlobalSetting(settingKey, false);
             }
         }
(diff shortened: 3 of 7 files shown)


nturinski:
> I still am not sure if this is the most discoverable since the button isn't the most noticeable but may be the best option.

Could you provide a screenshot of this? Makes it easier to contribute opinions 😉

motm32:
Haha yes sorry 😊
 
image

Here is an example. I think it may be hard to notice since the button only appears when you hover your mouse over a stack with a close EOL.

alexweininger (commented):
It doesn't seem like any QuickPick feature can properly present this information.

I lean towards using the description, however we cannot provide all needed info in there.

The button is great since we can provide a tooltip, but as Megan pointed out, it's super hard to discover which sorta defeats the purpose of this feature.

I'm thinking the best option might be to use `showWarningMessage` when a stack is selected. I would also want the ability to pick "Don't show this again" on the message since it could be annoying.

nturinski:
> I'm thinking the best option might be to use `showWarningMessage` when a stack is selected. I would also want the ability to pick "Don't show this again" on the message since it could be annoying.

We discussed this option earlier. And unless it's a modal, I think it'd largely go unnoticed by most users. 

Maybe we could prefix it some sort of warning icon (so it's part of the label?) Otherwise, I'd prefer to use the description to give a brief warning, and keep the button as a more-info solution.

alexweininger:
Oops, I forgot to specify that I meant to use `showWarningMessage` with `modal: true`.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -61,6 +62,16 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
                         break;
                 }
 
+                const endofLifeDate = minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
+                const sixMonthsFromNow = new Date();
+                sixMonthsFromNow.setMonth(sixMonthsFromNow.getMonth() + 6);
+                if (endofLifeDate) {
```
alexweininger, line 68: This logic could go in the helper method.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -176,3 +197,9 @@ function removeHiddenStacksAndProperties(stacks: FunctionAppStack[]): void {
         }
     }
 }
+
+export function compareDates(eolDate: string, sixMonthsDate: Date): boolean {
```
alexweininger, line 201: I think it'd be best if this function was a bit more specific to our use case. Something like `shouldShowEolWarning`. Will allow you to hide all of the EOL specific logic in there instead of before calling the function..

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -71,6 +82,16 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
         }
     }
 
+    if (hasEndOfLife) {
+        picks.push({
+            label: localize('endOfLife', `$(extensions-warning-message) Some stacks have an end of support deadline coming up. Learn more...`),
+            onPicked: async () => {
+                await openUrl('https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=azure-cli%2Cwindows%2Cin-process%2Cv4&pivots=programming-language-csharp://aka.ms/func-end-of-life');
```
alexweininger, line 89: We should turn this into an aka.ms link

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
                 context.version = await promptForFuncVersion(context);
             } else {
+                const endofLifeDate = result.minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
+                const sixMonthsFromNow = new Date();
+                sixMonthsFromNow.setMonth(sixMonthsFromNow.getMonth() + 6);
+                const settingKey: string = 'endOfLifeWarning';
+                if (getWorkspaceSetting<boolean>('endOfLifeWarning')) {
+                    if (endofLifeDate) {
+                        if (compareDates(endofLifeDate, sixMonthsFromNow)) {
+                            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. After the deadline, function apps can be created and deployed, and existing apps continue to run. However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
```
alexweininger, line 32: nit: line is really long 😭

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
             if (!result) {
                 context.version = await promptForFuncVersion(context);
             } else {
+                const endofLifeDate = result.minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
+                const sixMonthsFromNow = new Date();
+                sixMonthsFromNow.setMonth(sixMonthsFromNow.getMonth() + 6);
+                const settingKey: string = 'endOfLifeWarning';
+                if (getWorkspaceSetting<boolean>('endOfLifeWarning')) {
```
alexweininger, line 29:
```suggestion
                const settingKey: string = 'endOfLifeWarning';
                if (getWorkspaceSetting<boolean>(settingKey)) {
```

package.json
```diff
@@ -1064,7 +1069,7 @@
         "@types/ps-tree": "^1.1.0",
         "@types/request": "2.0.7",
         "@types/semver": "^5.5.0",
-        "@types/vscode": "1.57.0",
+        "@types/vscode": "^1.63.0",
```
alexweininger, line 1072: The minimum version of VS Code Functions supports (as marked by `engines.vscode` in package.json) is `1.66.0`. Should we just update the types to match that?

package.json
```diff
@@ -1032,6 +1032,11 @@
                         "scope": "resource",
                         "type": "string",
                         "description": "%azureFunctions.funcCliPath%"
+                    },
+                    "azureFunctions.endOfLifeWarning": {
+                        "type": "boolean",
+                        "description": "%azureFunctions.endOfLifeWarning%",
```
alexweininger, line 1038: This string needs to be added to the package.nls.json file

image

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
+                    if (endofLifeDate) {
+                        if (compareDates(endofLifeDate, sixMonthsFromNow)) {
+                            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. After the deadline, function apps can be created and deployed, and existing apps continue to run. However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
+                            const result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
```
alexweininger, line 33: Ideally selecting "Back" here would take the user back to the stack step, instead of back to the name step.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -176,3 +197,9 @@ function removeHiddenStacksAndProperties(stacks: FunctionAppStack[]): void {
         }
     }
 }
+
+export function compareDates(eolDate: string, sixMonthsDate: Date): boolean {
+    const endOfLife = new Date(eolDate).getTime();
+    const sixMonths = new Date(sixMonthsDate).getTime();
+    return endOfLife <= sixMonths;
```
alexweininger, line 204: You can compare `Date` objects with `<=` directly in JS without having to do `.getTime()`

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -71,6 +82,16 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
         }
     }
 
+    if (hasEndOfLife) {
+        picks.push({
+            label: localize('endOfLife', `$(extensions-warning-message) Some stacks have an end of support deadline coming up. Learn more...`),
+            onPicked: async () => {
+                await openUrl('https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=azure-cli%2Cwindows%2Cin-process%2Cv4&pivots=programming-language-csharp://aka.ms/func-end-of-life');
+            },
+            data: { stack: stacks[0], majorVersion: stacks[0].majorVersions[0], minorVersion: stacks[0].majorVersions[0].minorVersions[0] },
```
alexweininger, line 91: Hmm... seems sorta weird to include this here just to satisfy the type of picks. I think we should let the caller handle adding this pick instead of adding it here inside of `getStackPicks`. This pick isn't a stack and is just for UX purposes.

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
+                            if (result === DialogResponses.learnMore) {
+                                await openUrl('https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=azure-cli%2Cwindows%2Cin-process%2Cv4&pivots=programming-language-csharp');
```
alexweininger, line 35: Should be aka.ms link (extracted as constant and shared where it's used elsewhere). Aka.ms is our link shortening service. We do this so in the event we have to change a link later, we can do so without having to make code changes and do a release.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -61,6 +62,16 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
                         break;
                 }
 
+                const endofLifeDate = minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
```
alexweininger, line 65: 🤠 
```suggestion
                const endOfLifeDate = minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
```

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
motm32: Yeah I was also thinking that. I was using the verbiage that they said to use in the original issue, but it does make it quite long.

alexweininger: Yeah we should just split the string up into multiple lines

src/commands/createFunctionApp/stacks/getStackPicks.ts
alexweininger: Follow up: I think you can change the return type of `getStackPicks` to `Promise<IAzureQuickPickItem<FullFunctionAppStack | undefined>[]>` so you don't have to leave data as undefined here.

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
                 context.version = await promptForFuncVersion(context);
             } else {
+                const endOfLifeDate = result.minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
+                const settingKey: string = 'endOfLifeWarning';
+                if (getWorkspaceSetting<boolean>(settingKey)) {
+                    if (endOfLifeDate) {
+                        if (shouldShowEolWarning(endOfLifeDate)) {
+                            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
+                                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
+                                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
```
alexweininger, line 32: nit: I like using backticks for multiline strings

```suggestion
                            const message = localize('endOfLife', `The chosen runtime stack has an end of support deadline coming up. 
                                After the deadline, function apps can be created and deployed, and existing apps continue to run. 
                                However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them`);
```

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -61,6 +62,14 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
                         break;
                 }
 
+                const endOfLifeDate = minorVersion.stackSettings.linuxRuntimeSettings?.endOfLifeDate;
+                if (endOfLifeDate) {
```
alexweininger, line 66: I would make `shouldShowEolWarning` take `minorVersion` as an arg, so you can move these lines inside the function.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -72,6 +81,15 @@ export async function getStackPicks(context: IFunctionAppWizardContext): Promise
...
     return picks.sort((p1, p2) => {
+        if (hasEndOfLife) {
+            (picks as IAzureQuickPickItem<FullFunctionAppStack | undefined>[]).push({
+                label: localize('endOfLife', `$(extensions-warning-message) Some stacks have an end of support deadline coming up. Learn more...`),
+                onPicked: async () => {
+                    await openUrl('https://aka.ms/AA1tpij');
+                },
+                data: undefined
+            });
+        }
```
alexweininger, line 92: Doing this here will result in many of these picks being pushed into `picks`. Remove the return here, and push the pick in after sorting, then `return picks`.

src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -176,3 +194,10 @@ function removeHiddenStacksAndProperties(stacks: FunctionAppStack[]): void {
         }
     }
 }
+
+export function shouldShowEolWarning(eolDate: string): boolean {
+    const sixMonthsFromNow = new Date();
+    sixMonthsFromNow.setMonth(sixMonthsFromNow.getMonth() + 6);
```
alexweininger, line 200: If it's possible to do on one line, I say do it. Especially since the variable name makes it clear what is happening here.

```suggestion
    const sixMonthsFromNow = new Date(new Date().setMonth(new Date().getMonth() + 6));
```

package.nls.json
```diff
@@ -33,6 +33,7 @@
     "azureFunctions.enableJavaRemoteDebugging": "Enable remote debugging for Java Functions Apps running on Windows. (experimental)",
     "azureFunctions.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
     "azureFunctions.enableRemoteDebugging": "Enable remote debugging for Node.js Function Apps running on Linux App Service plans. Consumption plans are not supported. (experimental)",
+    "azureFunctions.endOfLifeWarning": "Enable End Of Life Warning",
```
alexweininger, line 36: Maybe something like "Show a warning when creating Function Apps with stacks within 6 months of their end of life date."?

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -20,6 +22,24 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
+                            const result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
+                            if (result === DialogResponses.learnMore) {
+                                await openUrl('https://aka.ms/AA1tpij');
```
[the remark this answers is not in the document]
motm32: Yeah I saw it was also used in `FuncVersion` so I'll update that one too.

package.nls.json
motm32: Do we know if stacks that are passed their end of life date still appear? If so, may want to modify the message to "...with stacks passed or within...".

alexweininger: Not sure, what about that one with an EOL of yesterday?

motm32: I actually read it wrong since the months are zero indexed so the actual EOL for Powershell 7.0 is 12/3. Whoops lol.

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
motm32: I added this suggestion and it seems to be doing something weird with the warning message.
image

alexweininger: 😱

My bad, of course the newlines will show up when using backticks. Change it back to how it was, or use the `\` character. 

example from docs:
```js
const longString = "This is a very long string which needs \
to wrap across multiple lines because \
otherwise my code is unreadable."
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#long_literal_strings

motm32: OK! With the `\` you can't include indents which I don't think looks too good so I just changed is back to how it was!

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -59,12 +40,17 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
...
+        const promptSteps: AzureWizardPromptStep<IFunctionAppWizardContext>[] = [];
+        if (shouldShowEolWarning(context.newSiteStack?.minorVersion)) {
+            promptSteps.push(new FunctionAppEOLWarningStep());
+        }
         if (context.newSiteOS === undefined) {
-            return { promptSteps: [new SiteOSStep()] };
+            promptSteps.push(new SiteOSStep())
         } else {
             await setLocationsTask(context);
             return undefined;
```
alexweininger, line 51: The warning step isn't showing up for me since we're returning undefined here.

motm32: Hmm that's weird it seems to show up for me.

alexweininger: If `context.newSiteOS` is defined it will enter the else block and always return `undefined`. We need to make sure we always return the EOL step if necessary.

alexweininger (commented):
Update the dev to the latest package to fix the compile errors that came from upgrading @types/vscode.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,31 @@
...
+import { funcVersionLink } from "../../../FuncVersion";
+import { localize } from "../../../localize";
+import { getWorkspaceSetting, updateGlobalSetting } from "../../../vsCodeConfig/settings";
+import { IFunctionAppWizardContext } from "../IFunctionAppWizardContext";
+
+export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAppWizardContext> {
+    public async prompt(context: IFunctionAppWizardContext): Promise<void> {
+        const settingKey: string = 'endOfLifeWarning';
+        if (getWorkspaceSetting<boolean>(settingKey)) {
+            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up." +
```
alexweininger, line 16: Needs spaces after the periods

image

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,31 @@
...
+                "After the deadline, function apps can be created and deployed, and existing apps continue to run." +
+                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them");
+            const result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
+            if (result === DialogResponses.learnMore) {
+                await openUrl(funcVersionLink);
```
alexweininger, line 21: Clicking learn more opens the link as expected, but it also advances to the next step. I think the preferred behavior would be to have the modal stay up after clicking learn more.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,31 @@
...
+            }
+            else if (result === DialogResponses.dontWarnAgain) {
```
alexweininger, line 23: Also, the only option the user has if they want to continue is to select "don't warn again". There should be a "Continue" button that allows for continuing, without suppressing future warnings.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
alexweininger: And add a period for the last sentence

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,34 @@
...
+export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAppWizardContext> {
+    public async prompt(context: IFunctionAppWizardContext): Promise<void> {
+        const settingKey: string = 'endOfLifeWarning';
+        if (getWorkspaceSetting<boolean>(settingKey)) {
+            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
+                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
+                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
+            const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
+            do {
```
alexweininger, line 21: Using a do-while here makes it so that selecting items other than Learn More result in the learn more link opening. I think you'll want to use a normal while loop here.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAppWizardContext> {
+    public async prompt(context: IFunctionAppWizardContext): Promise<void> {
+        const settingKey: string = 'endOfLifeWarning';
+        if (getWorkspaceSetting<boolean>(settingKey)) {
+            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
+                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
+                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
+            const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
```
alexweininger, line 20: nit: extra comma
```suggestion
            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain);
```

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -0,0 +1,34 @@
...
+            while (result === DialogResponses.learnMore) {
+                await openUrl(funcVersionLink);
+                result = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
```
alexweininger, line 23: To avoid code duplication I think it'd be best to move showing the warning message to a private helper method. That way you can use it here and when you declare `result` above.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -13,22 +13,27 @@ export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAp
     public async prompt(context: IFunctionAppWizardContext): Promise<void> {
         const settingKey: string = 'endOfLifeWarning';
         if (getWorkspaceSetting<boolean>(settingKey)) {
-            const message = localize('endOfLife', "The chosen runtime stack has an end of support deadline coming up. " +
-                "After the deadline, function apps can be created and deployed, and existing apps continue to run. " +
-                "However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them.");
-            const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
-            let result: MessageItem = await context.ui.showWarningMessage(message, { modal: true }, DialogResponses.learnMore, continueOn, DialogResponses.dontWarnAgain,);
-            while (result === DialogResponses.learnMore) {
+            let result: Promise<MessageItem> = this.getWarningMessage(context);
```
alexweininger, line 16: I would await the method here instead of inside the if expressions.

src/commands/createFunctionApp/stacks/FunctionAppEOLWarningStep.ts
```diff
@@ -13,22 +13,27 @@ export class FunctionAppEOLWarningStep extends AzureWizardPromptStep<IFunctionAp
...
                 await updateGlobalSetting(settingKey, false);
             }
         }
     }
+
     public shouldPrompt(context: IFunctionAppWizardContext): boolean {
         return !!context.newSiteStack;
     }
+
+    private async getWarningMessage(context: IFunctionAppWizardContext): Promise<MessageItem> {
```
alexweininger, line 32: I think `showWarningMessage` is a better name.

alexweininger, line 32: nit: technically this parameter only needs to be `IActionContext`.

alexweininger (approved):
💯

PR #3385, opened 2022-10-24
Don't write properties to function.json that are undefined

### LINKED ISSUE #3205: Adding CosmosDB Output binding through the wizard creates invalid json
Version of VS Code - 1.67.2
Your operating system - Windows 11
Version of the relevant VS Code extension - Azure Functions v1.7.2
Reproducible steps (1... 2... 3...) that cause the issue
- create a function project
- add a blob storage trigger
- add a cosmosdb output binding
    - follow the prompts but leave `partitionkey` and `collectionThrouthput` as 'undefined'
- try and run the function locally with F5

    What you expected to see, versus what you actually saw
    You get this error in the terminal:
    image

- If instead of running locally you deploy it to Azure, the function just doesn't trigger and I couldn't find errors anywhere to explain why

function.json file after adding the CosmosDB binding through the wizard:
```JSON
{
  "bindings": [
    {
      "name": "InputBlob",
      "type": "blobTrigger",
      "direction": "in",
      "path": "source2/{name}",
      "connection": "storagefundemo7_STORAGE"
    },
    {
      "type": "cosmosDB",
      "direction": "out",
      "name": "outputDocument",
      "databaseName": "grillenDb",
      "collectionName": "data",
      "createIfNotExists": false,
      "connectionStringSetting": "cosmosdbfunctiondemo_DOCUMENTDB",
      "partitionKey": "undefined",
      "collectionThroughput": "undefined"
    }
  ]
}
```

To fix the issue I had to remove these two keys (and the preceding comma) and redeploy the function to Azure.
```JSON
      "partitionKey": "undefined",
      "collectionThroughput": "undefined"
```

motm32:
Fixes #3205.

### DIFF
--- a/src/commands/addBinding/BindingCreateStep.ts
+++ b/src/commands/addBinding/BindingCreateStep.ts
@@ -27,7 +27,9 @@ export class BindingCreateStep extends AzureWizardExecuteStep<IBindingWizardCont
         binding.direction = bindingTemplate.direction;
 
         for (const b of bindingTemplate.settings) {
-            binding[b.name] = getBindingSetting(context, b);
+            if (getBindingSetting(context, b) !== "undefined") {
+                binding[b.name] = getBindingSetting(context, b);
+            }
         }
 
         await confirmEditJsonFile(context, context.functionJsonPath, (functionJson: IFunctionJson) => {
--- a/src/commands/addBinding/BindingCreateStep.ts
+++ b/src/commands/addBinding/BindingCreateStep.ts
@@ -27,8 +27,9 @@ export class BindingCreateStep extends AzureWizardExecuteStep<IBindingWizardCont
         binding.direction = bindingTemplate.direction;
 
         for (const b of bindingTemplate.settings) {
-            if (getBindingSetting(context, b) !== "undefined") {
-                binding[b.name] = getBindingSetting(context, b);
+            const bindingSetting = getBindingSetting(context, b);
+            if (bindingSetting !== "undefined" && bindingSetting !== "null") {
+                binding[b.name] = bindingSetting;
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/commands/addBinding/BindingCreateStep.ts
```diff
@@ -27,7 +27,9 @@ export class BindingCreateStep extends AzureWizardExecuteStep<IBindingWizardCont
         binding.direction = bindingTemplate.direction;
 
         for (const b of bindingTemplate.settings) {
-            binding[b.name] = getBindingSetting(context, b);
+            if (getBindingSetting(context, b) !== "undefined") {
```
bwateratmsft, line 30: I'd store the result from `getBindingSetting(...)` to avoid needing to call it twice. e.g.:

```typescript
const bindingSetting = getBindingSetting(context, b);
if (bindingSetting !== "undefined") {
    binding[b.name] = bindingSetting;
}
```

bwateratmsft, line 30: Is `"undefined"` (a string) the actual value or are you looking for `undefined` (the symbol)? Is `null` relevant as well?

motm32: I am pretty sure it should be a string in the case of `partitionKey` and `collectionThroughput` settings since we are getting the value from user input and from the binding settings it also seems to be a string. Another issue I noticed is that the current behavior of the input box has undefined as the placeholder for `partitionKey` and `collectionThroughput` which may also want to be changed since its an invalid input. I tried inputting null and it also caused issues so I will add that, but not sure if undefined the symbol is used anywhere.

nturinski (approved):
@bwateratmsft I have a feeling that you have a more clever way to check this... 🤔

bwateratmsft:
> @bwateratmsft I have a feeling that you have a more clever way to check this... 🤔

Nothing comes to mind :smile:

Issue #2883, opened 2021-07-14
Error: Cannot find module durable-functions


MaximRouiller:
Using the latest version of vscode and the extensions.

* Create a brand new Azure Functions JavaScript project.
* Create a Durable HTTP Starter function
* Create a Durable Orchestrator Function
* Create a Durable Activity Function called `Hello` (to match the orchestrator Function)
* Press F5 to debug.

**Expected result:** Azure Function running and waiting for an HTTP request
**Actual Result:**
![Screenshot](https://user-images.githubusercontent.com/209384/125675368-98d9dc57-7103-466d-89fb-7e4f151e641f.png)

**Root cause:**
While creating the different durable functions, no dependencies for `durable-functions` were ever added in `package.json`.

**Possible fix:**
When creating Durable Functions, ensure that the dependency has been added before/after the creation of the Function.

MicroFish91:
Fixed by the PR above

Issue #3272, opened 2022-08-02
Expose new Durable Functions backend options 


lilyjma:
Netherite (which uses Event Hubs and Azure Storage) and SQL should be available as storage providers when using Durable Functions. The default option currently is Azure Storage.

lilyjma:
(this is a feature request, but I couldn't add a label because of lack of permission)

nturinski:
Hi @lilyjma

Could you please share more information about this feature request?  We would need to know what the expected UX/UI design would be. Would we need to list user's Netherite and SQL connection strings, for example?

How are these normally hooked into Durable Functions?  Is there any documentation that you could provide so that we could understand this scenario better?

alexweininger:
Closing to merge with #2956

Issue #2956, opened 2021-09-08
Support new Durable Functions backends


anthonychu:
Durable Functions is adding a couple of new backends, in addition to the existing backend.

https://github.com/microsoft/durabletask-netherite
https://microsoft.github.io/durabletask-mssql

There are a couple of scenarios we could support with the VS Code extension:

* Selecting the backend when creating Durable Functions - This is likely the main one we want to tackle
* Switching between backends - Likely a less important use case, we can solve this with documentation

Our plan is for the DurableTask extension to include support for all backends, and the backends will automatically be available when extension bundles are used. 

* Storage backend is used by default (same as today). A customer can select another backend by modifying host.json.
* The Netherite backend uses Event Hubs and Storage behind the scenes. Customer needs to configure connection strings in settings.
* The SQL backend only depends on MSSQL. Customer needs to configure the SQL connection string in settings.

One potential way to support backend selection is to prompt the customer for the backend type on the first Durable Function that they create in a project. The default will continue to be Storage. If they select Netherite or SQL, we would need to prompt them for the connection string(s). Ideally we'd allow them to pull those from existing Azure resources, or perhaps even create new resources in the process.

davidmrdavid:
Just bumping this thread, as we're nearing the release. It seems Anthony provided most of the context here, with the exception that the backend we'll be releasing first is Netherite, so we should be prioritizing support for it first.

I agree with Anthony's suggestion on the expected flow:

> One potential way to support backend selection is to prompt the customer for the backend type on the first Durable Function that they create in a project. The default will continue to be Storage. If they select Netherite or SQL, we would need to prompt them for the connection string(s). Ideally we'd allow them to pull those from existing Azure resources, or perhaps even create new resources in the process.

MicroFish91:
@lilyjma

PR #3390, opened 2022-10-26
Connecting Durable Functions with Backends - Part I

MicroFish91:
...WIP do not review yet

We were tasked with adding Durable Backends via issues [3272](https://github.com/microsoft/vscode-azurefunctions/issues/3272) and [2956](https://github.com/microsoft/vscode-azurefunctions/issues/2956).  Anthony Chu in the second issue explains the high-level view of what we are trying to accomplish with this implementation.  

There are 3 main storage schemes we can choose from to store state:

1) Default: Requires storage only
2) [Netherite](https://microsoft.github.io/durabletask-netherite/#/README): Requires event hub and storage
3) [SQL](https://microsoft.github.io/durabletask-mssql/#/README?id=introduction): Requires SQL database and storage

The actual implementation required quite a bit of code due to how many entry points of verification we have in Functions. Since we give the user the option to defer certain steps at the beginning, depending on the later step we choose, we may need to verify and step the user through certain creation procedures based on the state of their current project, and this is variable depending on the step we're on (i.e. "Create Function...", "Debug...", "Create Function App in Azure...", "Deploy...").  Because of this, we have a lot of smart detection logic required to get this bad boy fully up and running...

### GitHub Merge Flow
I'm splitting up the work I've done in the [OG branch](https://github.com/microsoft/vscode-azurefunctions/tree/mwf/durable-backends) into smaller PRs with the hopes that it will be easier for you all to review.  This is the flow I'm envisioning to get all the work eventually merged into main:

<image src="https://user-images.githubusercontent.com/40250218/198110768-fba85a3b-6f5d-431f-8ee7-3b09b4b6d3ba.png" height=300>

### Part 1
This PR primarily covers the "Create Function..." command flow scenario pertaining to the following:
* Storage Schemes: **Default Storage** and **Netherite**
* Languages: JS/TS, C#, Python

There may be some other file changes in here that don't necessarily belong to those categories - they were likely only included because separating them out per file would be very tedious and difficult.  When in doubt, ask me directly or you can reference the original branch in the link above.

### Part 2
Coming Soon...

### Part 3
Coming Soon...

### Part 4
Coming Soon...

## Final Notes
I'll bring it up again in a future meeting or stand-up, but let me know if you guys would like a full demo of this feature in its current state.  As always, appreciate you guys taking the time to review my code.  Let me know if you have any questions or concerns, thanks again 😊


src/FuncVersion.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { IActionContext, IAzureQuickPickItem, IAzureQuickPickOptions } from '@microsoft/vscode-azext-utils';
-import { localize } from './localize';
+import { localize, recommended } from './localize';
```
bwateratmsft, line 7: Putting the localized strings in the `localize.ts` file isn't really the intended usage, but also, this contradicts with the way the new `l10n` API is going to be used. It will be easiest to keep all of the strings in their original locations in the source code, rather than clustered in a file full of constants.

MicroFish91: Hmm, so I only added the localize constants that were being used in multiple files here.  Previously localize constants that fit this bill were being shown in `constants.ts` but it was starting to look really cluttered to me so I figured placing the localize constants in a separate file might make more sense/be cleaner.

Out of curiosity, will the `l10n` API not be caching the localize string by ID (i.e. the first parameter of the localize call)?  Would it make sense for me to remove mine and all previous localize constants as well (previously part of the migration from `constants.ts`?  Thanks :)

bwateratmsft: Yeah, I think they should all be removed from `constants.ts` as well. The `l10n` API uses the string itself as the key, plus any comment you add for applying uniqueness as needed or clarifying things for the localizers. Not needing an ID I think is a good improvement; there will doubtless be some iteration though on fixing bugs that arise from loss of uniqueness in the keys.

MicroFish91: Okay, that makes a lot of sense, thank you for the explanation 🙂

nturinski: > Yeah, I think they should all be removed from `constants.ts` as well. The `l10n` API uses the string itself as the key, plus any comment you add for applying uniqueness as needed or clarifying things for the localizers. Not needing an ID I think is a good improvement; there will doubtless be some iteration though on fixing bugs that arise from loss of uniqueness in the keys.

I'm not following why we would move all of the localized strings out of `constants.ts`. Even if localize is using the string as the key, does its location in source get stored as metadata as well? Because if not, it might be easier to change all of our localized strings in one file (especially if some of those strings are reused, such as a "Yes" label).

Definitely not in `localize.ts`, but maybe in some file called `constants-nls.ts` or something?

bwateratmsft: The old `vscode-nls` way required a parallel file tree to your whole code base with a file containing translations for every file in source that had strings. Multiply that by language and you have a colossal number of files.

The new `l10n` way puts all those strings into one file (per language), using the original string from the source as a key (plus comment if applicable), with the value being whatever the localized string is.

With the new `l10n` API, putting the strings into `constants.ts` (or, `localize.ts`) is not going to make the actual process of localization any easier--but it is going to mean anytime I want to see what a string value is, I have to <kbd>F12</kbd> over to the constants file instead of being able to just read it inline.

bwateratmsft: That being said, for strings we expect to change frequently, and/or that we want to all stay in sync, I think it does make sense to use a file with constants (though I wouldn't use `constants.ts`, and not `localize.ts` either since that's `vscode-nls`-specific.)

src/constants-nls.ts
```diff
@@ -0,0 +1,25 @@
...
+export const previewDescription: string = localize('preview', '(Preview)');
+export const pythonNewModelPreview: string = localize('pythonNewModelPreview', 'Python (Programming Model V2)');
+export const hostJsonConfigFailed: string = localize('durableStorageConfigFailed', 'WARNING: Failed to configure your JSON files for durable storage, you may need to configure them manually or start from a clean project.');
+export const emptyWorkspace: string = localize('emptyWorkspace', 'Your workspace folder looks empty, please navigate to the root directory of your project.');
+export const useEmulator: string = localize('useEmulator', 'Use Local Emulator');
+export const skipForNow: string = localize('skipForNow', 'Skip For Now');
+export const overwriteRemoteConnection = (a: string) => localize('overwriteRemoteConnection', 'We detected a different local connection setting for "{0}" than what was previously used. Would you like to overwrite your remote setting?', a);
+
+export const invalidAlphanumericWithHyphens: string = localize('invalidAlphanumericOrHyphen', `A name must consist of alphanumeric characters or '-', and must start and end with an alphanumeric character.`);
```
MicroFish91, line 12: Just an FYI, but both of these "invalid" messages only are used in one place each currently; however, I put the constants here because I wanted them to be easily discoverable by devs since they are meant for reuse and match up with the common `validateUtils` methods I added.

src/utils/azureClients.ts
```diff
@@ -8,6 +8,7 @@ import { WebSiteManagementClient } from '@azure/arm-appservice';
 import { CosmosDBManagementClient } from '@azure/arm-cosmosdb';
 import { EventHubManagementClient } from '@azure/arm-eventhub';
 import { ServiceBusManagementClient } from '@azure/arm-servicebus';
+import { SqlManagementClient } from '@azure/arm-sql';
```
bwateratmsft, line 11: All of these `@azure` client imports should be `import type`, otherwise there's no gain by doing `await import` in the code below.

src/utils/azure.ts
```diff
@@ -3,16 +3,18 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
+import { AccessKeys, EventHubManagementClient } from '@azure/arm-eventhub';
 import { StorageAccount, StorageAccountListKeysResult, StorageManagementClient } from '@azure/arm-storage';
```
bwateratmsft, line 7:
```suggestion
import type { StorageAccount, StorageAccountListKeysResult, StorageManagementClient } from '@azure/arm-storage';
```

src/utils/durableUtils.ts
```diff
@@ -0,0 +1,288 @@
...
+
+    export async function getEventHubName(projectPath?: string): Promise<string | undefined> {
+        projectPath ??= getWorkspaceRootPath();
+        if (!projectPath) {
+            throw new Error(emptyWorkspace);
+        }
+
+        const hostJsonPath = path.join(projectPath, hostFileName);
+        if (!AzExtFsExtra.pathExists(hostJsonPath)) {
+            return;
```
bwateratmsft, line 189:
```suggestion
            return undefined;
```

src/utils/venvUtils.ts
```diff
@@ -18,6 +21,30 @@ export namespace venvUtils {
...
+            return;
+        }
+
+        const requirementsPath: string = path.join(projectPath, requirementsFileName);
+        if (await AzExtFsExtra.pathExists(requirementsPath)) {
+            try {
+                // Attempt to install packages so that users get Intellisense right away
+                await runCommandInVenv(`pip install -r ${requirementsFileName}`, <string>venvName, projectPath);
+            } catch {
+                ext.outputChannel.appendLog(localize('pipInstallFailure', 'WARNING: Failed to install packages in your virtual environment. Run "pip install" manually instead.'));
```
bwateratmsft, line 43: Should the error be shown? Does it need to be rethrown?

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -6,52 +6,63 @@
...
         }
 
-        context.telemetry.properties.azureWebJobsStorageType = context.azureWebJobsStorageType || 'skipForNow';
+        context.telemetry.properties.azureWebJobsStorageType = context.azureWebJobsStorageType;
     }
 
-    public shouldPrompt(context: T): boolean {
-        // Only should prompt if no storage account was selected
-        context.azureWebJobsStorageType = (!!context.storageAccount || !!context.newStorageAccountName) ? 'azure' : undefined;
+    public shouldPrompt(context: T & { eventHubConnectionType?: ConnectionTypeValues, sqlDbConnectionType?: ConnectionTypeValues }): boolean {
```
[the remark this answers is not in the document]
nturinski: The issue with doing the setting logic in `prompt()` is that even if we don't actually display anything, it messes with the step count.

I agree it's gross to have to set things in `shouldPrompt()`. In that case, would it make more sense to set `context.azureWebJobsStorageType`wherever `this._options?.preSelectedConnectionType` was set?

I think having a `prompt()` step that doesn't actually prompt is less intuitive than setting things in `shouldPrompt()` though if we want to consolidate the logic all to this file.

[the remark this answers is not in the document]
bwateratmsft: > The issue with doing the setting logic in `prompt()` is that even if we don't actually display anything, it messes with the step count.

Do users actually care if they skip from step 1 to 3? Or if we hide the step counter?

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -81,6 +85,37 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
...
+                case DurableBackend.SQL:
+                    hostJson.extensions.durableTask = sqlUtils.getDefaultSqlTaskConfig();
+                    setLocalAppSetting(context, context.projectPath, 'SQLDB_Connection', '', MismatchBehavior.Overwrite);
+                    break;
+                default:
+            }
+
+            await AzExtFsExtra.writeJSON(hostJsonPath, hostJson);
+        } catch {
+            ext.outputChannel.appendLog(hostJsonConfigFailed);
```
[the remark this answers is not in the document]
MicroFish91: Hmm, looking at this more closely, I'm thinking I can just remove the try/catch entirely and just do a check against 'pathExists`.  I think I was originally expecting more potential points for side effects, but upon closer inspection, it doesn't seem like any major ones except if `host.json` path doesn't exist.

src/commands/createFunction/FunctionSubWizard.ts
```diff
@@ -86,7 +90,25 @@ export class FunctionSubWizard {
...
+                    // Todo: Uncomment out in future PR
+                    // promptSteps.push(new SqlDatabaseConnectionPromptStep(), new SqlDatabaseListStep());
+                    // executeSteps.push(new SqlDatabaseConnectionExecuteStep());
+                    break;
+                case DurableBackend.Storage:
+                default:
+            }
+
+            if (
+                !await getLocalConnectionString(context, ConnectionKey.Storage, context.projectPath) &&   // If we already have a storage setup, don't re-prompt
```
MicroFish91, line 108: Would appreciate a second pair of eyes to make sure I preserved the logic properly here!  I did my best to make it more readable and to add comments where appropriate

src/commands/createFunction/FunctionSubWizard.ts
```diff
@@ -86,7 +90,25 @@ export class FunctionSubWizard {
...
+                (context.newDurableStorageType ||   // Always prompt for new durable storage setups unless a storage account has already been linked
+                    (!template.isHttpTrigger && !template.isSqlBindingTemplate && !canValidateAzureWebJobStorageOnDebug(context.language)))
+            ) {
```
MicroFish91, line 111: Would appreciate a second pair of 👀 here to make sure I preserved the logic properly!  I did my best to make this statement more readable and and tried to organize it in a way that makes sense.

src/commands/createFunction/FunctionSubWizard.ts
```diff
@@ -86,7 +90,21 @@ export class FunctionSubWizard {
...
+                case DurableBackend.SQL:
+                    // Todo: Uncomment out in future PR
+                    // promptSteps.push(new SqlDatabaseConnectionPromptStep(), new SqlDatabaseListStep());
+                    // executeSteps.push(new SqlDatabaseConnectionExecuteStep());
+                    break;
+                case DurableBackend.Storage:
+                default:
+            }
+
+            if (context.newDurableStorageType || !template.isHttpTrigger && !template.isSqlBindingTemplate && !canValidateAzureWebJobStorageOnDebug(context.language) && !await getLocalConnectionString(context, ConnectionKey.Storage, context.projectPath)) {
```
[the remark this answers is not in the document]
MicroFish91: Basically the latter, except when a local connection string already exists.  I'll re-group and add comments in a way that (hopefully) makes more sense.

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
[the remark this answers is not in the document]
nturinski: Hard to say. I think people would think it's a bug or that they somehow skipped a step and something went wrong though. 

We could hide the step counter, but I think it's nice to know when you're on the final step since there's currently no confirmation page.

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceNameStep.ts
```diff
@@ -0,0 +1,47 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { EventHubManagementClient } from '@azure/arm-eventhub';
+import { delay } from '@azure/ms-rest-js';
```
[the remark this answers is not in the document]
MicroFish91: Yeah, I'll look into replacing this with a more appropriate debounce util based on your suggestion [here](https://github.com/microsoft/vscode-azurefunctions/pull/3392#discussion_r1009535966)

src/commands/createFunction/dotnetSteps/DotnetFunctionCreateStep.ts
```diff
@@ -53,6 +56,32 @@ export class DotnetFunctionCreateStep extends FunctionCreateStepBase<IDotnetFunc
...
+                    await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', durableUtils.dotnetDfSqlPackage);
+                    break;
+                case DurableBackend.Storage:
+                default:
+            }
+
+            // Seems that the package arrives out-dated and needs to be updated
+            await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', durableUtils.dotnetDfBasePackage);
+        } catch {
+            ext.outputChannel.appendLog(localize('funcInstallFailed', 'WARNING: Failed to install and update Durable Functions NuGet packages to the .csproj project file. You may need to configure them manually or start from a clean project.'))
```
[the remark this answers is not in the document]
MicroFish91: Hmm good question, I'm open to either, but my initial thought is that we probably don't want to throw an error and halt the whole process.  I was considering the package install to be more of a 'nice-to-have' and not a 'need-to-have' enhancement since often times there are instructions in the template files guiding the user on how to install durable function dependencies.  

Your comment did make me re-consider how useful the current log message is to the user though, so I ended up re-coding it to list out all the packages that still need to be installed rather than just the all-in-one generic message that I'm using here, let me know if you feel like additional changes should be made or if you prefer the error message instead!

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -6,52 +6,64 @@
...
+        if (result === connectStorageButton) {
+            context.azureWebJobsStorageType = ConnectionType.Azure;
+        } else if (result === useEmulatorButton) {
+            context.azureWebJobsStorageType = ConnectionType.NonAzure;
+        } else {
+            context.azureWebJobsStorageType = ConnectionType.None;
         }
 
-        context.telemetry.properties.azureWebJobsStorageType = context.azureWebJobsStorageType || 'skipForNow';
+        context.telemetry.properties.azureWebJobsStorageType = context.azureWebJobsStorageType;
```
nturinski, line 43: Will the telemetry property for `skipForNow` be `None` after this change? 

This is a potential data point for us to see if people are dropping in the funnel here, so it'd be good if we could keep this consistent with how it used to be.

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
```diff
@@ -0,0 +1,39 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { EHNamespace, EventHubManagementClient } from '@azure/arm-eventhub';
+import { ext } from '@microsoft/vscode-azext-azureappservice/out/src/extensionVariables';
```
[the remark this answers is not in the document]
MicroFish91: Ohhh, good eye and thank you for catching that one, how annoying 😆 

I'll be on the lookout for this in the future from now on

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -6,52 +6,64 @@
...
     }
 
-    public shouldPrompt(context: T): boolean {
-        // Only should prompt if no storage account was selected
-        context.azureWebJobsStorageType = (!!context.storageAccount || !!context.newStorageAccountName) ? 'azure' : undefined;
+    public shouldPrompt(context: T & { eventHubConnectionType?: ConnectionTypeValues, sqlDbConnectionType?: ConnectionTypeValues }): boolean {
+        if (this._options?.preSelectedConnectionType) {
+            context.azureWebJobsStorageType = this._options.preSelectedConnectionType;
+        } else if (!!context.storageAccount || !!context.newStorageAccountName) {
+            context.azureWebJobsStorageType = ConnectionType.Azure;  // Only should prompt if no storage account was selected
```
nturinski, line 50: This comment should be moved to the top and updated, explaining what scenarios stifle the prompt.

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
```diff
@@ -6,52 +6,64 @@
...
+        } else if (context.eventHubConnectionType) {
+            context.azureWebJobsStorageType = context.eventHubConnectionType;
+        } else if (context.sqlDbConnectionType === ConnectionType.Azure || context.sqlDbConnectionType === ConnectionType.None) {
```
nturinski, line 53: In what cases would `context.sqlDbConnectionType` be set to `None`? It sounds like this would be more of an emulator/local server scenario.

src/commands/createFunction/durableSteps/netherite/NetheriteConfigureHostStep.ts
```diff
@@ -0,0 +1,40 @@
...
+import { AzExtFsExtra, AzureWizardExecuteStep, nonNullValue } from '@microsoft/vscode-azext-utils';
+import * as path from 'path';
+import { hostFileName } from '../../../../constants';
+import { hostJsonConfigFailed } from '../../../../constants-nls';
+import { IHostJsonV2, INetheriteTaskJson } from '../../../../funcConfig/host';
+import { netheriteUtils } from '../../../../utils/durableUtils';
+import { IEventHubsConnectionWizardContext } from '../../../appSettings/IEventHubsConnectionWizardContext';
+
+export class NetheriteConfigureHostStep<T extends IEventHubsConnectionWizardContext> extends AzureWizardExecuteStep<T> {
+    public priority: number = 250;
```
nturinski, line 14: This has the same priority as `OpenFolderStep`. We should avoid using the same priority number for execute steps. I'm not sure if they'd ever be used in the same wizard, but I think it's possible if you create a new project and your first function trigger is durable.

Even if it's not possible, it's probably still better practice to avoid conflicts. You can change this one to 245 or something, or increase `OpenFolderStep` to 300.

src/commands/createFunction/durableSteps/netherite/NetheriteConfigureHostStep.ts
```diff
@@ -0,0 +1,40 @@
...
+            const existingHubName: string | undefined = durableTask?.hubName;
+
+            hostJson.extensions ??= {};
+            hostJson.extensions.durableTask = netheriteUtils.getDefaultNetheriteTaskConfig(
+                nonNullValue(context.newEventHubName || existingHubName)
+            );
+
+            await AzExtFsExtra.writeJSON(hostJsonPath, hostJson);
+        } catch {
+            ext.outputChannel.appendLog(hostJsonConfigFailed);
```
nturinski: I think it would be helpful to output the error message as well. 

For example, if the error is just a JSON parse failed error, it'd be a lot easier for the user to diagnose and resolve.

It could just be in the output channel as well. Though, I think that the warning message itself should appear as a notification with a button that leads to the output channel for more information.

Their project straight up won't work if this fails, correct? So we should make sure the user notices this warning.

src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
```diff
@@ -56,8 +62,34 @@ export class ScriptFunctionCreateStep extends FunctionCreateStepBase<IScriptFunc
...
+                    await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'npm', 'install', durableUtils.nodeDfPackage);
+                    break;
+                case ProjectLanguage.Python:
+                    await pythonUtils.addDependencyToRequirements(durableUtils.pythonDfPackage);
+                    await venvUtils.runPipInstallCommandIfPossible();
+                    break;
+                default:
+            }
+        } catch {
+            ext.outputChannel.appendLog(dfDepInstallFailed);
```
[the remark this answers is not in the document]
MicroFish91: Hmmm, do we want to show a more detailed error message in the logs?  I don't think I'd want to re-throw the error in this cause just because there's already instructions in the template file on how to install the dependency, so I feel there's no need to block the user if at all possible since this is more of a 'nice-to-have' than a 'need-to-have' enhancement.

src/commands/createFunction/durableSteps/netherite/NetheriteEventHubNameStep.ts
```diff
@@ -0,0 +1,59 @@
...
+        if (!validateUtils.isValidLength(name, 1, 256)) {
+            return invalidLength('1', '256');
+        }
+        if (!validateUtils.isLowerCaseAlphanumericWithHypens(name)) {
+            return invalidLowerCaseAlphanumericWithHyphens;
+        }
+
+        const isNameAvailable: boolean = !this._eventHubs.some(eh => eh.name === name);
+        if (!isNameAvailable) {
+            return localize('eventHubExists', 'The event hub you entered already exists. Please enter a unique name.');
```
nturinski, line 54: To be more consistent with validation error message, I'd recommend something like this.

```suggestion
            return localize('eventHubExists', 'An event hub with the name "{0}" already exists.', name);
```

src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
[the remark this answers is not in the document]
nturinski: I said it in another comment, but I think it's still worth outputting the original error with the localized string as well.

src/utils/venvUtils.ts
MicroFish91: I assume whatever answer applies to [this](https://github.com/microsoft/vscode-azurefunctions/pull/3390/files#r1009927424) applies here as well.

src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
```diff
@@ -5,26 +5,34 @@
...
-    public async execute(context: IAzureWebJobsStorageWizardContext): Promise<void> {
+    public constructor(private readonly _setConnectionForDeploy?: boolean) {
+        super();
+    }
+
+    public async execute(context: T): Promise<void> {
         let value: string;
-        if (context.azureWebJobsStorageType === 'emulator') {
-            value = localEmulatorConnectionString;
+        if (context.azureWebJobsStorageType === ConnectionType.NonAzure) {
```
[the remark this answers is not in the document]
MicroFish91: My reasoning was that I felt the code flows essentially boiled down to 1 of 3 paths... we are either 1) stepping the user through to create the appropriate Azure resources to connect, 2) giving the user the option to connect to some sort of resource outside of Azure instead (whether its hosted outside or locally), or 3) deferring the connection setup to a later time.

For `storage` and `netherite`, `NonAzure` does refer to emulators, whereas with `SQL`, `NonAzure` refers to any database resource outside of Azure.

Splitting `NonAzure` into `SQL` and `Emulator` means I'd have to go through and write `Omit` logic wherever I use `ConnectionTypeValues` because technically `EventHub` connections don't have a `SQL` option and there are no `Emulator` options for SQL.  In the end I chose `NonAzure` because even though the names of connections are different, I felt that their flows were very close and I didn't feel it was worth adding that extra complexity throughout just for an improvement in specificity when I could umbrella both into the same flow.

Of course I'm open to changing it, but I just wanted to explain my initial though process...  Thoughts?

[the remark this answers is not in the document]
MicroFish91: I'll take a closer re-look at this later, but I originally merged `Emulator` and `SQL` under one umbrella because it made it a lot easier to adopt my existing validation and creation flows without having to write a whole bunch of omit logic for `ConnectionTypeValues`.  Omit logic would be necessary because EventHubs can't have a `SQL` type and SQL doesn't have an `Emulator` type.  I would have to revisit all of my code flows to ensure that nothing breaks when we have those two types separated out vs under one umbrella.

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
MicroFish91: That's your classic `skipForNow` scenario during the initial `Create Function...` command

src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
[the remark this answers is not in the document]
MicroFish91: It just made adopting my existing flows a lot cleaner/simpler because I didn't have to write out different `Omit` logic for `ConnectionTypeValues` (e.g. `Netherite` can't have `SQL` and SQL doesn't have an `Emulator` option).  I remember at the time thinking it might complicate my validation logic as well but I'd really have to revisit to jog my memory on how involved that change would be

src/commands/createFunction/dotnetSteps/DotnetFunctionCreateStep.ts
```diff
@@ -66,22 +66,29 @@ export class DotnetFunctionCreateStep extends FunctionCreateStepBase<IDotnetFunc
...
-                    await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', durableUtils.dotnetDfNetheritePackage);
-                    break;
-                case DurableBackend.SQL:
-                    await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', durableUtils.dotnetDfSqlPackage);
-                    break;
-                case DurableBackend.Storage:
-                default:
+            while (packages.length) {
+                await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', packages[packages.length - 1]);
+                packages.pop();  // Only pop off the package if we know running it did not cause an error
```
bwateratmsft, line 88: This could result in an infinite loop

src/commands/appSettings/AzureWebJobsStoragePromptStep.ts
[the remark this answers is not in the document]
bwateratmsft: (I'll close the other comments relating to the `shouldPrompt()` having effects so we can keep the discussion in one place)

src/commands/createFunction/dotnetSteps/DotnetFunctionCreateStep.ts
MicroFish91: Hmm, I'm not immediately seeing how.  If each pass is popping off a value and reducing the length by 1, and nothing is pushing values back into the array, it should eventually hit 0 right?  And 0 evaluates as false-y.  

Or am I missing something? 👀 😅

bwateratmsft: 🤦 I'm not thinking clearly. My brain was thinking that an error thrown would restart the loop without popping anything.

MicroFish91: Ahh so like if the entire try/catch were inside the while loop, makes sense.  All good 🙂

src/commands/createFunction/scriptSteps/ScriptFunctionCreateStep.ts
```diff
@@ -56,8 +62,34 @@ export class ScriptFunctionCreateStep extends FunctionCreateStepBase<IScriptFunc
...
+            switch (language) {
+                case ProjectLanguage.JavaScript:
+                case ProjectLanguage.TypeScript:
+                    await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'npm', 'install', durableUtils.nodeDfPackage);
+                    break;
+                case ProjectLanguage.Python:
+                    await pythonUtils.addDependencyToRequirements(durableUtils.pythonDfPackage);
+                    await venvUtils.runPipInstallCommandIfPossible();
+                    break;
+                default:
```
nturinski, line 87: Add the case `ProjectLanguage.Powershell` with a comment about having to wait for the blocker.

src/constants-nls.ts
```diff
@@ -0,0 +1,24 @@
...
+export const invalidLowerCaseAlphanumericWithHyphens: string = localize('invalidLowerAlphanumericOrHyphen', `A name must consist of lower-case alphanumeric characters or '-', and must start and end with a lower-case alphanumeric character.`);
+
+export const invalidLength = (lowerLimitIncl?: string, upperLimitIncl?: string) => {
```
nturinski, line 14: The name is pretty non-descript when you see it invoked in the code. Could we go with `getInvalidLengthError` or something along those lines? 

Also, I'd prefer if numbers were the parameters (which get casted into strings). It feels more intuitive that way since `isValidLength` accepts numbers as parameters.

src/constants.ts
```diff
@@ -96,7 +116,10 @@ export const localhost: string = '127.0.0.1';
 export const tsDefaultOutDir: string = 'dist';
 export const tsConfigFileName: string = 'tsconfig.json';
 
-export const localEmulatorConnectionString: string = 'UseDevelopmentStorage=true';
+export const localEventHubsEmulatorConnectionStringDefault: string = 'MemoryF';
+export const localEventHubsEmulatorConnectionStringAlternate: string = 'Memory';
+export const localStorageEmulatorConnectionString: string = 'UseDevelopmentStorage=true';
+export const localEventHubsEmulatorConnectionRegExp: RegExp = new RegExp(`${localEventHubsEmulatorConnectionStringDefault}|${localEventHubsEmulatorConnectionStringAlternate}`);
```
nturinski, line 122: Does case-sensitivity matter at all here?

src/funcConfig/local.settings.ts
```diff
@@ -17,16 +23,43 @@ export interface ILocalSettingsJson {
...
+    if (!projectPath) {
+        return undefined;
     }
 
     const settings: ILocalSettingsJson = await getLocalSettingsJson(context, path.join(projectPath, localSettingsFileName));
-    return settings.Values && settings.Values[azureWebJobsStorageKey];
+    return settings.Values && settings.Values[connectionKey];
+}
+
+export async function validateStorageConnection(context: IActionContext, options?: Omit<IValidateConnectionOptions, 'suppressSkipForNow'>, projectPath?: string): Promise<void> {
```
nturinski, line 41: Where will this be used? I noticed that it's not anywhere right now (not even in the commented out code)

nturinski (approved):

nturinski (changes requested):
lol that approval was on accident. Sorry about the false hope

alexweininger:
When this feature is done, it definitely needs automated tests. Functions has a lot of tests you can use as reference. When the changes are closer to being finalized, or in-between reviews it'd be a good idea to start adding them.

src/commands/appSettings/IEventHubsConnectionWizardContext.ts
```diff
@@ -0,0 +1,26 @@
...
+import { IActionContext, ISubscriptionContext } from "@microsoft/vscode-azext-utils";
+import { ConnectionTypeValues } from "../../constants";
+
+export interface IEventHubsConnectionWizardContext extends IActionContext, Partial<ISubscriptionContext> {
+    projectPath: string;
+
+    resourceGroup?: ResourceGroup;
+
+    // Connection Types
+    azureWebJobsStorageType?: ConnectionTypeValues;
```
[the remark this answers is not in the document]
MicroFish91: So I believe we do check for this attribute in the `shouldPrompt` of `EventHubsConnectionPromptStep`.

However, there's still a small case for removing it.  In the current create flow, we seem to always add the `AzureWebJobsStoragePromptStep` after the other connection prompt steps, so I can definitely see a case for removing this and not even bothering to check for it in the `shouldPrompt` of eventhubs and sql... like maybe it's really only worth doing that check in `AzureWebJobsConnectionPromptStep`.  

My counter argument to that though is that by doing it that way it tightly restricts the order in which these steps must be called which I'm not a huge fan of, so I decided to include the same general checks in all connection prompt steps to be order agnostic.

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -56,12 +57,17 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
     }
 
     public async getSubWizard(context: IFunctionWizardContext): Promise<IWizardOptions<IFunctionWizardContext> | undefined> {
-        const isV2PythonModel = isPythonV2Plus(context.language, context.languageModel);
+        const isV2PythonModel = pythonUtils.isV2Plus(context.language, context.languageModel);
+
+        const needsStorageSetup: boolean = !!context.functionTemplate && durableUtils.requiresDurableStorage(context.functionTemplate.id, context.language) && !context.hasDurableStorage;
+        if (needsStorageSetup) {
```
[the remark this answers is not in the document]
MicroFish91: I actually tried that first and remember it not working.  It's been a while so I can't remember exactly why, but I felt pretty forced to do it this way to get it working (it wasn't my first choice either).  I'll have to reinvestigate and get back with you

src/commands/createFunction/FunctionSubWizard.ts
MicroFish91: Yeah, I believe we should be doing it for all new DF storage types

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceNameStep.ts
```diff
@@ -0,0 +1,47 @@
...
+    private async _validateInput(name: string | undefined): Promise<string | undefined> {
+        name = name ? name.trim() : '';
+
+        if (!validateUtils.isValidLength(name, 6, 50)) {
+            return invalidLength('6', '50');
+        }
+        if (!validateUtils.isAlphanumericWithHypens(name)) {
+            return invalidAlphanumericWithHyphens;
+        }
+        await delay(500);
```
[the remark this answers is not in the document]
MicroFish91: This should be replaced.  I have plans to replace with a new debounce util based on Brandon's earlier recommendation

src/commands/createFunction/dotnetSteps/DotnetFunctionCreateStep.ts
```diff
@@ -53,6 +56,39 @@ export class DotnetFunctionCreateStep extends FunctionCreateStepBase<IDotnetFunc
...
+                packages.push(durableUtils.dotnetDfSqlPackage);
+                break;
+            case DurableBackend.Storage:
+            default:
+        }
+
+        // Seems that the package arrives out-dated and needs to be updated
+        packages.push(durableUtils.dotnetDfBasePackage);
+
+        try {
```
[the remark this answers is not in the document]
MicroFish91: You nailed it, and that's fair, I may opted for writing prettier code here in exchange for a slightly less optimal user experience.  I kind of figured if they were going to have to install a the package manually anyway, it wouldn't be really that much more difficult to copy paste an additional argument from the same error message.

src/funcConfig/local.settings.ts
MicroFish91: It will be used during deploy I believe, it wouldn't be in this PR yet

src/constants.ts
MicroFish91: I remember testing two of those cases and case-sensitivity seemed to matter

src/commands/createFunction/durableSteps/netherite/NetheriteConfigureHostStep.ts
MicroFish91: You bring up a good point, this is more of a critical error since the project won't work properly.  I'm thinking instead of logging it in the channel at all, I could just take out the try/catch and have the wizard throw the error instead.  How do you feel about that solution?

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -81,6 +84,37 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
...
+                hostJson.extensions.durableTask = netheriteUtils.getDefaultNetheriteTaskConfig();
+                setLocalAppSetting(context, context.projectPath, 'EventHubsConnection', '', MismatchBehavior.Overwrite);
+                break;
+            case DurableBackend.SQL:
+                hostJson.extensions.durableTask = sqlUtils.getDefaultSqlTaskConfig();
+                setLocalAppSetting(context, context.projectPath, 'SQLDB_Connection', '', MismatchBehavior.Overwrite);
+                break;
+            default:
+        }
+
```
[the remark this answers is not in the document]
MicroFish91: Will revisit this based on if you leave another comment for the one above

src/commands/createFunction/createFunction.ts
```diff
@@ -64,9 +65,10 @@ export async function createFunctionInternal(context: IActionContext, options: a
     }
 
     const { language, languageModel, version } = await verifyInitForVSCode(context, projectPath, options.language, /* TODO: languageModel: */ undefined, options.version);
+    const hasDurableStorage: boolean = await durableUtils.verifyHasDurableStorage(language, projectPath);
```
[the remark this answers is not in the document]
MicroFish91: I figured it was easier to manage/more flexible to just add this onto the `wizardContext` at the beginning of the create process since it can already be known so I don't have to worry about passing/managing a separate variable later on.

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
```diff
@@ -0,0 +1,39 @@
...
+            sku: {
+                name: 'Standard',
+                capacity: 1
+            },
+        };
+        context.eventHubsNamespace = await client.namespaces.beginCreateOrUpdateAndWait(rgName, newNamespaceName, defaultParams);
+    }
+
+    public shouldExecute(context: T): boolean {
+        return !context.eventHubsNamespace && !!context.resourceGroup && !!context.newEventHubsNamespaceName && LocationListStep.hasLocation(<ILocationWizardContext>context);
```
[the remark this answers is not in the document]
MicroFish91: > I appreciate how thorough this statement is, but it's kind of unnecessary.

Yeah, you're right.  I was going a bit hard on the &&'s huh? 😂

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceListStep.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        const promptSteps: AzureWizardPromptStep<T & ISubscriptionContext>[] = [];
+        const executeSteps: AzureWizardExecuteStep<T & ISubscriptionContext>[] = [];
+
+        promptSteps.push(new EventHubsNamespaceNameStep());
+        executeSteps.push(new EventHubsNamespaceCreateStep());
+
+        LocationListStep.addStep(context, promptSteps);
+        promptSteps.push(new ResourceGroupListStep());
```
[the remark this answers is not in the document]
MicroFish91: Ahh that makes sense, I can check to make sure they have a 1 to 1 overlap and switch to that logic.

src/utils/debounce.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const activeDebounces: { [id: string]: DisposableLike } = {};
+
+/*
+ *  To be used in validating user input through VS Code's showInputBox API
+ *  VS Code's validateInput callback can throw a silent timeout error when the delay is too long and defaults to a return of 'undefined' (causing false returns)
+ *  This function defaults to 1000ms for safety/consistency.
+ */
+export async function inputBoxDebounce<T>(delay: 500 | 750 | 1000, id: string, callback: (...args: unknown[]) => Promise<T> | T, ...args: unknown[]): Promise<T> {
```
MicroFish91, line 17: So I need some feedback here.  I ran into a weird issue where if a validateInput callback function has a delay that's too long, VS Code will throw a timeout error (thanks to @nturinski for helping me find that error).  VS Code seems to silently catch this error and then resolve validateInput to a false positive (I think `undefined`).

I do consider this to be a sort of janky behavior that we should avoid if possible, but the thing is, it doesn't really impact us all that much because false positives on intermediate keystrokes don't actually matter.  Only the last keystroke matters, and that works just fine.

I'd like to avoid the silent errors if possible, but it makes our code a little bit more complex.  I will post the alternate code below, but we can resolve to say undefined for intermediate keystrokes instead whenever a new keystroke is pressed, basically helping to alleviate delay buildup in intermediate keystrokes; however, now undefined will essentially become a `truthy` value for us which I find a little unintuitive.

Anyway, thoughts on which version to implement.

We will call this PR's code version A, this comments reference version B, and version C can be some entirely other implementation suggested by you.

Version B:
image

Example of calling it:

Version A:
image


Version B:
image

bwateratmsft: @nturinski can you think of a way to use the `InputBox`'s `onDidAccept` event to deal with slow validation steps like this? We'd want a way to avoid prompting twice (or else the framework gets confused about back buttons and such)--so I don't think putting a loop around it is the way to go. However, _within_ the wizard/UI framework, we could add a method for these slow validation steps that checks on accept, and forces the user to try again if needed, but internally knows that this is really one prompt.

bwateratmsft: I'm imagining an `async`, slow validation task that is checked with `onDidAccept`, and if it fails it reopens the prompt, noting the error, and supplying what the user entered as the already-inputted text (so they don't have to retype everything).

src/commands/createFunction/FunctionListStep.ts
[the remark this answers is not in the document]
MicroFish91: Bit of a nuanced issue, discussing offline

src/utils/azure.ts
```diff
@@ -66,6 +68,43 @@ export async function getStorageConnectionString(context: IStorageAccountWizardC
...
+export async function getEventHubsConnectionString(context: IEventHubsConnectionWizardContext & ISubscriptionContext): Promise<IResourceResult> {
+    const client: EventHubManagementClient = await createEventHubClient(context);
+    const resourceGroupName: string = parseAzureResourceId(nonNullValue(context.eventHubsNamespace?.id)).resourceGroup;
+    const namespaceName: string = nonNullValue(context.eventHubsNamespace?.name);
+    const authorizationRuleName: string = 'RootManageSharedAccessKey';
+
+    const accessKeys: AccessKeys = await client.namespaces.listKeys(resourceGroupName, namespaceName, authorizationRuleName);
+    return {
+        name: namespaceName,
+        connectionString: accessKeys.primaryConnectionString || ""
```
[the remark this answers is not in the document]
MicroFish91: Changed the logic quite a bit for this, but I think it should address your concerns here, if not let me know!

MicroFish91:
I think I was able to address most of the feedback so far, let me know if I missed anything, but this should be ready for a second major review pass.  I wanted to also point out that I added an edit note in the main section [above](https://github.com/microsoft/vscode-azurefunctions/pull/3390#issue-1424591833) describing a slight change to the PR review plan (see the bottom note related to eventual removal of the `skipForNow` pattern). Thanks again for all the PR review :)

src/utils/workspace.ts
```diff
@@ -15,19 +15,27 @@ export function isMultiRootWorkspace(): boolean {
         && vscode.workspace.name !== vscode.workspace.workspaceFolders[0].name; // multi-root workspaces always have something like "(Workspace)" appended to their name
 }
 
+export function getWorkspaceRootPath(): string | undefined {
+    if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 0) {
+        return vscode.workspace.workspaceFolders[0].uri.fsPath;
+    } else {
+        return;
```
bwateratmsft:
```suggestion
        return undefined;
```

src/utils/azure.ts
```diff
@@ -66,6 +69,66 @@ export async function getStorageConnectionString(context: IStorageAccountWizardC
...
+export async function getEventHubsConnectionString(context: IEventHubsConnectionWizardContext & ISubscriptionContext): Promise<IResourceResult> {
+    const client: EventHubManagementClient = await createEventHubClient(context);
+    const resourceGroupName: string = getResourceGroupFromId(nonNullValue(context.eventHubsNamespace?.id));
+    const namespaceName: string = nonNullValue(context.eventHubsNamespace?.name);
+
+    const authRulesIterable = client.namespaces.listAuthorizationRules(resourceGroupName, namespaceName);
+    const authRules: AuthorizationRule[] = await uiUtils.listAllIterator(authRulesIterable);
+
+    let authRule: string;
+    if (!authRules.length) {
```
MicroFish91, line 81: I think this probably works for now, but I want to revisit this logic in a future PR before we merge the final into main.  I have an open issue here for tracking: https://github.com/microsoft/vscode-azurefunctions/issues/3445

src/commands/appSettings/AzureWebJobsStorageExecuteStep.ts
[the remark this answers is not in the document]
MicroFish91: Let's revisit this discussion next week: https://learn.microsoft.com/en-us/azure/azure-sql/database/local-dev-experience-sql-database-emulator?view=azuresql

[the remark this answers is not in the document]
MicroFish91: New issue to cover the new connection types:

https://github.com/microsoft/vscode-azurefunctions/issues/3448

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -81,6 +84,37 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
...
+            return;
+        }
+
+        const hostJson: IHostJsonV2 = await AzExtFsExtra.readJSON(hostJsonPath) as IHostJsonV2;
+        hostJson.extensions ??= {};
+
+        switch (context.newDurableStorageType) {
+            case DurableBackend.Storage:
+                hostJson.extensions.durableTask = durableUtils.getDefaultStorageTaskConfig();
+                break;
```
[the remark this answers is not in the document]
nturinski: Hmm, I might be looking at this wrong, but it only sets it for `DotnetProjectCreate` so may be good to set it here.

src/utils/debounce.ts
MicroFish91: I'll open an issue to investigate this further in `AzureTools`.  I would assume that either debounce implementation would still benefit from having a solution like the one Brandon mentioned above.

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
```diff
@@ -0,0 +1,43 @@
...
+    private readonly _functionSettings: { [key: string]: string | undefined };
+
+    public constructor(functionSettings?: { [key: string]: string | undefined }) {
+        super();
+        this._functionSettings = functionSettings || {};
+    }
+
+    public async prompt(context: T): Promise<void> {
+        const durableStorageLabels: string[] = [
+            'Durable Functions Orchestration using Storage',
```
nturinski, line 22: nit: Could localize these labels.

src/utils/debounce.ts
MicroFish91: Resume discussions here:  https://github.com/microsoft/vscode-azuretools/issues/1275

src/constants.ts
nturinski: Then I think it'd be good to add the `i` flag just in case.

src/utils/debounce.ts
nturinski: > I'm imagining an `async`, slow validation task that is checked after `onDidAccept`, and if it fails it reopens the prompt, noting the error, and supplying what the user entered as the already-inputted text (so they don't have to retype everything).

I think that this would work. We could make the slow validation task property optional as well so that this wouldn't be a breaking change.

src/commands/appSettings/IEventHubsConnectionWizardContext.ts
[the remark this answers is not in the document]
MicroFish91: Ahh I see what you mean now.  We don't ever `set` `azureWebJobsStorageType` in this flow, but we do check for it.

image

src/constants.ts
MicroFish91: I think you read that as case-insensitivity then 🙂

src/commands/createFunction/FunctionCreateStepBase.ts
```diff
@@ -53,6 +56,8 @@ export abstract class FunctionCreateStepBase<T extends IFunctionWizardContext> e
         progress.report({ message: localize('creatingFunction', 'Creating new {0}...', template.name) });
 
         const newFilePath: string = await this.executeCore(context);
+        await this._configureForDurableStorageIfNeeded(context);
```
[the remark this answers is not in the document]
MicroFish91: Oh maybe I mis-understood what you were wanting implemented in your last comment then.  Did you want something more along the lines of the previous implementation?

src/commands/createFunction/FunctionCreateStepBase.ts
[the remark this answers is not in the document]
MicroFish91: Also here:

image

MicroFish91:
Added an info link for durable storage providers:

image

src/commands/createFunction/FunctionCreateStepBase.ts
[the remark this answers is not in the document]
MicroFish91: Let's discuss this one offline

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
```diff
@@ -0,0 +1,54 @@
...
+            { label: durableStorageLabels[1], data: DurableBackend.Netherite },
+            { label: durableStorageLabels[2], data: DurableBackend.SQL },
+            { label: durableStorageInfo, data: undefined, suppressPersistence: true }
+        ];
+
+        let pick: DurableBackendValues | undefined;
+        while (!pick) {
+            pick = (await context.ui.showQuickPick(picks, { placeHolder })).data;
+            if (!pick) {
+                await openUrl('https://aka.ms/durable-storage-providers');
```
MicroFish91, line 41: new info link code here...

src/constants.ts
MicroFish91: I think you might have read that backwards? :)

src/commands/createFunction/durableSteps/ConfigureDurableProjectStep.ts
```diff
@@ -0,0 +1,127 @@
...
+import { IHostJsonV2 } from '../../../funcConfig/host';
+import { MismatchBehavior, setLocalAppSetting } from '../../../funcConfig/local.settings';
+import { localize } from '../../../localize';
+import { cpUtils } from '../../../utils/cpUtils';
+import { durableUtils, netheriteUtils, sqlUtils } from '../../../utils/durableUtils';
+import { pythonUtils } from '../../../utils/pythonUtils';
+import { venvUtils } from '../../../utils/venvUtils';
+import { IFunctionWizardContext } from '../IFunctionWizardContext';
+
+export class ConfigureDurableProjectStep<T extends IFunctionWizardContext> extends AzureWizardExecuteStep<T> {
```
nturinski, line 20: nit: Prefer that we call this `DurableProjectConfigureStep.ts`. Most files in our repo follow the pattern of `NounVerbStep`

src/commands/createFunction/durableSteps/ConfigureDurableProjectStep.ts
```diff
@@ -0,0 +1,127 @@
...
+
+        await this._configureHostAndLocalSettingsJson(context);
+        await this._installDurableDependencies(context);
+    }
+
+    public shouldExecute(context: T): boolean {
+        return !!context.newDurableStorageType;
+    }
+
+    private async _configureHostAndLocalSettingsJson(context: T): Promise<void> {
```
nturinski, line 35: I'm not sure about this, but wanted to bring it up:

Rather than failing the entire `createFunction` wizard, would it be better to just output the error at the end, but continue creating anyway?

I feel like this will block users in a way that they may not be able to fix easily. Like, maybe the path is janky or we don't have access to it, ect.

But if we just alert them to the error, they can manually resolve it themselves.

src/commands/createFunction/durableSteps/ConfigureDurableProjectStep.ts
```diff
@@ -0,0 +1,127 @@
...
+            const message: string = localize('failedToFindHost', 'Failed to find "{0}" in your project root.', hostFileName);
+            throw new Error(message);
+        }
+
+        const hostJson: IHostJsonV2 = await AzExtFsExtra.readJSON(hostJsonPath) as IHostJsonV2;
+        hostJson.extensions ??= {};
+
+        switch (context.newDurableStorageType) {
+            case DurableBackend.Storage:
+                hostJson.extensions.durableTask = durableUtils.getDefaultStorageTaskConfig();
```
nturinski, line 47: Could you leave a comment that this is set in `AzureWebJobsExecuteStep`? I think someone would assume that this is a mistaken omission.

src/commands/createFunction/durableSteps/ConfigureDurableProjectStep.ts
```diff
@@ -0,0 +1,127 @@
...
+                hostJson.extensions.durableTask = sqlUtils.getDefaultSqlTaskConfig();
+                await setLocalAppSetting(context, context.projectPath, ConnectionKey.SQL, '', MismatchBehavior.Overwrite);
+                break;
+            default:
+        }
+
+        await AzExtFsExtra.writeJSON(hostJsonPath, hostJson);
+    }
+
+    private async _installDurableDependencies(context: T): Promise<void> {
```
nturinski, line 63: Bit of a nit: Since this can technically fail, it's nice to prefix this with try just so people are aware that its success is not guaranteed (or won't error out). Since it's a private function, it doesn't _really_ matter which is why it's more of a nit, but thought I'd point it out anyway.

```suggestion
    private async _tryInstallDurableDependencies(context: T): Promise<void> {
```

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
```diff
@@ -0,0 +1,38 @@
...
+    public async execute(context: T, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const rgName: string = nonNullValue(context.resourceGroup?.name);
+        const newNamespaceName: string = nonNullValue(context.newEventHubsNamespaceName);
+        const creating: string = localize('creatingEventHubsNamespace', 'Creating new event hubs namespace "{0}"...', newNamespaceName);
+        ext.outputChannel.appendLog(creating);
+        progress.report({ message: creating });
+
+        const client: EventHubManagementClient = await createEventHubClient(<T & ISubscriptionContext>context);
+        const defaultParams: EHNamespace = {
+            location: nonNullValue(context.resourceGroup?.location),
```
nturinski, line 26: Discussed offline: Use `LocationListStep.getLocation` here after invoking `LocationListStep.setLocation` based off the resource group's location.

src/commands/createFunction/durableSteps/ConfigureDurableProjectStep.ts
MicroFish91: Yeah we could do it that way.  I think if we go this route, I should also go into `NetheriteConfigureHostStep` and make sure it won't error when the host is missing then.  I originally wasn't trying to prevent this for the same reason as why I was throwing the error in this configure step.

src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private async _configureHostAndLocalSettingsJson(context: T): Promise<void> {
+        const hostJsonPath: string = path.join(context.projectPath, hostFileName);
+
+        if (!await AzExtFsExtra.pathExists(hostJsonPath)) {
+            const message: string = localize('durableHostConfigFailed', 'Unable to find and configure "{0}" in your project root. You may need to configure your durable function settings manually.', hostFileName);
+            ext.outputChannel.appendLog(message);
+
+            await callWithTelemetryAndErrorHandling('durableHostConfigFailed', async () => {
```
nturinski, line 43: Rather than creating a new telemetry event for us to track, I think it'd be easier if you just set telemetry on the context.

Here's an example how that's done:
`context.telemetry.properties.durableHostConfigFailed = true`

src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
```diff
@@ -0,0 +1,141 @@
...
+                const notification: string = localize('failedToConfigureHost', 'Failed to configure your "{0}"', hostFileName);
+                void window.showInformationMessage(notification, viewOutput).then(result => {
```
nturinski, line 45: Not 100% sure that it's worth showing a notification for this, but if we do, we should use a warning message since it will cause the project to fail.

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
```diff
@@ -22,8 +23,9 @@ export class EventHubsNamespaceCreateStep<T extends IEventHubsConnectionWizardCo
         progress.report({ message: creating });
 
         const client: EventHubManagementClient = await createEventHubClient(<T & ISubscriptionContext>context);
+        const location: AzExtLocation = await LocationListStep.getLocation(<T & ISubscriptionContext>context);
```
nturinski, line 26: Does this work? I wasn't sure if you had to `LocationListStep.setLocation` first, after getting it from the `context.resourceGroup.location`.

MicroFish91: Yeah it works, I looked at `ResourceGroupListStep` and found that it actually sets it for you:

image

src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
MicroFish91: Do we want a more silent failure where we only post the issue to the `ext.outputChannel`?

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
MicroFish91: I believe it works because the location is set with `StorageAccountListStep`.  Let's talk offline about this, there's some questions I want to address now that I'm looking into the Azure Wizard Step utils more closely.

src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
nturinski: I don't know, generally we try to avoid notifications when possible, but I agree that this is a breaking problem so we should be more obvious about it.

If we ever move it into the activity log, we can remove the notification then since that is more obvious than the output channel.

src/commands/createFunction/durableSteps/netherite/EventHubsNamespaceCreateStep.ts
MicroFish91: Discussed:  Sounds like we can probably keep this `getLocation` call and investigate whether or not it makes sense to add a `setLocation` in `ResourceGroupListStep` as a part of `Azure Tools`.

MicroFish91:
🥹🫡

MicroFish91:
Also closes #2883

PR #3391, opened 2022-10-26
Remove deprecated `createFunctionApp` API

alexweininger:
In an effort to remove usages of `findTreeItem`, we can remove this deprecated part of `createFunctionApp`.

We deprecated this API in 2020 https://github.com/microsoft/vscode-azurefunctions/pull/2077

### DIFF
--- a/src/commands/createFunctionApp/createFunctionApp.ts
+++ b/src/commands/createFunctionApp/createFunctionApp.ts
@@ -5,19 +5,13 @@
 
 import { AzExtParentTreeItem, IActionContext } from '@microsoft/vscode-azext-utils';
 import { ext } from '../../extensionVariables';
-import { localize } from '../../localize';
 import { SlotTreeItem } from '../../tree/SlotTreeItem';
 import { ICreateFunctionAppContext, SubscriptionTreeItem } from '../../tree/SubscriptionTreeItem';
 import { ISiteCreatedOptions } from './showSiteCreated';
 
-export async function createFunctionApp(context: IActionContext & Partial<ICreateFunctionAppContext>, subscription?: AzExtParentTreeItem | string, newResourceGroupName?: string): Promise<string> {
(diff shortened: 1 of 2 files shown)


nturinski (commented):
I think we have to actually delete the entry points `createFunctionFromCommand` and `createNewProjectFromCommand` to actually remove these deprecated commands, otherwise the users will have odd errors.

We could replace the implementation with something that throws an error stating to use the `createFunctionApp` API call instead?

nturinski:
Actually, these two commands aren't related either. `createFunctionApp` is not deprecated, `createFunctionFromCommand` and `createNewProjectFromCommand` are what are deprecated.

Both of these commands refer to local function projects whereas `createFunctionApp` is a way to create Function Apps in Azure programmatically. I don't think that we can just remove this without creating a new API endpoint for creating Function Apps.

alexweininger:
> Actually, these two commands aren't related either. `createFunctionApp` is not deprecated, `createFunctionFromCommand` and `createNewProjectFromCommand` are what are deprecated.
> 
> Both of these commands refer to local function projects whereas `createFunctionApp` is a way to create Function Apps in Azure programmatically. I don't think that we can just remove this without creating a new API endpoint for creating Function Apps.

I'm not removing `createFunctionApp`, just removing the ability to pass in the subscription argument as a string. Which according to [this unit test](https://github.com/microsoft/vscode-azurefunctions/pull/3391/files#diff-a21a2a361409a83170ebfe53dc61da7bc7eb025ca49a8b44977e8ea87dea294fL67) is deprecated.

Do we know who may be using this command passing in a subscription id as a string? I couldn't find any in my searches on GitHub.

ejizba:
It was some sort of IoT team - I can dig through my email if you really care. A few places IoT is mentioned:
- https://github.com/microsoft/vscode-azurefunctions/pull/728
- https://github.com/microsoft/vscode-azurefunctions/blob/main/test/createFunction/createFunction.CSharp.test.ts#L166

PR #3406, opened 2022-10-28
Durable Functions with Backends - Part IV

MicroFish91:

## Pull Request - Part 4
This PR primarily covers:
* Creating a Function App in Azure

#### <ins>General Flow and What to Inspect</ins>
Since we can technically create a Function App without having a project initialized yet, we do not force the allocation of any storage assets until we do a final verification on deploy...

If we can detect any durable storage features in the current project, we will add those relevant details to the function app create as necessary, but it is not required to get started.

image

#### 1.  No Durable Project

image

#### 2. Default Storage

Subtle difference from the above scenario, but we don't auto-name the assets the same by default when a durable project is detected (since the assets may already exist, we want to list those out and allow a selection instead)...

image

#### 3. Netherite

image

#### 4.  SQL

image


src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -9,8 +9,9 @@ import { LocationListStep } from '@microsoft/vscode-azext-azureutils';
 import { AzureWizardExecuteStep, parseError } from '@microsoft/vscode-azext-utils';
```
bwateratmsft, line 9: Similar to other PRs, the `@azure` import should be `import type`.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -12,14 +12,16 @@ import { FunctionAppCreateStep } from '../commands/createFunctionApp/FunctionApp
 import { FunctionAppHostingPlanStep, setConsumptionPlanProperties } from '../commands/createFunctionApp/FunctionAppHostingPlanStep';
```
bwateratmsft, line 12: Similar to other PRs, the `@azure` import should be `import type`.

MicroFish91:
A lot of the `Create Function App` changes could be revisited in a non-Durable PR.  I'll open a separate issue and we can close this in favor of something more comprehensive in the future.

PR #3407, opened 2022-10-28
Durable Functions with Backends - Part V

MicroFish91:

## Pull Request - Part 5
This PR primarily covers:
* Deploying local project to Function app


src/commands/deploy/deploy.ts
```diff
@@ -3,17 +3,19 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { SiteConfigResource } from '@azure/arm-appservice';
-import { deploy as innerDeploy, getDeployFsPath, getDeployNode, IDeployContext, IDeployPaths, showDeployConfirmation } from '@microsoft/vscode-azext-azureappservice';
+import { SiteConfigResource, StringDictionary } from '@azure/arm-appservice';
```
bwateratmsft, line 6:
```suggestion
import type { SiteConfigResource, StringDictionary } from '@azure/arm-appservice';
```

src/commands/deploy/verifyAppSettings.ts
```diff
@@ -7,19 +7,20 @@ import { StringDictionary } from '@azure/arm-appservice';
 import { ParsedSite } from '@microsoft/vscode-azext-azureappservice';
```
bwateratmsft, line 7: Similar to other PRs, the `@azure` import should be `import type`.

src/commands/deploy/deploy.ts
```diff
@@ -92,13 +124,12 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         void validateGlobSettings(context, context.effectiveDeployFsPath);
     }
 
-    if (language === ProjectLanguage.CSharp && !node.site.isLinux) {
-        await updateWorkerProcessTo64BitIfRequired(context, siteConfig, node, language);
+    if (language === ProjectLanguage.CSharp && !node.site.isLinux || durableStorageType) {
```
nturinski, line 127: Does the durable storage only work with 64 bit? 

Looking below, it seems like it may be necessary for Netherite? I think it makes more sense to check for Netherite here rather than in the function definition. 

I'm assuming that you put the check in the function itself so that the warning message wouldn't come up, but is that not a valid warning even if they're using Netherite?

src/commands/deploy/shouldValidateConnection.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    return { shouldValidateStorage, shouldValidateEventHubs, shouldValidateSqlDb };
+}
+
+async function promptShouldOverwrite(context: IActionContext, key: ConnectionKeyValues): Promise<boolean> {
+    const overwriteButton: vscode.MessageItem = { title: localize('overwrite', 'Overwrite') };
+    const skipButton: vscode.MessageItem = { title: localize('skip', 'Skip') };
+    const buttons: vscode.MessageItem[] = [overwriteButton, skipButton];
+
+    const message: string = localize('overwriteRemoteConnection', 'We detected a different local connection setting for "{0}" than what was previously used. Would you like to overwrite your remote setting?', key);
```
nturinski, line 64: In general, I feel like our prompts are very concise and sound like a computer is telling you something. This warning message has a very "human-written" feeling to me rather than a "system" warning in my opinion. I had a couple of suggestions, but if you prefer what you have, then don't worry about changing it.

`Local and remote application setting "{0}" do not match. Overwrite the remote setting?`

`The "{0}" setting in the "local.settings.json" does not match the remote application setting. Overwrite application setting?`

src/commands/deploy/deploy.ts
```diff
@@ -82,6 +94,26 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         context.deployMethod = 'storage';
     }
 
+    // Preliminary local validation done to ensure all required resources have been created and are available. Final deploy writes are made in 'verifyAppSettings'
+    switch (durableStorageType) {
+        case DurableBackend.Netherite:
```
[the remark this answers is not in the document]
MicroFish91: Good call, and I think I can do you one even better by removing the check on durableStorageType altogether (I just remembered the booleans themselves already include a check for that 🤭)

src/commands/deploy/shouldValidateConnection.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    return { shouldValidateStorage, shouldValidateEventHubs, shouldValidateSqlDb };
+}
+
+async function promptShouldOverwrite(context: IActionContext, key: ConnectionKeyValues): Promise<boolean> {
+    const overwriteButton: vscode.MessageItem = { title: localize('overwrite', 'Overwrite') };
+    const skipButton: vscode.MessageItem = { title: localize('skip', 'Skip') };
+    const buttons: vscode.MessageItem[] = [overwriteButton, skipButton];
+
+    const message: string = localize('overwriteRemoteConnection', 'The "{0}" setting in the "local.settings.json" does not match the remote application setting. Overwrite remote setting?', key);
```
nturinski, line 64: nit: We have a constant for "local.settings.json" that you can use.

nturinski (approved):

Issue #2977, opened 2021-09-28
Add "Durable Functions entity"  templates triggers for Python, JS, and TS to verified templates


nturinski:
We need to verify manually that templates work and add test cases to suite to be considered "Verified".

nturinski:
Fixed by https://github.com/microsoft/vscode-azurefunctions/pull/3412

PR #3412, opened 2022-11-03
Verify Durable Functions Entity templates

davidmrdavid:
_This PR is still in progress._

This PR adds the necessary tests to add the Durable Functions Entity templates into the Verified filter.
Currently, the JS, TS, and Python templates have been verified as per this PR.
Unfortunately, the C# templates have not as their template ID is not recognized. This is still being investigated.

One caveat is that it appears the verification has only taken place for Functions runtime V4. I need to validate that these templates would show up in other Functions versions as well.


nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/2977

davidmrdavid:
Just a heads up - this PR may explain why we were not seeing the C# templates in our tests. For now, let's focus on just verifying the JS, TS, and Python templates.

test/createFunction/createFunction.Script.v2.test.ts
```diff
@@ -146,6 +146,16 @@ function addSuite(tester: FunctionTesterBase): void {
...
         },
+        {
+            functionName: 'Durable Functions entity',
+            inputs: [],
+            skip: tester.language === ProjectLanguage.PowerShell || tester.language === ProjectLanguage.Java
+        },
+        {
+            functionName: 'Durable Functions Entity HTTP starter',
+            inputs: tester.language === ProjectLanguage.JavaScript ? [] : [getRotatingAuthLevel()],
+            skip: tester.language === ProjectLanguage.PowerShell || tester.language === ProjectLanguage.Java || tester.language === ProjectLanguage.Python
```
nturinski, line 157: Does this not exist for Python? I'm wondering why you decided to skip testing this one.

src/templates/script/getScriptVerifiedTemplateIds.ts
```diff
@@ -26,6 +26,8 @@ export function getScriptVerifiedTemplateIds(version: string): (string | RegExp)
         verifiedTemplateIds = verifiedTemplateIds.concat([
             'CosmosDBTrigger',
             'DurableFunctionsActivity',
+            'DurableFunctionsEntity',
+            'DurableFunctionsEntityHttpStart',
```
nturinski, line 30: We would have to remove this from the verified template id for Python specifically until we don't skip the test.

test/createFunction/createFunction.Script.v2.test.ts
davidmrdavid: Yes, it doesn't exist yet. I'll be adding it later, apologies for the inconsistency across languages. We did not notice this was missing, in part, because the templates don't show up by default :)

src/templates/script/getScriptVerifiedTemplateIds.ts
davidmrdavid: @nturinski: does this latest commit [here](https://github.com/microsoft/vscode-azurefunctions/pull/3412/commits/12306666105bdc23598681b420cb51a3c26b9cf2) match what you're looking for?

davidmrdavid:
@nturinski: Please note that we still have the open question as to why I only had to modify the expected template count for Functions V4. As shown here, in the case of Python: https://github.com/microsoft/vscode-azurefunctions/blob/12306666105bdc23598681b420cb51a3c26b9cf2/test/templateCount.test.ts#L28-L30

Do you know if I'm missing anything to enable this in Functions V3 as well? In any case, I don't mind fixing this for Functions V4 first and then investigating further for Functions V3, but it would be ideal to do it all in one go.

nturinski:
> Do you know if I'm missing anything to enable this in Functions V3 as well? In any case, I don't mind fixing this for Functions V4 first and then investigating further for Functions V3, but it would be ideal to do it all in one go.

Yeah, I figured out why it seemed like they were being skipped.  It's cuz they were. 😅 Thanks to this line in `templateCount.test.ts`
```
if ((source === TemplateSource.Staging && skipStagingTemplateSource) || shouldSkipVersion(version)) {
                    this.skip();
                }
```

We only test v4 with local tests. The other versions are tested in our "long running tests". Unfortunately, our long running tests seem to be broken right now so I wasn't able to verify the other function runtimes.

In order to test the other versions, could you please go to the `global.test.ts` file and have `isLongRunningVersion` always return `false`.  That way it should test all Function runtime versions and the verified templates will be tested for every runtime.

nturinski:
In light of this announcement:

> Beginning on December 3, 2022, function apps running on versions 2.x and 3.x of the Azure Functions runtime can no longer be supported. Before that time, please test, verify, and migrate your function apps to version 4.x of the Functions runtime. For more information, see [Migrate apps from Azure Functions version 3.x to version 4.x](https://learn.microsoft.com/en-us/azure/azure-functions/migrate-version-3-version-4). After the deadline, function apps can be created and deployed, and existing apps continue to run. However, your apps won't be eligible for new features, security patches, performance optimizations, and support until you upgrade them to version 4.x.
> 
> End of support for these runtime versions is due to the ending of support for .NET Core 3.1, which is required by these older runtime versions. This requirement affects all Azure Functions runtime languages.
> Functions version 1.x is still supported for C# function apps that require the .NET Framework. Preview support is now available in Functions 4.x to [run C# functions on .NET Framework 4.8](https://learn.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide#supported-versions).
-- https://learn.microsoft.com/en-us/azure/azure-functions/functions-versions?tabs=v4&pivots=programming-language-csharp

I think it's safe to ignore the v3 and v2 runtimes. Don't worry about running those tests.  Do you know if Durable Entity templates apply to v1?

PR #3414, opened 2022-11-04
Streamline project creation when no active project is detected via `Create Function` 

### LINKED ISSUE #3265: Make it easier to create a brand new project from the workspace UI
I find that with the new unified view, it's much more difficult to create a new functions project. Here are two things that I think could help

# Make the button more visible

Screen Shot 2022-07-28 at 4 48 13 PM

Can we move "Create function" out of the submenu and into a top-level button on the workspace view? I was a big fan of the lightning bolt button because I thought it was very recognizable as Functions. I think putting the action in a submenu is less discoverable and more annoying because it requires an extra click. If a user has the Functions extension installed, I think it's fair to assume they will use this action and it deserves a top-level button.

# Reduce the number of prompts

If I'm starting without a project I get one of these prompts when I do "Create function" based on if I have a folder open:

Screen Shot 2022-07-28 at 4 47 55 PMScreen Shot 2022-07-28 at 4 50 14 PM

This is pretty much the quintessential "new user" getting started experience, and I don't think immediately showing them a warning is a good look. Can we either
1. Add the explicit "Create new project" button back to the UI (Right now it's possible in the command palette, but I feel like plenty of people never use the command palette)
2. Keep it to just one button, but add some auto-magic functionality that skips the above prompts and just goes right into the 'create project' wizard if it doesn't think you have a project already. I'm pretty sure SWA already does something like this.

Note: The other problem with the first warning shown above is it doesn't let me pick a new folder to create a project in - aka it forces me to create the project in my open folder which may already have stuff.

MicroFish91:
Fixes #3265 

Though we may still need to revisit this again when we address [buttons lost to time](https://github.com/microsoft/vscode-azureresourcegroups/issues/316).

Based on the recommendations in the issue, I removed the following prompts from the `Create Function` task when we don't think the user has a project ready.

image

image

Instead, if the workspace is not selected or is empty, we put the user directly into the `Create New Project` flow.

### DIFF
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -56,6 +56,8 @@ export async function createFunctionInternal(context: IActionContext, options: a
         workspaceFolder = getContainingWorkspace(workspacePath);
     }
 
+    addLocalFuncTelemetry(context, workspacePath);
+
     let projectPath: string | undefined = await verifyProjectPath(context, workspaceFolder || workspacePath);
     if (!projectPath) {
         // If we cannot find a valid Functions project, we need to put the user into the 'Create New Project' flow..
@@ -63,8 +65,6 @@ export async function createFunctionInternal(context: IActionContext, options: a
         await createNewProjectInternal(context, options);
         return;
     }
-
-    addLocalFuncTelemetry(context, workspacePath);
 
     const { language, languageModel, version } = await verifyInitForVSCode(context, projectPath, options.language, /* TODO: languageModel: */ undefined, options.version);
 
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -6,7 +6,6 @@
 import { AzureWizard, IActionContext, UserCancelledError } from '@microsoft/vscode-azext-utils';
 import { window, workspace, WorkspaceFolder } from 'vscode';
 import { ProjectLanguage, projectTemplateKeySetting } from '../../constants';
-import { NoWorkspaceError } from '../../errors';
 import { addLocalFuncTelemetry } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
 import { FuncVersion } from '../../FuncVersion';
 import { localize } from '../../localize';
@@ -62,15 +61,7 @@ export async function createFunctionInternal(context: IActionContext, options: a
         // If we cannot find a valid Functions project, we need to put the user into the 'Create New Project' flow..
         context.telemetry.properties.noWorkspaceResult = 'createNewProject';
         await createNewProjectInternal(context, options);
-
-        if (!workspace.workspaceFolders || workspace.workspaceFolders.length === 0) {
-            // Theoretically shouldn't trigger if 'Create New Project' was successful...
-            throw new NoWorkspaceError();
-        }
-
-        workspaceFolder = workspace.workspaceFolders[0];
-        workspacePath = workspaceFolder.uri.fsPath;
-        projectPath = workspacePath;
+        return;
     }
 
     addLocalFuncTelemetry(context, workspacePath);
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -52,7 +52,7 @@ export async function createFunctionInternal(context: IActionContext, options: a
 
     if (workspacePath === undefined) {
         workspaceFolder = await getWorkspaceFolder();
-        workspacePath = workspaceFolder ? workspaceFolder.uri.fsPath : undefined;
+        workspacePath = workspaceFolder?.uri.fsPath;
     } else {
         workspaceFolder = getContainingWorkspace(workspacePath);
     }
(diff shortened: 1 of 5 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -49,20 +49,32 @@ export async function createFunctionFromCommand(
 export async function createFunctionInternal(context: IActionContext, options: api.ICreateFunctionOptions): Promise<void> {
     let workspaceFolder: WorkspaceFolder | undefined;
     let workspacePath: string | undefined = options.folderPath;
+
     if (workspacePath === undefined) {
-        workspaceFolder = await getWorkspaceFolder(context, options);
-        workspacePath = workspaceFolder.uri.fsPath;
+        workspaceFolder = await getWorkspaceFolder();
+        workspacePath = workspaceFolder ? workspaceFolder.uri.fsPath : undefined;
```
bwateratmsft, line 55: Optional chaining is the best thing since sliced bread

```suggestion
        workspacePath = workspaceFolder?.uri.fsPath;
```

MicroFish91: Oh duh... 🤦‍♂️

Thanks for catching that!

bwateratmsft (dismissed):

nturinski:
There's a weird bug where when you create a new function in an empty folder, it will create the new project, a new function, and then prompt me for another new http trigger.

src/commands/createFunction/createFunction.ts
```diff
@@ -49,20 +49,32 @@ export async function createFunctionFromCommand(
...
-
-    const projectPath: string | undefined = await verifyAndPromptToCreateProject(context, workspaceFolder || workspacePath, options);
+    let projectPath: string | undefined = await verifyProjectPath(context, workspaceFolder || workspacePath);
     if (!projectPath) {
-        return;
+        // If we cannot find a valid Functions project, we need to put the user into the 'Create New Project' flow..
+        context.telemetry.properties.noWorkspaceResult = 'createNewProject';
+        await createNewProjectInternal(context, options);
+
+        if (!workspace.workspaceFolders || workspace.workspaceFolders.length === 0) {
```
nturinski, line 66: Part of creating a new project is creating a new function, so it should probably just return here.

src/commands/createFunction/createFunction.ts
```diff
@@ -49,20 +48,24 @@ export async function createFunctionFromCommand(
...
-    const projectPath: string | undefined = await verifyAndPromptToCreateProject(context, workspaceFolder || workspacePath, options);
+    let projectPath: string | undefined = await verifyProjectPath(context, workspaceFolder || workspacePath);
     if (!projectPath) {
+        // If we cannot find a valid Functions project, we need to put the user into the 'Create New Project' flow..
+        context.telemetry.properties.noWorkspaceResult = 'createNewProject';
+        await createNewProjectInternal(context, options);
         return;
     }
 
+    addLocalFuncTelemetry(context, workspacePath);
```
nturinski, line 67: I think you should move this back up to where it was to capture telemetry.

src/commands/createNewProject/verifyIsProject.ts
```diff
@@ -81,23 +79,14 @@ async function promptForProjectSubpath(context: IActionContext, workspacePath: s
...
-    const projectPath: string | undefined = await tryGetFunctionProjectRoot(context, workspaceFolder, 'modalPrompt');
-    if (!projectPath) {
-        if (!options.suppressCreateProjectPrompt) {
-            const message: string = localize('notFunctionApp', 'The selected folder is not a function project. Create new project?');
-            // No need to check result - cancel will throw a UserCancelledError
-            await context.ui.showWarningMessage(message, { modal: true, stepName: 'notAProject' }, DialogResponses.yes);
+export async function verifyProjectPath(context: IActionContext, workspaceFolder?: WorkspaceFolder | string): Promise<string | undefined> {
+    if (workspaceFolder) {
+        const projectPath: string | undefined = await tryGetFunctionProjectRoot(context, workspaceFolder, 'modalPrompt');
+        if (projectPath) {
```
nturinski, line 87: nit: You don't need to check `projectPath`, you could just do

`return await tryGetFunctionProjectRoot(context, workspaceFolder, 'modalPrompt');`

src/commands/createFunction/createFunction.ts
MicroFish91: Because of the change I've made in `getWorkspaceFolder` workspacePath may sometimes be `undefined` which is why I moved it down to a spot where I know `workspacePath` will always be defined.  Does this matter for our purposes in telemetry or is `undefined` an okay value to record?

nturinski: Actually, don't worry about it. It's already done in `createNewProjectInternal`. But to answer your question, yes, it's fine as the function definition is as follows:

`export function addLocalFuncTelemetry(context: IActionContext, workspacePath: string | undefined)`

MicroFish91: I ended up moving it back before `createNewProjectInternal`.  Let me know if you prefer it back down after `createNewProjectInternal`.

alexweininger (approved):

nturinski (approved):

PR #3422, opened 2022-11-09
Set always on = on when creating Azure function

### LINKED ISSUE #3037: Azure function created with always on = off
- VS Code Version: latest
- Azure Account Extension Version:
- OS Version: win 10

Steps to Reproduce:

1. Create an azure function via VS Code extension.
2. Deploy.
3. Login into Azure Portal and find the created function.
4. Go to Configuration -> General Settings.

Result: Always on = off
image

motm32:
Fixes #3037. 

Only applicable to app service plans.

### DIFF
--- a/src/commands/createFunctionApp/FunctionAppCreateStep.ts
+++ b/src/commands/createFunctionApp/FunctionAppCreateStep.ts
@@ -80,11 +80,13 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
-        const isNotFree = context.plan?.sku?.family?.toLowerCase() !== 'u';
-        const isNotElasticPremium = context.plan?.sku?.tier?.toLowerCase() !== 'ep';
-        const isNotConsumption: boolean = context.plan?.sku?.family?.toLowerCase() !== 'y';
-        if (isNotFree && isNotElasticPremium && isNotConsumption) {
-            nonNullProp(site, 'siteConfig').alwaysOn = true;
+        if (context.plan?.sku?.family !== undefined) {
+            const isNotFree = context.plan?.sku?.family?.toLowerCase() !== 'u';
+            const isNotElasticPremium = context.plan?.sku?.family?.toLowerCase() !== 'ep';
+            const isNotConsumption: boolean = context.plan?.sku?.family?.toLowerCase() !== 'y';
+            if (isNotFree && isNotElasticPremium && isNotConsumption) {
+                nonNullProp(site, 'siteConfig').alwaysOn = true;
+            }
         }
 
         return site;
--- a/src/commands/createFunctionApp/FunctionAppCreateStep.ts
+++ b/src/commands/createFunctionApp/FunctionAppCreateStep.ts
@@ -80,6 +80,7 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
+        // Always on setting added for App Service plans excluding the free tier https://github.com/microsoft/vscode-azurefunctions/issues/3037
         if (context.plan?.sku?.family !== undefined) {
             const isNotFree = context.plan.sku.family.toLowerCase() !== 'u';
             const isNotElasticPremium = context.plan.sku.family.toLowerCase() !== 'ep';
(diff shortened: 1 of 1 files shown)


src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -80,6 +80,13 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
+        const isNotFree = context.plan?.sku?.family?.toLowerCase() !== 'u';
```
bwateratmsft, line 83: This applies to all three. If the `family` / `tier` are `undefined`, these booleans will evaluate to `true`. Is that intentional?

motm32: Hmm I am not totally sure, but probably not since I think always on should only be set to on when an app service plan that is not in the free tier is chosen.

nturinski: Why are we excluding the free tier?

motm32: Based off what happens in Azure the free tier does not allow the allow on feature to be changed, so I wasn't sure if we would enable that automatically if users aren't able to change it. But that's the only reason I was excluding it for now. 

image

bwateratmsft: Maybe I'm totally off but I thought that you _couldn't_ have always-on functions on the free tier, and that was an inherent part of the "agreement" for free tier (i.e. we won't give you any idle CPU time for free).

alexweininger: > Maybe I'm totally off but I thought that you couldn't have always-on functions on the free tier, and that was an inherent part of the "agreement" for free tier (i.e. we won't give you any idle CPU time for free).

Correct, in order to enable always-on functions, it has to be running on top of an App Service Plan which isn't available for the free tier.

https://learn.microsoft.com/en-us/azure/azure-functions/dedicated-plan#always-on

nturinski: Is there an error when you try to enable it for a free sku? Or does it just ignore the setting?

If it does ignore it, then I think we might as well just always set it on for anything that's not a consumption plan.

motm32: This is the error I got when I tried to enable it for a free sku. 

image

src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -80,11 +80,13 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
-        const isNotFree = context.plan?.sku?.family?.toLowerCase() !== 'u';
-        const isNotElasticPremium = context.plan?.sku?.tier?.toLowerCase() !== 'ep';
-        const isNotConsumption: boolean = context.plan?.sku?.family?.toLowerCase() !== 'y';
-        if (isNotFree && isNotElasticPremium && isNotConsumption) {
-            nonNullProp(site, 'siteConfig').alwaysOn = true;
+        if (context.plan?.sku?.family !== undefined) {
+            const isNotFree = context.plan?.sku?.family?.toLowerCase() !== 'u';
```
bwateratmsft, line 84: Inside this `if` block you no longer need to use the `?.`, since `context.plan.sku.family` has already been determined to be defined. TypeScript also knows this so it won't complain.

src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -80,6 +80,15 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
+        if (context.plan?.sku?.family !== undefined) {
```
nturinski, line 83: Could you leave a little comment about what/why we're doing this? 😅

src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -80,6 +80,16 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
             this.addCustomLocationProperties(site, context.customLocation);
         }
 
+        // Always on setting added for App Service plans excluding the free tier https://github.com/microsoft/vscode-azurefunctions/issues/3037
+        if (context.plan?.sku?.family !== undefined) {
```
alexweininger, line 84: I believe we can remove the `!== undefined`

```suggestion
        if (context.plan?.sku?.family) {
```

nturinski (dismissed):

alexweininger (approved):

PR #3427, opened 2022-11-10
Default to English if the current language templates are not available

### LINKED ISSUE #3279: Fallback to default language if current language pack is not available in backup templates
For create function from functions template, but I got follow message, and failed.

it appears pop up with 'Failed to get latest templates: CloseBraceExpected near line "328", column "4"'
and output of Azure Functions is below

> 9:01:50: Failed to get latest templates: Retrieving templates timed out. Modify setting "azureFunctions.requestTimeout" if you want to extend the timeout.
> 9:01:50: Failed to get backup templates: ENOENT: no such file or directory, open '/Users/billy_k/.vscode/extensions/ms-azuretools.vscode-azurefunctions-1.7.4/resources/backupTemplates/java/~4/resources/Resources.ja-JP.json'
> 9:01:50: Error: Failed to get latest templates: Retrieving templates timed out. Modify setting "azureFunctions.requestTimeout" if you want to extend the timeout.

### Steps.
0. (install Japanese Language Pack for VisualStudio Code(ms-ceintl.vscode-language-pack-ja))
1. create workspace from workspace menu of azure functions 
2. select "Create Function" from add button of workspace menu

### Temporary Measure
Uninstall language pack, and restart VSCode

### Versions
VSCode: 1.70.0
Azure Functions Extension: 1.7.4

motm32:
Fixes #3279.

### DIFF
--- a/src/templates/CentralTemplateProvider.ts
+++ b/src/templates/CentralTemplateProvider.ts
@@ -21,7 +21,7 @@ import { IFunctionTemplate, TemplateCategory } from './IFunctionTemplate';
 import { ITemplates } from './ITemplates';
 import { getJavaVerifiedTemplateIds } from './java/getJavaVerifiedTemplateIds';
 import { JavaTemplateProvider } from './java/JavaTemplateProvider';
-import { getScriptResourcesLanguage } from './script/getScriptResourcesLanguage';
+import { english, getScriptResourcesLanguage } from './script/getScriptResourcesLanguage';
 import { getScriptVerifiedTemplateIds } from './script/getScriptVerifiedTemplateIds';
 import { IScriptFunctionTemplate } from './script/parseScriptTemplates';
 import { PysteinTemplateProvider } from './script/PysteinTemplateProvider';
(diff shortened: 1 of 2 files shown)


src/templates/CentralTemplateProvider.ts
```diff
@@ -258,12 +259,24 @@ export class CentralTemplateProvider implements Disposable {
...
+                context.telemetry.properties.templateSource = 'latest';
+                const result: ITemplates = await provider.getLatestTemplates(context, latestTemplateVersion);
+                await provider.cacheTemplateMetadata(latestTemplateVersion);
+                await provider.cacheTemplates(context);
+                return result;
+            }
+        } catch (error) {
+            // If we failed to get templates for a non-English language, try English
+            const language = provider.resourcesLanguage || getScriptResourcesLanguage();
+            if (language !== 'en-US') {
```
alexweininger, line 273: We should import and use this constant for this string since it's already declared.

https://github.com/microsoft/vscode-azurefunctions/blob/4d43353ad99c360a0ad09cd980b31986fe8e00d2/src/templates/script/getScriptResourcesLanguage.ts#L8

nturinski (commented):
I think this is good, but we're not covering the case where their internet may not work (proxy or CDN problems). So if both of these attempts fail, then when we get to backupTemplates, that will fail again with no default result.

src/templates/CentralTemplateProvider.ts
```diff
@@ -258,12 +259,26 @@ export class CentralTemplateProvider implements Disposable {
...
+                return result;
+            }
+        } catch (error) {
+            // If we failed to get Java templates for a non-English language, try English
+            if (provider.language === 'Java') {
+                const language = provider.resourcesLanguage || getScriptResourcesLanguage();
+                if (language !== english) {
+                    provider.resourcesLanguage = english;
+                    const result = provider.getLatestTemplates(context, latestTemplateVersion);
+                    const message: string = localize('languageTemplateWarning', 'The templates for the current language are not supported yet. Defaulting to English templates.');
```
nturinski, line 277: nit: I think it would be nice to include the `language` in the error message.
```suggestion
                    const message: string = localize('languageTemplateWarning', '{0} templates are currently not supported. Defaulting to English templates.', language);
```

nturinski: Don't hate me, but I think we should actually have this logic in `getScriptResourcesLanguage`. It seems like we could add a property to check to see if the project is Java or not and just default to English if it is.

motm32: 🫠  Haha ok I will move it!

nturinski (approved):
Hooray! From 25 lines to 6 lines of changes 🥳

Issue #3350, opened 2022-09-21
Having multiple create buttons is confusing


fiveisprime:
When a customer has both Functions and SWA, it's confusing to see this:

<img width="509" alt="Screen Shot 2022-09-21 at 9 06 09 AM" src="https://user-images.githubusercontent.com/1186948/191582968-d6abee11-de4d-4618-bda2-c99209151f89.png">

I know that the commands are contributed independently, but what's possible to simplify how this is exposed?

pamelafox:
Hm, it's more clear from the command menu as it's prefixed with the extension:

<img width="862" alt="MicrosoftTeams-image (1)" src="https://user-images.githubusercontent.com/297042/191583948-ed8a9a7f-d03b-4234-896b-c72cdf41aa69.png">

<img width="712" alt="MicrosoftTeams-image" src="https://user-images.githubusercontent.com/297042/191583954-2c77e1df-edcd-4a1c-958f-2a4a5434b4cc.png">

but I imagine that would be quite wordy for the dropdown, might get truncated?

I have to admit, I'm now confused (as a user) as to why Static Web Apps need to create HTTP functions, but that's because I haven't explored SWAs at all. I might have a better suggestion when I learn more about SWAs. 

cc @kjaymiller (who took those screenshots)

kjaymiller:
You can create a static web app from an azure function. I haven't tried this to see the difference but to my knowledge there are some javascript frameworks that can deploy a static site with a command. My assumption would be that this is to support that workflow.

alexweininger:
We've been brainstorming solutions to this issue over on https://github.com/microsoft/vscode-azureresourcegroups/issues/316, feel free to chime in.

<img src="https://media4.giphy.com/media/VbEuHLBUPQm55MyqJg/giphy.gif"/>

nturinski:
Temporarily, we can change the labels to be more explicit about which project it is adding to.

PR #3430, opened 2022-11-11
Remove 'Create Function' from + button

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3350

Nice to have you back, lighting bolt button. Since it's also doing double duty for `Create New Project...` and `Create Function...`, I think it makes sense to just extract back to its own button.

Should be a temporary measure until we get an actions menu implemented.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -356,13 +356,6 @@
             }
         ],
         "menus": {
-            "azureWorkspaceCreate": [
-                {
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

alexweininger:
:shipit:

PR #3431, opened 2022-11-12
Durable Functions with Backends - Part III

MicroFish91:

## Pull Request - Part 3
This PR covers removing the `skipForNow` connection pattern and also strips most of the Azure resource creation logic out of `Create Function`.


MicroFish91:
Just going to combine parts 3 and 4 together.

PR #3437, opened 2022-11-15
Skip all automated tests

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3413

For now, I'm skipping every test due to the some of the unit tests being broken.

Currently, long-running tests are broken for every extension that do testing against Azure so we'll be skipping those as well.

The priority is to fix the unit tests first, and then get the Azure tests working again.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
 gulpfile.ts
 .eslintrc.js
+test/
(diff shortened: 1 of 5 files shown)


alexweininger (approved):
Make sure to create an issue to remind us to fix these in the future

alexweininger (approved):

alexweininger (approved):

PR #3453, opened 2022-11-29
Add warning messages for Azure Functions Core Tools EOL and mismatched versions

### LINKED ISSUE #2985: Node: Prompt or detect upgrading users from azure-functions-core-tools@2 to azure-functions-core-tools@3 (or other major version)
Been a user of this extension for a long time (thanks!). I recently used it to create some new TypeScript/Node V3 Functions for a new project. When I went to run/debug the Functions app, I noticed it was saying I was running Functions Runtime v2 😮

It turns out, when I installed this extension, v2 was the major version. But now it's V3. Initially, it installed `azure-functions-core-tools@2` globally.  However, it wasn't detecting that I was _trying_ to run a V3 app yet my `azure-functions-core-tools` npm global package is outdated/doesn't match. AFAIK for Node apps, there's not an explicit declaration I'm using V3 in the `host.json` or anything, so in a way that kinda makes sense.

The command `Install Or Update Azure Function Core Tools` also doesn't appear to prompt me to upgrade. It only detects I'm on v2 currently and just updates to `@2` instead of `@3`.

It took me awhile to understand how this worked. I finally fixed it by installing `@3` myself globally 😄 

**Proposed Solution**

In my workflow, the best place this would have been addressed is when I was originally prompted to create a V3 Functions Node App. At that point, it could have checked to see if I was running outdated Core Tools and then initiated the update.

Alternatively, the next best place would be when I tried to run `func host start` (via Debugger). The functions runtime is set in my `.vscode/settings.json`:

```json
{
  "azureFunctions.deploySubpath": ".",
  "azureFunctions.postDeployTask": "npm install (functions)",
  "azureFunctions.projectLanguage": "TypeScript",
  "azureFunctions.projectRuntime": "~3",
  "debug.internalConsoleOptions": "neverOpen",
  "azureFunctions.preDeployTask": "npm prune (functions)"
}
```

Lastly, `Install or Update Azure Function Core Tools` could perform this check but I only discovered that looking at the Extension Points.

I don't know how this would be addressed if running a V2 app. There is only one globally installed Core Tools npm package (at the moment) so another possible enhancement might be to find a way to detect it and choose the version.

**Extension Output**

When running `Install Or Update Azure Function Core Tools` when I was on v2.

```sh
9:24:04 PM: Running command: "npm install -g azure-functions-core-tools@2"...
C:\Program Files\nodejs\func -> C:\Program Files\nodejs\node_modules\azure-functions-core-tools\lib\main.js
C:\Program Files\nodejs\azfun -> C:\Program Files\nodejs\node_modules\azure-functions-core-tools\lib\main.js
C:\Program Files\nodejs\azurefunctions -> C:\Program Files\nodejs\node_modules\azure-functions-core-tools\lib\main.js

> azure-functions-core-tools@2.7.3188 postinstall C:\Program Files\nodejs\node_modules\azure-functions-core-tools
> node lib/install.js

attempting to GET "https://functionscdn.azureedge.net/public/2.7.3188/Azure.Functions.Cli.win-x64.2.7.3188.zip"

+ azure-functions-core-tools@2.7.3188
updated 1 package in 28.29s
9:24:34 PM: Finished running command: "npm install -g azure-functions-core-tools@2".
```

### LINKED ISSUE #3346: Show warning/notification to customers deploying to V3 Function apps to upgrade as V3 will be EOL by December
image

motm32:
Fixes #2985. 
Fixes #3346.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -5,18 +5,19 @@
 
 import { SiteConfigResource } from '@azure/arm-appservice';
 import { deploy as innerDeploy, getDeployFsPath, getDeployNode, IDeployContext, IDeployPaths, showDeployConfirmation } from '@microsoft/vscode-azext-azureappservice';
-import { DialogResponses, IActionContext } from '@microsoft/vscode-azext-utils';
+import { DialogResponses, IActionContext, nonNullValue, openUrl } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 import { deploySubpathSetting, functionFilter, ProjectLanguage, remoteBuildSetting, ScmType } from '../../constants';
 import { ext } from '../../extensionVariables';
-import { addLocalFuncTelemetry } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
+import { addLocalFuncTelemetry, getLocalFuncCoreToolsVersion } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
+import { FuncVersion, funcVersionLink, tryParseFuncVersion } from '../../FuncVersion';
 import { localize } from '../../localize';
 import { ResolvedFunctionAppResource } from '../../tree/ResolvedFunctionAppResource';
 import { SlotTreeItem } from '../../tree/SlotTreeItem';
 import { dotnetUtils } from '../../utils/dotnetUtils';
 import { isPathEqual } from '../../utils/fs';
 import { treeUtils } from '../../utils/treeUtils';
-import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
+import { getWorkspaceSetting, updateGlobalSetting } from '../../vsCodeConfig/settings';
 import { verifyInitForVSCode } from '../../vsCodeConfig/verifyInitForVSCode';
 import { tryGetFunctionProjectRoot } from '../createNewProject/verifyIsProject';
 import { notifyDeployComplete } from './notifyDeployComplete';
@@ -60,6 +61,35 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     if (language === ProjectLanguage.Python && !node.site.isLinux) {
         context.errorHandling.suppressReportIssue = true;
         throw new Error(localize('pythonNotAvailableOnWindows', 'Python projects are not supported on Windows Function Apps. Deploy to a Linux Function App instead.'));
+    }
+
+    const showCoreToolsWarningKey: string = 'showCoreToolsWarning';
+    if (getWorkspaceSetting<boolean>(showCoreToolsWarningKey)) {
+        const coreToolsVersion = await getLocalFuncCoreToolsVersion(context, undefined)
+        if (coreToolsVersion !== null) {
+            const localVersion = nonNullValue(tryParseFuncVersion(coreToolsVersion), 'localCoreToolsVersion');
+            let result: vscode.MessageItem | undefined;
+
+            if (localVersion === FuncVersion.v2 || localVersion === FuncVersion.v3) {
+                result = await showCoreToolsEOLWarning(context, localVersion);
+                while (result === DialogResponses.learnMore) {
+                    await openUrl(funcVersionLink);
+                    result = await showCoreToolsEOLWarning(context, localVersion);
+                }
+                if (result === DialogResponses.dontWarnAgain) {
+                    await updateGlobalSetting(showCoreToolsWarningKey, false);
+                }
+            } else if (localVersion !== version) {
+                result = await showCoreToolsMismatchWarning(context, localVersion, version);
+                while (result === DialogResponses.learnMore) {
+                    await openUrl(funcVersionLink);
+                    result = await showCoreToolsMismatchWarning(context, localVersion, version);
+                }
+                if (result === DialogResponses.dontWarnAgain) {
+                    await updateGlobalSetting(showCoreToolsWarningKey, false);
+                }
+            }
+        }
     }
 
     const client = await node.site.createClient(actionContext);
@@ -160,4 +190,23 @@ async function validateGlobSettings(context: IActionContext, fsPath: string): Pr
         const message: string = localize('globSettingRemoved', '"{0}" and "{1}" settings are no longer supported. Instead, place a ".funcignore" file at the root of your repo, using the same syntax as a ".gitignore" file.', includeKey, excludeKey);
         await context.ui.showWarningMessage(message, { stepName: 'globSettingRemoved' });
     }
+}
+
+async function showCoreToolsEOLWarning(context: IActionContext, localVersion: FuncVersion): Promise<vscode.MessageItem> {
+    const message = localize(
+        'outdatedFunctionRuntime',
+        'Your Azure Functions Core Tools Version ({0}) is past its end of life. Update to the latest version for the best experience.',
+        localVersion
+    );
+    return await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain)
+}
+
+async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion): Promise<vscode.MessageItem> {
+    const message = localize(
+        'mismatchedFunctionRuntime',
+        'Your Azure Functions Core Tools Version ({0}) does not match your Azure Functions runtime version ({1}). Update to the latest version for the best experience.',
+        localVersion,
+        projectVersion
+    );
+    return await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain)
 }
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -5,23 +5,23 @@
 
 import { SiteConfigResource } from '@azure/arm-appservice';
 import { deploy as innerDeploy, getDeployFsPath, getDeployNode, IDeployContext, IDeployPaths, showDeployConfirmation } from '@microsoft/vscode-azext-azureappservice';
-import { DialogResponses, IActionContext, nonNullValue, openUrl } from '@microsoft/vscode-azext-utils';
+import { DialogResponses, IActionContext } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
 import { deploySubpathSetting, functionFilter, ProjectLanguage, remoteBuildSetting, ScmType } from '../../constants';
 import { ext } from '../../extensionVariables';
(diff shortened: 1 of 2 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -62,6 +63,35 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         throw new Error(localize('pythonNotAvailableOnWindows', 'Python projects are not supported on Windows Function Apps. Deploy to a Linux Function App instead.'));
     }
 
+    const showCoreToolsWarningKey: string = 'showCoreToolsWarning';
```
bwateratmsft, line 66: I think a `Memento` is a better place for this, i.e. `ext.context.globalState`. This would not sync across machines (by default) (which I think is the _correct_ behavior, different machines likely have different versions), and would also not pollute user `settings.json`.

nturinski: This isn't a reason not to use the `ext.context.globalState` but `showCoreToolsWarning` is setting that already exists that Megan was leveraging. The idea was that if someone had already ignored the warning when their version was out of date, then it would remained ignored.

Also, when using `ext.context.globalState`, it can be difficult for users to "un-Don't Show Again" in the event of a mis-click whereas the setting is pretty easy to discover/delete since it'll show up in SCM.

bwateratmsft: Ah, I didn't realize a setting already existed. I think that's reasonable then.

src/commands/deploy/deploy.ts
```diff
@@ -62,6 +63,35 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
...
+    if (getWorkspaceSetting<boolean>(showCoreToolsWarningKey)) {
```
nturinski, line 67: I would prefer that you move all of this code in a separate file. Maybe something like `showCoreToolsWarning.ts` that also exists in the deploy folder.

`deploy.ts` is already complicated code to look at so moving this out can help new viewers digest the code a lot easier.

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -0,0 +1,61 @@
...
+            const localVersion = nonNullValue(tryParseFuncVersion(coreToolsVersion), 'localCoreToolsVersion');
+            let result: MessageItem | undefined;
+
+            if (localVersion === FuncVersion.v2 || localVersion === FuncVersion.v3) {
+                result = await showCoreToolsEOLWarning(context, localVersion);
+                while (result === DialogResponses.learnMore) {
+                    await openUrl(funcVersionLink);
+                    result = await showCoreToolsEOLWarning(context, localVersion);
+                }
+                if (result === DialogResponses.dontWarnAgain) {
```
nturinski, line 27: I believe that you can consolidate this with L32. If you move it out of the conditional and then just check the result at the end (as in on L40), you should only have to check once.

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -0,0 +1,61 @@
...
+                    await updateGlobalSetting(showCoreToolsWarningKey, false);
+                }
+            } else if (localVersion !== projectVersion) {
+                result = await showCoreToolsMismatchWarning(context, localVersion, projectVersion);
+                while (result === DialogResponses.learnMore) {
```
nturinski, line 32: You should be able to put this logic in the `showCoreTools...Warning` methods. I think that'll clean up this conditional a little bit and make it a bit easier to read :)

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -0,0 +1,61 @@
...
+                }
+                if (result === DialogResponses.dontWarnAgain) {
+                    await updateGlobalSetting(showCoreToolsWarningKey, false);
+                }
+            }
+        }
+    }
+}
+
+async function showCoreToolsEOLWarning(context: IActionContext, localVersion: FuncVersion): Promise<MessageItem> {
```
nturinski, line 44: If you add the while loop logic to this function, I think it'd make sense to have another function that has all the showWarning logic that gets called by these two functions.

It's a bit confusing to explain, but what I mean is something along the lines of this:

```
async function showCoreToolsWarning(context: IActionContext, message: string): Promise<MessageItem> {
   let result;
   while(result === DialogResponse.learnMore) {
      result = await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
   }

   if (result === DialogResponses.dontWarnAgain) {
      await updateGlobalSetting(showCoreToolsWarningKey, false);
   }
   return result;
}

async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion): Promise<MessageItem> {
    const message = localize(
        'mismatchedFunctionRuntime',
        'Your Azure Functions Core Tools Version ({0}) does not match your Azure Functions runtime version ({1}). Update to the latest version for the best experience.',
        localVersion,
        projectVersion
    );
   return await showCoreToolsWarning(context, message)
}
```

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -0,0 +1,61 @@
...
+        'Your Azure Functions Core Tools Version ({0}) is past its end of life. Update to the latest version for the best experience.',
+        localVersion
+    );
+    return await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain)
+}
+
+async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion): Promise<MessageItem> {
+    const message = localize(
+        'mismatchedFunctionRuntime',
+        'Your Azure Functions Core Tools Version ({0}) does not match your Azure Functions runtime version ({1}). Update to the latest version for the best experience.',
```
nturinski, line 56: I think we should be more specific with this error. Maybe something along the lines of 
> The local Azure Functions Core Tools version ({0}) does not match function app "{1}"'s Azure Functions runtime version ({2}). Ensure that the versions match for the best experience.

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -10,52 +10,52 @@ import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVer
 import { localize } from "../../localize";
 import { getWorkspaceSetting, updateGlobalSetting } from "../../vsCodeConfig/settings";
 
-export async function showCoreToolsWarning(context: IActionContext, projectVersion: FuncVersion): Promise<void> {
+export async function showCoreToolsWarning(context: IActionContext, projectVersion: FuncVersion, projectName: string): Promise<void> {
```
nturinski, line 13: nit: Technically, this isn't the project (which refers to the local project) but the function app itself.

```suggestion
export async function showCoreToolsWarning(context: IActionContext, projectVersion: FuncVersion, siteName: string): Promise<void> {
```

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -10,52 +10,52 @@ import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVer
...
-    return await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain)
+    return await showCoreToolsWarningHelper(context, message);
 }
 
-async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion): Promise<MessageItem> {
+async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion, projectName: string): Promise<MessageItem> {
     const message = localize(
         'mismatchedFunctionRuntime',
-        'Your Azure Functions Core Tools Version ({0}) does not match your Azure Functions runtime version ({1}). Update to the latest version for the best experience.',
+        'The local Azure Functions Core Tools Version ({0}) does not match the Azure Functions runtime version ({2}) of function app "{1}". Ensure that the versions match for the best experience.',
```
nturinski, line 55: nit: lol it's so confusing to have the numbers not be in order. That being said, you listed them correctly in the function call, so I won't bother you to change it 😋

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -10,52 +10,52 @@ import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVer
...
         'outdatedFunctionRuntime',
         'Your Azure Functions Core Tools Version ({0}) is past its end of life. Update to the latest version for the best experience.',
         localVersion
     );
-    return await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain)
+    return await showCoreToolsWarningHelper(context, message);
 }
 
-async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion): Promise<MessageItem> {
+async function showCoreToolsMismatchWarning(context: IActionContext, localVersion: FuncVersion, projectVersion: FuncVersion, projectName: string): Promise<MessageItem> {
```
nturinski, line 52: nit: `projectVersion` is technically more the `runtimeVersion`. I had a similar comment as to why that's kind of confusing above.

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -10,52 +10,52 @@ import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVer
...
+                await showCoreToolsMismatchWarning(context, localVersion, projectVersion, projectName);
             }
         }
     }
 }
 
+async function showCoreToolsWarningHelper(context: IActionContext, message: string): Promise<MessageItem> {
+    let result: MessageItem;
+    const showCoreToolsWarningKey: string = 'showCoreToolsWarning';
+    result = await context.ui.showWarningMessage(message, DialogResponses.learnMore, DialogResponses.dontWarnAgain);
```
nturinski, line 32: nit: if you follow this design pattern, it's a little bit cleaner:
https://github.com/microsoft/vscode-azurefunctions/blob/main/src/funcCoreTools/validateFuncCoreToolsIsLatest.ts#L91-L102

Not really necessary, but either way you can remove `let result: MessageItem;` declaration and just declare and assign it here.

src/commands/deploy/showCoreToolsWarning.ts
motm32: Lol I had it originally the way you had it in the comment, but decided to change it and forgot to change the numbers 🫠

nturinski (approved):
🧼🫧🧼
SafemoonMrCleanGIF

Issue #3285, opened 2022-08-15
Support new Node.js Programming Model


ejizba:
Related to https://github.com/Azure/azure-functions-nodejs-worker/issues/480

We will basically want the same functionality as the new Python programming model work Phil is doing in https://github.com/microsoft/vscode-azurefunctions/pull/3235

Quick summary:
1. There will be new templates for projects/functions. The biggest highlight is there will be no more "function.json" file
2. Because there are no "function.json" files, a few features will have to change like listing functions in the UI

ejizba:
A few updates. First this wiki has background and instructions for trying out the new model: https://aka.ms/AzFuncNodeV4

Now here is a tiny spec on how I see these features working:

## Create project

Here are the key differences from old projects:

- The `@azure/functions` package needs to be in the `dependencies` section of `package.json` (not `devDependencies`) and reference `^4.0.0-alpha.2` as the version
- The "main" field needs to be set in package.json. There's a lot of options for this, but I propose `dist/src/functions/*.js` for TypeScript and `src/functions/*.js` for JavaScript. That will allow us to create a new file for each function which simplifies some of the tooling logic and IMO is a more realistic folder structure for an app than say putting everything in one file. Open to suggestions on this
- `@types/node` should reference v18
- In host.json the bundle should be `"version": "[3.15.0, 4.0.0)"`

## Create function

I have a first draft of templates [here](https://gist.github.com/ejizba/4780c79193acf5e0082a9bdb50bde4a3) based on [this branch](https://github.com/Azure/azure-functions-templates/tree/ej/new-nodejs). The new templates are suffixed with "-4.x" in the id. Unlike Python, this is essentially the 4th version of our programming model - it just happens that the first 3 didn't have any major changes that would've affected tooling templates. Just like Python, I think this needs to be hard-coded into the extension for now because it may take us time to decide on an official new schema for these models.

The existing wizard should remain the exact same. However, the result will be different
- The function name should be used to replace the `%functionName%` string in both the file name and file contents
- The binding parameters (for example "connection" or "queueName") should replace the relevant string (`%connection%` or `%queueName%`) in the file contents
- The file should be put in the `src/functions` folder. This assumes a project setup like above. I think this is fine for public preview, but we could try and parse the users "package.json" "main" field if we want this to be smarter

[...]

PR #3456, opened 2022-12-01
Node.js Programming Model (Preview)

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3285

@ejizba Please review at your leisure!

A few known issues:
- Since the new programming model isn't available to roll out on Azure, keeping this feature behind a feature flag.
- Any name/file with `NodeProgrammingModel...` is because I wasn't creative enough to think of anything better. Is there a better way that I should denote this? Maybe just `NodejsV4...`?
- The template provider is relying solely on the backup templates for now. Will have to change this whenever the templates are available from the feed.
- Still need to support running/adding functions to a single index.js file. This will change a lot of logic in getting a unique name, how to start debugging, and adding new functions so figured it'd be worth doing in another PR.


package.json
```diff
@@ -1035,6 +1035,17 @@
                         "type": "boolean",
                         "description": "%azureFunctions.endOfLifeWarning%",
                         "default": true
+                    },
+                    "azureFunctions.functionSubpath": {
+                        "type": "string",
+                        "default": "src/functions",
+                        "description": "%azureFunctions.functionSubpath%"
+                    },
+                    "azureFunctions.showNodeProgrammingModel": {
```
bwateratmsft, line 1044: Should this be done as an experiment instead?

src/utils/pythonUtils.ts
```diff
@@ -8,3 +8,8 @@ import { ProjectLanguage } from "../constants";
 export function isPythonV2Plus(language: string | undefined, model: number | undefined): boolean {
     return language === ProjectLanguage.Python && model !== undefined && model > 1;
 }
+
+export function isNodeV4Plus(language: string | undefined, model: number | undefined): boolean {
```
bwateratmsft, line 12: This file should be renamed if this method is going to stay here

src/utils/bundleFeedUtils.ts
```diff
@@ -100,4 +100,11 @@ export namespace bundleFeedUtils {
 
         return feedUtils.getJsonFeed(context, url);
     }
+
+    export function overwriteExtensionBundleVersion(hostJson: IHostJsonV2, expectedRange: string, newRange: string): void {
+        if (hostJson.extensionBundle) {
+            if (hostJson.extensionBundle.version === expectedRange)
```
bwateratmsft, line 106: Can we use braces here? Or change the condition in the outer `if` to `hostJson.extensionBundle?.version === expectedRange` and remove the inner `if`? Non-block-style `if`s too often lead to hard-to-spot bugs...

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -82,6 +88,8 @@ __azurite_db*__.json`));
         };
 
         await bundleFeedUtils.addDefaultBundle(context, hostJson);
+        if (context.languageModel === nodejsNewModelVersion)
```
bwateratmsft, line 91: Can we use braces here? Non-block-style `if`s too often lead to hard-to-spot bugs...

src/commands/createNewProject/ProgrammingModelStep.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const options: QuickPickOptions = { placeHolder: localize('selectLanguage', 'Select a {0} programming model', context.language) };
+        const result = (await context.ui.showQuickPick(modelsPick, options)).data;
+
+        context.languageModel = result;
+    }
+
+    public shouldPrompt(context: IProjectWizardContext): boolean {
+        // this only impacts node.js for now so only check the feature flag for node.js
+        return context.languageModel === undefined && !!getWorkspaceSetting(nodeProgrammingModelSetting);
```
bwateratmsft, line 39: Should there be an additional check to see if there is more than one item in `this._models`? The prompt is unnecessary if there's only one, right?

package.json
nturinski: Similar to the Python v2 model, we're using the flag because the local experience totally works, but the Azure rollout isn't complete yet. Once that happens, we can remove the flag, but we wanted to give users/internals a chance to use the local experience and offer feedback.

bwateratmsft: The same can be accomplished with a setting-based experiment, though. Users _can_ override the experimental setting that they are assigned, FWIW.

nturinski: Fair enough! And then we could remove it without a release. I'm not sure how much effort it really is to pilot an experiment, but if you think it's worth it, let's do it.

I'll definitely need some assistance with that though.

bwateratmsft: I guess the best way to answer that is, do we want more or less people to see the feature? If we go with off-by-default and no experiment, then very few folks will discover it. An experiment could greatly increase the number of folks that see the feature.

bwateratmsft: Yeah. And naturally, you can control not just the exposure, but _when_ you turn it on. So we could leave it all hidden (including the setting) until we're ready. You can also turn it off if things go sideways.

src/commands/createFunction/scriptSteps/NodeProgrammingModelFunctionCreateStep.ts
```diff
@@ -0,0 +1,38 @@
...
+import { AzExtFsExtra, nonNullProp } from '@microsoft/vscode-azext-utils';
+import * as path from 'path';
+import { functionSubpathSetting, ProjectLanguage } from '../../../constants';
+import { IScriptFunctionTemplate } from '../../../templates/script/parseScriptTemplates';
+import { getWorkspaceSetting } from '../../../vsCodeConfig/settings';
+import { FunctionCreateStepBase } from '../FunctionCreateStepBase';
+import { IScriptFunctionWizardContext } from './IScriptFunctionWizardContext';
+import { getScriptFileNameFromLanguage } from './ScriptFunctionCreateStep';
+
+export class NodeProgrammingModelFunctionCreateStep extends FunctionCreateStepBase<IScriptFunctionWizardContext> {
```
ejizba, line 15: > Any name/file with NodeProgrammingModel... is because I wasn't creative enough to think of anything better. Is there a better way that I should denote this? Maybe just NodejsV4...?

I would go with `NodeV4`

src/constants.ts
```diff
@@ -108,10 +110,22 @@ export const viewOutput: string = localize('viewOutput', 'View Output');
 export const previewDescription: string = localize('preview', '(Preview)');
 export const pythonNewModelPreview: string = localize('pythonNewModelPreview', 'Python (Programming Model V2)');
 
+/**
+ * The "current" Node.js model is 3 (and assumed, if the number is omitted).
+ * The new Node.js model is 4.
+ * Any significantly changed new model should use an incremented number.
+ */
+export const nodejsNewModelVersion: number = 4;
+export const nodejsNewModelLabel: string = localize('nodejsNewModelPreview', 'Programming Model V4 (Preview)')
```
ejizba, line 119: Screen Shot 2022-12-01 at 4 02 22 PM

A few suggestions:
1. Can we remove "Programming Model" from both picks? I just don't like the duplicated stuff
2. I'd remove "(Default)" from v3. I think it looks too similar to "(Preview)" and I think "preview" is more important because people can deduce "v3" is the default since it's at the top
3. Can you add `https://aka.ms/AzFuncNodeV4` as a learn more link

src/commands/createFunction/scriptSteps/NodeProgrammingModelFunctionCreateStep.ts
```diff
@@ -0,0 +1,38 @@
...
+            for (const setting of template.userPromptedSettings) {
+                // the setting name keys are lowercased
+                contents = contents.replace(new RegExp(`%${setting.name}%`, 'g'), context[setting.name.toLowerCase()]);
+            }
+
+            await AzExtFsExtra.writeFile(path.join(functionPath, filename), contents);
+        }));
+
+        const language: ProjectLanguage = nonNullProp(context, 'language');
+        const fileName = getScriptFileNameFromLanguage(language) as string;
```
ejizba, line 35: This doesn't work. It's using `index.js` instead of something like `httpTrigger1.js`. The consequence is that the new function file isn't being opened in the editor after it's created

src/commands/createFunction/scriptSteps/NodeProgrammingModelFunctionNameStep.ts
```diff
@@ -0,0 +1,33 @@
...
+        const template: IScriptFunctionTemplate = nonNullProp(context, 'functionTemplate');
+        const functionSubpath: string = getWorkspaceSetting(functionSubpathSetting, context.projectPath) as string;
+        return await this.getUniqueFsPath(
+            path.join(context.projectPath, functionSubpath),
+            template.defaultFunctionName,
+            context.language === ProjectLanguage.TypeScript ? '.ts' : '.js');
+    }
+
+    protected async validateFunctionNameCore(context: IScriptFunctionWizardContext, name: string): Promise<string | undefined> {
+        if (await AzExtFsExtra.pathExists(path.join(context.projectPath, name))) {
```
ejizba, line 27: This needs to take into account the `functionSubpath`

src/commands/createFunction/FunctionSubWizard.ts
```diff
@@ -9,7 +9,7 @@ import { canValidateAzureWebJobStorageOnDebug } from '../../debug/validatePreDeb
 import { getAzureWebJobsStorage } from '../../funcConfig/local.settings';
```
ejizba, line 9: Can we hide the OpenAPI one at the bottom for the new model? I'm almost certain it's not compatible

Screen Shot 2022-12-01 at 4 23 42 PM

src/constants.ts
nturinski: I assume that you want an actual "learn more" quick pick item appended to the bottom, but just in case, did you mean/want this little ? button?

image

ejizba: Yeah I wanted a real quick pick. Not sure i'd ever see that little butterino...

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -136,7 +136,10 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
                 data: <IFunctionTemplate | TemplatePromptResult><unknown>undefined,
                 onPicked: () => { /* do nothing */ }
             })
-        } else if (language === ProjectLanguage.CSharp || language === ProjectLanguage.Java || (language === ProjectLanguage.Python && !isPythonV2Plus(language, languageModel)) || language === ProjectLanguage.TypeScript) {
+        } else if (language === ProjectLanguage.CSharp ||
+            language === ProjectLanguage.Java ||
+            (language === ProjectLanguage.Python && !isPythonV2Plus(language, languageModel)) ||
+            (language === ProjectLanguage.TypeScript) && !isNodeV4Plus(context)) {
```
ejizba, line 142: Not sure it makes a difference, but I think you meant to do this:
```suggestion
            (language === ProjectLanguage.TypeScript && !isNodeV4Plus(context))) {
```

package.json
```diff
@@ -1035,6 +1035,17 @@
                         "type": "boolean",
                         "description": "%azureFunctions.endOfLifeWarning%",
                         "default": true
+                    },
+                    "azureFunctions.functionSubpath": {
```
ejizba, line 1039: This one should be `"scope": "resource"` and the other one should not. Basically you want that if the setting can be specific to a folder/project and otherwise you don't need it.

package.nls.json
```diff
@@ -101,5 +101,7 @@
     "azureFunctions.validateFuncCoreTools": "Validate the Azure Functions Core Tools is installed before debugging.",
     "azureFunctions.viewCommitInGitHub": "View Commit in GitHub",
     "azureFunctions.viewDeploymentLogs": "View Deployment Logs",
-    "azureFunctions.viewProperties": "View Properties"
+    "azureFunctions.viewProperties": "View Properties",
+    "azureFunctions.functionSubpath": "The default subpath to create new functions.",
+    "azureFunctions.showNodeProgrammingModel": "Enable Node.js (New Model Preview)"
```
ejizba, line 106: "Enable Node.js" makes it sound like we don't support Node.js at all without the setting haha

```suggestion
    "azureFunctions.showNodeProgrammingModel": "Enable preview Node.js programming model"
```

package.nls.json
```diff
@@ -101,5 +101,7 @@
     "azureFunctions.validateFuncCoreTools": "Validate the Azure Functions Core Tools is installed before debugging.",
     "azureFunctions.viewCommitInGitHub": "View Commit in GitHub",
     "azureFunctions.viewDeploymentLogs": "View Deployment Logs",
-    "azureFunctions.viewProperties": "View Properties"
+    "azureFunctions.viewProperties": "View Properties",
+    "azureFunctions.functionSubpath": "The default subpath to create new functions.",
```
ejizba, line 105: How about adding this:
```suggestion
    "azureFunctions.functionSubpath": "The default subpath to create new functions. Currently, this only applies to Node.js programming model v4+.",
```

src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
```diff
@@ -50,8 +45,47 @@ export class JavaScriptProjectCreateStep extends ScriptProjectCreateStep {
...
+        return {
+            start: 'func start',
+            test: 'echo \"No tests yet...\"'
+        }
+    }
+
+    protected getPackageJsonDeps(context: IProjectWizardContext): { [key: string]: string } {
+        const deps: { [key: string]: string } = {};
+        if (isNodeV4Plus(context))
+            deps[azureFunctionsDependency] = azureFunctionsDependencyVersion;
```
ejizba, line 77: I'm with Brandon - get those braces added you hooligan

src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
```diff
@@ -50,8 +45,47 @@ export class JavaScriptProjectCreateStep extends ScriptProjectCreateStep {
...
+
+        return deps;
+    }
+
+    protected getPackageJsonDevDeps(context: IProjectWizardContext): { [key: string]: string } {
+        const devDeps: { [key: string]: string } = {};
+        if (isNodeV4Plus(context)) {
+            devDeps['@types/node'] = '^18.0.0';
```
ejizba, line 85: This is only needed for TypeScript projects, not JavaScript

src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
bwateratmsft: inb4 new Heartbleed

src/commands/createNewProject/ProjectCreateStep/JavaScriptProjectCreateStep.ts
nturinski: Would `^18.x` work for the node version for TypeScript?

ejizba: Yeah I think `^18.x` and `^18.0.0` are pretty much the same.

ejizba (dismissed):
yay

nturinski:
image

ejizba (approved):
Lol did it do that automatically or did you do it on accident?

nturinski:
Yeeeeah... new repo policy that we must abide by 😭

PR #3459, opened 2022-12-01
Improve Core Tools install experience for Linux

### LINKED ISSUE #2745: Improve Core Tools install experience in Linux
This is primarily feedback from running in a Codespace but might apply to Linux in general.

Currently customers are sent to the GitHub readme which has a [list of commands](https://github.com/Azure/azure-functions-core-tools#linux):

image

Some ideas for improvement:

* In the prompt, can we tell the customer what distro and version they're running?
* Can we detect the distro and version, and send them to a dedicated page/location with very specific instructions? (Functions team can maintain this)
* If npm is available, as is the case of the Codespaces kitchen sink container, can we provide a new default option, which is to install via npm? If the install fails, then send them to the docs (or tell them to try rerunning the command using sudo).

MicroFish91:
Closes #2745 

Ideas for improvements addressed from original issue:
- Idea 1: In the prompt, can we tell the customer what distro and version they're running?

   Solution:  Yes, if we either a) have no package manager available, or b) fail to install via package manager, the distro and version will be shown in the notice to the user

- Idea 3:  If npm is available, as is the case of the Codespaces kitchen sink container, can we provide a new default option, which is to install via npm? If the install fails, then send them to the docs (or tell them to try rerunning the command using sudo).

  Solution:  Pretty much implemented a change that is doing this exact thing

Note: For figuring out the Linux distro and version, I have defaulted to using the `/etc/os-release` file as it seems like this file is supported in the majority of recent Linux distributions

See:  https://gist.github.com/natefoo/814c5bf936922dad97ff

If no release file is found, it will still work; it will just show less info to the user.


src/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -44,9 +48,28 @@ export async function validateFuncCoreToolsInstalled(context: IActionContext, me
                 items.push(DialogResponses.learnMore);
             }
 
+            if (os.platform() === 'linux') {
+                const linuxDistroTag: LinuxDistroTag | undefined = await getLinuxDistroTag();
+                if (linuxDistroTag) {
+                    let linuxMessage: string = '';
+                    if (linuxDistroTag['PRETTY_NAME']) {
```
nturinski, line 55: Does the pretty name include a version as well? Could you give examples of what the expected output is?

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -4,20 +4,24 @@
...
 import { getWorkspaceSetting } from '../vsCodeConfig/settings';
 import { getFuncCliPath, hasFuncCliSetting } from './getFuncCliPath';
 import { getFuncPackageManagers } from './getFuncPackageManagers';
+import { getLinuxDistroTag, LinuxDistroTag } from './getLinuxDistroTag';
 import { installFuncCoreTools } from './installFuncCoreTools';
 
 export async function validateFuncCoreToolsInstalled(context: IActionContext, message: string, workspacePath: string): Promise<boolean> {
     let input: MessageItem | undefined;
     let installed: boolean = false;
+    let failedInstall: string = localize('failedInstallFuncTools', 'Core Tools installation has failed and will need to be installed manually. Try re-running the last command with "sudo" or click "Learn more" for more options.');
```
nturinski, line 24: I think it could be confusing for the failure message to include
>Try re-running the last command with "sudo" or click "Learn more" for more options.

Could we just concatenate this at the end of the Linux failure scenario?

MicroFish91: We could but then I'll have to re-scope the variables above because they are not part of the same block.

So what exactly would you want the full verbiage for the failure message to look like?

nturinski: Could you do something like this on L52?
```
let linuxMessage = 'Try re-running the last command with "sudo" or click "Learn more" for more options.'
```

And then everything else would get appended as it currently does now.

MicroFish91: Example:

image

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
MicroFish91: Yeah, my observation has been that if there is both a version and pretty_name, the pretty_name will include the version.

See example below: https://github.com/microsoft/vscode-azurefunctions/pull/3459#discussion_r1037639930

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
nturinski: Yeah, I think that would be fine. Can't remember if there is anyway we can add line breaks to the notification messages.

MicroFish91: Oh I see what you mean, will do.

MicroFish91: > Yeah, I think that would be fine. Can't remember if there is anyway we can add line breaks to the notification messages. Would be nice after `Note:`

I tried, but VS Code seems to strip away new lines

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
MicroFish91: Also here is the default message when no package manager is detected on Linux

image

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
bwateratmsft: Can we put in a button to copy that command to the clipboard?

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -61,7 +85,8 @@ export async function validateFuncCoreToolsInstalled(context: IActionContext, me
 
     // validate that Func Tools was installed only if user confirmed
     if (input === install && !installed) {
-        if (await context.ui.showWarningMessage(localize('failedInstallFuncTools', 'The Azure Functions Core Tools installation has failed and will have to be installed manually.'), DialogResponses.learnMore) === DialogResponses.learnMore) {
+        ext.outputChannel.show();  // Make it easier for the user to find the last command in case they want to re-run with 'sudo'
+        if (await context.ui.showWarningMessage(failedInstall, DialogResponses.learnMore) === DialogResponses.learnMore) {
```
bwateratmsft, line 89: Adding the comment here too since the other thread was resolved;

Can we add a button to copy the command to the clipboard?

MicroFish91: Updated: 

image

src/funcCoreTools/installFuncCoreTools.ts
```diff
@@ -20,10 +22,12 @@ export async function installFuncCoreTools(context: IActionContext, packageManag
     switch (packageManagers[0]) {
         case PackageManager.npm:
             const distTag: INpmDistTag = await getNpmDistTag(context, version);
+            lastCoreToolsInstallCommand = `npm install -g ${funcPackageName}@${distTag.tag}`;
```
nturinski, line 25: Can you make this an array of args, and then use a spread operation on the cpUtils.executeCommand call?

```
lastCoreToolsInstallCommand = ['npm', 'install', '-g', `${funcPackageName}@${distTag.tag}`]
await cpUtils.executeCommand(ext.outputChannel, undefined, ...lastCoreToolsInstallCommand);
```

I think that should work. And then for the copy command, you can just do `lastCoreToolsInstallCommand.join(' ');`

Added benefit of copying the `brew tap` command as well in case that is the one that errored out instead of `brew install`

src/funcCoreTools/validateFuncCoreToolsInstalled.ts
```diff
@@ -44,9 +48,29 @@ export async function validateFuncCoreToolsInstalled(context: IActionContext, me
                 items.push(DialogResponses.learnMore);
             }
 
+            if (os.platform() === 'linux') {
```
nturinski, line 51: Could you extract this out to its own function?

Something like `generateLinuxrrorMessage`? It could return the string that should be appended to the `failedInstall` string.

It's just a lot of code in the middle of the validation that makes it harder to follow the flow of logic imo

nturinski (approved):

PR #3468, opened 2022-12-08
Durable Functions with Backends

### LINKED ISSUE #2956: Support new Durable Functions backends

MicroFish91:
🤞

Closes #2956


MicroFish91:
Note: Remove _'s from private methods.

https://github.com/microsoft/vscode-azurefunctions/pull/3392#pullrequestreview-1209272589

MicroFish91:
Todo:
- Move `validateConnection` type logic out of `durableUtils`.
- Remove debounce logic and consolidate async task logic into the new prompt methods introduced in utils

MicroFish91:
Now that Java Durable Orchestration is showing up, we might be able to get that included as well, time permitting...

For now `Maven` only.

src/commands/appSettings/connectionSettings/getLocalConnectionSetting.ts
```diff
@@ -0,0 +1,102 @@
...
+
+        authRuleName = (await context.ui.showQuickPick(manageAccessRules.map(authRule => {
+            return { label: nonNullProp(authRule, 'name'), suppressPersistence: true, description: authRule.name === rootKeyName ? defaultDescription : '' };
+        }), { placeHolder })).label;
+    }
+
+    const accessKeys: AccessKeys = await client.namespaces.listKeys(rgName, namespaceName, authRuleName);
+    if (!accessKeys.primaryConnectionString && !accessKeys.secondaryConnectionString) {
+        const learnMoreLink: string = 'https://aka.ms/event-hubs-connection-string';
+        const message: string = localize('missingEventHubsConnectionString', 'There are no connection strings available on your namespace\'s shared access policy. Locate a valid access policy and add the connection string to "{0}".', 'local.settings.json');
```
nturinski, line 70: nit: Use the constant for local settings json

nturinski, line 70: nit: Use the constant for local settings json

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -64,7 +66,12 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
                 promptSteps: [new PythonLocationStep(this._functionSettings)]
             };
         } else {
-            return await FunctionSubWizard.createSubWizard(context, this._functionSettings);
+            const needsDurableStorageSetup: boolean = !!context.functionTemplate && durableUtils.requiresDurableStorage(context.functionTemplate.id, context.language) && !context.hasDurableStorage;
```
nturinski, line 69: I think it's cleaner to move all the other conditions into the `requiresDurableStorage` function. If you just pass the entire context, you can do all of these checks in the function.

src/commands/createFunction/FunctionSubWizard.ts
```diff
@@ -94,9 +91,8 @@ export class FunctionSubWizard {
                 }
             }
 
-            if ((!template.isHttpTrigger && !template.isSqlBindingTemplate) && !canValidateAzureWebJobStorageOnDebug(context.language) && !await getAzureWebJobsStorage(context, context.projectPath)) {
-                promptSteps.push(new AzureWebJobsStoragePromptStep());
-                executeSteps.push(new AzureWebJobsStorageExecuteStep());
+            if (context.newDurableStorageType) {
```
nturinski, line 94: Wait, any idea why this was changed? I don't feel like this is equivalent.

nturinski, line 94: Wait, any idea why this was changed? I don't feel like this is equivalent.

MicroFish91: Because we removed the `skipForNow` pattern we no longer need to create these resources during `createFunction`.

MicroFish91: See above!

nturinski: Sorry for all the double comments. My internet was very spotty in the car

MicroFish91: Haha yeah I figured, no worries!

nturinski (approved):
Okay, took a bit to revisit everything. I think we're good.... I mean, some of this review was done months ago, so also trusting our pasts selves a bit here 🤣

MicroFish91:
Going to wait before merging until we get some final feedback after standup

PR #3472, opened 2022-12-13
Add .NET 7.0 to netVersions array

### LINKED ISSUE #3374: Pop up a notification to install the .NET Core SDK when creating a .NET 7 project
**OS:** M1 & Win10
**Build Version:** 20221015.2
**.NET Version:** 7.0.100-rc.2.22477.23

**Repro Steps:**
1. Make sure only .NET 7 installed.
2. F1 -> Azure Function: Create New Project... -> Select a folder ->Select C#.
3. Check whether continue to select .NET runtime.

**Expect:**
Continue to select .NET runtime.

**Actual:**
Pop up a notification to install the .NET Core SDK.
image

**More Info:**
1. More details as below after running `dotnet --list-runtimes` and `dotnet --list-sdks` in the terminal.
image
2. This issue doesn't reproduce when installing the .NET 6 at the same time.

motm32:
Fixes #3374.

### DIFF
--- a/src/templates/dotnet/executeDotnetTemplateCommand.ts
+++ b/src/templates/dotnet/executeDotnetTemplateCommand.ts
@@ -61,7 +61,7 @@ async function getFramework(context: IActionContext, workingDirectory: string | 
         }
 
         // Prioritize "LTS", then "Current", then "Preview"
-        const netVersions: string[] = ['7.0', '6.0', '5.0', '3.1'];
+        const netVersions: string[] = ['6.0', '7.0', '5.0'];
         const semVersions: SemVer[] = netVersions.map(v => semVerCoerce(v) as SemVer);
 
         let pickedVersion: SemVer | undefined;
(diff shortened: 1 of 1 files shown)


src/templates/dotnet/executeDotnetTemplateCommand.ts
```diff
@@ -61,7 +61,7 @@ async function getFramework(context: IActionContext, workingDirectory: string |
         }
 
         // Prioritize "LTS", then "Current", then "Preview"
-        const netVersions: string[] = ['6.0', '5.0', '3.1'];
+        const netVersions: string[] = ['7.0', '6.0', '5.0', '3.1'];
```
bwateratmsft, line 64: .NET Core 3.1 is EOL today. We may as well remove it. Also, 6.0 is LTS, so based on the comment shouldn't it be 6.0, 7.0, 5.0?

motm32: Ahh yes I will make that change and remove 3.1 😊

bwateratmsft (approved):

nturinski:
This doesn't actually fix the problem. We need to add .NET 7 to the JsonCliTool. You can test it by only having the .NET 7 SDK installed.

motm32:
Hmm ok I had tested it while only having the .NET 7 SDK installed and it seemed to load the templates correctly, but I will look into the JsonCliTool!

nturinski:
https://github.com/microsoft/vscode-azurefunctions/pull/2923/files

image

PR #3480, opened 2022-12-16
Fix error when executing a service bus queue trigger with a JSON request body

### LINKED ISSUE #3003: Executing a local service bus queue trigger gives "Unexpected status code: 400" error with JSON request body
### Issue

Passing in valid JSON into the request body of a service bus queue trigger results in a "Unexpected status code: 400" error, and the function does not execute.

Any valid JSON seems to cause this error; I tried passing in a JSON service bus message with the payload in `body` as a string and object but received the same error.

### Expected Behavior

When valid a valid JSON payload is entered in the service bus message payload prompt, the payload should be passed as a deserialized object to the service bus message trigger function argument. This is what appears to happen when the deployed function is triggered regardless of `ContentType`.

### Steps to Reproduce

1. Create function
2. Select "Azure Service Bus Queue trigger"
3. Select defaults for remaining prompts
4. Run `func start`
5. Execute function
6. Select created function
7. Enter valid JSON in request body prompt
8. Receive error

### Context

- Windows 10 OS
- TypeScript Function Project Runtime
- VSCode v1.62.2
- Azure Functions Extension v1.6.0
- Azure Functions Core Tools v3.0.3904
- Function Runtime v3.3.1.0

motm32:
Fixes #3003.

### DIFF
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -48,6 +48,7 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         const data: string = await context.ui.showInputBox({ prompt, value, stepName: 'requestBody' });
         try {
             functionInput = <{}>JSON.parse(data);
+            // stringify JSON object to match the format in the portal
             functionInput = <{}>JSON.stringify(functionInput, undefined, 2);
         } catch {
             functionInput = data;
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -48,6 +48,7 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         const data: string = await context.ui.showInputBox({ prompt, value, stepName: 'requestBody' });
         try {
             functionInput = <{}>JSON.parse(data);
+            functionInput = <{}>JSON.stringify(functionInput, undefined, 2);
         } catch {
             functionInput = data;
         }


nturinski (approved):
Two weeks of work for one line of code

src/commands/executeFunction.ts
```diff
@@ -48,6 +48,8 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         const data: string = await context.ui.showInputBox({ prompt, value, stepName: 'requestBody' });
         try {
             functionInput = <{}>JSON.parse(data);
+            // stringify JSON object to match the format in the portal
+            functionInput = <{}>JSON.stringify(functionInput, undefined, 2);
```
ejizba, line 52: The whole point of the try/catch is to `JSON.parse` the data, so if you're just going to re-stringify it then you could get rid of the whole try/catch.

More importantly though, I believe http triggers need this to be JSON.parsed and that's probably why I added this try/catch in the first place. The default request body is `{ "name": "Azure" }` and it should return "Hello, Azure! ..."

ejizba: Ftr, Nathan's gif is why I looked at this PR 👀

nturinski: We were experiencing issues with users of the service bus triggers. The service bus queue triggers weren't accepting JSON objects.  Apparently the portal takes the object and stringifies it instead.

That being said, Eric has a point about the HTTP trigger expecting a JSON object and that working. So all of this seems pretty opaque 😐

ejizba: The default request body is empty in the portal for an http trigger, but if I put `{ "name": "Azure" }` then it correctly responds with `Hello, Azure...`. So there's obviously something else the portal is doing here that VS Code isn't

Also ftr, I would prioritize maintaining http support over fixing a long-standing service bus bug

nturinski: I think we could either:

- Check to see if it's an HTTP trigger and skip this behavior
- Try the original method first-- if there's a 400 error, try stringifying it

motm32: I think I may prefer checking for an HTTP trigger since that would be simpler code, but I could see arguments for both.

nturinski: Though one drawback is that we may be breaking more triggers than just HTTP.

PR #3482, opened 2022-12-16
Add more robust logic for shared access keys

### LINKED ISSUE #3445: Add more robust logic for event hub connection string shared access keys
Phil Allen dropped by the office the other day and gave us some insights on additional edge case handling we probably should be doing when building out our event hub connection strings.  Things we should revisit:

- Filtering keys based on 'manage' access level
- Offer a way to create a shared access key if filtered keys === 0

https://learn.microsoft.com/en-us/azure/event-hubs/event-hubs-get-connection-string

MicroFish91:
Closes #3445

### DIFF
--- a/src/utils/azure.ts
+++ b/src/utils/azure.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import type { AccessKeys, AuthorizationRule, EventHubManagementClient } from '@azure/arm-eventhub';
+import { AccessKeys, AuthorizationRule, EventHubManagementClient, KnownAccessRights } from '@azure/arm-eventhub';
 import type { StorageAccount, StorageAccountListKeysResult, StorageManagementClient } from '@azure/arm-storage';
 import { AppKind, IAppServiceWizardContext } from '@microsoft/vscode-azext-azureappservice';
 import { getResourceGroupFromId, IStorageAccountWizardContext, uiUtils, VerifyProvidersStep } from '@microsoft/vscode-azext-azureutils';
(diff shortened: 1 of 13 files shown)


src/utils/azure.ts
```diff
@@ -72,30 +73,31 @@ export async function getStorageConnectionString(context: IStorageAccountWizardC
...
+    } else if (manageAccessRules.length === 1) {
+        authRuleName = nonNullProp(authRules[0], 'name');
     } else {
         const rootKeyName: string = 'RootManageSharedAccessKey';
         const placeHolder: string = localize('chooseSharedAccessPolicy', 'Choose a shared access policy.');
 
-        authRule = (await context.ui.showQuickPick(authRules.map(authRule => {
-            return { label: nonNullProp(authRule, 'name'), description: authRule.name === rootKeyName ? localize('default', '(Default)') : '' };
+        authRuleName = (await context.ui.showQuickPick(manageAccessRules.map(authRule => {
+            return { label: nonNullProp(authRule, 'name'), suppressPersistence: true, description: authRule.name === rootKeyName ? localize('default', '(default)') : '' };
```
nturinski, line 93: Why did lower-case default? Haha I prefer it with the uppercase. That's also how we do it for the (Preview) label.

src/utils/azure.ts
```diff
@@ -72,30 +74,31 @@ export async function getStorageConnectionString(context: IStorageAccountWizardC
 
 export async function getEventHubsConnectionString(context: IEventHubsConnectionWizardContext & ISubscriptionContext): Promise<IResourceResult> {
```
nturinski, line 75: Could you move all these `get____ConnectionString` functions out of `azure.ts`? I think it would make sense if they were in `appSettings/connectionSettings` as its own file for getting connection strings.

MicroFish91: Trueee 🙌

nturinski (approved):

PR #3485, opened 2022-12-20
Refresh local workspace after creating a function project

### LINKED ISSUE #3454: Refresh local workspace when creating a function project
It seems like you have to manually refresh the local workspace for a project to show up after creating a function project.

motm32:
Fixes #3454.


alexweininger (approved):
In v2, we'll should try to just refresh the local project tree item and not the whole workspace view.

PR #3490, opened 2022-12-22
Fix "Start Remote Debugging" action for Node Linux function apps

### LINKED ISSUE #3417: Fail to execute "Start Remote Debugging" action for the Node Linux function app
**OS:** Win10
**Build Version:** 20221107.1

**Repro Steps:**
1. Enable the setting `Azure Functions: Enable Remote Debugging`.
2. Create a **Node Linux** function app with **App Service Plan** hosting plan.
3. Deploy the JavaScript project to the function app.
4. Right click the above function app -> Start Remote Debugging.
5. Check whether succeeds to execute "Start Remote Debugging" action for the Node Linux function app.

**Expect:**
Succeed to execute "Start Remote Debugging" action for the Node Linux function app.

**Actual:**
Fail to execute "Start Remote Debugging" action for the Node Linux function app.
image

motm32:
Fixes #3417.

A couple of notes:

Consumption apps are still not supported but the `linuxFxVersion` value is no longer empty in this case, so I added a check for that. I also found that if changes are made to the function apps configuration settings the `linuxFxVersion` value will be 'NODE|...' instead of the expected 'Node|...' so that's why I lowercased the value 😄.

### DIFF
--- a/src/commands/remoteDebug/getRemoteDebugLanguage.ts
+++ b/src/commands/remoteDebug/getRemoteDebugLanguage.ts
@@ -14,7 +14,7 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfig, appServicePlan?: 
     //      or it will contain language information, e.g. "Node|14"
     const isNotConsumption: boolean = appServicePlan?.toLowerCase() !== 'y';
     if (isNotConsumption) {
-        if (siteConfig.linuxFxVersion && (/^(DOCKER\|(mcr\.microsoft\.com)\/(azure-functions)\/(node)|Node\|)/i).test(siteConfig.linuxFxVersion)) {
+        if (siteConfig.linuxFxVersion && (/(DOCKER\|mcr\.microsoft\.com\/azure-functions\/node)|(Node\|)/i).test(siteConfig.linuxFxVersion)) {
             return RemoteDebugLanguage.Node;
         }
     }
--- a/src/commands/remoteDebug/getRemoteDebugLanguage.ts
+++ b/src/commands/remoteDebug/getRemoteDebugLanguage.ts
@@ -12,11 +12,10 @@ export function getRemoteDebugLanguage(siteConfig: SiteConfig, appServicePlan?: 
     //   If the Function App is running on Linux consumption plan remote debugging is not supported
     //   If the Function App is running on a Linux App Service plan, it will contain Docker registry information, e.g. "DOCKER|repo.azurecr.io/image:tag"
     //      or it will contain language information, e.g. "Node|14"
-    if (appServicePlan) {
-        if (/^[^yY]$/.test(appServicePlan)) {
-            if (siteConfig.linuxFxVersion && (siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node') || siteConfig.linuxFxVersion.toLowerCase().startsWith('node|'))) {
-                return RemoteDebugLanguage.Node;
-            }
+    const isNotConsumption: boolean = appServicePlan?.toLowerCase() !== 'y';
+    if (isNotConsumption) {
+        if (siteConfig.linuxFxVersion && (/^(DOCKER\|(mcr\.microsoft\.com)\/(azure-functions)\/(node)|Node\|)/i).test(siteConfig.linuxFxVersion)) {
+            return RemoteDebugLanguage.Node;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -6,14 +6,17 @@
...
     //   If the Function App is running on Windows, it will be empty
-    //   If the Function App is running on Linux consumption, it will be empty
+    //   If the Function App is running on Linux consumption plan remote debugging is not supported
     //   If the Function App is running on a Linux App Service plan, it will contain Docker registry information, e.g. "DOCKER|repo.azurecr.io/image:tag"
-    //      The blessed Node.js Docker image will be something like "DOCKER|mcr.microsoft.com/azure-functions/node:2.0-node8-appservice"
-    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node')) {
-        return RemoteDebugLanguage.Node;
+    //      or it will contain language information, e.g. "Node|14"
+    const isNotConsumption: boolean = appServicePlan?.toLowerCase() !== 'y';
+    if (isNotConsumption) {
```
nturinski, line 16: I think we can upgrade this to an case-insensitive regex expression. I know you're up for the task 👩‍🎓

motm32: image

src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -6,15 +6,20 @@
...
-    //   If the Function App is running on Linux consumption, it will be empty
+    //   If the Function App is running on Linux consumption plan remote debugging is not supported
     //   If the Function App is running on a Linux App Service plan, it will contain Docker registry information, e.g. "DOCKER|repo.azurecr.io/image:tag"
-    //      The blessed Node.js Docker image will be something like "DOCKER|mcr.microsoft.com/azure-functions/node:2.0-node8-appservice"
-    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node')) {
-        return RemoteDebugLanguage.Node;
+    //      or it will contain language information, e.g. "Node|14"
+    if (appServicePlan) {
+        if (/^[^yY]$/.test(appServicePlan)) {
+            if (siteConfig.linuxFxVersion && (siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node') || siteConfig.linuxFxVersion.toLowerCase().startsWith('node|'))) {
```
nturinski, line 17: lmao I actually meant _this_ line should be used as a regex, not the consumption plan check 🤣😂😹

nturinski: I actually preferred your original way of checking for consumption plans. 😅

motm32: Lol I was confused for a second why you would want it for the consumption plan 😄, but I will change that back and put a regex for the other line.

src/commands/remoteDebug/getRemoteDebugLanguage.ts
```diff
@@ -6,15 +6,19 @@
...
-    //   If the Function App is running on Linux consumption, it will be empty
+    //   If the Function App is running on Linux consumption plan remote debugging is not supported
     //   If the Function App is running on a Linux App Service plan, it will contain Docker registry information, e.g. "DOCKER|repo.azurecr.io/image:tag"
-    //      The blessed Node.js Docker image will be something like "DOCKER|mcr.microsoft.com/azure-functions/node:2.0-node8-appservice"
-    if (siteConfig.linuxFxVersion && siteConfig.linuxFxVersion.startsWith('DOCKER|mcr.microsoft.com/azure-functions/node')) {
-        return RemoteDebugLanguage.Node;
+    //      or it will contain language information, e.g. "Node|14"
+    const isNotConsumption: boolean = appServicePlan?.toLowerCase() !== 'y';
+    if (isNotConsumption) {
+        if (siteConfig.linuxFxVersion && (/^(DOCKER\|(mcr\.microsoft\.com)\/(azure-functions)\/(node)|Node\|)/i).test(siteConfig.linuxFxVersion)) {
```
nturinski, line 17: That wasn't working for me in regex101, but you don't need to group the different parts of the DOCKER string in parens anyway. I think that this should work since we're expecting an exact match on the DOCKER string.

```suggestion
        if (siteConfig.linuxFxVersion && (/(DOCKER\|mcr\.microsoft\.com\/azure-functions\/node)|(Node\|)/i).test(siteConfig.linuxFxVersion)) {
```

nturinski (approved):

PR #3504, opened 2022-12-27
Fix multi-root support for downloading remote app settings and debugging

### LINKED ISSUE #3452: [Question]: Download remote settings for multiple functions in a workspace
Hi,

I had a question in regards to downloading remote settings for a function app.

My workspace contains multiple function apps.
I am able to deploy, test it, it all seems to work fine.

But when I try to download remote settings, it will prompt if `local.settings.json` doesn't exist for my first function app.
But if I try to select another function app and download it, I no longer have the prompt to select the target folder/app to download it to. It always overwrites the first one.

How do I force it to ask me every time what app to download the remote settings for?

Hope you can help, thanks!

Here is a screenshot of what I mean:

image

MicroFish91:
Fixes #3452 

This fixes both the downloading of remote app settings and local debugging for multi-root projects.

For the debug configuration fix, I changed this:

image
 to instead be tested with a RegExp.

Reason being, with multi-root, since the `preLaunchTask` config for projects that aren't the first one usually have a port assignment as part of the string, they will not match exactly the string `func: host start`.  Because of this, the local projects that come later were skipping `preDebugValidate`.  This should prevent them from being skipped now.

As per the wiki:
image

### DIFF
--- a/src/debug/FuncDebugProviderBase.ts
+++ b/src/debug/FuncDebugProviderBase.ts
@@ -6,7 +6,7 @@
 import { callWithTelemetryAndErrorHandling, IActionContext } from '@microsoft/vscode-azext-utils';
 import { CancellationToken, DebugConfiguration, DebugConfigurationProvider, WorkspaceFolder } from 'vscode';
 import { isFunctionProject } from '../commands/createNewProject/verifyIsProject';
-import { hostStartTaskName } from '../constants';
+import { hostStartTaskNameRegExp } from '../constants';
 import { IPreDebugValidateResult, preDebugValidate } from './validatePreDebug';
 
 export abstract class FuncDebugProviderBase implements DebugConfigurationProvider {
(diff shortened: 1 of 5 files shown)


src/debug/FuncDebugProviderBase.ts
```diff
@@ -46,7 +46,7 @@ export abstract class FuncDebugProviderBase implements DebugConfigurationProvide
             context.telemetry.suppressIfSuccessful = true;
 
             this._debugPorts.set(folder, <number | undefined>debugConfiguration.port);
-            if (debugConfiguration.preLaunchTask === hostStartTaskName) {
+            if (hostStartTaskNameRegExp.test(debugConfiguration.preLaunchTask)) {
```
nturinski, line 49: I'm okay with this change, but I don't think people should have been changing the name of their `preLaunchTask` to include the port. Where did you see that this was an issue? I didn't notice it skimming through the original issue.

nturinski (approved):

src/debug/FuncDebugProviderBase.ts
MicroFish91: User mentioned: 

`The issue is only with the local.settings.json, it only seems to look for/pick the local.settings.json file of the first function app in the list.`

When I started digging into the screenshot and testing out the different configs, I stumbled across this as the likely culprit.

PR #3505, opened 2022-12-28
Fix HTTP Trigger error with JSON response body

motm32:
This PR contains context for this error #3480. 

Refactored so we would only stringify the JSON body when we get a _"Unexpected status code: 400"_ error back to make sure we didn't break any of the other trigger types.

### DIFF
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -48,8 +48,6 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
         const data: string = await context.ui.showInputBox({ prompt, value, stepName: 'requestBody' });
         try {
             functionInput = <{}>JSON.parse(data);
-            // stringify JSON object to match the format in the portal
-            functionInput = <{}>JSON.stringify(functionInput, undefined, 2);
         } catch {
             functionInput = data;
         }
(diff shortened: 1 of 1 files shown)


src/commands/executeFunction.ts
```diff
@@ -79,6 +77,11 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
             if (!client && parseError(error).errorType === 'ECONNREFUSED') {
                 context.errorHandling.suppressReportIssue = true;
                 throw new Error(localize('failedToConnect', 'Failed to connect. Make sure your project is [running locally](https://aka.ms/AA76v2d).'));
+            } else if (parseError(error).message === 'Unexpected status code: 400') {
```
nturinski, line 80: Is there something else you can use to identify the error? Maybe `errorType` or something? Messages aren't always reliable because depending on the language, it might be different. Wording will also change sometimes.

MicroFish91: Since we're now using `parseError(error)` in two places, can you store the value in a variable and access it through that instead?

Something like:
```javascript
const pError = parseError(error);

if (pError.errorType ....
if (pError.message ....
```

nturinski (approved):

PR #3508, opened 2023-01-04
Declare limited support for virtual workspaces

### LINKED ISSUE #2793: Support virtual workspaces
👋 Hi there, Martin here, from the VS Code team.

Recently we've announced the [Remote Repository feature](https://code.visualstudio.com/updates/v1_56#_remote-repositories-remotehub) that lets you browse and edit files and folders directly on GitHub.

`Open Remote Repository...` opens VSCode on a folder or workspace located on a virtual file system. We call this a **virtual workspace**. We observed that not all extension support this well, either because they can not, or they haven't thought about it.

It would be fantastic if you could test whether your extension can handle virtual workspaces:

Check out the [Virtual Workspaces Extension Author Guide](https://github.com/microsoft/vscode/wiki/Virtual-Workspaces) on how to do that.

When done, set the new `virtualWorkspaces` capability in your 'package.json'.

```
{
  "capabilities": {
    "virtualWorkspaces": true | false
  }
}
```

- Use `"virtualWorkspaces": true` if your extension is prepared for virtual workspaces
- Use `"virtualWorkspaces": false` if your extension should be disabled when a virtual workspace is opened

For questions and comments please use the [Virtual Workspaces Tracking Issue](https://github.com/microsoft/vscode/issues/123115).

Thanks for the support and the great work! ❤️

motm32:
Fixes #2793.

Was able to test creating/deleting function apps and slots. Unable to deploy or create a new function project since access to the local file system is limited. Debugging also seems to not be supported based on the limitations listed [here.](https://marketplace.visualstudio.com/items?itemName=GitHub.remotehub#:~:text=in%20GitHub%20Codespaces.-,Limitations,-When%20you%20work)


bwateratmsft (dismissed):

nturinski:
Do you not have to declare which commands are and aren't compatible in the `package.json`? What happens when you try to run a command that's not supported in a virtual workspace?

motm32:
> Do you not have to declare which commands are and aren't compatible in the `package.json`? What happens when you try to run a command that's not supported in a virtual workspace?

Yes, you are correct. I just saw that in the document. Creating new function projects and deploying seem to be somewhat intuitive since you have to choose a project folder and you can't do that if you have a virtual workspace open. But it probably be good to declare those commands as incompatible. I am not completely sure if debugging would need to be declared since based off the documentation that should be a limitation for all virtual workspaces?

nturinski:
I was able to find a way to deploy via the cloud button and through the command palette which results in an error
image

But that being said, I don't know if I prefer making the commands disappear completely from context menus. It makes discoverability really difficult. Also, users would have no idea how much functionality that is absent in the virtual workspace compared to local extensions.

SWA declared limited support for virtual workspaces in a different part of the package.json that makes it look like this:
image

I will say I don't like that there's no hover-over tooltip about _why_ they're disabled, but that's something that VS Code could fix.

This is how that is implemented:
https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/package.json#L137-L144

I'm up for taking opinions on it, but I think we should be consistent between our extensions in how we declare limited support. I don't think we ever had a full discussion about it, but here are the guidelines Alex was following: https://github.com/microsoft/vscode-azurestaticwebapps/issues/380#issuecomment-896308988

Here's Alex's full PR for more context: https://github.com/microsoft/vscode-azurestaticwebapps/pull/489

nturinski:
> I will say I don't like that there's no hover-over tooltip about why they're disabled, but that's something that VS Code could fix.

https://github.com/microsoft/vscode/issues/131205

Apparently the onus is on us by using title information. We can probably use conditional when clauses for that?

motm32:
I agree, I think being consistent between extensions makes the most sense.

nturinski:
Some other commands that don't work:
```
azureFunctions.addBinding
azureFunctions.appSettings.decrypt
azureFunctions.appSettings.encrypt
azureFunctions.initProjectForVSCode
azureFunctions.installOrUpdateFuncCoreTools
azureFunctions.startJavaRemoteDebug
azureFunctions.startRemoteDebug
azureFunctions.uninstallFuncCoreTools
```

nturinski (approved):

PR #3518, opened 2023-01-12
Don't overwrite `func` tasks with a label different from `"func: host start"`

### LINKED ISSUE #3469: Confusing prompt to overwrite tasks upon workspace open results in non-runnable app
## Environment
VS Code 1.74.0
Mac OS Ventura 13.0.1
Functions extension v1.9.0

## Repro steps
1. Install prerequisites for Azure Dev CLI-enabled application "ToDo App with CSharp Functions and Azure SQL": https://github.com/Azure-Samples/todo-csharp-sql-swa-func#prerequisites
2. Create new empty folder and run `azd init -t todo-csharp-sql-swa-func` there
3. Open the folder in VS Code.

## Expected
Apart from the prompt to restore the .NET API (middle tier) project, there should be no prompts/warnings.

## Actual
There is a prompt that says
> Detected an Azure Functions Project in folder 'blah' that may have been created outside of VS Code. 
Initialize for optimal use with VS Code?

Who doesn't want optimal use? You say yes and see another prompt

> This will overwrite the following tasks in your tasks.json: "func: host start"

This is confusing: there is no task "func: host start" in `tasks.json`. What the extension refers to is a task that is part of the template, which is labeled ["API Functions Run"](https://github.com/Azure-Samples/todo-csharp-sql-swa-func/blob/main/.vscode/tasks.json#L43). If a task to be overwritten has a label, the extension should use the label for prompting.

If the user says yes, the task will be overwritten. Unfortunately this task is used by other tasks, and the end result is that the user loses ability to run the app locally.

## What I really want
An ability for this template to "tell" Azure Functions extension that it is complete with regards to  Azure Functions-related tasks, and that it should not offer to update/overwrite `tasks.json` file. Is there such a mechanism?

motm32:
Fixes #3469.

### DIFF
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
@@ -133,7 +133,7 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
         const matchingTaskLabels: string[] = [];
         for (const existingTask of existingTasks) {
             const existingLabel = this.getTaskLabel(existingTask);
-            if (existingLabel && newTasks.some(newTask => existingLabel === this.getTaskLabel(newTask))) {
+            if (existingLabel && newTasks.some(newTask => existingLabel === this.getTaskLabel(newTask) && existingTask.type !== 'func')) {
                 matchingTaskLabels.push(existingLabel);
             } else {
                 nonMatchingTasks.push(existingTask);
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
@@ -153,9 +153,10 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
         switch (task.type) {
             case func:
                 if (!task.label) {
-                    return task.type
+                    return task.type;
+                } else {
+                    return task.label;
                 }
-            // eslint-disable-next-line no-fallthrough
             case 'shell':
             case 'process':
                 return task.label;
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
@@ -133,7 +133,7 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
         const matchingTaskLabels: string[] = [];
         for (const existingTask of existingTasks) {
             const existingLabel = this.getTaskLabel(existingTask);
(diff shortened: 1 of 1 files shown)


src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
```diff
@@ -133,7 +133,7 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
         const matchingTaskLabels: string[] = [];
         for (const existingTask of existingTasks) {
             const existingLabel = this.getTaskLabel(existingTask);
-            if (existingLabel && newTasks.some(newTask => existingLabel === this.getTaskLabel(newTask))) {
+            if (existingLabel && newTasks.some(newTask => existingLabel === this.getTaskLabel(newTask) && existingTask.type !== 'func')) {
```
nturinski, line 136: This doesn't do quite exactly what I was expecting.

This [code](https://github.com/microsoft/vscode-azurefunctions/blob/bf447689cfe91df031e97ce2e1261106f59dd290/src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts#L152-L162) will make any task with `type` `func` return `func: host start`. 

We want to respect the label property (if it exists).  Then we should compare it to `export const hostStartTaskName: string = `${func}: ${hostStartCommand}`;` that lives in the `constants.ts` file. If it matches, then we should push it into `matchingTaskLabels`

Also, we don't need to check the `type` on the existingTask. There are times that we will want to replace `func` tasks. Like, if somebody changed their `func: start host` command and are trying to re-initialize their project to get it back to how it was.

alexweininger:
This could be a prime candidate for some automated tests...

nturinski (dismissed):

src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
```diff
@@ -152,7 +152,10 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
     private getTaskLabel(task: ITask): string | undefined {
         switch (task.type) {
             case func:
-                return `${task.type}: ${task.command}`;
+                if (!task.label) {
+                    return task.type
+                }
+            // eslint-disable-next-line no-fallthrough
```
bwateratmsft, line 158: Rather than suppress this, I'd add an `else` to retain clarity about what should happen.

src/commands/initProjectForVSCode/InitVSCodeStep/InitVSCodeStepBase.ts
```diff
@@ -152,7 +152,10 @@ export abstract class InitVSCodeStepBase extends AzureWizardExecuteStep<IProject
     private getTaskLabel(task: ITask): string | undefined {
         switch (task.type) {
             case func:
-                return `${task.type}: ${task.command}`;
+                if (!task.label) {
+                    return task.type
```
bwateratmsft, line 156: Nit: missing semicolon

nturinski (approved):

nturinski (changes requested):

nturinski (approved):

PR #3528, opened 2023-01-25
Remove .NET (Core) 2, 3.1, and 5 from `Create New Project` and `Create Function App in Azure...`

### LINKED ISSUE #3474: .NET Core 3.1 end-of-life
I'm not super familiar with all the features of the Functions extension, but with .NET Core 3.1's end-of-life being yesterday, we should remove any facilities for creating new projects like this (which #3472 does some of), and we can also remove any code that is in place specifically to support 3.1 and/or below. In addition, branding changes should be made since, after .NET Core 3.1, all versions are just called ".NET".

There may or may not be changes to still make after #3472, someone more familiar with the extension will have to look :smile:

MicroFish91:
Closes #3474 

See title.

There are a number of tests that include references to the above versions.  We will need to revisit and look into fixing these eventually.  I will put in an issue for the engineering work on this to be completed for when we allocate team time to fix the test suite.

### DIFF
--- a/src/commands/createFunctionApp/stacks/backupStacks.ts
+++ b/src/commands/createFunctionApp/stacks/backupStacks.ts
@@ -3,790 +3,1330 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-// Last updated on 2021-05-24
+// Last updated on 2023-01-09
 export const backupStacks: string = `{
     "value": [
-        {
-            "id": null,
-            "name": "dotnet",
-            "type": "Microsoft.Web/functionAppStacks?stackOsType=All",
-            "properties": {
-                "displayText": ".NET",
-                "value": "dotnet",
-                "preferredOs": "windows",
-                "majorVersions": [
-                    {
-                        "displayText": ".NET 5 (non-LTS)",
-                        "value": "dotnet5",
-                        "minorVersions": [
-                            {
-                                "displayText": ".NET 5 (non-LTS)",
-                                "value": "5 (non-LTS)",
-                                "stackSettings": {
-                                    "windowsRuntimeSettings": {
-                                        "runtimeVersion": "v5.0",
-                                        "isHidden": true,
-                                        "remoteDebuggingSupported": false,
-                                        "appInsightsSettings": {
-                                            "isSupported": true
-                                        },
-                                        "gitHubActionSettings": {
-                                            "isSupported": true,
-                                            "supportedVersion": "5.0.x"
-                                        },
-                                        "appSettingsDictionary": {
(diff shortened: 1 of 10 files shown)


src/commands/createFunctionApp/stacks/backupStacks.ts
```diff
@@ -270,63 +270,6 @@ export const backupStacks: string = `{
...
-                      ],
-                      "endOfLifeDate": "2022-12-03T00:00:00Z",
-                      "isDeprecated": true
-                    }
-                  }
-                }
-              ]
-            },
             {
               "displayText": ".NET Core 2",
```
bwateratmsft, line 274: We should take .NET Core 2 out as well.

MicroFish91: I thought so too, but the Create Function App options are still showing .NET Core 2 even with the `removeDeprecatedStacks` option set to true.

image

Should we just remove it anyway rather than wait for the API to catch up?

bwateratmsft: It might be because of the API version? In any case, we should _absolutely never_ recommend customers use EOL stuff, so I'd say we do our best to hide it regardless of whether the API is...

MicroFish91: Hmm it looks like the same thing is happening with version `2022-03-01` 🙁

I will filter it out manually.

MicroFish91:
Summary

I ended up making changes to remove .NET 2, .NET 3.1, and .NET 5 based on [this](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core).

For `Create New Project`:

I'm not even showing the ability to change Azure Function versions anymore because each of the versions listed under those options will only show deprecated runtimes which we would have to strip out.  I could keep the gear and just show that no runtimes are available, but that seemed like a bad user experience.  The portal just shows .NET 6 and 7, so I'm doing roughly the same thing.  Here's the view now:

image

For `Create Function App in Azure...`:

Even with `removeDeprecatedStacks` set to true, some of the deprecated stacks like .NET 2 were still coming through, so I manually had to filter them out.  I kept the backup stacks because we have a workspace setting called `showDeprecatedStacks`, and I wanted to preserve that functionality.  

Since the runtimes are all expanded out together, I added some functionality to add a QP that tells the user when no supporting runtimes are available (otherwise, it's kind of hard to tell quickly since the .NET or other grouping just disappears).  See example when selecting Azure Functions v3:

image

MicroFish91:
I wasn't completely sure if this section needed to be changed.  It seems like this section is hit whenever we are trying to search for function templates via `Create Function..`, so really mostly affects existing projects.  My initial thought was to keep it but just don't allow building anything new with `Create New Project` and `Create Function App in Azure`.

image

bwateratmsft (approved):

Issue #3534, opened 2023-01-28
`SingleHost` Netherite emulator string only works with C#


MicroFish91:
Was given a request before final merge to change `MemoryF` to `SingleHost`.  I did some additional testing and recently found that `SingleHost` is currently only working for C# projects.  I think the reason it might not be working with non-C# may have to do with the fact that our extension bundles aren't always updated in lock step with the latest C# package updates.

I've given CTI a build that has the changes using `MemoryF` again in the mean time.

MicroFish91:
The Bundles team is planning to make a new release.  We will monitor and see if it comes in time for our release, if so we can keep as `SingleHost`, else we'll have to revert back to `MemoryF`.

PR #3539, opened 2023-01-30
Revert `Netherite` emulator string from `SingleHost` to `MemoryF`

MicroFish91:
Current extension bundle brings in `Netherite` v1.1.1.  I believe `SingleHost` was introduced in v1.2, so only works with C# projects.  We should revert back and put #3534 in tracking...

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -140,10 +140,10 @@ export const tsDefaultOutDir: string = 'dist';
 export const tsConfigFileName: string = 'tsconfig.json';
 
 export const localEventHubsEmulatorConnectionStringDefault: string = 'MemoryF';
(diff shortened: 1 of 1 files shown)


bwateratmsft (dismissed):

bwateratmsft (approved):

bwateratmsft (approved):

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3534?

MicroFish91:
Yeah @nturinski, but I was going to keep it open so we could keep tracking for when we eventually swap everything back to `SingleHost` though (when the extension bundles are updated with the new Netherite package)

PR #3546, opened 2023-02-03
Change file structure for typescript functions

horihiro:
By default, functions in TypeScript (Model V4) are created in `src/functions` directory , then transpiled function code are generated in `dist/src/functions` directory as below.

```
├── src
│   └── functions
│       └──httpTrigger1.ts
├── dist # generated by tsc
│   └── src
│       └── functions
│           ├──httpTrigger1.js
│           └──httpTrigger1.map.js
```

In the above structure, `src` in `dist` does not seem to be needed.

This PR changes default settings (including generated `package.json`, `tsconfig.json` etc) for the structure from above to below.

```
├── src
│   └── functions
│       └──httpTrigger1.ts
├── dist # generated by tsc
│   └── functions
│       ├──httpTrigger1.js
│       └──httpTrigger1.map.js
```

Additionally, structure for JavaScript functions also is changed because transpilation is not needed.
```
├── functions
│   └──httpTrigger1.js
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1057,7 +1057,7 @@
                     "azureFunctions.functionSubpath": {
                         "scope": "resource",
                         "type": "string",
-                        "default": "src/functions",
+                        "default": "functions",
                         "description": "%azureFunctions.functionSubpath%"
                     },
                     "azureFunctions.showNodeProgrammingModel": {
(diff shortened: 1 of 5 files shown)


horihiro:
Thanks @ejizba !
I understand the reason of the structure clearly, so I close this PR.

\# I would like you to consider about including `test` templates next to `src`.

PR #3558, opened 2023-02-06
Use tryGetLocalFuncVersion

### LINKED ISSUE #3556: Deployment from VSC 1.10 to Azure functions failed: 'func' is not recognized
Deployment from VSCode v1.09 to Azure functions was working well until VSCode was updated automatically to v1.10 a few days ago (Feb 2, 2023). Now deployment yields an error: 

Error: Command "func --version" failed with exit code "1": 'func' is not recognized as an internal or external command, operable program or batch file.

I reverted back to VSCode v1.09 and deployment is working again. Could you pls advise if and when this is fixed?

Many thanks

nturinski:
Fixes #3556

### DIFF
--- a/src/commands/deploy/showCoreToolsWarning.ts
+++ b/src/commands/deploy/showCoreToolsWarning.ts
@@ -5,16 +5,16 @@
 
 import { DialogResponses, IActionContext, nonNullValue, openUrl } from "@microsoft/vscode-azext-utils";
 import { MessageItem } from "vscode";
-import { getLocalFuncCoreToolsVersion } from "../../funcCoreTools/getLocalFuncCoreToolsVersion";
 import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVersion";
+import { tryGetLocalFuncVersion } from "../../funcCoreTools/tryGetLocalFuncVersion";
 import { localize } from "../../localize";
 import { getWorkspaceSetting, updateGlobalSetting } from "../../vsCodeConfig/settings";
 
 export async function showCoreToolsWarning(context: IActionContext, runtimeVersion: FuncVersion, siteName: string): Promise<void> {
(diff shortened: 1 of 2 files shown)


alexweininger (dismissed):

bwateratmsft (dismissed):

src/commands/deploy/showCoreToolsWarning.ts
```diff
@@ -5,16 +5,16 @@
 
 import { DialogResponses, IActionContext, nonNullValue, openUrl } from "@microsoft/vscode-azext-utils";
 import { MessageItem } from "vscode";
-import { getLocalFuncCoreToolsVersion } from "../../funcCoreTools/getLocalFuncCoreToolsVersion";
 import { FuncVersion, funcVersionLink, tryParseFuncVersion } from "../../FuncVersion";
+import { tryGetLocalFuncVersion } from "../../funcCoreTools/tryGetLocalFuncVersion";
 import { localize } from "../../localize";
 import { getWorkspaceSetting, updateGlobalSetting } from "../../vsCodeConfig/settings";
 
 export async function showCoreToolsWarning(context: IActionContext, runtimeVersion: FuncVersion, siteName: string): Promise<void> {
```
ejizba, line 13: I would advocate that a warning like this shouldn't block deploy regardless of error/success, meaning this call probably shouldn't be awaited:

https://github.com/microsoft/vscode-azurefunctions/blob/main/src/commands/deploy/deploy.ts#L75

nturinski: This seems fair enough. We show a warning, but we don't give any actional buttons that would cancel the deployment, so there doesn't seem to be a reason to stop the deployment. 

It's not a modal either, so it may not be obvious that's why your deployment is being blocked.

alexweininger (approved):

PR #3592, opened 2023-02-23
Fix .NET isolated v4 template counts

alexweininger:
I'm guessing an .NET isolated template was added for v4?


bwateratmsft (approved):

nturinski (approved):

alexweininger:
Closing this since the template was only added to Staging, so this actually breaks more tests than it fixes.

Issue #3566, opened 2023-02-10
An error occurs when debugging a C# durable project created with "Netherite/MSSQL" durable storage type


v-xinda:
**OS:** M1 (Ventura 13.2) 
**Build Version:** 20230207.6
**.NET Version:** 7.0.100-rc.2.22477.23
**Regression:** Not a regression

**Repro Steps:**

1. Create and open a C# project with "Durable Function orchestrator" function and select "MSSQL" durable storage type.
2. F5 -> Select "Connect Azure SQL Database".
3. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
![image](https://user-images.githubusercontent.com/57662131/218058334-97732654-1bd4-4e46-ad72-5dce35d6bb68.png)
Details in Output:
![image](https://user-images.githubusercontent.com/57662131/218058498-c23a038b-f422-4910-bc82-d102e79cede7.png)

**More Info:**

1. This issue also reproduces for the C# project created with "Netherite" durable storage type.
2. This issue doesn't reproduce for the C# project created with "Azure Storage" durable storage type.
3. This issue doesn't reproduce on Windows and Linux.

MicroFish91:
Interesting, looks like this specifically reproduces with .NET 7

![image](https://user-images.githubusercontent.com/40250218/221255883-6d9b72b2-9d73-44d1-af51-3d7f5cd149a9.png)

MicroFish91:
Really good catch, I have a fix coming for this soon 🙂

MicroFish91:
Rest of the solution will need to come upstream.  Tracking: https://github.com/Azure/azure-functions-durable-extension/issues/2392

v-xinda:
Verified this issue using build 20230304.7, it **doesn't reproduce for the .NET 6 .0 (LTS)** project. 
But **an error also occurs when debug a .NET 6.0 Isolated or .NET 7.0 Isolated project.** The error details as below:
<img width="650" alt="Screenshot 2023-03-06 at 15 20 57" src="https://user-images.githubusercontent.com/57662131/223609689-435a651c-2ee1-4ac9-b12e-f924bd2ea14a.png">
Details in Terminal window:
```
 *  Executing task: dotnet clean /property:GenerateFullPaths=true /consoleloggerparameters:NoSummary 

MSBuild version 17.5.0-preview-23061-01+040e2a90e for .NET
Build started 3/8/2023 10:21:19 AM.
[...]

PR #3597, opened 2023-02-28
Add Dotnet Isolated runtime specific dependencies and cleanup `durableUtils`

MicroFish91:
Add .NET Isolated runtime specific dependencies, specifically for `Netherite` and `MSSQL` storage types.

Partially addresses the error message shown in #3566.

### DIFF
--- a/src/utils/durableUtils.ts
+++ b/src/utils/durableUtils.ts
@@ -19,11 +19,11 @@ import { venvUtils } from "./venvUtils";
 import { findFiles } from "./workspace";
 
 export namespace durableUtils {
-    const dotnetLtsDfSqlPackage: string = 'Microsoft.DurableTask.SqlServer.AzureFunctions';
+    const dotnetInProcDfSqlPackage: string = 'Microsoft.DurableTask.SqlServer.AzureFunctions';
     const dotnetIsolatedDfSqlPackage: string = 'Microsoft.Azure.Functions.Worker.Extensions.DurableTask.SqlServer';
-    const dotnetLtsDfNetheritePackage: string = 'Microsoft.Azure.DurableTask.Netherite.AzureFunctions';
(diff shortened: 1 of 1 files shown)


src/utils/durableUtils.ts
```diff
@@ -163,19 +155,28 @@ export namespace durableUtils {
...
             case DurableBackend.Storage:
             default:
         }
 
-        // Seems that the package arrives out-dated and needs to be updated
-        packageNames.push(durableUtils.dotnetDfBasePackage);
+        // Seems that the package arrives out-dated and needs to be updated to at least 2.9.1;
+        // otherwise, error appears when running with sql backend
+        if (!isDotnetIsolated) {
+            packageNames.push(dotnetLtsDfBasePackage);
```
MicroFish91, line 178: Not 100% sure if this is the best way to handle this; let me know if anyone has opinions on if this should be handled differently.

src/utils/durableUtils.ts
```diff
@@ -111,19 +113,9 @@ export namespace durableUtils {
                     // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
                     let packageReferences = result?.['Project']?.['ItemGroup']?.[0]?.PackageReference ?? [];
```
alexweininger, line 114: Unrelated to this PR, but there is a bug here if the package references are split into multiple item groups since we're only using the first item group. I can fix this in a separate PR.

MicroFish91: Ahh good catch! It should probably loop through the 'ItemGroup' and check for the 'PackageReference' instead

alexweininger (approved):

PR #3604, opened 2023-03-01
Fix TypeScript project language detection

alexweininger:
Changes in https://github.com/microsoft/vscode-azurefunctions/pull/3586 changed TypeScript language detection in that JavaScript would also be detected, which results in prompting the user to choose a language in cases that didn't before #3586 was merged. I found this while fixing the unit tests.

These changes fix the unit tests so that if TypeScript is detected (tsconfig found, etc.), then JavaScript won't also be detected.

### DIFF
--- a/src/commands/initProjectForVSCode/detectProjectLanguage.ts
+++ b/src/commands/initProjectForVSCode/detectProjectLanguage.ts
@@ -24,7 +24,10 @@ export async function detectProjectLanguage(context: IActionContext, projectPath
             if (await isTypeScriptProject(projectPath)) {
                 detectedLangs.push(ProjectLanguage.TypeScript);
             } else {
-                detectedLangs.push(ProjectLanguage.JavaScript);
+                // don't add JavaScript if TypeScript is already detected
+                if (!detectedLangs.includes(ProjectLanguage.TypeScript)) {
+                    detectedLangs.push(ProjectLanguage.JavaScript);
+                }
             }
         }
 


src/commands/initProjectForVSCode/detectProjectLanguage.ts
```diff
@@ -24,7 +24,10 @@ export async function detectProjectLanguage(context: IActionContext, projectPath
             if (await isTypeScriptProject(projectPath)) {
                 detectedLangs.push(ProjectLanguage.TypeScript);
             } else {
-                detectedLangs.push(ProjectLanguage.JavaScript);
+                // don't add JavaScript if TypeScript is already detected
```
ejizba, line 27: > if TypeScript is detected (tsconfig found, etc.)

I don't think it's checking for tsconfig right now, but that's an interesting idea

src/commands/initProjectForVSCode/detectProjectLanguage.ts
```diff
@@ -24,7 +24,10 @@ export async function detectProjectLanguage(context: IActionContext, projectPath
...
+                if (!detectedLangs.includes(ProjectLanguage.TypeScript)) {
```
ejizba, line 28: `isTypeScriptProject` currently checks for `typescript` listed as a dev dependency. Isn't that required for TypeScript projects? At the end of the day, it seems fine if we prompt for JS vs TS for weird projects - and I think the unit tests might have weird projects

alexweininger: If there's a `*/index.ts` file, but `typescript` is not a dev dependency, then we will detect JS and TS. Whereas before we only detected TS.

Although the test case is weird, if we find a `*/index.ts` file, I think it's totally safe to assume the project language isn't JS.

src/commands/initProjectForVSCode/detectProjectLanguage.ts
alexweininger: My bad, I thought we were doing that. Adding a check for a tsconfig would also satisfy these tests since a tsconfig is created upon initializing the test folder.

src/commands/initProjectForVSCode/detectProjectLanguage.ts
ejizba: No strong preference from me 🤷‍♂️

ejizba (approved):

PR #3609, opened 2023-03-03
Change learn more label for programming model step

### LINKED ISSUE #3547: Node v4 model: Clarify what "Learn more" is about
image

**Request:** Change to "Learn more about Model V4"

nturinski:
Fixes #3547 

image

### DIFF
--- a/src/commands/createNewProject/ProgrammingModelStep.ts
+++ b/src/commands/createNewProject/ProgrammingModelStep.ts
@@ -31,7 +31,6 @@ export class ProgrammingModelStep extends AzureWizardPromptStep<IProjectWizardCo
 
         const learnMoreQp = { label: localize('learnMore', '$(link-external) Learn more about Model V4...'), description: '', data: undefined };
         if (this._learnMoreLink) {
-
             modelsPick.push(learnMoreQp);
         }
 
--- a/src/commands/createNewProject/ProgrammingModelStep.ts
+++ b/src/commands/createNewProject/ProgrammingModelStep.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizardPromptStep, IAzureQuickPickItem, IAzureQuickPickOptions, nonNullValue, openUrl } from '@microsoft/vscode-azext-utils';
-import { learnMoreQp, nodeProgrammingModelSetting } from '../../constants';
+import { nodeProgrammingModelSetting } from '../../constants';
 import { localize } from '../../localize';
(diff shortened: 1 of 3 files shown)


src/commands/createNewProject/ProgrammingModelStep.ts
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizardPromptStep, IAzureQuickPickItem, IAzureQuickPickOptions, nonNullValue, openUrl } from '@microsoft/vscode-azext-utils';
-import { learnMoreQp, nodeProgrammingModelSetting } from '../../constants';
```
bwateratmsft, line 7: Is the `learnMoreQp` constant still referenced by anything else or can it be removed?

src/commands/createNewProject/ProgrammingModelStep.ts
```diff
@@ -29,7 +29,9 @@ export class ProgrammingModelStep extends AzureWizardPromptStep<IProjectWizardCo
             }
         });
 
+        const learnMoreQp = { label: localize('learnMore', '$(link-external) Learn more about Model V4...'), description: '', data: undefined };
```
bwateratmsft, line 32: Does the `IAzureQuickPick` lack a way to override the "Learn More" text? That would be a good addition to the API that could simplify this code.

src/commands/createNewProject/ProgrammingModelStep.ts
```diff
@@ -29,7 +29,9 @@ export class ProgrammingModelStep extends AzureWizardPromptStep<IProjectWizardCo
...
         if (this._learnMoreLink) {
+
```
bwateratmsft, line 34: Nit: Spurious newline

bwateratmsft (approved):

src/commands/createNewProject/ProgrammingModelStep.ts
nturinski: Still used, but only in one other place. I'll just move it to there.

src/commands/createNewProject/ProgrammingModelStep.ts
nturinski: I think at the moment we only have a Learn More `MessageItem` for notification dialogs and a Learn More `QuickInputButton` which is that little ? button on the top right of the quickpick window: 
image

Today, we don't have a standardized way of adding `learn more...` to quick picks, but I think it makes sense to add that to the API along with a way to override the learn more text.

bwateratmsft (approved):

PR #3611, opened 2023-03-03
Make skip for now message more clear when there are no resources available

### LINKED ISSUE #3550: Node v4 model: Connection info not populated correctly for Cosmos and Service Bus triggers
Cosmos trigger: `connectionStringSetting` is undefined, though user is prompted to create local app setting. 

**Request:** 
1. ~~Should _not_ prompt user about this if can't fill in the info correctly - simply use empty string as value for the property.~~ Guide user to create resource if doesn't already exist, just like Storage today. 
2. **OR** - if we can't do the above now, change the "skip for now" (2nd screenshot) to "skip for now because no resource is found" 

Current experience:
1. Choosing to create new local app setting 
  image
2. After logging into Azure, only option is to skip
  image

**Similar observation as above for Service Bus Queue and Topic trigger.** 

```[tasklist]
### Tasks
- [ ] Allow creating EventHub Namespaces
- [ ] Allow creating Service Bus Queues
```

nturinski:
Partially fixes #3550 

This is the old skip for now. The behavior is the same as before:
image

If no picks were found, the message is edited to reflect that and hopefully makes it more clear for users. Ideally, we could create these resources for the user here later, but this is more of a gapfixer.
image

### DIFF
--- a/src/utils/azure.ts
+++ b/src/utils/azure.ts
@@ -23,7 +23,7 @@ export async function promptForResource<T extends IBaseResourceWithName>(context
             .map((r: T) => r.name ? { data: r, label: r.name, description: r._description } : undefined)
             .filter((p: IAzureQuickPickItem<T> | undefined) => p));
         picks.push({
-            label: localize('skipForNow', '$(clock) Skip for now'),
+            label: picks.length ? localize('skipForNow', '$(clock) Skip for now') : localize('skipForNow', '$(warning) Skip because no resource was found'),
             data: undefined,
             suppressPersistence: true
         });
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):

ejizba:
I feel like I wouldn't know I can click the new one because it doesn't seem actionable. I think "Skipped" is the main word confusing me because it's past tense

What about:
> No matching resources found. Skip for now

nturinski:
...dang it, I didn't see this comment until now. I'll change this if we aren't able to resolve this properly by allowing users to create a new resource directly.

PR #3647, opened 2023-04-14
Remove `azureWebJobsStorage` validation on deploy

### LINKED ISSUE #3638: AzureWebJobsStorage configuration setting is created on deployment, even when AzureWebJobsStorage__accountName already exists
It seems I have the same issue described in #3575 , even though I use v1.10.4, that according to the last comment in the issue should fix it.

MicroFish91:
Fixes #3638 

As part of the Durable work, I introduced a remote resource/connection string validation pattern on deploy because these were required for netherite and sql backends.  I went ahead and applied this pattern to storage to maintain consistency across the connection strings.  

AFAICT, this was not being done before, and in adding this, I introduced an unintended consequence - see: https://github.com/microsoft/vscode-azurefunctions/issues/3638#issuecomment-1502080143 
Basically, I was not aware of and did not take this into account: https://learn.microsoft.com/en-us/azure/azure-functions/functions-identity-based-connections-tutorial#edit-the-azurewebjobsstorage-configuration

Since we already connect storage accounts during Function App creation, it's probably not super necessary to do this validation anyway.  I am removing `azureWebJobsStorage` validation on deploy to be consistent with how we used to things and to fix the above issue.

### DIFF
--- a/src/commands/appSettings/connectionSettings/azureWebJobsStorage/validateStorageConnection.ts
+++ b/src/commands/appSettings/connectionSettings/azureWebJobsStorage/validateStorageConnection.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizard } from "@microsoft/vscode-azext-utils";
-import { CodeAction, ConnectionKey, localStorageEmulatorConnectionString } from "../../../../constants";
+import { CodeAction, ConnectionKey } from "../../../../constants";
 import { getLocalSettingsConnectionString } from "../../../../funcConfig/local.settings";
 import { IConnectionPromptOptions } from "../IConnectionPromptOptions";
 import { ISetConnectionSettingContext } from "../ISetConnectionSettingContext";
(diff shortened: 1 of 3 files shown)


src/commands/appSettings/connectionSettings/azureWebJobsStorage/validateStorageConnection.ts
```diff
@@ -4,29 +4,24 @@
...
 import { getLocalSettingsConnectionString } from "../../../../funcConfig/local.settings";
 import { IConnectionPromptOptions } from "../IConnectionPromptOptions";
 import { ISetConnectionSettingContext } from "../ISetConnectionSettingContext";
 import { AzureWebJobsStorageExecuteStep } from "./AzureWebJobsStorageExecuteStep";
 import { AzureWebJobsStoragePromptStep } from "./AzureWebJobsStoragePromptStep";
 import { IAzureWebJobsStorageWizardContext } from "./IAzureWebJobsStorageWizardContext";
 
-// Supports validation on both 'debug' and 'deploy'
 export async function validateStorageConnection(context: Omit<ISetConnectionSettingContext, 'projectPath'>, projectPath: string, options?: IConnectionPromptOptions): Promise<void> {
+    if (context.action === CodeAction.Deploy) {
```
nturinski, line 16: Do we need even need this?` validateStorageConnection` never gets called in the deploy scenario anyway, right? I think it's only in pre-debugging.

nturinski (approved):

src/commands/appSettings/connectionSettings/azureWebJobsStorage/validateStorageConnection.ts
MicroFish91: Correct, I wanted to add it in there so that it's extremely explicit, especially since the rest of the connection type validations can happen in both debug or deploy

PR #3654, opened 2023-04-21
Add clean script to TypeScript projects

ejizba:
In model v4, the entrypoint is a glob (`dist/src/functions/*.js`) meaning if you delete a ts file in `src`, you also need to delete the js file in `dist` otherwise the function will still show up. This can be confusing, so probably worth adding a clean script.

A few notes:
- rimraf v5 works in [Node.js v14+](https://github.com/isaacs/rimraf/blob/main/package.json#L79), but I feel like that's fine since Node 12 was EOL Oct 2020
- There's not as much value when adding this to the old model, but I thought it was easier to keep them consistent
- I'm _pretty sure_ my test changes work, but obviously I wasn't able to do a full test run given the state of the tests


nturinski:
> I'm pretty sure my test changes work, but obviously I wasn't able to do a full test run given the state of the tests

nturinski (approved):

PR #3678, opened 2023-05-17
Support Pystein V2 Templates without implementing v2 Schema

nturinski:
Note that this is a hardcoded approach to support the V2 schema templates! We are adding all of the V2 assets into our `backupTemplate` folder and using that to overwrite template files.

This approach is to leverage the new V2 templates without having to support the entire V2 schema which includes a bunch of new useful design patterns.

Part of supporting the v2 templates includes:
- Prompting users for parameters and replacing the content in the files
- Removing (preview) anywhere it was present
- Allowing users create their project with a function trigger rather than the boilerplate code that was commented out
- Replacing the file content of the V1 templates with the V2 templates

Known issue:
The prompts are a little wonky because we try to prompt for a local.settings.json storage connection. This is due to the fact that we always prompt users for Python projects on V1. I should be able to fix that when I return on Monday, so please just ignore that bug for now.


src/templates/script/parseScriptTemplates.ts
```diff
@@ -142,6 +143,43 @@ function parseScriptSetting(data: object, resources: IResources, variables: IVar
...
+            enums.push({
+                value: getResourceValue(resources, ev.value),
+                displayName: getResourceValue(resources, ev.display)
+            });
+        }
+    }
+
+    return {
+        name: rawSetting ? getResourceValue(resources, rawSetting.name) : input.paramId,
+        resourceType: rawSetting ? rawSetting.resource : undefined,
```
bwateratmsft, line 161: Why not this for these and below?
```suggestion
        resourceType: rawSetting?.resource,
```

src/templates/script/parseScriptTemplates.ts
```diff
@@ -142,6 +143,43 @@ function parseScriptSetting(data: object, resources: IResources, variables: IVar
...
+        valueType: rawSetting ? rawSetting.value : undefined,
+        description: rawSetting && rawSetting.help ? replaceHtmlLinkWithMarkdown(getResourceValue(resources, rawSetting.help)) : undefined,
+        defaultValue: rawSetting && rawSetting.defaultValue ? rawSetting.defaultValue : input.defaultValue,
```
bwateratmsft, line 164: e.g.
```suggestion
        defaultValue: rawSetting?.defaultValue || input.defaultValue,
```

bwateratmsft (approved):

PR #3700, opened 2023-06-07
Add SQL input and output binding verified templates

VasuBhog:
Adding SQL input and output bindings to the verified templates list as we have now GA'd. 

We support 
C#, C#-Script (no template), Javascript, Typescript, Python, PowerShell, and Java.


src/templates/dotnet/getDotnetVerifiedTemplateIds.ts
```diff
@@ -33,6 +35,6 @@ export function getDotnetVerifiedTemplateIds(version: string): RegExp[] {
     }
 
     return verifiedTemplateIds.map(id => {
-        return new RegExp(`^azure\\.function\\.csharp\\.(?:isolated\\.|)${id}\\.[0-9]+\\.x$`, 'i');
+        return new RegExp(`^azure\\.function\\.csharp\\.(?:isolated\\.|)${id}(\\.[0-9]+\\.x|)$`, 'i');
```
VasuBhog, line 38: After discussions with @alexweininger and @nturinski, we found that this regex specifically only looks for template `identity` names that include `.x.Version.x` alongside it to be included for the verified templates list. However, after looking over the templates.json of many extensions we found most include either `.1.x` or `.2.x` and it came to our assumption that this was the function runtimes to keep track of what is supported. However, since this seems like an extremely tedious task, it was forgotten and no longer updated for many extensions as such. 

As such, I would like to know if we could change the regex to **not** include the versioning requirement in order for GA'd SQL input/output binding can be shown on the verified lists.

src/templates/dotnet/getDotnetVerifiedTemplateIds.ts
```diff
@@ -16,7 +16,9 @@ export function getDotnetVerifiedTemplateIds(version: string): RegExp[] {
         'ServiceBusTopicTrigger',
         'ServiceBusQueueTrigger',
         'CosmosDBTrigger',
-        'EventGridTrigger'
+        'EventGridTrigger',
+        'SQLInputBinding',
+        'SQLOutputBinding'
```
bwateratmsft, line 21: Nit, for one less line-change in the future:
```suggestion
        'SQLOutputBinding',
```

src/templates/script/getScriptVerifiedTemplateIds.ts
```diff
@@ -10,7 +10,9 @@ export function getScriptVerifiedTemplateIds(version: string): (string | RegExp)
         'BlobTrigger',
         'HttpTrigger',
         'QueueTrigger',
-        'TimerTrigger'
+        'TimerTrigger',
+        'SQLInputBinding',
+        'SQLOutputBinding'
```
bwateratmsft, line 15: Nit, for one less line-change in the future:
```suggestion
        'SQLOutputBinding',
```

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurefunctions/pull/4358

Sorry that this sort of got stuck in limbo.

PR #3707, opened 2023-06-12
Update to utils v2

alexweininger:
Fully converts extension to Track 2 SDKs.

* Updated min version of VS Code to 1.79.0.
* Removed `@azure/ms-rest-js` dependency
* Added `@azure/core-client`, `@azure/core-rest-pipeline`, `@microsoft/vscode-azureresources-api`, and `@microsoft/vscode-azext-azureappsettings`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "icon": "resources/azure-functions.png",
     "aiKey": "0c6ae279ed8443289764825290e4f9e2-1a736e7c-1324-4338-be46-fc2a58ae4d14-7255",
     "engines": {
-        "vscode": "^1.78.0"
+        "vscode": "^1.79.0"
     },
     "repository": {
         "type": "git",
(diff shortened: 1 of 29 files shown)


package.json
```diff
@@ -1182,18 +1182,21 @@
         "webpack-cli": "^4.6.0"
     },
     "dependencies": {
-        "@azure/arm-appinsights": "^4.0.0",
```
alexweininger, line 1185: Sadly the App Insights SDK hasn't GA'd their Track 2 version. So we need to use an alpha build.

bwateratmsft (dismissed):

nturinski:
Not really sure why webpack is freaking out. Unless you specify the "webworker" target, it should only be targeting node.

alexweininger:
> Not really sure why webpack is freaking out. Unless you specify the "webworker" target, it should only be targeting node.

It also freaked out in VMs. I had to explicitly specific target: node to fix it. Very weird.

bwateratmsft (approved):

bwateratmsft (approved):

PR #3721, opened 2023-06-20
Fix SWA create function for node model v4

ejizba:
When picking the template based on a templateId (which is how SWA creates a function [here](https://github.com/microsoft/vscode-azurestaticwebapps/blob/main/src/commands/createHttpFunction.ts#L34)), the functions extension always uses the v3 template even if v4 was selected by the user. I adjusted the filtering in `FunctionListStep` but it didn't quite work because we were calling `FunctionListStep.create` _before_ the model prompt. I switched to use the fancy new `configureBeforePrompt` method which selects the template _after_ the model prompt and worked for me.

Fixes https://github.com/microsoft/vscode-azurestaticwebapps/issues/829

### DIFF
--- a/src/templates/script/getScriptVerifiedTemplateIds.ts
+++ b/src/templates/script/getScriptVerifiedTemplateIds.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { FuncVersion } from '../../FuncVersion';
+import { nodeV4Suffix } from '../../utils/programmingModelUtils';
 
 export function getScriptVerifiedTemplateIds(version: string): (string | RegExp)[] {
     let verifiedTemplateIds: string[] = [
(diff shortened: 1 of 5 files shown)


src/templates/script/getScriptVerifiedTemplateIds.ts
```diff
@@ -39,11 +40,11 @@ export function getScriptVerifiedTemplateIds(version: string): (string | RegExp)
         // These languages are only supported in v2+ - same functions as JavaScript, with a few minor exceptions that aren't worth distinguishing here
         // NOTE: The Python Preview IDs are only temporary.
         // NOTE: The Node Programming Model IDs include -4.x as a suffix
-        const regExps = verifiedTemplateIds.map(t => new RegExp(`^${t}-(JavaScript(-4.x)?|TypeScript(-4.x)?|Python|PowerShell|Custom|Python-Preview|Python-Preview-Append)$`, 'i'));
+        const regExps = verifiedTemplateIds.map(t => new RegExp(`^${t}-(JavaScript(${nodeV4Suffix})?|TypeScript(${nodeV4Suffix})?|Python|PowerShell|Custom|Python-Preview|Python-Preview-Append)$`, 'i'));
```
bwateratmsft, line 43: Template strings in a regular expression? You're braver than me. :smile:

bwateratmsft (approved):

PR #3726, opened 2023-06-23
Resolve using list instead of get

### LINKED ISSUE #3723: Slow loading time when listing function apps
Hi all.

We have been experiencing issues with the listing of Azure Functions in vs-code for the last couple of weeks. 

When we attempt to deploy our code to Azure Functions, the process to list all of our function apps from Azure is significantly slow, taking around 45-50 seconds. This is considerably longer than expected and previous experiences. 

Any ideas on what could be wrong?

alexweininger:
Fixes #3723 

Making a HTTP request for each resource is really hurting performance. Ideally we can delay these requests until the user expands the resource or invokes a command on it, but for now I think this is the best solution.

The cache only valid for 3 seconds, long enough to make sure the entire batch of resolve calls can hit the cache. I didn't do anything fancy since I want to validate with the customer that this speeds things up. I figure we can make a util and share this across extensions later.

### DIFF
--- a/src/FunctionAppResolver.ts
+++ b/src/FunctionAppResolver.ts
@@ -15,6 +15,7 @@ export class FunctionAppResolver implements AppResourceResolver {
             const client = await createWebSiteClient({ ...context, ...subContext });
 
             if (this.siteCacheLastUpdated < Date.now() - 1000 * 3) {
+                this.siteCache.clear();
                 const sites = await uiUtils.listAllIterator(client.webApps.list());
                 sites.forEach(site => this.siteCache.set(nonNullProp(site, 'id').toLowerCase(), site));
                 this.siteCacheLastUpdated = Date.now();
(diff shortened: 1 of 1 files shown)


nturinski (approved):

nturinski (changes requested):
Actually, you never actually clear the cache. I'm worried that if we delete a function app, the siteCache will still be in the map because we're doing a list, and then a for each over the returned sites list.

We could just save the array instead and delete that whenever we reset the cache? It might be easier to maintain, though that's not necessarily better than a map or anything.

alexweininger:
> Actually, you never actually clear the cache. I'm worried that if we delete a function app, the siteCache will still be in the map because we're doing a list, and then a for each over the returned sites list.

Good catche 😄 , it's probably a good idea to clear the cache.

I do want to note that technically delete would work OK since RGs will always re-list the resources and only resolve what it knows about. So having deleted things in the cache wouldn't cause issues for that particular case.

I do think it might cause issues at other times, so it's still a good idea.

nturinski (approved):

alexweininger:
Since the release is delayed maybe we merge this and do another round of testing.

PR #3737, opened 2023-06-30
Support ZRS enabled storage accounts in advanced create

### LINKED ISSUE #3735: Storage account with ZRS replication not found
OS: All
Version: 1.11.0

Repro Steps:
I have some storage accounts StorageV2 (general purpose v2) with Standard Performance and different replication

1. Create Function App in Azure... (Advanced)
2. Written unique name for the function app
3. Selected any runtime stack
4. Selected any OS
5. Selected resource group in which my storage account with ZRS replication enabled is
6. Selected any hosting plan
7. My storage accounts with ZRS replication enabled are not present. Only storage accounts with LRS replication and RA-GRS replication are listed.

Expect:
On step 7, I expect to see all my storage accounts, included the ones with ZRS replication 

Actual:
The storage accounts with ZRS replication are not listed by vsc. Doing the procedure through the Azure portal is successful even with the storage account not listed by vsc.

alexweininger:
Fixes #3735 

ZRS enabled storage accounts were being deliberately excluded from the list. However, I wasn't able to find out why and I verified that the FA can be created and works as expected.

I also added some comments on the other items in the filter.


nturinski (approved):
Yeah, I don't see any reason why we shouldn't include ZRS. It seems recommended even: https://learn.microsoft.com/en-us/azure/storage/common/storage-redundancy#redundancy-in-the-primary-region

PR #3742, opened 2023-07-10
Nat/v2 actions handler

nturinski:
Built off the template provider, this will actually create the projects. It takes the inputs and turns them into prompt steps and takes the actions and turns them into execute steps.

So far it seems like it works relatively well. There have been some small hiccups with the templates which I manually corrected.

The template feed is also still not live, so can't do testing with the real thing just yet.


src/templates/script/parseScriptTemplatesV2.ts
```diff
@@ -70,38 +80,35 @@ interface IRawUserPrompt {
...
+
 }
 
-export interface IParsedJob extends IRawJob {
-    prompts: IParsedInputPrompt[];
-    // doesn't seem like there's a need to parse these yet
-    executes: IParsedAction[];
+export interface ParsedJob extends RawJob {
+    prompts: ParsedInput[];
+    executes: ParsedAction[];
```
alexweininger, line 94: I'm not a huge fan of the mismatched property name and type names here. I know that each input will later correspond to a prompt step and each action an execute step. However, I don't love that the translation is done here in the property names before the inputs and actions are actually made into steps.

I'd rather have the property names match the types like so:
```suggestion
    parsedInputs: ParsedInput[];
    parsedActions: ParsedAction[];
```

I think this makes even more sense taking into account the prompt step factory definition is:
```ts
function promptStepFactory<T>(input: ParsedInput): AzureWizardPromptStep<T>;
```

src/commands/createFunction/promptStepsV2/promptStepFactory.ts
```diff
@@ -0,0 +1,29 @@
...
+            switch (input.resource) {
+                case ResourceType.ExistingFile:
+                    return new ExistingFileStep(input);
+                case ResourceType.NewFile:
+                    return new NewFileStep(input);
+                default:
+                    return new StringInputStep(input);
+            }
+    }
+}
```
alexweininger, line 29: Nested switch statements make me sad but this one seems pretty much unavoidable

src/commands/createFunction/JobsListStep.ts
```diff
@@ -3,49 +3,71 @@
...
-        // create steps based on the job
+    public async getSubWizard(context: FunctionWizardV2Context): Promise<IWizardOptions<FunctionWizardV2Context> | undefined> {
         if (context.job) {
-            const promptSteps: AzureWizardPromptStep<IFunctionWizardContext>[] = [];
-            context.job.prompts.map(_p => {
-                promptSteps.push();
+            const promptSteps: AzureWizardPromptStep<FunctionWizardV2Context>[] = [];
+            context.job.prompts.map(p => {
+                promptSteps.push(promptStepFactory(p));
             });
```
alexweininger, line 37: This code will also make more sense IMO if `prompts` is renamed to `parsedInputs`:

```suggestion
            context.job.parsedInputs.map(parsedInput => {
                promptSteps.push(promptStepFactory(parsedInput));
            });
```

src/commands/createFunction/FunctionV2WizardContext.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { FunctionTemplateV2 } from '../../templates/FunctionTemplateV2';
+import { ParsedJob } from "../../templates/script/parseScriptTemplatesV2";
+import { IFunctionWizardContext } from './IFunctionWizardContext';
+
+export interface FunctionWizardV2Context extends IFunctionWizardContext {
```
alexweininger, line 10: What is your opinion on naming this `V2FunctionWizardContext`. It looks worse, but reads better IMO. I'm on the fence 😄 

Also whatever we choose I'd want to be consistent and use v2 as a prefix everywhere.

src/commands/createFunction/actionStepsV2/ReplaceTokensInTextExecuteStep.ts
```diff
@@ -0,0 +1,36 @@
...
+        await this.writeToFile(context, content);
+    }
+}
+
+//     {
+//       name: "replaceText_FunctionBody",
+//       type: "ReplaceTokensInText",
+//       assignTo: "$(TIMER_FUNCTION_BODY)",
+//       source: "$(TIMER_FUNCTION_BODY)",
+//     }
```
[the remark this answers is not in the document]
nturinski: Oh, these comments were just references for me to know what the object looks like.

src/commands/createFunction/promptStepsV2/promptStepFactory.ts
bwateratmsft: It's either that or a big if-else block

alexweininger: Yeah, and I don't think I like that more

src/commands/createFunction/FunctionV2WizardContext.ts
nturinski: I don't really like having the V2 prefixed because it makes it kind of harder to read for me.

Though actually, what about `FunctionV2WizardContext`? It's actually the Function that's V2, not the FunctionWizard. WizardContext should be one word anyway.

alexweininger: Yeah, that's better.

src/commands/createFunction/actionStepsV2/ShowMarkdownPreviewExecuteStep.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { nonNullProp } from "@microsoft/vscode-azext-utils";
+import { showMarkdownPreviewContent } from "../../../utils/textUtils";
+import { FunctionWizardV2Context } from "../FunctionV2WizardContext";
+import { ActionSchemaBaseStep } from "./ActionSchemaBaseStep";
+
+export class ShowMarkdownPreviewExecuteStep<T extends FunctionWizardV2Context> extends ActionSchemaBaseStep<T> {
```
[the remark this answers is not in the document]
nturinski: I'd opt for `showMarkdownPreviewContent` going into utils if we wanted something shared. This step relies on very specific Functions schema template objects to "function".

src/commands/createFunction/promptStepsV2/StringInputStep.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public shouldPrompt(context: T): boolean {
+        return !context[this.input.assignTo];
+    }
+
+    // value used to determine the prompt (string = input, boolean/enum = quickpick)
+    // if required is false, then add a skip for now (for quickpick only)
+
+    // resource types
```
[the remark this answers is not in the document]
nturinski: Yeah, was using them as a reference for prompt behaviors. Removed.

src/commands/createFunction/promptStepsV2/promptStepFactory.ts
nturinski: image

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurefunctions/pull/3799

PR #3754, opened 2023-07-12
Fix attaching to C# func process

### LINKED ISSUE #3748: An error occurs when debugging a C# project
**OS:** All
**Build Version:** 20230711.5
**Regression:** Preview Release

**Repro Steps:**

1. Create a C# project with "HttpTrigger" template.
2. Set a breakpoint.
3. F5
4. Check whether succeeds to debug the project.

**Expect:**
Succeed to debug the project.

**Actual:**
An error occurs.
image

**More Info:**
This issue also reproduces when debugging C# project with "Durable Function orchestrator" function.

alexweininger:
Fix #3748 

The doc string on this property states: 
image

Which is sorta confusing because it makes it sound like *they* set it to true when http is used. But turns out it just defaults to false.

It would help me a lot if you could reproing #3748 and then make sure this fixes it for you. It does for me 😄

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -97,6 +97,8 @@ async function startFuncTask(context: IActionContext, workspaceFolder: vscode.Wo
                     const statusRequest: AzExtRequestPrepareOptions = { url: `${scheme}://localhost:${funcPort}/admin/host/status`, method: 'GET' };
                     if (scheme === 'https') {
                         statusRequest.rejectUnauthorized = false;
+                    } else {
+                        statusRequest.allowInsecureConnection = true;
                     }
 
                     try {


nturinski:
I can reproduce this on our main branch and confirmed that this fixes it.

Just to clarify, the project still debugs, but never attaches. So the func host is running and the trigger will be execute, but your debugger never attaches and therefore cannot hit breakpoints.

However, listing local triggers isn't fixed due to the same error. We make this request separately for local tree items, so I'd just add it to `'LocalFunctionsTreeItem.getChildrenForHostedProjects` as well. Or maybe we could extract it out as a util function? There are a few places that use the `rejectedUnauthorized` property, though I didn't confirm exactly how they are used 🤷
image

nturinski:
Shouldn't we do that instead of merging this then?

alexweininger:
> Shouldn't we do that instead of merging this then?

Yeah I'll do it

PR #3758, opened 2023-07-13
Fix execute function command

### LINKED ISSUE #3716: Fail to execute the "Execute Function Now..." action for the deployed function
**OS:** Win 10
**Build Version:** 20230614.4

**Repro Steps:**
1. Create a JavaScript project with "HttpTrigger" template.
2. Deploy this project to a Node Function App.
3. Expand the above Function App -> Execute "Execute Function Now..." action for the "HttpTrigger" function with the default value.
4. Check whether succeeds to execute the "Execute Function Now..." action or not.

**Expect:**
Succeed to execute the "Execute Function Now..." action.
image

**Actual:**
Fail to execute the "Execute Function Now..." action with an error.
image

**More Info:**
1. This issue also reproduces for other runtime projects.
2. This issue does not reproduce on the current released extension v1.11.0.

alexweininger:
Fixes #3716 

Firstly, the new http client throws an error if the body isn't a string or stream or buffer. So I am stringifying the body.

Pasted Graphic

And also the content type needs to be set in order for service bus queue triggers to work. 

These things must have been happening automatically with the older client, so we'll need to investigate applying a better fix later.

### DIFF
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -68,12 +68,14 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
 
     let responseText: string | null | undefined;
     await node.runWithTemporaryDescription(context, localize('executing', 'Executing...'), async () => {
-        const headers = createHttpHeaders();
+        const headers = createHttpHeaders({
+            'Content-Type': 'application/json',
+        });
         if (client) {
             headers['x-functions-key'] = (await client.listHostKeys()).masterKey;
         }
         try {
-            responseText = (await requestUtils.sendRequestWithExtTimeout(context, { method: 'POST', ...triggerRequest, headers, body })).bodyAsText;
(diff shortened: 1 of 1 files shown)


alexweininger:
> These things must have been happening automatically with the older client, so we'll need to investigate applying a better fix later.

I can confirm that these things were done automatically when we used ms-rest-js.

Body is stringified by default: https://github.com/Azure/ms-rest-js/blob/master/lib/webResource.ts#L473-L475
Content type is set to JSON by default: https://github.com/Azure/ms-rest-js/blob/master/lib/webResource.ts#L450-L452

alexweininger:
So yeah new client has a default serialization policy but it's only executed when there is an operation spec. I think this means it will only be applied if the client is being used in an SDK request.

alexweininger:
It might really be worth creating our own custom request pipeline without any of the default pipeline policies. We make a lot of requests to non-Azure Resource Management API endpoints yet we are using a client with a ton of Azure SDK specific pipeline policies attached to it. 

This makes it hard to know what policies are doing what, since when they behave differently depending on if they have an operation spec.

nturinski (approved):
Add it to the backlog, I guess 🙄

Issue #3762, opened 2023-07-18
Executing a local service bus topic trigger gives "Unexpected status code: 400" error with JSON request body


TechTrailblazerX:
## Issue
Passing in valid JSON into the request body of a service bus topic trigger results in a "Unexpected status code: 400" error, and the function does not execute.
This regression appeared in version `v1.12.0`. When downgrading to `v1.11.0`, it does work as expected.

## Expected Behavior
When valid a valid JSON payload is entered in the service bus message payload prompt, the payload should be passed as a deserialized object to the service bus message trigger function argument. 
Note that passing the object in single or double quotes (for example `'{"key":"value"}'` instead of `{"key":"value"}` does work, but the json is then not parsed.

## Steps to Reproduce
1. Create function
2. Select "Azure Service Bus Topic trigger"
3. Select defaults for remaining prompts
4. Run func start
5. Execute function
6. Select created function
7. Enter valid JSON in request body prompt (`{"key":"value"}`)
8. Receive error

alexweininger:
Thank you for bringing this to our attention. Looks like this was caused by #3758.

alexweininger:
@TechTrailblazerX fix has been released in v1.12.1

PR #3763, opened 2023-07-18
Fix executing service bus topic trigger

alexweininger:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3762

Using https://www.npmjs.com/package/cross-fetch here because our own clients simply cannot form the body correctly for some reason.

cross-fetch only depends on node-fetch which we already depend on via the Azure SDKs. This just gives us easy access to it.

### DIFF
--- a/src/commands/executeFunction.ts
+++ b/src/commands/executeFunction.ts
@@ -6,6 +6,7 @@
 import { createHttpHeaders } from '@azure/core-rest-pipeline';
 import { SiteClient } from '@microsoft/vscode-azext-azureappservice';
 import { IActionContext, parseError } from '@microsoft/vscode-azext-utils';
+import fetch from 'cross-fetch';
 import { window } from 'vscode';
 import { FuncVersion } from '../FuncVersion';
 import { functionFilter } from '../constants';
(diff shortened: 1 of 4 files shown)


src/commands/executeFunction.ts
```diff
@@ -82,10 +83,15 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
                 context.errorHandling.suppressReportIssue = true;
                 throw new Error(localize('failedToConnect', 'Failed to connect. Make sure your project is [running locally](https://aka.ms/AA76v2d).'));
             } else if (errorType === '400') {
-                // stringify JSON object to match the format in the portal
-                functionInput = <{}>JSON.stringify(functionInput, undefined, 2);
-                body = { input: functionInput };
-                responseText = (await requestUtils.sendRequestWithExtTimeout(context, { method: 'POST', ...triggerRequest, headers, body: JSON.stringify(body) })).bodyAsText;
+                const response = await fetch(triggerRequest.url, {
+                    method: 'POST',
+                    body: JSON.stringify({
```
nturinski, line 88: Stringify the stringify, ehh?

nturinski (approved):

PR #3764, opened 2023-07-19
Print a warning if the function project root can't be found on deploy

mkfrey:
Inform the user that the project root could not be found instead of throwing a generic null value error and suggest a possible cause.

Currently, if no hosts.json exists in the project directory (e.g. because it was accidentally deleted or not commited), the right click -> "Deploy to ..." functionality will just return the cryptic message `Internal error: Expected value to be neither null nor undefined` on the error console.

This commit will let the extension show a warning message instead indicating that no project root could be found.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -5,9 +5,9 @@
 
 import type { SiteConfigResource } from '@azure/arm-appservice';
 import { IDeployContext, IDeployPaths, getDeployFsPath, getDeployNode, deploy as innerDeploy, showDeployConfirmation } from '@microsoft/vscode-azext-azureappservice';
-import { DialogResponses, IActionContext, nonNullValue } from '@microsoft/vscode-azext-utils';
+import { DialogResponses, IActionContext } from '@microsoft/vscode-azext-utils';
 import * as vscode from 'vscode';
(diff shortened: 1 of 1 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -45,7 +45,13 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 
     addLocalFuncTelemetry(actionContext, deployPaths.workspaceFolder.uri.fsPath);
 
-    const projectPath: string = nonNullValue(await tryGetFunctionProjectRoot(actionContext, deployPaths.workspaceFolder));
+    const projectPath: string | undefined = await tryGetFunctionProjectRoot(actionContext, deployPaths.workspaceFolder);
+    if (projectPath === undefined) {
+        const message: string = localize('functionProjectRootNotFound', 'No azure function project root could be found. This can be caused by a missing {0} file.', hostFileName);
+        await actionContext.ui.showWarningMessage(message, { stepName: "functionProjectRootNotFound" })
+        return
```
alexweininger, line 52:
```suggestion
        throw new Error(message);
```

Throwing an exception here would be best as it stops execution, and will automatically show an error notification.

alexweininger (approved):
Thank you for your contribution!

alexweininger:
@mkfrey, these change have been released in version 1.12.1 of the extension.

PR #3799, opened 2023-08-10
TemplateProvider for python v2 templates

nturinski:
Combines the feedback from both https://github.com/microsoft/vscode-azurefunctions/pull/3729 and https://github.com/microsoft/vscode-azurefunctions/pull/3742

This is a massive PR, but hopefully shouldn't touch any existing logic with our old code paths. I'll delete the readme file before I merge, but I left it in here for some context as to what/why I'm doing for parsing the new templates.


src/How to parse v2 Templates.md
alexweininger, line 1: I think we can leave this file in the codebase. It's useful. Perhaps clean it up to get it into proper "documentation" shape? Remove the personal notes, etc.

src/commands/createFunction/actionStepsV2/GetTemplateFileContentExecuteStep.ts
```diff
@@ -0,0 +1,27 @@
...
+import { nonNullProp } from "@microsoft/vscode-azext-utils";
+import { FunctionV2WizardContext } from "../FunctionV2WizardContext";
+import { ActionSchemaBaseStep } from "./ActionSchemaBaseStep";
+
+export class GetTemplateFileContentExecuteStep<T extends FunctionV2WizardContext> extends ActionSchemaBaseStep<T> {
+    public async executeAction(context: T): Promise<void> {
+        const filePath = nonNullProp(this.action, 'filePath');
+        let source: string | undefined = context.functionTemplateV2?.files[filePath];
+        if (!source) {
+            throw new Error(`Could not find file "${filePath}" in template`);
```
alexweininger, line 15: localize

src/commands/createFunction/actionStepsV2/GetTemplateFileContentExecuteStep.ts
```diff
@@ -0,0 +1,27 @@
...
+        }
+
+        // all tokens saved on the context are prefixed with '$('
+        const assignToTokens = Object.keys(context).filter(k => k.startsWith('$('));
+        for (const token of assignToTokens) {
+            source = source.replace(new RegExp(this.escapeRegExp(token), 'g'), context[token] as string);
+        }
```
alexweininger, line 22: This code is duplicated in `ReplaceTokensInTextExecuteStep`.

src/commands/createFunction/actionStepsV2/ShowMarkdownPreviewExecuteStep.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { nonNullProp } from "@microsoft/vscode-azext-utils";
+import { showMarkdownPreviewContent } from "../../../utils/textUtils";
+import { FunctionV2WizardContext } from "../FunctionV2WizardContext";
+import { ActionSchemaBaseStep } from "./ActionSchemaBaseStep";
+
+export class ShowMarkdownPreviewExecuteStep<T extends FunctionV2WizardContext> extends ActionSchemaBaseStep<T> {
+    public async executeAction(context: T): Promise<void> {
+        const filename = nonNullProp(this.action, 'filePath');
+        const content = context.functionTemplateV2?.files[filename] || '';
```
alexweininger, line 14: Any specific reason using `||` over `??` here? Just curious.

src/commands/createFunction/promptStepsV2/PromptSchemaBaseStep.ts
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+
+        // special case for existingfile/new file
+
+        return undefined;
+    }
+
+    public shouldPrompt(context: T): boolean {
+        return !context[this.input.assignTo];
```
alexweininger, line 45: If the input type is boolean, and the user inputs `false` and `context[this.input.assignTo] === false` then `shouldPrompt()` will return `true` when it should return `false` since the user has already completed this prompt.

Certainly an edge case, but I could see it causing issues.

We probably should just check if the value is not null or undefined.

src/commands/createFunction/promptStepsV2/StringInputStep.ts
```diff
@@ -0,0 +1,49 @@
...
+                if (!new RegExp(validator.expression).test(input)) {
+                    return validator.errorText;
+                }
+            }
+        }
+
+        return undefined;
+    }
+
+    public shouldPrompt(context: T): boolean {
```
alexweininger, line 46: This implementation is the same as the extended base class.

src/commands/createFunction/actionStepsV2/ActionSchemaBaseStep.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardExecuteStep } from "@microsoft/vscode-azext-utils";
+import { ParsedAction } from "../../../templates/script/parseScriptTemplatesV2";
+import { FunctionV2WizardContext } from "../FunctionV2WizardContext";
+
+export abstract class ActionSchemaBaseStep<T extends FunctionV2WizardContext> extends AzureWizardExecuteStep<T> {
```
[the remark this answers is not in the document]
nturinski: Yeah, you're right. I was thinking of it as the base for the step, but it's actually the step's base.

English 💯

src/commands/createFunction/actionStepsV2/GetTemplateFileContentExecuteStep.ts
nturinski: Yeah, I wasn't sure what to do. Technically, this step is supposed to exist, but I don't actually see it in any of the jobs. Ideally, I could remove this code and it would be done in `ReplaceTokenInTextExecuteStep`, but since it never seems to be a step, I'll just delete this with a comment in GetTemplateFileContentExecuteStep`.

src/commands/createFunction/actionStepsV2/ShowMarkdownPreviewExecuteStep.ts
nturinski: Just an old habit 😅

src/commands/createFunction/promptStepsV2/QuickPickInputStep.ts
```diff
@@ -0,0 +1,27 @@
...
+import { localize } from "../../../localize";
+import { FunctionV2WizardContext } from "../FunctionV2WizardContext";
+import { PromptSchemaStepBase } from "./PromptSchemaStepBase";
+
+export abstract class QuickPickInputStep<T extends FunctionV2WizardContext> extends PromptSchemaStepBase<T> {
+    protected async promptAction<U>(context: T): Promise<U | undefined> {
+        const picks: IAzureQuickPickItem<U | undefined>[] = await this.getPicks<U>(context);
+
+        if (!this.input.required) {
+            picks.push({
```
alexweininger, line 16: Can you utilize the skip for now constant here too?

nturinski: No, unfortunately the data types are different and I didn't think it was worth extracting it out as a constant. The QuickInputStep.ts needs to use a string, but the FunctionListStep.ts has an internal type called `TemplatePromptResult` that are specific strings that we are reading for.

It didn't feel like it was worth it to make a constant for the `Skip for now` due to that,.

src/templates/CentralTemplateProvider.ts
```diff
@@ -96,6 +97,22 @@ export class CentralTemplateProvider implements Disposable {
         }
     }
 
+
+    public async getFunctionV2Templates(context: IActionContext, projectPath: string | undefined, language: ProjectLanguage, languageModel: number | undefined, version: FuncVersion, templateFilter: TemplateFilter, projectTemplateKey: string | undefined): Promise<FunctionV2Template[]> {
+        const templates: ITemplates = await this.getTemplates(context, projectPath, language, languageModel, version, projectTemplateKey);
+        switch (templateFilter) {
+            case TemplateFilter.All:
+                return templates.functionTemplatesV2;
+            // TODO: Verify if schema V2 templates have categories
```
alexweininger, line 106: status on this todo?

src/templates/script/PysteinTemplateProvider.ts
```diff
@@ -5,172 +5,89 @@
...
-
-    public async getLatestTemplates(_context: IActionContext, _latestTemplateVersion: string): Promise<ITemplates> {
-        return await this.getBackupTemplates();
+    protected _resources: { en: { [key: string]: string } };
+    protected _rawTemplates: RawTemplateV2[];
+    protected _rawBindings: object[];
+    protected _language: string;
+
+    public async getLatestTemplates(context: IActionContext, latestTemplateVersion: string): Promise<ITemplates> {
+        // just use backup templates until template feed deploys v2 templates
```
alexweininger, line 36: update this comment

src/templates/script/parseScriptTemplatesV2.ts
```diff
@@ -0,0 +1,149 @@
...
+export function parseScriptTemplates(rawTemplates: RawTemplateV2[], rawBindings: object[], resources: Resources): FunctionV2Template[] {
+    const userPrompts: RawUserPrompt[] = parseUserPrompts(rawBindings, resources);
+    const templates: FunctionV2Template[] = [];
+    for (const templateV2 of rawTemplates) {
+        // look into jobs-- each job can be an Azure Wizard. Name is the title
+        const parsedJobs: ParsedJob[] = [];
+        for (const job of templateV2.jobs) {
+            const parsedInputs: ParsedInput[] = [];
+            job.inputs.forEach(input => {
+                // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
alexweininger, line 108: this still needed?

src/templates/script/parseScriptTemplatesV2.ts
```diff
@@ -0,0 +1,149 @@
...
+export interface ParsedJob extends RawJob {
+    parsedInputs: ParsedInput[];
+    parsedActions: ParsedAction[];
+}
+
+export function parseScriptTemplates(rawTemplates: RawTemplateV2[], rawBindings: object[], resources: Resources): FunctionV2Template[] {
+    const userPrompts: RawUserPrompt[] = parseUserPrompts(rawBindings, resources);
+    const templates: FunctionV2Template[] = [];
+    for (const templateV2 of rawTemplates) {
+        // look into jobs-- each job can be an Azure Wizard. Name is the title
```
alexweininger, line 103: update/remote

src/utils/bundleFeedUtils.ts
```diff
@@ -26,22 +28,34 @@ export namespace bundleFeedUtils {
...
     }
 
-    export async function getLatestTemplateVersion(context: IActionContext, bundleMetadata: IBundleMetadata | undefined): Promise<string> {
+    export interface ITemplatesReleaseV1 extends ITemplatesReleaseBase {
+        bindings: string;
+    }
+
+    export interface ITemplatesReleaseV2 extends ITemplatesReleaseBase {
+        userPrompts: string;
+        // it is not supposed to exist in the v2 schema, but sometimes userPrompts accidentally gets replaced with bindings
```
alexweininger, line 50: link to a GitHub issue for this?

src/utils/bundleFeedUtils.ts
```diff
@@ -50,16 +64,34 @@ export namespace bundleFeedUtils {
...
 
-    export function isBundleTemplate(template: IFunctionTemplate | IBindingTemplate): boolean {
+    export async function getReleaseV2(context: IActionContext, bundleMetadata: IBundleMetadata | undefined, templateVersion: string): Promise<ITemplatesReleaseV2> {
+        const feed: IBundleFeed = await getBundleFeed(context, bundleMetadata);
+        return feed.templates.v2[templateVersion];
+    }
+
+    export function isBundleTemplate(template: IFunctionTemplate | IBindingTemplate | FunctionV2Template): boolean {
         const bundleTemplateTypes: string[] = ['durable', 'signalr'];
+        // I don't think v2 templates are ever a bundle template? Will have to verify this
```
alexweininger, line 79: What's the verdict

src/templates/CentralTemplateProvider.ts
nturinski: They do not. We could still have All vs. Verified (once we add tests) but I'm not sure if there's much value in that. I may just remove the filter for V2

src/utils/bundleFeedUtils.ts
ejizba: gif caught my attention 😅

Yeah they should be bundle templates. For example, we should have a different template for a Cosmos DB trigger depending if it's a newer bundle or old bundle. Here are examples of how the code is different:
- New bundle: https://github.com/Azure/azure-functions-nodejs-e2e-tests/blob/main/app/v4/src/functions/cosmosDBTrigger1.ts
- Old bundle: https://github.com/Azure/azure-functions-nodejs-e2e-tests/blob/main/app/v4-oldBundle/src/functions/cosmosDBTrigger1.ts

If this isn't the case, we can have a chat with Naren and I'm sure he'll sort us out.

nturinski: Sorry, what I meant is, I assume that they are all "true" and since it wouldn't be included in the v1 runtime, we don't ever call it in the first place.

So nah, we don't need this.

alexweininger (dismissed):

alexweininger (approved):

PR #3831, opened 2023-09-07
Use Node 18.15

alexweininger:
VS Code has updated to Node 18.15: https://code.visualstudio.com/updates/v1_82#_electron-25-update


bwateratmsft:
We should raise the minimum VSCode version in package.json as well

.azure-pipelines/common/build.yml
```diff
@@ -1,8 +1,8 @@
 steps:
 - task: NodeTool@0
-  displayName: 'Use Node 16.x'
+  displayName: 'Use Node 18.x'
   inputs:
-    versionSpec: 16.x
+    versionSpec: 18.x
```
bwateratmsft, line 5: An alternative that puts version into exactly one place (.nvmrc):
steps:
```yaml
  - task: NodeTool@0
    displayName: Using Node.js
    inputs:
      versionSource: fromFile
      versionFilePath: .nvmrc
```

alexweininger: Oh nice. Is that new? I thought I looked for this option when I added the .nvmrc files

bwateratmsft: I'm not sure

package.json
```diff
@@ -1150,7 +1150,7 @@
         "@types/node": "^16.18.36",
         "@types/ps-tree": "^1.1.0",
         "@types/semver": "^5.5.0",
-        "@types/vscode": "1.79.0",
+        "@types/vscode": "^1.81.0",
```
alexweininger, line 1153: This is the latest types version. Not sure why 1.82.0 doesn't exist.

bwateratmsft: I've seen that before where the `@types/vscode` package lags behind VSCode itself

bwateratmsft (approved):

Issue #3839, opened 2023-09-12
Fail to debug Python Model V2 project


v-chenzha:
**OS:** All
**Build Version:** 20230911.21
**Regression:** Previous release (1.12.4)

**Repro Steps:**

1. Create and open a Python Model V2 project in VS Code.
2. F5.
3. Check whether succeeds to debug.

**Expect:**
Succeed to debug.

**Actual:**
Fail to debug.
<img width="279" alt="image" src="https://github.com/microsoft/vscode-azurefunctions/assets/107091828/fec69499-5aa1-4939-af3b-335562aacc40">
<img width="941" alt="image" src="https://github.com/microsoft/vscode-azurefunctions/assets/107091828/1e546a2f-e296-4912-868e-f7ecd0da5e9e">

Issue #3838, opened 2023-09-12
The "CosmosDB trigger" template shows in the templates list when creating a Python V2 project


v-xinda:
**OS:** All
**Build Version:** 20230911.21
**Regression:** Not a regression

**Repro Steps:**
1. F1 -> Azure Functions: Create New Project... -> Select a folder -> Select "Python" language -> Select Model V2 -> Select a template -> Type default name -> Select "ANONYMOUS" -> Open in current window.
2. F1 -> Select "Azure Functions: Create Function...".
3. Check whether the "CosmosDB trigger" template doesn't show in the templates list.

**Expect:**
The "CosmosDB trigger" template doesn't show in the templates list.

**Actual:**
The "CosmosDB trigger" template shows in the templates list.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/57662131/d568966d-14c4-4f39-a6da-538709db2bfb)

**More Info:**
This issue also reproduces when creating a new python Model V2 project, if a python Model V2 project opened in the VSCode previously.

Issue #3837, opened 2023-09-12
No template shows in the dropdown list when creating a JavaScript/TypeScript Model V4 project


v-xinda:
**OS:** All
**Build Version:** 20230911.21
**Regression:** Previous release (1.12.4)

**Repro Steps:**
1. F1 -> Azure Functions: Create New Project... -> Select a folder -> Select "JavaScript" language -> Select Model V4.
2. Check whether the templates show in the dropdown list.

**Expect:**
The templates show in the dropdown list.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/57662131/4b53e2ad-ad6c-4153-b32c-5ca90d802e95)

**Actual:**
No template shows in the dropdown list.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/57662131/ee551409-db9d-44b3-8f05-874598fd6220)

**More Info:**
This issue also reproduces for the TypeScript Model V4 project.

PR #3842, opened 2023-09-12
Refactor template providers to check template schema, not language model

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3837
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3839
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3838

I'll be honest, when I first started this, I thought this was a much better idea. But the implementation isn't as clean as I was hoping, but I think it's an important distinction.

`Language model =/= Template Schema`!!

Originally, a lot of the checks were just looking at language model, but there isn't actually a 1:1 correlation between template schema and language models. 

This became a problem when some programming models were on 3/4 while others are at 1/2. You can't do a numeric check anyway, because Node.js programming model 3/4 is higher than 1, but Python programming model 2 needs to use v2 schemas.

This is obvious now when you look at the issue above where we're using the new programming model for Node.js, but it's still on v1 templates. I hope we don't have to do deal with this, but there also may be a case where templates provide both v1 and v2 schemas for a runtime (god save us all).

Anyway, the idea is with project creation, we know the language and language model. Right now it's simple-- if it's not Python on v2, then it is a v1 schema language. As more templates adopt the new schema, we can just update that mapping and hopefully the code will just work. 🤞

Also, deleted some old Python project create code that was specifically for the hardcoded v2 approach. They aren't used anymore.


src/commands/createFunction/FunctionListStep.ts
```diff
@@ -62,6 +66,10 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
     }
 
     public async getSubWizard(context: IFunctionWizardContext): Promise<IWizardOptions<IFunctionWizardContext> | undefined> {
+        if (context.functionTemplate?.templateSchemaVersion === TemplateSchemaVersion.v2) {
```
nturinski, line 69: It actually makes more sense for this do be pushed in here. The `JobsListStep` actually depends on there being a function template to iterate its `jobs` array.

alexweininger (approved):

PR #3871, opened 2023-09-26
Update changelog from patch releases

alexweininger:


nturinski (approved):

nturinski:
Note: we also need to bump the appservice package from the patch version

PR #3873, opened 2023-09-29
Add `listLocalProjects` function

alexweininger:
Goals:
* Separate the local project model from the local project tree item
* Create a standalone function that lists local projects that can be used later for the Limelight API

Non goals:
* Change any behavior

Created a `LocalProject` class that can live underneath `LocalProjectTreeItem`. We can later on pass this class to the Limelight extension so they can wrap it in their own UI.

I will follow up this PR with the same code changes but for local functions.

### DIFF
--- a/src/workspace/listLocalProjects.ts
+++ b/src/workspace/listLocalProjects.ts
@@ -1,0 +1,119 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IActionContext, callWithTelemetryAndErrorHandling } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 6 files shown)


src/workspace/listLocalProjects.ts
```diff
@@ -0,0 +1,119 @@
...
+interface UnitializedLocalProject {
+    workspaceFolder: vscode.WorkspaceFolder;
+    projectPath: string;
+}
+
+interface InvalidLocalProject extends UnitializedLocalProject {
+    error: unknown;
+}
+
+interface ListWorkspaceProjectsResult {
```
nturinski, line 27: It's not immediately obvious to me that these are all exclusive arrays. I think it'd be better to rename it to `initalizedProjects` or maybe have another array that _is_ all of the projects, regardless if it is invalid or uninitalized.

nturinski (approved):
Looks very clean. Like the way that you did this quite a bit!

PR #3905, opened 2023-11-21
Update backup templates

hossam-nasr:
As in title, update the backup templates to catch up with latest templates.


nturinski (approved):
Looks good 👍

nturinski:
We know that the tests failing are a false positive right now, so I'm going to just force merge this one.

PR #3908, opened 2023-11-23
Add version numbers to duplicate template IDs

nturinski:
Specifically, there are 3 types of CosmosDB triggers for C# floating around. Staging templates have 2.x and 4.x and backup/latest have 3.x and 4.x.

It's a bit of a pain to handle, so I'm only showing 4.x templates in verified. Part of this decision was driven that we should probably only recommend the latest version, but it also made our testing suite hard to deal with when there were different choices across latest/staging/backup.


alexweininger:
Does this address a specific issue? Can you link it if so?

alexweininger (approved):

nturinski:
I'm closing this for now. I'm not consistently seeing this reproduce anymore...

PR #3916, opened 2023-11-29
Add EventGrid Cloud Event templates for .NET to verified list

hossam-nasr:
As in title, add the new "EventGrid Cloud Event" templates to the verified list. The only tests failing now should be the TypeScript template count and the .NET isolated back up template count, which do not appear to be related to my changes.


alexweininger (approved):
Nice!

nturinski (approved):
The tests that are currently failing are unrelated to Hossam's changes. Latest version of TypeScript templates don't include the CosmosDB trigger for some reason. I've reached out to the partner team about it already.

PR #3924, opened 2023-12-13
Improve local settings when creating templates

nturinski:
I added the value to the list of local settings. I think it's helpful to see what those are to give more context.  Also letting users select emulator vs. remote when adding a new local setting for a storage connection string.

image

I still need to do this for the v2 schema since they don't use bindings anymore. It should be pretty similar, but I'll do it in a different PR.

Also `setAzureWebJobsStorage` is never actually invoked anywhere, so I went ahead and deleted that.


src/commands/addBinding/settingSteps/AzureConnectionCreateStepBase.ts
```diff
@@ -33,7 +33,10 @@ export abstract class AzureConnectionCreateStepBase<T extends IBindingWizardCont
         progress.report({ message: localize('retrieving', 'Retrieving connection string...') });
 
         const result: IConnection = await this.getConnection(context);
-        let appSettingKey: string = `${result.name}_${nonNullProp(this._setting, 'resourceType').toUpperCase()}`;
+        let appSettingKey: string = context.useStorageEmulator ?
+            // if using the storage emulator, don't append the resource type to the app setting key
```
alexweininger, line 37: nit: The comment explains what we're doing but not why. I think it's more helpful if the why was included.

alexweininger (approved):

src/commands/addBinding/settingSteps/AzureConnectionCreateStepBase.ts
nturinski: To be honest, I'm not really sure why @ejizba did this in the first place.

https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#azurewebjobsstorage

Here's the documentation about the setting `AzureWebJobsStorage` which is the default app setting linking to a storage account.

The old behavior would take the name of the azure resource and append the resource type (so like, naturins-storage_STORAGE) as the key. I'm wondering if `AzureWebJobsStorage` didn't exist at that time so he was just finding a common convention for the key?

package.json
```diff
@@ -276,11 +276,6 @@
                 "title": "%azureFunctions.restartFunctionApp%",
                 "category": "Azure Functions"
             },
-            {
-                "command": "azureFunctions.setAzureWebJobsStorage",
```
ejizba, line 280: > Also setAzureWebJobsStorage is never actually invoked anywhere, so I went ahead and deleted that.

I use it a decent amount please don't remove it 🥺 If you clone a function app repo or just accidentally delete your "local.settings.json" it's a fast way to get the file back

src/commands/addBinding/settingSteps/LocalAppSettingListStep.ts
```diff
@@ -33,10 +32,11 @@ export class LocalAppSettingListStep extends BindingSettingStepBase {
     public async promptCore(context: IBindingWizardContext): Promise<BindingSettingValue> {
         const localSettingsPath: string = path.join(context.projectPath, localSettingsFileName);
         const settings: ILocalSettingsJson = await getLocalSettingsJson(context, localSettingsPath);
-        const existingSettings: string[] = settings.Values ? Object.keys(settings.Values) : [];
+        const existingSettings: [string, string][] = settings.Values ? Object.entries(settings.Values) : [];
+
         let picks: IAzureQuickPickItem<string | undefined>[] = [{ label: localize('newAppSetting', '$(plus) Create new local app setting'), data: undefined }];
-        picks = picks.concat(existingSettings.map((s: string) => { return { data: s, label: s }; }));
-        const placeHolder: string = localize('selectAppSetting', 'Select setting from "{0}"', localSettingsFileName);
+        picks = picks.concat(existingSettings.map((s: [string, string]) => { return { data: s[0], label: s[0], description: s[1] }; }));
```
ejizba, line 38: > I added the value to the list of local settings.

The value can be a secret, though. You could add an option or something to show values, but they should be hidden by default just to be safe. Just like how app settings are currently displayed in the tree view

src/commands/addBinding/settingSteps/AzureConnectionCreateStepBase.ts
ejizba: Not sure if he's talking about what I did or what you did @nturinski. The "_STORAGE" thing was just a random naming convention and I feel like I was probably just following an existing pattern from docs or the portal

In any case, I'm confused about what you're doing just like Alex probably is. Like, what does the emulator have to do with appending "_STORAGE"? Seems like both "AzureWebJobsStorage" and "randomkey_STORAGE" could use or not use the emulator, right?

nturinski: The issue is when I pass in `AzureWebJobsStorage` as the app key, it'll automatically append _STORAGE to the end of it.  You can see it in my screenshot where that happened.

Is it pretty common to have other storage connection strings outside of `AzureWebJobsStorage`?

src/commands/addBinding/settingSteps/LocalAppSettingListStep.ts
nturinski: image
image

package.json
nturinski: Hahaha okay. I figured the only entry point was through the command palette, so who could even be using it...? But I guess you are 1 of 85 people per month.

image

nturinski:
Not sure who's approval I'm seeking now...

src/commands/registerCommands.ts
```diff
@@ -93,7 +92,6 @@ export function registerCommands(): void {
     registerCommandWithTreeNodeUnwrapping('azureFunctions.pickProcess', pickFuncProcess);
     registerCommandWithTreeNodeUnwrapping('azureFunctions.redeploy', redeployDeployment);
     registerCommandWithTreeNodeUnwrapping('azureFunctions.restartFunctionApp', restartFunctionApp);
-    registerCommandWithTreeNodeUnwrapping('azureFunctions.setAzureWebJobsStorage', setAzureWebJobsStorage);
```
ejizba, line 96: Don't we still need this for the command palette to work?

src/commands/appSettings/connectionSettings/getLocalConnectionSetting.ts
```diff
@@ -21,11 +21,17 @@ export interface IResourceResult {
     connectionString: string;
 }
 
-export async function getStorageConnectionString(context: IStorageAccountWizardContext): Promise<IResourceResult> {
+export async function getStorageConnectionString(context: IStorageAccountWizardContext & { useStorageEmulator?: boolean }): Promise<IResourceResult> {
+    if (context.useStorageEmulator) {
+        return {
+            name: 'AZURITE',
```
ejizba, line 27: I feel like "AZURITE" as a name isn't ideal because the same name is used when you deploy to Azure and you obviously can't use Azurite in Azure. I'd pick a more generic name that makes sense for both local and remote scenarios

Some random ideas:

- connection1_STORAGE
- {randomHex}_STORAGE (for example a8b382_STORAGE)
- {functionName}Connection (for example blobTrigger1Connection)
- storageConnection1

ejizba (commented):
> Not sure who's approval I'm seeking now...

Idk, but don't block on me. I'm skim-reviewing not actual-reviewing

src/commands/appSettings/connectionSettings/getLocalConnectionSetting.ts
ejizba: To be clear: my suggestions were for the end result of the app setting name, not necessarily this particular "name" variable as I'm not exactly sure how it's used.

src/commands/registerCommands.ts
nturinski: You are absolutely correct. Forgot to revert that one. 🤦

src/commands/appSettings/connectionSettings/getLocalConnectionSetting.ts
nturinski: Going with randomHex because it's easiest and don't need to keep track of duplicates or anything like that,

src/commands/appSettings/connectionSettings/getLocalConnectionSetting.ts
```diff
@@ -21,11 +21,17 @@ export interface IResourceResult {
     connectionString: string;
 }
 
-export async function getStorageConnectionString(context: IStorageAccountWizardContext): Promise<IResourceResult> {
+export async function getStorageConnectionString(context: IStorageAccountWizardContext & { useStorageEmulator?: boolean }): Promise<IResourceResult> {
+    if (context.useStorageEmulator) {
+        return {
+            name: randomUtils.getRandomHexString(6),
```
alexweininger, line 27: I think I prefer Eric's suggestion where we do random hex string + `_STORAGE`. I think it's worth adding the label on there so it's more clear what this is for.

nturinski: It does do that still. It's confusing, but it happens here: https://github.com/microsoft/vscode-azurefunctions/pull/3924/files/ab23bf1afd5053a427989abb73574fbc13ad9094#diff-50e86af002b096abb16ed6118f7678e5f2a914a1985c9c4706b8674a77047c87R39

alexweininger (approved):

Issue #3523, opened 2023-01-20
VSCode support for deploying Functions on App Environment (Centauri)


raorugan:
Provide VSCode tooling support for creating and managing  Function apps on Managed App environment(Centauri) which is a Azure container apps based platform.

VScode to support for following experiences - 
Create/View/Delete function app on App environment
Showing the resource - Function App container created
Create Functions - (Triggers/Bindings)

Mocks will be updated soon!

MicroFish91:
@mistymadonna Can you please work with raorugan to clarify the requirements and prioritize this request for our next semester

mistymadonna:
@raorugan @MicroFish91 Just emailed for a status update on this issue.

PR #3929, opened 2023-12-20
Add support for creating functions projects with dockerfiles 

motm32:
Contributes to #3523 

Currently the entry point is next to the current "Create New Project..." command in the workspace view: 
image


src/commands/createNewProject/dockerfileSteps/IDockerfileProjectContext.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type ExecuteActivityContext, type IActionContext } from "@microsoft/vscode-azext-utils";
+import { type IProjectWizardContext } from "../IProjectWizardContext";
+
+export interface IDockerfileProjectContext extends IActionContext, ExecuteActivityContext, IProjectWizardContext {
+    projectLanguage?: string;
```
nturinski, line 10: `IProjectWizardContext` already should have a `language: ProjectLanguage` property that you should be able to leverage.

src/commands/createNewProject/dockerfileSteps/DockefileProjectLanguageStep.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, type IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import { ProjectLanguage } from "../../../constants";
+import { type IDockerfileProjectContext } from "./IDockerfileProjectContext";
+
+export class DockerfileProjectLanguageStep extends AzureWizardPromptStep<IDockerfileProjectContext> {
```
nturinski, line 10: You could probably just use the `NewProjectLanguageStep` and add some filtering in the case for `DockerfileProjectLanguageStep` instead of creating a new file that does largely the same thing.

src/commands/createNewProject/dockerfileSteps/DockefileProjectLanguageStep.ts
```diff
@@ -0,0 +1,26 @@
...
+    public async prompt(context: IDockerfileProjectContext): Promise<void> {
+        const language: IAzureQuickPickItem<string>[] = [
+            { label: ProjectLanguage.JavaScript, data: 'node' },
```
MicroFish91, line 13: Should `data` be `javascript` here?

In which case would it be appropriate to use the `ProjectLanguage` enum value as the data value as well?

motm32: Yeah it should be javascript! On the docs they also had node (javascript) not sure exactly but I'll use javascript since they use that in the examples. The problem with the `ProjectLanguage` enum value is since I am using the data property directly on this [line](https://github.com/microsoft/vscode-azurefunctions/pull/3929/files#diff-7ff0c87d39e4f588780460a9ca5278a3004ee49ae26c04b28b6b7fa1fdc0f958:~:text=await%20cpUtils.executeCommand(ext.outputChannel%2C%20nonNullValueAndProp(context%2C%20%27projectPath%27)%2C%20%22func%22%2C%20%22init%22%2C%20%22%2D%2Dworker%2Druntime%22%2C%20nonNullValueAndProp(context%2C%20%27projectLanguage%27)%2C%20%22%2D%2Ddocker%22)) using the data value won't work since they are different.

src/commands/createNewProject/dockerfileSteps/DockefileProjectLanguageStep.ts
motm32: Actually update since I am not using the NewProjectLanguageStep I don't think I can use the `language: ProjectLanguage` since the data types don't match up.

nturinski: Yeah but you should be able to just change the type instead of using a string.

nturinski: Looking at this more, I think we should just speak offline about this.

package.nls.json
```diff
@@ -18,6 +18,7 @@
     "azureFunctions.createFunctionAppAdvanced": "Create Function App in Azure... (Advanced)",
     "azureFunctions.createFunctionAppDetail": "For serverless, event driven apps and automation.",
     "azureFunctions.createNewProject": "Create New Project...",
+    "azureFunctions.createNewProjectWithDockerfile": "Create New Project With Dockerfile...",
```
nturinski, line 21: Maybe we should call it  `Create New Containerized Project...`?

I don't have a strong preference, but the learn document does refer to it as a containerized function app.

https://learn.microsoft.com/en-us/azure/azure-functions/functions-deploy-container?tabs=acr%2Cbash%2Cazure-cli&pivots=programming-language-javascript

src/commands/createNewProject/IProjectWizardContext.ts
```diff
@@ -29,6 +29,8 @@ export interface IProjectWizardContext extends IActionContext {
     openApiSpecificationFile?: Uri[];
 
     targetFramework?: string | string[];
+
+    dockerfile?: boolean;
```
nturinski, line 33: Property name is a bit odd for a boolean. If I see `dockerfile`, I think it's the dockerfile path or Uri. I would have called it `addDockerfile` or something like that.

But that being said, I don't think you need this property anymore. I'm assuming it used to be used to determine whether or not to add the `CreateDockerfileProjectStep` but we've covered that in another way.

src/commands/createNewProject/dockerfileSteps/CreateDockerfileProjectStep.ts
```diff
@@ -0,0 +1,33 @@
...
+import { validateFuncCoreToolsInstalled } from "../../../funcCoreTools/validateFuncCoreToolsInstalled";
+import { localize } from "../../../localize";
+import { cpUtils } from "../../../utils/cpUtils";
+import { type IFunctionWizardContext } from "../../createFunction/IFunctionWizardContext";
+
+export class CreateDockerfileProjectStep extends AzureWizardExecuteStep<IFunctionWizardContext>{
+    public priority: number = 100;
+
+    public async execute(context: IFunctionWizardContext): Promise<void> {
+        const message: string = localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to run this command.');
```
nturinski, line 17: I mentioned this in another PR, but you should front-load this error. Nothing more annoying than going through a project wizard and then having it fail right at the end.

src/commands/createNewProject/createNewProject.ts
```diff
@@ -54,6 +56,13 @@ export async function createNewProjectInternal(context: IActionContext, options:
     const wizardContext: Partial<IFunctionWizardContext> & IActionContext = Object.assign(context, options, { language, version: tryParseFuncVersion(version), projectTemplateKey });
     const optionalExecuteStep = options.executeStep;
 
+    if (optionalExecuteStep instanceof CreateDockerfileProjectStep) {
+        const message: string = localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to run this command.');
```
nturinski, line 60: nit: Since we know what command they are running, I think we should tell them specifically that they need to have the core tools installed to create a containerized function project or whatever.

nturinski (approved):
It's a just a nit, so if you disagree, feel free to just merge! Looks good!

Issue #3932, opened 2024-01-02
Typos and issues in blueprint template for python (HttpTrigger-python) 


prise6:
Version of extension: v1.13.1

Step to reproduce (for example):
1. Create Functions
2. HTTP Trigger
3. Create New Blueprint file 
4. Name: `name_of_file.py`
5. Name of function: `name_of_function`
6. Authorization: FUNCTION 

This will generate somethin wrong : 

```python
# Register this blueprint by adding the following line of code 
# to your entry point file.  
# app.register_functions(name_of_file) 
# 
# Please refer to https://aka.ms/azure-functions-python-blueprints


import azure.functions as func
import logging

name_of_file = func.Blueprint()


@name_of_file.route(route="name_of_file, auth_level=func.AuthLevel.FUNCTION")
def name_of_file(req: func.HttpRequest) -> func.HttpResponse:
# ...
```

[Looking at the template](https://github.com/microsoft/vscode-azurefunctions/blob/c30d8cecafc2968c7c6b19b594eee98d73f7a0bd/resources/backupTemplates/pystein/templates-v2/templates.json#L965C1-L966C1), in key `blueprint_body.py`:

* `$(BLUEPRINT_FILENAME)` is used as definition of function instead of `$(FUNCTION_NAME_INPUT)`
* final quote for `route` argument is placed after `auth_level` argument resulting in a error like above example.

Final result should be, i guess:

```python
# Register this blueprint by adding the following line of code 
# to your entry point file.  
# app.register_functions(name_of_file) 
# 
# Please refer to https://aka.ms/azure-functions-python-blueprints


import azure.functions as func
import logging

name_of_file = func.Blueprint()


@name_of_file.route(route="name_of_function", auth_level=func.AuthLevel.FUNCTION)
def name_of_function(req: func.HttpRequest) -> func.HttpResponse:
# ...
```
 
So template could be replaced by:
```json
[...]

PR #3933, opened 2024-01-02
Fix typo template in python v2 (HttpTrigger)

prise6:
Proposal fixing for issue  #3932


resources/backupTemplates/pystein/templates-v2/templates.json
```diff
@@ -962,7 +962,7 @@
...
+            "blueprint_body.py": "\r\n@$(BLUEPRINT_FILENAME).route(route=\"$(FUNCTION_NAME_INPUT)\", auth_level=func.AuthLevel.$(AUTHLEVEL_INPUT))\r\ndef $(FUNCTION_NAME_INPUT)(req: func.HttpRequest) -> func.HttpResponse:\r\n    logging.info('Python HTTP trigger function processed a request.')\r\n\r\n    name = req.params.get('name')\r\n    if not name:\r\n        try:\r\n            req_body = req.get_json()\r\n        except ValueError:\r\n            pass\r\n        else:\r\n            name = req_body.get('name')\r\n\r\n    if name:\r\n        return func.HttpResponse(f\"Hello, {name}. This HTTP triggered function executed successfully.\")\r\n    else:\r\n        return func.HttpResponse(\r\n             \"This HTTP triggered function executed successfully. Pass a name in the query string or in the request body for a personalized response.\",\r\n             status_code=200\r\n        )",
```
nturinski, line 965: This is correct, but the issue is that we're retrieving these templates from a feed here:  https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/index-v2.json

These templates are just backups in case we couldn't retrieve the templates. We don't typically manually edit these files. We have a script that caches everything so the next time we do that, this fix would get overwritten.

It seems like that this should be fixed now: https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates-v2/HttpTrigger-Python/blueprint_body.py#L3

But maybe they haven't released the new templates from their repo yet.

nturinski (commented):
This is correct, but the issue is that we're retrieving these templates from a feed here:  https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/index-v2.json

These templates are just backups in case we couldn't retrieve the templates. We don't typically manually edit these files. We have a script that caches everything so the next time we do that, this fix would get overwritten.

It seems like that this should be fixed now: https://github.com/Azure/azure-functions-templates/blob/dev/Functions.Templates/Templates-v2/HttpTrigger-Python/blueprint_body.py#L3

But maybe they haven't released the new templates from their repo yet.

nturinski:
I'm going to ahead and close this since we don't usually update the backup templates manually like this.

PR #3943, opened 2024-01-19
Add containerized function app tree items

motm32:
Contributes to #3523 

Since containerized function apps sites are different enough from regular function apps we decided it made more sense to make seperate tree items that do not utilize `ParsedSite`.

Here is how the tree looks for these items:
image

The functionality for the tree items should be similar except:

-  log streaming functions which is not supported.
- clicking a function to view properties is not supported but right click view properties does work

Note: In `SubscriptionTreeItem` the dockerfilepath property was added in order for their not to be typing errors but it is not used in this PR. This property will be used in the


src/tree/containerizedFunctionApp/FunctionTreeItem.ts
```diff
@@ -0,0 +1,54 @@
...
+import { DialogResponses, nonNullValueAndProp, type IActionContext } from "@microsoft/vscode-azext-utils";
+import { type ViewPropertiesModel } from "@microsoft/vscode-azureresources-api";
+import { ProgressLocation, window } from "vscode";
+import { ext } from "../../extensionVariables";
+import { localize } from "../../localize";
+import { FunctionTreeItemBase } from "../FunctionTreeItemBase";
+import { type FunctionItem } from "./FunctionItem";
+import { type FunctionsTreeItem } from "./FunctionsTreeItem";
+
+export class FunctionTreeItem extends FunctionTreeItemBase {
```
nturinski, line 16: We should name this something like `ContainerFunctionTreeItem` or something to help identify what it is. 

Of course, you can look at the folder name and figure that out, but I think it's still helpful to specify so we can tell immediately. `LocalFunctionTreeItem` and `RemoteFunctionTreeItem` do the same thing.

src/tree/containerizedFunctionApp/FunctionsTreeItem.ts
```diff
@@ -0,0 +1,104 @@
...
+import * as retry from 'p-retry';
+import { ParsedFunctionJson } from "../../funcConfig/function";
+import { localize } from "../../localize";
+import { FunctionsTreeItemBase } from "../FunctionsTreeItemBase";
+import { getFunctionNameFromId } from "../remoteProject/RemoteFunctionTreeItem";
+import { type ContainerTreeItem } from "./ContainerTreeItem";
+import { FunctionItem } from "./FunctionItem";
+import { FunctionTreeItem } from "./FunctionTreeItem";
+
+export class FunctionsTreeItem extends FunctionsTreeItemBase {
```
nturinski, line 19: Same as above.

src/tree/containerizedFunctionApp/FunctionsTreeItem.ts
```diff
@@ -0,0 +1,104 @@
...
+        await ti.refreshImpl(context);
+        return ti;
+    }
+
+    public hasMoreChildrenImpl(): boolean {
+        return !!this._nextLink;
+    }
+
+    public async refreshImpl(context: IActionContext): Promise<void> {
+        this.isReadOnly = await this.parent.isReadOnly(context);
```
nturinski, line 39: You probably don't need to do this as the containerized functions should always be read-only, right?

src/tree/containerizedFunctionApp/FunctionsTreeItem.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    public async loadMoreChildrenImpl(clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        if (clearCache) {
+            this._nextLink = undefined;
+        }
+        /*
+            Related to issue: https://github.com/microsoft/vscode-azurefunctions/issues/3179
```
nturinski, line 47: Do you know if you actually need all of this extra retry logic here? I'm not sure if the containers have the same "BadGateway" errors and/or have a warm-up function.

It doesn't really hurt to have it, but it'd be nice to be able to clean up this file a bit.

src/tree/containerizedFunctionApp/ImageTreeItem.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        return [
+            new GenericTreeItem(this, {
+                label: localize('registry', 'Registry'),
+                description: loginServer,
+                contextValue: 'registry',
+                iconPath: new ThemeIcon('dash')
+            }),
+            new GenericTreeItem(this, {
+                label: localize('imageandTag', 'Image and Tag'),
```
nturinski, line 29: nit: I'm a little conflicted about this... I personally hate `Image and Tag` as the label (fully recognizing that this is what the portal does in the settings)

image

What do you think about just calling this node the "ID" or something? The image ID + tag is another identifier to point to a specific image (sort of like a commit) so I think that also makes sense.

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase.ts
```diff
@@ -0,0 +1,150 @@
...
+        this.appSettingsTreeItem = new AppSettingsTreeItem(proxyTree, new ContainerAppSettingsClientProvider(proxyTree, proxyTree.subscription), ext.prefix, {
+            contextValuesToAdd: ['azFunc']
+        });
+
+        const children: AzExtTreeItem[] = [this._functionsTreeItem, this._imageTreeItem, this.appSettingsTreeItem]
+
+        return children;
+    }
+
+    public async isReadOnly(context: IActionContext): Promise<boolean> {
```
nturinski, line 90: You probably don't need to implement this the same way. I assume that this can just be true.

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase.ts
```diff
@@ -0,0 +1,150 @@
...
+    private _functionsTreeItem: FunctionsTreeItem;
+    private _imageTreeItem: ImageTreeItem;
+
+    private _cachedVersion: FuncVersion | undefined;
+    private _cachedHostJson: IParsedHostJson | undefined;
+
+    public readonly source: ProjectSource = ProjectSource.Remote;
+
+    public constructor(site: Site) {
+        this.site = site;
```
nturinski, line 39: There's a bunch of things that we mask when we create the `ResolvedFunctionAppResource`. I believe that you should be able to add most of thoss properties here as well, but you'll have to get it from the `Site` rather than `parsedSite`.

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase.ts
```diff
@@ -0,0 +1,150 @@
...
+import { treeUtils } from "../../utils/treeUtils";
+import { type ApplicationSettings, type FuncHostRequest } from "../IProjectTreeItem";
+import { ProjectSource } from "../projectContextValues";
+import { ContainerAppSettingsClientProvider } from "./AppSettingsClient";
+import { type ContainerTreeItem } from "./ContainerTreeItem";
+import { FunctionsTreeItem } from "./FunctionsTreeItem";
+import { ImageTreeItem } from "./ImageTreeItem";
+
+export class ResolvedContainerizedFunctionAppResource implements ResolvedAppResourceBase {
+    public site: Site;
```
nturinski, line 26: I recommend keeping this as the `data` property rather than naming it `site`.

Some of our common functions rely on a `data` property existing (it's usually the JSON representation of an Azure resource).

Also, because this is so similar to `ResolvedFunctionAppResource`, I think it'd be easier to understand if the properties were named the same and just forego having a `site` property.

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase.ts
```diff
@@ -0,0 +1,150 @@
...
+        const client = await (new ContainerAppSettingsClientProvider(proxyTree, proxyTree.subscription).createClient(context));
+        const appSettings: StringDictionary = await client.listApplicationSettings();
+        if (!appSettings.properties) {
+            appSettings.properties = {};
+        }
+        appSettings.properties[key] = value;
+        await client.updateApplicationSettings(appSettings);
+    }
+}
+
```
nturinski, line 150: You'll need to implement `pickTreeItemImpl` for when people do things in the command palette.

nturinski (commented):
I'm okay to keep the files the way that they are. The app settings refactoring was the most important to me (and you did a great job with it) and I'm glad that you were able to implement the `IProjectTreeItem`.

But there's still quite a bit of repeated code specifically in the `ResolvedContainerizedFunctionAppResourceBase` and `ContainerTreeItem` that we could _probably_ eliminate if you wanted to continue chipping away at it.

I'm okay to leave it as it is though, since it may not be worth the effort since we haven't allocated any more bandwidth to continue the work, but if you're not sick and tired of this yet and want to try, I'm happy to sit down and help you with it.

src/commands/browseWebsite.ts
```diff
@@ -16,5 +16,10 @@ export async function browseWebsite(context: IActionContext, node?: SlotTreeItem
         });
     }
 
-    await openUrl(node.site.defaultHostUrl);
+    try {
+        await openUrl(node.site.defaultHostUrl);
+    } catch {
+        // containerized function apps don't have a default host url
+        await openUrl(`https://${node.site.defaultHostName}`)
```
nturinski, line 23: I think we should just create the `defaultHostUrl` the same way the `ParsedSite` does so that you don't need this.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -171,6 +171,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         if (context.dockerfilePath) {
             const resolved = new ResolvedContainerizedFunctionAppResource(nonNullProp(wizardContext, 'site'))
+            await ext.rgApi.tree.refresh(context);
```
nturinski, line 174: nit:

Small refactor so we don't maintain two paths:

```
let node: AzExtTreeItem;

if (context.dockerFilePath) {
  const resolved = new ResolvedContainerizedFunctionAppResource(nonNullProp(wizardContext, 'site'))
  node = new ContainerTreeItem(subscription, resolved);
} else {
  const resolved = new ResolvedFunctionAppResource(subscription.subscription, nonNullProp(wizardContext, 'site'));
  node = new SlotTreeItem(subscription, resolved);
}

await ext.rgApi.tree.refresh(context);
return node;
```

src/tree/containerizedFunctionApp/FunctionsTreeItem.ts
motm32: Hmm not totally sure if we need it and also not sure about the errors/warm up function but I can take a look into it.

src/tree/containerizedFunctionApp/ImageTreeItem.ts
motm32: After in person discussion changing this to "Name" since Image is the title of the grouping.

src/tree/ResolvedFunctionAppBase.ts
```diff
@@ -0,0 +1,29 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type ParsedSite } from "@microsoft/vscode-azext-azureappservice";
+import { nonNullProp, nonNullValueAndProp } from "@microsoft/vscode-azext-utils";
+import { type ResolvedAppResourceBase } from "@microsoft/vscode-azext-utils/hostapi";
+import { type FuncHostRequest } from "./IProjectTreeItem";
+import { type ContainerSite } from "./containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase";
+
+export abstract class ResolvedFunctionAppBase implements ResolvedAppResourceBase {
```
nturinski, line 12: This is pretty bare bones still. I think there are quite a few more shared properties/functionality that you can add here.

Also abstract classes have the benefit of being able to define functions that are required on the class, but let each class that extends it implement it.

A good use case for this is something like `getApplicationSettings` where the parameters and the return result are the same, but the implementation is different between `ResolvedFunctionAppResource` and `ResolvedContainerizedFunctionAppResource`.

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResourceBase.ts
```diff
@@ -0,0 +1,186 @@
...
+import { matchContextValue } from "../ResolvedFunctionAppResource";
+import { ProjectResource, ProjectSource, matchesAnyPart } from "../projectContextValues";
+import { ContainerAppSettingsClientProvider } from "./AppSettingsClient";
+import { ContainerFunctionsTreeItem } from "./ContainerFunctionsTreeItem";
+import { type ContainerTreeItem } from "./ContainerTreeItem";
+import { ImageTreeItem } from "./ImageTreeItem";
+
+export type ContainerSite = Site & { defaultHostUrl?: string; fullName?: string; isSlot?: boolean };
+
+export class ResolvedContainerizedFunctionAppResource extends ResolvedFunctionAppBase implements ResolvedAppResourceBase {
```
nturinski, line 29: Name and class doesn't match.

src/commands/browseWebsite.ts
```diff
@@ -3,18 +3,19 @@
...
-export async function browseWebsite(context: IActionContext, node?: SlotTreeItem): Promise<void> {
+export async function browseWebsite(context: IActionContext, node?: SlotTreeItem | ContainerTreeItem): Promise<void> {
     if (!node) {
         node = await ext.rgApi.pickAppResource<SlotTreeItem>(context, {
             filter: functionFilter,
         });
     }
 
-    await openUrl(node.site.defaultHostUrl);
+    await openUrl(nonNullValueAndProp(node.site, 'defaultHostUrl'));
```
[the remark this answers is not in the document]
motm32: defaultHostUrl is a optional value this is set under `ContainerSite` in the ResolvedContainerizedFunctionAppResourceBase` file.

src/tree/ResolvedFunctionAppBase.ts
nturinski: Looks much better btw 🙂

nturinski (approved):

nturinski (approved):

nturinski (approved):

Issue #3954, opened 2024-01-26
There are two HTTP trigger templates in the list when creating a .NET 7.0 project


v-xinda:
**OS:** All
**Build Version:** 20240126.14
**Regression:** Previous release (1.13.1)

**Repro Steps:**

1. F1 -> Azure Functions: Create New Project... -> Select a folder -> C# -> .NET 7.0 Isolate.
2. Check whether there is one HTTP trigger template in the list.

**Expect:**
There is one HTTP trigger template in the list.

**Actual:**
There are two HTTP trigger templates in the list.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/57662131/7a9f4f20-3a25-46db-863b-5a01851246c5)

**More Info:**
This issue doesn't reproduce for the .NET 6.0/8.0.

nturinski:
I don't believe that this is a regression. The template feed got updated so now I think you are seeing double. If you "Reload Templates", it should resolve this.

I'll put a fix out for this issue, but it won't help any existing customers already seeing double. If they "Reload Templates" and delete their cache, it should resolve it for them.

PR #3958, opened 2024-01-29
Delete the correct .NET project templates on reload

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3954 (hopefully)

I think the reason that this appeared during testing is due to the templates being cached, but on first reload, the wrong cache was getting deleted. 

The dotnet template provider looked for a csproj setting or an internal `projectTemplateKey` for which .NET templates to clear. However, when creating a new project, this setting would either a.) not exist, or b.) was potentially incorrect.

For example, if you were creating a net8.0 project, but your last project was net6.0, your old settings would have picked that up and deleted the wrong cache on reloading the templates. By default, the net6.0 templates were getting deleted, so if you tried to reload net7.0-isolated without a project opened, the reload wouldn't actually do anything.

I'm able to reproduce the duplicate templates if I do some combination of the following steps:
1. Have an empty folder opened
2. Get the latest templates from the feed
3. Change my template source to "Staging"
4. Reload VS Code
5. Get the latest templates from the staging feed
6. Change my setting back to "" (which defaults to latest)
7. Reload VS Code
8. Try to create a .NET 7 Isolated project
9. It should list multiple HTTP functions

Note that my fix shouldn't prevent the duplicates from showing up, but makes the reload and `clearCachedTemplates` command work properly.  That being said, I haven't seen it reproducing anymore so maybe it also fixed it in ways I don't understand 🤷 

Edit: Confirmed that if I revert my changes, it starts reproducing again. I'm assuming the `getProjKey` is being used somewhere else that I'm not realizing and is what's causing duplicating templates.

### DIFF
--- a/src/templates/TemplateProviderBase.ts
+++ b/src/templates/TemplateProviderBase.ts
@@ -7,6 +7,7 @@ import { AzExtFsExtra, type IActionContext } from '@microsoft/vscode-azext-utils
 import * as path from 'path';
 import { Disposable, env } from 'vscode';
 import { FuncVersion } from '../FuncVersion';
+import { type IProjectWizardContext } from '../commands/createNewProject/IProjectWizardContext';
 import { type ProjectLanguage } from '../constants';
 import { NotImplementedError } from '../errors';
 import { ext } from '../extensionVariables';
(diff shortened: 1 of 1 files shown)


nturinski:
Gonna go ahead and merge this. Please let me know if anyone sees any issues

PR #3959, opened 2024-01-29
Change PickFunctionApp to distinguish between containerized function apps and regular ones 

motm32:
Contributes to #3523. 

Also refer to #3943 for UI changes. 

When using the command palette actions that don't pertain to containerized function apps you are still ablt to pick them because the `functionFilter` value is the same for both. To bypass this we will throw an error for the commands that don't pertain and have a seperate function `pickContainerOrFunctionApp` that allows either to be chosen.


alexweininger:
Just want to clarify: the reason we can't filter between the two is that our `PickAppResource` function can only filter the top level resources by `type` and `kind`, which for these resources are the exact same. That's a real bummer.

nturinski:
> Just want to clarify: the reason we can't filter between the two is that our `PickAppResource` function can only filter the top level resources by `type` and `kind`, which for these resources are the exact same. That's a real bummer.
> 
> I bet we could add a way to filter top level resources by context value, but for now I think it's fine.

Yeah, this is how I recommended Megan do it because it would be kind of a pain to implement a `pickAppResource` that can filter by context value. I think that it could be useful to have at some point, but right now there hasn't really been a compelling reason to do this work yet. 

If we have more resources that need this or if containerized function app usage explodes, we should probably do it then.

nturinski (dismissed):

nturinski (dismissed):

nturinski (approved):

PR #3964, opened 2024-01-31
Add create containerized function app capabilities 

motm32:
Contributes to https://github.com/microsoft/vscode-azurefunctions/issues/3523

Our entry point for this command is through the existing `Create Function App in Azure...` command. We do some smart detection and if we find a dockerfile we ask the user if they would like to create a containerized function app: 

image

If the click no we will create a regular function app.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,7 +26,7 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionDevelopmentPath=${workspaceFolder}/../vscode-azurecontainerapps"
+                "--extensionDevelopmentPath=${workspaceFolder}/../vscode-azureresourcegroups"
             ],
             "outFiles": [
                 "${workspaceFolder}/out/**/*.js"
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -26,7 +26,7 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionDevelopmentPath=${workspaceFolder}/../vscode-azureresourcegroups"
+                "--extensionDevelopmentPath=${workspaceFolder}/../vscode-azurecontainerapps"
             ],
             "outFiles": [
                 "${workspaceFolder}/out/**/*.js"
--- a/src/commands/createFunctionApp/containerImage/ContainerizedFunctionAppCreateStep.ts
+++ b/src/commands/createFunctionApp/containerImage/ContainerizedFunctionAppCreateStep.ts
@@ -1,0 +1,85 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type Site, type WebSiteManagementClient } from "@azure/arm-appservice";
+import { LocationListStep } from "@microsoft/vscode-azext-azureutils";
+import { AzureWizardExecuteStep, nonNullProp } from "@microsoft/vscode-azext-utils";
+import { type AppResource } from "@microsoft/vscode-azext-utils/hostapi";
+import { type Progress } from "vscode";
+import { webProvider } from "../../../constants";
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { createWebSiteClient } from "../../../utils/azureClients";
+import { getStorageConnectionString } from "../../appSettings/connectionSettings/getLocalConnectionSetting";
+import { type IFunctionAppWizardContext } from "../IFunctionAppWizardContext";
+import { showSiteCreated } from "../showSiteCreated";
+
+export class ContainerizedFunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWizardContext> {
+    public priority: number = 140;
+
+    public async execute(context: IFunctionAppWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const message: string = localize('creatingNewApp', 'Creating new function app "{0}"...', context.newSiteName);
+        ext.outputChannel.appendLog(message);
+        progress.report({ message });
+
+        if (!context.deployWorkspaceResult?.registryLoginServer || !context.deployWorkspaceResult?.imageName) {
+            throw new Error(localize('failToCreateApp', 'Failed to create function app. There was an error creating the necessary container resources.'));
+        }
+
+        const siteName: string = nonNullProp(context, 'newSiteName');
+        const rgName: string = nonNullProp(nonNullProp(context, 'resourceGroup'), 'name');
+        const client: WebSiteManagementClient = await createWebSiteClient(context);
+        context.site = await client.webApps.beginCreateOrUpdateAndWait(rgName, siteName, await this.getNewSite(context));
+        context.activityResult = context.site as AppResource;
+
+        const containerSite = Object.assign(context.site, { defaultHostUrl: `https://${context.site.defaultHostName}`, fullName: context.site.name, isSlot: false });
+
+        showSiteCreated(containerSite, context);
+    }
+
+    public shouldExecute(context: IFunctionAppWizardContext): boolean {
+        return !!context.dockerfilePath
+    }
+
+    private async getNewSite(context: IFunctionAppWizardContext): Promise<Site> {
+        const location = await LocationListStep.getLocation(context, webProvider);
+        const site: Site = {
+            name: context.newSiteName,
+            kind: 'functionapp',
+            location: nonNullProp(location, 'name'),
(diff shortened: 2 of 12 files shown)


src/getExtensionApi.ts
```diff
@@ -15,3 +17,22 @@ export async function getResourceGroupsApi(): Promise<AzureHostExtensionApi> {
         throw new Error(localize('noResourceGroupExt', 'Could not find the Azure Resource Groups extension'));
     }
 }
+
+/**
+ * @param installMessage Override default message shown if extension is not installed.
+ */
+export async function getAzureContainerAppsApi(context: IActionContext, installMessage?: string): Promise<acaApi.AzureContainerAppsExtensionApi> {
```
nturinski, line 24: You should call this way earlier in the create containerized function app flow. You get through creating every resource and then it fails at the end.

src/getExtensionApi.ts
```diff
@@ -15,3 +17,22 @@ export async function getResourceGroupsApi(): Promise<AzureHostExtensionApi> {
...
+    const acaExtensionId: string = 'ms-azuretools.vscode-azurecontainerapps';
+    const acaExtension: apiUtils.AzureExtensionApiProvider | undefined = await apiUtils.getExtensionExports(acaExtensionId);
+
+    if (acaExtension) {
+        return acaExtension.getApi<acaApi.AzureContainerAppsExtensionApi>('^0.0.1');
+    }
+
+    await context.ui.showWarningMessage(installMessage ?? localize('acaInstall', 'You must have the "Azure Container Apps" extension installed to perform this operation.'), { title: 'Install', stepName: 'installContainerApps' });
```
nturinski, line 32: Install should be localized.

src/commands/createFunctionApp/containerImage/DeployWorkspaceProjectStep.ts
```diff
@@ -0,0 +1,37 @@
...
+import { ext } from "../../../extensionVariables";
+import { getAzureContainerAppsApi } from "../../../getExtensionApi";
+import { localize } from "../../../localize";
+import { type IFunctionAppWizardContext } from "../IFunctionAppWizardContext";
+
+export class DeployWorkspaceProjectStep extends AzureWizardExecuteStep<IFunctionAppWizardContext> {
+    public priority: number = 137;
+
+    public async execute(context: IFunctionAppWizardContext, progress: Progress<{ message?: string; increment?: number }>): Promise<void> {
+        const message: string = localize('creatingCAResources', 'Creating container resources this may take a few minutes...');
```
nturinski, line 17:
```suggestion
        const message: string = localize('creatingCAResources', 'Creating container resources; this may take a few minutes...');
```

src/commands/createFunctionApp/containerImage/ContainerizedFunctionAppCreateStep.ts
```diff
@@ -0,0 +1,85 @@
...
+                    {
+                        name: 'AzureWebJobsStorage',
+                        value: (await getStorageConnectionString(context)).connectionString
+                    },
+                    {
+                        name: 'APPLICATIONINSIGHTS_CONNECTION_STRING',
+                        value: `InstrumentationKey=${context.appInsightsComponent?.instrumentationKey}`
+                    },
+                    {
+                        name: 'FUNCTIONS_EXTENSION_VERSION',
```
nturinski, line 65: I'm not sure if you have to define this one?

src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -0,0 +1,49 @@
...
+            const files = (await findFiles(context.workspaceFolder, `*/${hostFileName}`));
+            hostPath = path.dirname(files[0].fsPath);
+        }
+
+        // check if dockerfile exists in the same folder as the host.json
+        context.dockerfilePath = (await findFiles(hostPath, dockerfileGlobPattern))[0].fsPath;
+
+        // prompt user to proceed with containerized function app creation
+        const placeHolder: string = 'Dockerfile detected. Would you like to continue creating a containerized function app?';
+        const result: vscode.MessageItem = await context.ui.showWarningMessage(placeHolder, { modal: true }, DialogResponses.yes, DialogResponses.no);
```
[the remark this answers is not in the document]
motm32: I thought about changing this since I also wasn't sure if the portal team wan't to brand them as "containerized function apps". In the portal when creating a containerized function app you choose to deploy with an image: 
image
We could do a similar message like : 
"Dockerfile detected. Would you like to continue creating a function app that deploys an image? Click yes to continue, click no to create a function app that deploys code."

src/tree/SubscriptionTreeItem.ts
```diff
@@ -207,3 +178,69 @@ async function getDefaultFuncVersion(context: ICreateFunctionAppContext): Promis
...
+}
+
+async function createContainerizedFunctionAppChild(wizardContext: IFunctionAppWizardContext, promptSteps: AzureWizardPromptStep<IAppServiceWizardContext>[], executeSteps: AzureWizardExecuteStep<IAppServiceWizardContext>[]) {
+    const storageAccountCreateOptions: INewStorageAccountDefaults = {
+        kind: StorageAccountKind.Storage,
+        performance: StorageAccountPerformance.Standard,
+        replication: StorageAccountReplication.LRS
+    };
+
+    LocationListStep.addStep(wizardContext, promptSteps);
```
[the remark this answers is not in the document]
motm32: The locations listed in the prompt also appear on the locations list in the portal when you choose to deploy from image when creating a function app. I would think that would mean they should work?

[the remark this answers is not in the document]
nturinski: Well, I tried West India and it didn't successfully create. We may not be able to filter it if the service doesn't recognize it will fail until runtime though.

src/commands/createFunctionApp/containerImage/detectDockerfile.ts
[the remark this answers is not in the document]
nturinski: Maybe we could do something like.

`Dockerfile with Azure Functions image detected. What would you like to deploy?`
`Code` `Container image`

.vscode/launch.json
```diff
@@ -26,7 +26,7 @@
             "runtimeExecutable": "${execPath}",
             "args": [
                 "--extensionDevelopmentPath=${workspaceFolder}",
-                "--extensionDevelopmentPath=${workspaceFolder}/../vscode-azureresourcegroups"
```
nturinski, line 29: You shouldn't remove `vscode-azureresourcegroups`. The host always implies the Resources extension.

If anything, you should just make a new launch task that includes container apps for testing the API. Launch Extension + API or something like that.

src/commands/createFunctionApp/containerImage/ContainerizedFunctionAppCreateStep.ts
```diff
@@ -0,0 +1,85 @@
...
+        showSiteCreated(containerSite, context);
+    }
+
+    public shouldExecute(context: IFunctionAppWizardContext): boolean {
+        return !!context.dockerfilePath
+    }
+
+    private async getNewSite(context: IFunctionAppWizardContext): Promise<Site> {
+        const location = await LocationListStep.getLocation(context, webProvider);
+        const site: Site = {
```
nturinski, line 48: nit: Just return the object here rather than setting it as a site.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -100,28 +106,26 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
...
-            wizardContext.stackFilter = getRootFunctionsWorkerRuntime(language);
+            wizardContext.stackFilter = getRootFunctionsWorkerRuntime(wizardContext.language);
             executeSteps.push(new ResourceGroupCreateStep());
-            executeSteps.push(new AppServicePlanCreateStep());
             executeSteps.push(new StorageAccountCreateStep(storageAccountCreateOptions));
-            executeSteps.push(new LogAnalyticsCreateStep());
             executeSteps.push(new AppInsightsCreateStep());
+            if (!context.dockerfilePath) {
+                executeSteps.push(new AppServicePlanCreateStep());
+                executeSteps.push(new LogAnalyticsCreateStep());
```
nturinski, line 127: This is used for the `AppInsightsCreateStep` to make it an workspace enabled AI. I think it should stay in the common steps.

motm32: Hmm ok. I originally put this here since the CA api created the log so if I am to add this back then we would need to change the api

nturinski: We should probably verify if there are duplicate steps and remove them if there are. This might have to be an azuretools change though.

motm32: After discussion offline is seems like the App Insights just checks to see if a log analytics exists in the same region and if not it creates one.

nturinski: I'll go ahead and approve this, but I'll end up consolidating everything into one command as discussed offline.

nturinski (approved):

PR #3968, opened 2024-02-01
Bump azureutils and appservice packages

alexweininger:
Consume downstream changes https://github.com/microsoft/vscode-azuretools/pull/1677, https://github.com/microsoft/vscode-azuretools/pull/1676, that prevent https://github.com/microsoft/vscode-azuretools/pull/1669 from breaking kudu calls

Also added this to the change log. We should try updating the change log as we make the changes this time around and see how it goes. I want to do a better job of reflecting the changes we are consuming from the shared packages in the change log too, as I often find myself having to search for when changes from the shared package made it into the extensions.


alexweininger:
Converted to draft because I'm still seeing issues today.

PR #3979, opened 2024-02-08
Starter integration with Azure agent extension

MRayermannMSFT:
If you haven't already read it, I highly recommend reading [this doc](https://github.com/microsoft/vscode-azure-agent/blob/main/docs/Architecture.md) that gives a high level overview of the Azure Agent architecture. You should/can also take a look at the agent integration typings in `@microsoft/vscode-azext-utils`.

Some initial work on integrating the Azure Functions extension with the Azure Agent extension.

- Add `agentIntegration.ts` which serves as the initial entry point for where the integration lives. Contains the implementations of the various `package.json.agentMetadata` commands.
- Expose create function app as a `WizardCommandConfig`, which also includes:
   - Adding `ui: IAzureAgentInput;` on the `IFunctionAppWizardContext` so the associated wizard step is forced by the compiler to use `IAzureAgentInput` related types.
   - Switching to `IAzureAgentInput` related types in `FunctionAppStackStep.ts`
   - Switching to `IAzureAgentInput` related types in `getStackPicks.ts`
   - Using the new option on `SiteNameStep` to give the step appropriate agent metadata information for its UI prompts
   - Appropriately using the new `skipExecute` option when creating the `AzureWizard` for the command
   - Appropriately setting the `showLoadingPrompt` option when creating the `AzureWizard` for the command
- Expose create function project and deploy to function app as `SimpleCommandConfig`s (quick way to get them exposed via agent, ideally they should be exposed as `WizardCommandConfig`s instead, what lucky dev will get to do that work? 🙃)
- Write some sample benchmarks for Azure Functions extension related functionality (note: there's no verification of buttons because buttons are in a messed up state right now, please don't ask).

### DIFF
--- a/src/agent/agentIntegration.ts
+++ b/src/agent/agentIntegration.ts
@@ -1,0 +1,129 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtUserInputWithInputQueue, callWithTelemetryAndErrorHandling, type AgentBenchmarkConfig, type AzureUserInputQueue, type IAzureUserInput, type SimpleCommandConfig, type WizardCommandConfig } from '@microsoft/vscode-azext-utils';
+import { createFunctionApp } from '../commands/createFunctionApp/createFunctionApp';
+
+const createFunctionAppCommandName = "createFunctionApp";
+const createFunctionProjectCommandName = "createFunctionProject";
+const deployToFunctionAppCommandName = "deployToFunctionApp";
+
+export async function getCommands(): Promise<(WizardCommandConfig | SimpleCommandConfig)[]> {
+    return [
+        {
+            type: "wizard",
+            name: createFunctionAppCommandName,
+            commandId: "azureFunctions.createFunctionApp",
+            displayName: "Create Function App",
+            intentDescription: "This is best when users ask to create a Function App resource in Azure. They may refer to a Function App as 'Function App', 'function', 'function resource', 'function app resource', 'function app' etc. This command is not useful if the user is asking how to do something, or if something is possible.",
+            requiresAzureLogin: true,
+        },
+        {
+            type: "simple",
+            name: createFunctionProjectCommandName,
+            commandId: "azureFunctions.createNewProject",
+            displayName: "Create Function Project",
+            intentDescription: "This is best when users ask to create a new function project in VS Code. They may also refer to creating a function project by asking to create a project based upon a function project template.",
+            requiresAzureLogin: true,
+        },
+        {
+            type: "simple",
+            name: deployToFunctionAppCommandName,
+            commandId: "azureFunctions.deploy",
+            displayName: "Deploy to Function App",
+            intentDescription: "This is best when users ask to create a deploy their function project or their code to a function app. They may refer to a Function App as 'Function App', 'function', 'function resource', 'function app resource', 'function app' etc. This is not best if they ask to deploy to a slot.",
+            requiresAzureLogin: true,
+        },
+    ];
+}
+
+export async function runWizardCommandWithoutExecution(command: WizardCommandConfig, ui: IAzureUserInput): Promise<void> {
+    if (command.commandId === 'azureFunctions.createFunctionApp') {
+        await callWithTelemetryAndErrorHandling('azureFunctions.createFunctionAppViaAgent', async (context) => {
+            return await createFunctionApp({ ...context, ui: ui, skipExecute: true });
+        });
+    } else {
+        throw new Error('Unknown command: ' + command.commandId);
+    }
+}
+
+export async function runWizardCommandWithInputs(command: WizardCommandConfig, inputsQueue: AzureUserInputQueue): Promise<void> {
+    if (command.commandId === 'azureFunctions.createFunctionApp') {
+        await callWithTelemetryAndErrorHandling('azureFunctions.createFunctionAppViaAgent', async (context) => {
+            const azureUserInput = new AzExtUserInputWithInputQueue(context, inputsQueue);
+            return await createFunctionApp({ ...context, ui: azureUserInput });
+        });
+    } else {
+        throw new Error('Unknown command: ' + command.commandId);
+    }
+}
+
+export function getAgentBenchmarkConfigs(): AgentBenchmarkConfig[] {
+    return [
+        {
+            name: "Learn 1 - How to run code on blob change",
+            prompt: "How can I use azure functions to run code whenever a blob is changed?",
+            acceptableHandlerChains: [
+                ["functions", "learn"],
+            ],
+        },
+        {
+            name: "Learn 2 - Functions vs WebApps",
+            prompt: "How do I write a lambda in Azure?",
+            acceptableHandlerChains: [
+                ["functions", "learn"],
+            ],
+        },
+        {
+            name: "Create Function App 1",
+            prompt: "I want to create a function app",
+            acceptableHandlerChains: [
+                ["functions", createFunctionAppCommandName],
+            ],
+        },
+        {
+            name: "Create Function App 2",
+            prompt: "I want to create a func app",
+            acceptableHandlerChains: [
+                ["functions", createFunctionAppCommandName],
+            ],
+        },
+        {
+            name: "Create Function App 3",
+            prompt: "I want to create a resource that lets me run serverless code",
+            acceptableHandlerChains: [
+                ["functions", createFunctionAppCommandName],
+            ],
+        },
+        {
+            name: "Create Function App 4",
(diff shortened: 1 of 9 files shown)


alexweininger (approved):

src/agent/agentIntegration.ts
```diff
@@ -0,0 +1,129 @@
...
+            intentDescription: "This is best when users ask to create a deploy their function project or their code to a function app. They may refer to a Function App as 'Function App', 'function', 'function resource', 'function app resource', 'function app' etc. This is not best if they ask to deploy to a slot.",
+            requiresAzureLogin: true,
+        },
+    ];
+}
+
+export async function runWizardCommandWithoutExecution(command: WizardCommandConfig, ui: IAzureUserInput): Promise<void> {
+    if (command.commandId === 'azureFunctions.createFunctionApp') {
+        await callWithTelemetryAndErrorHandling('azureFunctions.createFunctionAppViaAgent', async (context) => {
+            return await createFunctionApp({ ...context, ui: ui, skipExecute: true });
```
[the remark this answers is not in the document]
MRayermannMSFT: I prefer being explicit with property name/value syntax. :)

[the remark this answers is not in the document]
alexweininger: Ok, that's fine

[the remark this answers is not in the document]
nturinski: > I prefer being explicit with property name/value syntax. :)

_This_ is where you prefer to be explicit? 🤔

nturinski:
Two questions: when are you planning on releasing this and would it be okay to release this in a few weeks?

The more important question is the second one-- if we need to do a release for Functions (since we have deliverables in end of Feb and in March), would we need to revert these changes or does none of this have any impact _until_ the azure agent stuff is released/installed?

MRayermannMSFT:
> Two questions: when are you planning on releasing this and would it be okay to release this in a few weeks?

Yes, it would be ok to release this anytime in a few weeks. Unless someone has the agent extension installed, this is dormant. I don't think there's any date in the next two months where I would say "this code needs to be released by that date".

> ...if we need to do a release for Functions... would we need to revert these changes or does none of this have any impact until the azure agent stuff is released/installed?

No revert is needed. If this code has an effect on non-agent usage of the functions extension, then that is a bug in this code.

src/agent/agentIntegration.ts
```diff
@@ -0,0 +1,129 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { AzExtUserInputWithInputQueue, callWithTelemetryAndErrorHandling, type AgentBenchmarkConfig, type AzureUserInputQueue, type IAzureUserInput, type SimpleCommandConfig, type WizardCommandConfig } from '@microsoft/vscode-azext-utils';
+import { createFunctionApp } from '../commands/createFunctionApp/createFunctionApp';
+
+const createFunctionAppCommandName = "createFunctionApp";
+const createFunctionProjectCommandName = "createFunctionProject";
+const deployToFunctionAppCommandName = "deployToFunctionApp";
+
+export async function getCommands(): Promise<(WizardCommandConfig | SimpleCommandConfig)[]> {
```
nturinski, line 13: Note for later: Putting all of the commands `WizardCommandConfigs` in here will definitely get bulky quickly. I'm thinking for organization, we have a commands folder and each file is just a command with all of this information.

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
```diff
@@ -18,7 +18,16 @@ export class FunctionAppStackStep extends AzureWizardPromptStep<IFunctionAppWiza
 
         let result: FullFunctionAppStack | undefined;
         while (true) {
-            result = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder, enableGrouping: true })).data;
+            const options: AgentQuickPickOptions = {
```
nturinski, line 21: I would prefer if the rest of the extension code basically had no idea about the `agentMetaData` stuff. Would we pass it in as an option instead when you invoke it here? https://github.com/microsoft/vscode-azurefunctions/pull/3979/files#diff-ed50eea82ea7bad017f14ed53416d4b50d441da3ce4240af90885e92c175a1b8R56

src/commands/registerCommands.ts
```diff
@@ -54,6 +55,12 @@ import { disableFunction, enableFunction } from './updateDisabledState';
 import { viewProperties } from './viewProperties';
 
 export function registerCommands(): void {
+
+    commands.registerCommand('azureFunctions.getCommands', getCommands);
```
nturinski, line 59: Is this a temporary measure? I'm not sure if we want to register the command with our custom wrapping logic, though I don't think it would hurt?  Also, I think it'd be helpful to have a prefix to easily distinguish between copilot commands and our normal extension commands.

src/tree/SubscriptionTreeItem.ts
```diff
@@ -97,7 +98,7 @@ export class SubscriptionTreeItem extends SubscriptionTreeItemBase {
 
         const promptSteps: AzureWizardPromptStep<IAppServiceWizardContext>[] = [];
         const executeSteps: AzureWizardExecuteStep<IAppServiceWizardContext>[] = [];
-        promptSteps.push(new SiteNameStep());
+        promptSteps.push(new SiteNameStep({ parameterDisplayTitle: "Function App Name", parameterDisplayDescription: "The name of the new function app." }));
```
nturinski, line 101: I know that there are cases where `SiteNameStep` is used for Functions or for Web Apps, but I still think it might be better to define these property in the step class itself rather than passing them as options?

To distinguish between the two, I think we can pass an option for web app or function app, but having to add the parameter information each time we create a new step seems like it'd be cumbersome especially for other steps (assuming that we follow that pattern for all steps)

MRayermannMSFT: Ok. This will require changes in azuretools.

src/commands/registerCommands.ts
MRayermannMSFT: > Is this a temporary measure?

Is it? I wasn't intendeding for it to be temporary, at least not anymore than code already is (it exists so long as it dose not need to exist).

> I'm not sure if we want to register the command with our custom wrapping logic, though I don't think it would hurt?

If you want it we can do it. If you don't know if you want it, then does waiting for a decision block this PR?

> Also, I think it'd be helpful to have a prefix to easily distinguish between copilot commands and our normal extension commands.

Ok.

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
MRayermannMSFT: > I would prefer if the rest of the extension code basically had no idea about the agentMetaData stuff. Would we pass it in as an option instead when you invoke it here?

Every call to a UI prompt needs to have various pieces of agent metadata added on. I can't think of a way to not accomplish that without sprinkling the agent metadata at each spot a UI prompt is done. Can you be more specific about what alternative ideas you have?

src/agent/agentIntegration.ts
MRayermannMSFT: Sure. We can keep an eye on it.

src/tree/SubscriptionTreeItem.ts
MRayermannMSFT: https://github.com/microsoft/vscode-azuretools/pull/1694

src/commands/registerCommands.ts
MRayermannMSFT: I added `.agent.` after `azurefunctions`. If you really want `agent.azurefunctions` I can do that but it felt weird....

nturinski: > If you want it we can do it. If you don't know if you want it, then does waiting for a decision block this PR?

We don't need to block on that decision, but just something to consider since it has all of our custom logic for error handling (such as ignoring user cancelled errors, for example)

src/commands/createFunctionApp/stacks/FunctionAppStackStep.ts
nturinski: Here specifically, I was originally think we could have had `options.agentMetadata` get passed into the constructor of `FunctionAppStackStep`. But that's just moving the agent code somewhere else still in the extension code base which isn't really alienating the agent code then...

And since each quick pick item itself needs to also have `agentMetadata` there probably isn't anyway to get around it. 

Oh well, no biggie. Resolving the conversation-- if I have any ideas later, I'll bring it up then.

nturinski (approved):

src/commands/registerCommands.ts
MRayermannMSFT: > We don't need to block on that decision

Let's defer for now then. :)

src/tree/SubscriptionTreeItem.ts
MRayermannMSFT: Updated.

nturinski (approved):

PR #3984, opened 2024-02-20
Improve Event Grid Local Development Experience

hossam-nasr:
Improve the local development experience for EventGrid functions.

### What this PR does:

- [x] When customer clicks on "Execute Function Now..." button, if it's an EventGrid function, the user is prompted to choose an event type from a list of supported file types:
image

- [x] The customer then selects one of sample event types under that event source, pulled from [this repo](https://github.com/Azure/azure-rest-api-specs/tree/master/specification/eventgrid/data-plane):

image

- [x] After selecting an event type, user is prompted to either modify the sample file or send as-is:

image

- [x] If "send sample request" is selected, EventGrid function gets triggered with sample request.

- [x] If "modify sample request" is selected, a new window opens with the sample request:

image

- [x] Once the user makes any necessary modifications, clicking on the "save and send request..." button on the top right of the editor, or on the "Save and send request" code lens would save the modified contents of the sample file and trigger the function. The file is then closed, and the temp file created is deleted after.

image

image

### Missing / Next steps:

- [ ] Try remove the styling contributions this PR makes. For some reason, I couldn't find a `.prettierrc` file configuration that would keeping the existing styling as-is. Any help there is appreciated.
- [ ] Customize and refine the experience specifically for EventGrid-based blob trigger functions
- [ ] Double check the "source of truth" for supported EG event sources and sample requests for each event type under each event source.
- [ ] Come up with a mechanism to streamline the process if the user is doing the same thing over and over again, and remember the modifications from last execution. Right now, nothing is remembered besides the choices for the different prompts.

### Open questions:

❓Do I need to do something extra to make localization work? I followed the pattern I saw of using the `localize` function, but I'm not sure how the code I used as the first argument gets used.
❓Right now, I create the temp files that store these sample files in the `.vscode` directory. Would it be better if I saved it elsewhere, such as the OS temp file directory, by default?
❓Because of the way the button and code lens need to be registered at the extension level, they appear in any `.eventgrid.json` file, even if no function is currently being executed or the function app is even running. Is there a way to stop them from showing unless the function app is running and a function is being executed?


src/commands/executeFunction/EventGridCodeLensProvider.ts
```diff
@@ -0,0 +1,14 @@
+import { CodeLens, CodeLensProvider, Range } from 'vscode';
```
nturinski, line 1: Every new file that you created needs the license at the top. You can refer to any of the files in the Functions repo for it.

package.json
```diff
@@ -687,6 +693,13 @@
                     "when": "resourceFilename==function.json",
                     "group": "zzz_binding@1"
                 }
+            ],
+            "editor/title": [
+                {
+                    "command": "azureFunctions.eventGrid.sendMockRequest",
+                    "when": "resourceFilename=~/.*.eventgrid.json$/",
```
nturinski, line 700: > Because of the way the button and code lens need to be registered at the extension level, they appear in any .eventgrid.json file, even if no function is currently being executed or the function app is even running. Is there a way to stop them from showing unless the function app is running and a function is being executed?

Regarding this question: I think there are a couple of ways we can handle this.

1. You should be able to define a extension context value that tracks whether or not the function host is running. Then, you would use that context value and check for it in this `when` clause to only show the button when an app is active.
2. When we execute the trigger, part of the implementation of execute should verify if the local host is running. If it isn't, then we can start it for the user and then execute the trigger.

package.json
```diff
@@ -350,6 +350,12 @@
                 "command": "azureFunctions.viewProperties",
                 "title": "%azureFunctions.viewProperties%",
                 "category": "Azure Functions"
+            },
+            {
+                "command": "azureFunctions.eventGrid.sendMockRequest",
```
nturinski, line 355: Is there anyway we can define the order of the buttons? 
image

Because it's in the middle, I find it a little less obvious a new button was contributed to this ribbon.

src/commands/executeFunction/EventGridCodeLensProvider.ts
```diff
@@ -0,0 +1,18 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { CodeLens, Range, type CodeLensProvider } from 'vscode';
+import { localize } from '../../localize';
+
+export class EventGridCodeLensProvider implements CodeLensProvider {
+    public provideCodeLenses(): CodeLens[] {
+        const sendRequestLens = new CodeLens(new Range(0, 0, 0, 0));
```
nturinski, line 10: As mentioned offline, it might be helpful to have this code lens also displayed at the bottom of the file. I think it would make it more obvious to the user, as well be convenient if you've scrolled to the bottom of the file and don't know that you need to scroll back to the top.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+    type: string;
+    _links: {
+        self: string;
+        git: string;
+        html: string;
+    };
+};
+
+const sampleFilesUrl =
+    'https://api.github.com/repos/Azure/azure-rest-api-specs/contents/specification/eventgrid/data-plane/' +
```
nturinski, line 36: We should use an aka.ms link for this in case they ever change the repo or the repo formatting. That allows us to change this without having to do a new release.

I haven't tried it personally, but we should be able to just leave `{eventSource}` in the aka.ms link, and then just replace it like how you're already doing.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+    // Prompt for event source
+    const eventGridSourcePicks: IAzureQuickPickItem<EventGridSource | undefined>[] = supportedEventGridSources.map((source: EventGridSource) => {
+        return {
+            label: supportedEventGridSourceLabels.get(source) || source,
+            data: source,
+        };
+    });
+    const eventSource: EventGridSource =
+        (
+            await context.ui.showQuickPick(eventGridSourcePicks, {
```
nturinski, line 50: All of these quickPick calls should be broken up into separate wizard steps. This gives us a lot more flexibility in terms of where and how we call these steps. I can see some reusability for when we deploy to a function app with an event grid trigger, for example.

It also gives the user the ability to go back if they accidentally select the wrong pick rather than having to start the entire wizard over.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+    const selectedFileUrl = sampleFiles.find((fileMetadata) => fileMetadata.name === selectedFileName)?.download_url || sampleFiles[0].download_url;
+    const selectedFileContents: {} = await feedUtils.getJsonFeed(context, selectedFileUrl);
+
+    // Prompt for whether to send or modify the sample request
+    const shouldModify: boolean =
+        (
+            await context.ui.showQuickPick(
+                [
+                    {
+                        label: localize('sendSample', 'Send sample request'),
```
nturinski, line 89: @alexweininger brought it up offline, and I was thinking something kind of similar: are there many scenarios where the user would want to just send the sample request without editing it? It seems like there are typically a lot of properties that need to be edited for it to actually emulate an Event Grid response.

I'm kind of leaning towards just skipping this prompt and always spit them out to the modify request.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+                } finally {
+                    disposable.dispose();
+                }
+            }
+        });
+    });
+}
+
+async function createTempSampleFile(eventSource: string, fileName: string, contents: {}): Promise<string> {
+    const samplesDirPath = await createSamplesDirIfNotExists(eventSource);
```
nturinski, line 136: > Right now, I create the temp files that store these sample files in the .vscode directory. Would it be better if I saved it elsewhere, such as the OS temp file directory, by default?

I'd prefer to write this in the OS temp file directory personally.

It's a pretty large task, but I think long term, we should probably have a virtual file system provider similar to container apps or container apps so that we don't have to write anything in temp.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+    const sampleFileName = fileName.replace(/\.json$/, '.eventgrid.json');
+    const filePath: string = path.join(samplesDirPath, sampleFileName);
+
+    await fs.writeFile(filePath, JSON.stringify(contents, undefined, 2));
```
nturinski, line 140: Use `AzExtFsExtra.writeFile` for any file system provider calls. `AzExtFsExtra` uses all of the `vscode.workspace.fs` APIs so this will help our extensions stay compatible with all the different VS Code platforms.

src/commands/executeFunction/executeEventGridFunction.ts
```diff
@@ -0,0 +1,156 @@
...
+    return filePath;
+}
+
+async function createSamplesDirIfNotExists(eventSource: string): Promise<string> {
+    const baseDir: string = vscode.workspace.workspaceFolders?.[0]?.uri?.fsPath || path.join(os.tmpdir(), 'vscode', 'azureFunctions');
+
+    // Create the path to the directory
+    const dirPath = path.join(baseDir, '.vscode', 'eventGridSamples', eventSource);
+    // Create the directory if it doesn't already exist
+    if (!(await fs.pathExists(dirPath))) {
```
nturinski, line 151: There are `AzExtFsExtra` equivalents of this as well.

src/commands/executeFunction/executeFunction.ts
```diff
@@ -113,4 +141,7 @@ export async function executeFunction(context: IActionContext, node?: FunctionTr
 
     const message: string = responseText ? localize('executedWithResponse', 'Executed function "{0}". Response: "{1}"', func.name, responseText) : localize('executed', 'Executed function "{0}"', func.name);
     void window.showInformationMessage(message);
+
+    ext.isExecutingFunction = false;
```
nturinski, line 145: This whole function needs to be wrapped with a `try/finally`. If there's ever an error when executing the function, I believe that these extension values will not get reset.

src/commands/executeFunction/sendEventGridRequest.ts
```diff
@@ -0,0 +1,33 @@
...
+    if (!activeEditor) {
+        const errorMsg = localize('noActiveTextEditor', 'No active text editor found.');
+        throw new Error(errorMsg);
+    }
+    const document = activeEditor.document;
+    await document.save();
+    const requestContent: string = document.getText();
+
+    await executeFunctionWithInput(context, requestContent, ext.currentExecutingFunctionNode);
+    await vscode.commands.executeCommand('workbench.action.closeActiveEditor');
```
nturinski, line 32: I actually don't like the experience of this closing on me automatically after executing the request.

package.nls.json
```diff
@@ -105,6 +105,7 @@
     "azureFunctions.viewCommitInGitHub": "View Commit in GitHub",
     "azureFunctions.viewDeploymentLogs": "View Deployment Logs",
     "azureFunctions.viewProperties": "View Properties",
+    "azureFunctions.eventGrid.sendMockRequest": "Save and send request...",
```
nturinski, line 108: Wondering if we should call this `Save and execute`? I know technically it's sending the request to the local host, but the entry point is call `Execute Function Now...` so it may be somewhat confusing verbiage.

nturinski (commented):
For files that are edited due to your prettier, I think you can unstage those changes and just not push them up. I can't be sure, but it looks like there are some files that are _only_ formatting changes, right?

As far as directory structure goes, I think there are enough EventGrid specific files that we should just create an `executeFunction/eventGrid` folder.

package.json
hossam-nasr: 1. How do I check for a context value in the `when` clause?
2. I already do that, but if the host isn't running, I simply just throw an error. The challenge of starting it up for them is which function exactly should I run? In the example that I showed, there was only one eventgrid function in the entire function app, but what if there were multiple? So not only do we need the host to be running, but also a specific function should be in the process of being executed I think.

src/commands/executeFunction/executeEventGridFunction.ts
alexweininger: > I'm kind of leaning towards just skipping this prompt and always spit them out to the modify request.

I'm leaning towards this as well

package.json
alexweininger: I second this. I believe there should be a way to move it to the left.

src/commands/executeFunction/eventGrid/eventGridSources.ts
```diff
@@ -0,0 +1,64 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
```
alexweininger, line 5: Can you add a comment here that links to where you got all these from?

src/tree/SubscriptionTreeItem.ts
alexweininger, line 1: I know it's from your formatter, but could you revert the formatting changes in these unrelated files?

src/commands/executeFunction/eventGrid/EventGridCodeLensProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+import { CodeLens, Range, type CodeLensProvider, type TextDocument } from 'vscode';
+import { localize } from '../../../localize';
+
+export class EventGridCodeLensProvider implements CodeLensProvider {
+    public provideCodeLenses(document: TextDocument): CodeLens[] {
+        const firstLineLens = new CodeLens(new Range(0, 0, 0, 0));
+        const lastLine = document.lineAt(document.lineCount - 1);
```
nturinski, line 11: As discussed offline, let's just remove this and have it at the top only.

src/commands/executeFunction/eventGrid/EventGridEventSourceStep.ts
```diff
@@ -0,0 +1,31 @@
+import { AzureWizardPromptStep, type IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
```
nturinski, line 1: Need license

src/commands/executeFunction/eventGrid/EventGridCodeLensProvider.ts
nturinski: For reference, this is what it looks like: 
image

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { feedUtils } from "../../../utils/feedUtils";
+import { type ExecuteEventGridFunctionContext } from "./ExecuteEventGridFunctionContext";
+
+export class OpenEventGridFileStep extends AzureWizardExecuteStep<ExecuteEventGridFunctionContext> {
+    public priority: number;
+
+    public async execute(context: ExecuteEventGridFunctionContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined; }>): Promise<void> {
+        if (!context.eventSource || !context.selectedFileName || !context.selectedFileUrl) {
```
nturinski, line 15: We have a helper function that will throw an internal error like this automatically if it's null.

You could do something like
```
const eventSource = nonNullProp(context, 'eventSource');
const selectedFileName  = nonNullProp(context, 'selectedFileName ');
const selectedFileUrl = nonNullProp(context, 'selectedFileUrl ');
```

That way you can use these variables with confidence that they're defined and if they're not, we'll know exactly which one is throwing the error.

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+            throw new Error('Internal error: event source or event type is missing');
+        }
+
+        // Get selected contents of sample request
+        const downloadingMsg: string = localize('downloadingSample', 'Downloading sample request...');
+        progress.report({ message: downloadingMsg });
+        context.selectedFileContent = await feedUtils.getJsonFeed(context, context.selectedFileUrl);
```
nturinski, line 22: You probably don't need to set this on the context since it's only used in this file.

Context values are typically reserved for values that get used across multiple steps.

You could then get rid of it from the typing.

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+        const tempFilePath: string = await createTempSampleFile(context.eventSource, context.selectedFileName, context.selectedFileContent);
+        const document: vscode.TextDocument = await vscode.workspace.openTextDocument(tempFilePath);
+        await vscode.window.showTextDocument(document, {
+            preview: false,
+        });
+        ext.fileToFunctionNodeMap.set(document.fileName, nonNullProp(ext, 'currentExecutingFunctionNode'));
+        context.fileOpened = true;
+
+        // Request will be sent when the user clicks on the button or on the codelens link
+        const doneMsg = localize('modifyFile', "You can modify the file and then click the 'Save and execute' button to send the request.");
```
nturinski, line 36: I think this notification is helpful, but it seems like it could get annoying if you were executing your trigger multiple times.

Maybe we could either a.) have a "don't show again" button or b.) have it timed to only show once a session or something like that

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+    return filePath;
+}
+
+async function createSamplesDirIfNotExists(eventSource: string): Promise<string> {
+    const baseDir: string = path.join(os.tmpdir(), 'vscode', 'azureFunctions');
+
+    // Create the path to the directory
+    const dirPath = path.join(baseDir, 'eventGridSamples', eventSource);
+    // Create the directory if it doesn't already exist
+    if (!(await AzExtFsExtra.pathExists(dirPath))) {
```
nturinski, line 79: small nit: You can just do `await AzExtFsExtra.ensureDir(dirPath)` since it will only do anything if the folder doesn't exist.

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+    const samplesDirPath = await createSamplesDirIfNotExists(eventSource);
+    const sampleFileName = fileName.replace(/\.json$/, '.eventgrid.json');
+    const filePath: string = path.join(samplesDirPath, sampleFileName);
+
+    await AzExtFsExtra.writeJSON(filePath, contents);
+
+    return filePath;
+}
+
+async function createSamplesDirIfNotExists(eventSource: string): Promise<string> {
```
nturinski, line 73: nit: Name of the function threw me off a little bit. Maybe name it `getSamplesDirPath`? I'm not sure if creating it if it doesn't exist is the important part of the function here and the comments in the function make it clear that it will ensure the directory exists before returning the path.

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
+import { AzExtFsExtra, AzureWizardExecuteStep, nonNullProp } from "@microsoft/vscode-azext-utils";
```
nturinski, line 1: License

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
```diff
@@ -0,0 +1,85 @@
...
+import * as os from 'os';
+import * as path from "path";
+import * as vscode from 'vscode';
+import { type Progress } from "vscode";
+import { ext } from "../../../extensionVariables";
+import { localize } from "../../../localize";
+import { feedUtils } from "../../../utils/feedUtils";
+import { type ExecuteEventGridFunctionContext } from "./ExecuteEventGridFunctionContext";
+
+export class OpenEventGridFileStep extends AzureWizardExecuteStep<ExecuteEventGridFunctionContext> {
```
nturinski, line 11: nit/suggestion: To keep things congruent, it might be better to name this `EventGridFileOpenStep` just so all the steps have the `EventGrid` prefix, making it easier to associate on first glance.

src/commands/executeFunction/eventGrid/ExecuteEventGridFunctionContext.ts
```diff
@@ -0,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type IActionContext } from "@microsoft/vscode-azext-utils";
+import { type EventGridSource } from "./eventGridSources";
+
+
+export interface ExecuteEventGridFunctionContext extends IActionContext {
```
nturinski, line 10: Similar to what I was saying about `OpenEventGridFileStep`, I think naming this `EventGridExecuteFunctionContext` just makes it easier for other people to associate this with all the other `eventGrid` files.

package.json
hossam-nasr: Discussed offline. Changed it such that the file stays open after a request is sent, and the button sends the request to the last function that was executed, until the file is closed or another function is executed.

package.json
hossam-nasr: Discussed offline with @nturinski. I agree and I tried to make it as far left as possible, I'm not aware of any ways to ensure that's the case. I think it may be related to which extensions load first, because in my experience, the "save and send request" button appears a few seconds before the git diff button. Do you know of a way to ensure it's always on the farthest left?

src/commands/executeFunction/EventGridCodeLensProvider.ts
hossam-nasr: Discussed offline. Ended up adding a line at the end then removing it 😛

src/commands/executeFunction/executeEventGridFunction.ts
hossam-nasr: Discussed offline. Tried using an `aka.ms` link, but turns out there isn't an easy way to do this parameterization with them. Keeping it like this for now.

src/commands/executeFunction/executeEventGridFunction.ts
hossam-nasr: Removed.

src/commands/executeFunction/executeEventGridFunction.ts
hossam-nasr: Done

src/commands/executeFunction/executeEventGridFunction.ts
hossam-nasr: Done

src/commands/executeFunction/executeFunction.ts
hossam-nasr: Done

package.nls.json
hossam-nasr: Done.

src/commands/executeFunction/eventGrid/eventGridSources.ts
hossam-nasr: Done!

src/tree/SubscriptionTreeItem.ts
hossam-nasr: Done, thanks

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
hossam-nasr: Done, thanks!

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
hossam-nasr: Made it such that I set a flag on `ext.context.workspaceState` so that it shows only once per workspace. I couldn't find a way to make it show only once per _session_, but let me know if you do! (I guess I could reset the flag to `false` in the `activate` function?)

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
hossam-nasr: Renamed to `getSamplesDirPath`

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
hossam-nasr: Renamed

src/commands/executeFunction/eventGrid/EventGridExecuteFunctionContext.ts
```diff
@@ -0,0 +1,16 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type IActionContext } from "@microsoft/vscode-azext-utils";
+import { type EventGridSource } from "./eventGridSources";
+
+
```
nturinski, line 9: nit: extra linebreak

package.json
nturinski: I was messing with it for a bit too and didn't see anyway to guarantee it. I was looking at [this](https://code.visualstudio.com/api/references/contribution-points#Sorting-of-groups) and it didn't really seem like it had anything useful.

src/commands/executeFunction/eventGrid/EventGridFileOpenStep.ts
```diff
@@ -0,0 +1,95 @@
...
+        context.fileOpened = true;
+
+        // Request will be sent when the user clicks on the button or on the codelens link
+        // Show the message only once per workspace
+        if (!ext.context.workspaceState.get('didShowEventGridFileOpenMsg')) {
+            const doneMsg = localize('modifyFile', "You can modify the file and then click the 'Save and execute' button to send the request.");
+            void vscode.window.showInformationMessage(doneMsg);
+            await ext.context.workspaceState.update('didShowEventGridFileOpenMsg', true);
+        }
+
```
nturinski, line 47: nit: extra linebreak

src/commands/executeFunction/eventGrid/EventGridFileOpenStep.ts
```diff
@@ -0,0 +1,95 @@
...
+        // Set a listener to delete the temp file after it's closed
+        void new Promise<void>((resolve, reject) => {
+            const disposable = vscode.workspace.onDidCloseTextDocument(async (closedDocument) => {
+                if (closedDocument.fileName === document.fileName) {
+                    try {
+                        ext.fileToFunctionNodeMap.delete(document.fileName);
+                        await AzExtFsExtra.deleteResource(tempFilePath);
+                        resolve();
+                    } catch (error) {
+                        reject(error);
```
nturinski, line 58: We probably just want to swallow the error here.

You can add this line to do so:
`context.errorHandling.suppressDisplay = true;`

src/commands/executeFunction/eventGrid/EventGridSourceStep.ts
```diff
@@ -0,0 +1,36 @@
...
+                data: source,
+            };
+        });
+        const eventSource: EventGridSource =
+            (
+                await context.ui.showQuickPick(eventGridSourcePicks, {
+                    placeHolder: localize('selectEventSource', 'Select the event source'),
+                    stepName: 'eventGridSource',
+                })
+            ).data ?? 'Microsoft.Storage';
```
nturinski, line 28: Why fallback to Microsoft.Storage? Do we expect the data to return `undefined` for any reason?

src/commands/executeFunction/eventGrid/EventGridTypeStep.ts
```diff
@@ -0,0 +1,72 @@
...
+        html: string;
+    };
+};
+
+export class EventGridTypeStep extends AzureWizardPromptStep<EventGridExecuteFunctionContext> {
+    public hideStepCount: boolean = false;
+
+    public async prompt(context: EventGridExecuteFunctionContext): Promise<void> {
+        if (!context.eventSource) {
+            throw new Error('Event source is required');
```
nturinski, line 38: nit: localize, or just use `nonNullProp` here.

src/commands/executeFunction/eventGrid/EventGridTypeStep.ts
```diff
@@ -0,0 +1,72 @@
...
+                .join(' '),
+        }));
+
+        context.selectedFileName =
+            (
+                await context.ui.showQuickPick(eventTypePicks, {
+                    placeHolder: localize('selectEventType', 'Select the event type'),
+                    stepName: 'eventType',
+                })
+            ).data ?? 'blob_created.json';
```
nturinski, line 63: Similar question as above.

src/commands/executeFunction/eventGrid/sendEventGridRequest.ts
```diff
@@ -0,0 +1,35 @@
...
+    const requestContent: string = document.getText();
+
+    const node = ext.fileToFunctionNodeMap.get(document.fileName);
+
+    if (!node) {
+        const errorMsg = localize(
+            'noFunctionBeingExecuted',
+            'No function is currently being executed. ' +
+            'This command is intended to be run while an EventGrid function is being executed. ' +
+            'Please make to execute your EventGrid function.',
```
nturinski, line 29:
```suggestion
            'Please make sure to execute your EventGrid function.',
```

src/commands/executeFunction/executeFunction.ts
```diff
@@ -8,54 +8,79 @@ import { type SiteClient } from '@microsoft/vscode-azext-azureappservice';
...
-            functionInput = <{}>JSON.parse(data);
-        } catch {
-            functionInput = data;
+            const data: string = await context.ui.showInputBox({
+                prompt,
+                value,
+                stepName: 'requestBody',
+            });
+            try {
+                functionInput = <{}>JSON.parse(data);
```
nturinski, line 68: Is the data sometimes a JSON object and other times not? I'd assume that we would want to throw if it wasn't since that seems like the sample event would be corrupted.

nturinski (commented):
Overall looking great. Most of my comments are just nits or small questions. Nice work!

src/commands/executeFunction/eventGrid/EventGridSourceStep.ts
hossam-nasr: I'm not actually sure what scenario causes this to be undefined, but I saw that it's a possible scenario in the typing. This is a leftover from a rougher prototype tbh. What should be the appropriate scenario here?

src/commands/executeFunction/eventGrid/EventGridTypeStep.ts
hossam-nasr: Similar answer as above haha

src/commands/executeFunction/executeFunction.ts
hossam-nasr: This is actually the old `executeFunction` code, I didn't modify this specific line. I'm not sure in which cases is it a JSON and which cases it is a straight string

src/commands/executeFunction/eventGrid/EventGridSourceStep.ts
nturinski: We should probably just throw an error here honestly since it isn't likely to happen.

I think ignoring the undefined and autofilling it with storage would be a really weird behavior for the user to decipher what the heck just happened lol

package.json
hossam-nasr: Resolving for now unless someone can suggest a way to ensure it is also at the farthest left

src/commands/executeFunction/eventGrid/EventGridFileOpenStep.ts
```diff
@@ -45,23 +45,47 @@ export class EventGridFileOpenStep extends AzureWizardExecuteStep<EventGridExecu
             await ext.context.workspaceState.update('didShowEventGridFileOpenMsg', true);
         }
 
+        // Set a listener to track whether the file was modified before the request is sent
+        let modifiedListenerDisposable: vscode.Disposable;
+        void new Promise<void>((resolve, reject) => {
+            modifiedListenerDisposable = vscode.workspace.onDidChangeTextDocument(async (event) => {
```
nturinski, line 51: This looks good and seems to work. One thing to consider though:
1. Do we want to fire an event every time a user edits the same trigger?
2. If we do, I'd say it's probably more valuable to keep a running count of how many times the trigger was edited before sent.

nturinski: Actually, it might be better for this to modify a context and then add it to the context of `executeFunctionWithInput` if possible. That way we just have a running count and the event is only tracked once.

package.json
```diff
@@ -697,6 +697,7 @@
             "editor/title": [
                 {
                     "command": "azureFunctions.eventGrid.sendMockRequest",
+                    "arguments": ["${file}"],
```
nturinski, line 700: Very nice find on this and the codelens one!

src/commands/executeFunction/eventGrid/EventGridCodeLensProvider.ts
```diff
@@ -12,6 +12,7 @@ export class EventGridCodeLensProvider implements CodeLensProvider {
         firstLineLens.command = {
             title: localize('saveExecute', 'Save and execute'),
             command: 'azureFunctions.eventGrid.sendMockRequest',
+            arguments: ['codeLens']
```
nturinski, line 15: It's worth to just make `codeLens` a constant in our `constant.ts` file and importing everywhere it's used.

src/commands/executeFunction/eventGrid/OpenEventGridFileStep.ts
nturinski: I think that will be helpful enough!

src/commands/executeFunction/eventGrid/EventGridFileOpenStep.ts
nturinski: Spoken offline: disregard this comment. I'd still prefer it gets added to the context of `executeFunctionWithInput` but not a blocker.

nturinski (approved):

PR #3989, opened 2024-02-26
UX: Change command labels, remove entry point for `Create new project...`, add entry points for `Create Function...`

nturinski:
This changes the Functions UI a bit to match what we discussed offline and insights from user studies.

- Go back to removing `Create New Project...`. Replace `Create Function...` with `Create function locally...` (Not sure what looks best in terms of casing)
If you're unfamiliar, `Create Function...` handles the following scenarios:
1. Local project exists? Add a new function template.
2. Local project exists but it's not initialized for VS Code? Prompts user to initialize for VS Code.
3. Local project doesn't exist? Create a new project _and_ ask what function template user wants (option to skip)

- Added two entry points to `Create function locally...`. One on the `Functions` node indicated by the Function + button. Another as a context menu action. I'm debating moving it to the local project node.
image

- Mentioned renaming `Create Function...` but also renamed `Deploy to Function App in Azure...` to `Deploy function to Azure...`. It makes more intuitive sense because your local project is being deployed to Azure, and we have to prompt for which app is being deployed to.
image

By that logic, I've kept `Deploy to Function App...` the same  in the `Resources` view though.
image

- Added `Create New Project...` to workspace view if there's no project in the workspace. It actually does call `azureFunctions.createNewProject` because we know that we need to start from scratch. `createFunction` has to do some checks before it ultimately cancels and calls `createNewProject` so performance is slightly better to just call it directly.
image

Still to do:
- Remove `Create Function App...` option in the local workspace view
  - This will be replaced with just `Deploy function to Azure...`. My plan is to include a `+ Create new function app` option as a quickpick when we prompt for the app
- Merge `Create New Containerized Project...` into `Create function locally...` and add it as a prompt at the beginning. It's annoying, but also a one-time cost to the user that can be easily entered through.

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,10 +175,6 @@
                 "command": "azureFunctions.createNewProject",
                 "title": "%azureFunctions.createNewProject%",
                 "category": "Azure Functions",
-                "icon": {
-                    "light": "resources/light/CreateNewProject.svg",
-                    "dark": "resources/dark/CreateNewProject.svg"
-                },
                 "enablement": "!virtualWorkspace"
             },
             {
@@ -369,10 +365,6 @@
                     "group": "1_projects@1"
                 },
                 {
-                    "command": "azureFunctions.createNewProject",
-                    "group": "1_projects@2"
-                },
-                {
                     "command": "azureFunctions.createNewProjectWithDockerfile",
                     "group": "1_projects@3"
                 },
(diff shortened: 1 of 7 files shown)


nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurefunctions/pull/4042

Issue #3940, opened 2024-01-16
Extension fails when TargetFramework is centralised to solution level


DaRosenberg:
For easier maintenance, we have centralised a number of project properties to a `Directory.Build.props` file at the solution level, including the `TargetFramework` property, like so:

```xml
<Project>
  <PropertyGroup>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <EnableNETAnalyzers>true</EnableNETAnalyzers>
    <AnalysisMode>AllEnabledByDefault</AnalysisMode>
    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
  </PropertyGroup>
</Project>
```

This makes the Azure Functions extension fail; apparently it is trying to extract this property from the function app `.csproj` file, assuming it will always be there:

<img width="571" alt="image" src="https://github.com/microsoft/vscode-azurefunctions/assets/3889090/fe2d2200-2058-4a2a-820c-7d18315e02fb">

This is an incorrect assumption, and the `TargetFramework` property value, if it is indeed really needed by the extension, should instead be determined through MSBuild (perhaps [this new feature in the .NET 8 SDK](https://learn.microsoft.com/en-us/dotnet/core/whats-new/dotnet-8#cli-based-project-evaluation) could be utilised?).

DaRosenberg:
@nturinski I think it's generally not a good idea to try to duplicate MSBuild functionality, because it leads to exactly these kinds of issues when those semantics are expanded or changed.

I do understand that it might be tricky to rely on .NET 8 SDK though.

I don't think there is a "clean" way to determine from simply inspecting a `.csproj` file that some of its properties are inherited from a `.props` file. As far as I understand, when building a `.csproj` file directly, MSBuild will always go looking for `.props` files by traversing the directory structure all the way up to a directory that contains a solution that happens to include the project being built.

However, in your case maybe it would be "good enough" to simply determine that the `.csproj` does not include the `<TargetFramework>` element, and in that case look for the appropriate `.props` file instead and try to get it from there. Many things do rely on the `<TargetFramework>` being specified in at least one of those two locations, so it's a fairly safe assumption that, if it's not in one, it will be in the other.

The `ManagePackageVersionsCentrally` property you mentioned is unrelated, it has to do with centrally managing NuGet package versions, not MSBuild properties.

PR #4013, opened 2024-03-05
Check for a `Directory.Build.props` file

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/3940

If we can't find the property we're looking for in the csproj file, we should check the Directory.Build.props file. It's a standard file name in .NET. More about that here: https://learn.microsoft.com/en-us/visualstudio/msbuild/customize-by-directory?view=vs-2022#directorybuildprops-and-directorybuildtargets

But basically, `dotnet` commands know to check the props file and apply that to other projects.

### DIFF
--- a/src/utils/dotnetUtils.ts
+++ b/src/utils/dotnetUtils.ts
@@ -5,6 +5,7 @@
 
 import { AzExtFsExtra, type IActionContext } from '@microsoft/vscode-azext-utils';
 import * as path from 'path';
+import { workspace } from 'vscode';
 import { FuncVersion } from '../FuncVersion';
 import { ProjectLanguage } from '../constants';
 import { localize } from "../localize";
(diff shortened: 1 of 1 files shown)


src/utils/dotnetUtils.ts
```diff
@@ -80,6 +81,17 @@ export namespace dotnetUtils {
         const regExp: RegExp = new RegExp(`<${prop}>(.*)<\\/${prop}>`);
         const matches: RegExpMatchArray | null = (await projFile.getContents()).match(regExp);
         if (!matches) {
+            // if we can't find the targetFramework in the csproj, there may be a `Directory.Build.props` file that sets it
+            const buildPropsFile: string = 'Directory.Build.props';
+            const buildPropsFiles = await workspace.findFiles(buildPropsFile, null);
+            if (buildPropsFiles.length > 0) {
+                const buildPropsContents = (await AzExtFsExtra.readFile(buildPropsFiles[0])).toString();
```
MicroFish91, line 88: Is the `.toString()` necessary here / doesn't `readFile` already return a string?

MicroFish91 (dismissed):
Otherwise LGTM, comment was just a question / nit.

src/utils/dotnetUtils.ts
nturinski: You are correct. I have a feeling copilot added that lol

motm32 (approved):

PR #4033, opened 2024-03-11
Fix no activity log when deleting containerized function app

### LINKED ISSUE #4004: There is no activity log in the "AZURE" window when deleting a containerized function app
**OS:** All
**Build Version:** 20240229.2

**Repro Steps:**
1. Delete a containerized function app.
2. Check whether there is an activity log in the "AZURE" window.

**Expect:**
There is an activity log in the "AZURE" window, like the one when deleting a regular function app.
image

**Actual:**
There is no activity log in the "AZURE" window.

**More Info:**
This issue does not reproduce for regular function app.

motm32:
Fixes #4004. 

Can't reuse the deleteSite steps within the appservice tools package since we cannot create a client the same way.

### DIFF
--- a/src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
+++ b/src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
@@ -1,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { createWebSiteClient } from "@microsoft/vscode-azext-azureappservice";
+import { AzureWizardExecuteStep, nonNullValueAndProp } from "@microsoft/vscode-azext-utils";
+import { ProgressLocation, window } from "vscode";
+import { ext } from "../../extensionVariables";
+import { localize } from "../../localize";
+import { type DeleteFunctionappWizardContext } from "./DeleteFunctionAppWizardContext";
+
+export class DeleteContainerizedFunctionappStep extends AzureWizardExecuteStep<DeleteFunctionappWizardContext> {
+    public priority: number = 100;
+
+    public async execute(context: DeleteFunctionappWizardContext): Promise<void> {
+        const deleting: string = localize('DeletingFunctionApp', 'Deleting function app "{0}"...', context.site.name);
+        const deleteSucceeded: string = localize('DeleteFunctionAppSucceeded', 'Successfully deleted function app "{0}".', context.site.name);
+
+        await window.withProgress({ location: ProgressLocation.Notification, title: deleting }, async (): Promise<void> => {
+            ext.outputChannel.appendLog(deleting);
+            const client = await createWebSiteClient([context, context.proxyTree.subscription]);
+            await client.webApps.delete(nonNullValueAndProp(context.site, 'resourceGroup'), nonNullValueAndProp(context.site, 'name'));
+            void window.showInformationMessage(deleteSucceeded);
+            ext.outputChannel.appendLog(deleteSucceeded);
+        });
+    }
+
+    public shouldExecute(): boolean {
+        return true;
+    }
+}
--- a/src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
+++ b/src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
@@ -5,7 +5,7 @@
 
 import { createWebSiteClient } from "@microsoft/vscode-azext-azureappservice";
 import { AzureWizardExecuteStep, nonNullValueAndProp } from "@microsoft/vscode-azext-utils";
-import { ProgressLocation, window } from "vscode";
+import { window } from "vscode";
 import { ext } from "../../extensionVariables";
 import { localize } from "../../localize";
 import { type DeleteFunctionappWizardContext } from "./DeleteFunctionAppWizardContext";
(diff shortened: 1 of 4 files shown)


src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResource.ts
```diff
@@ -139,18 +141,24 @@ export class ResolvedContainerizedFunctionAppResource extends ResolvedFunctionAp
...
-        const deleting: string = localize('DeletingFunctionApp', 'Deleting function app "{0}"...', this.site.name);
-        const deleteSucceeded: string = localize('DeleteFunctionAppSucceeded', 'Successfully deleted function app "{0}".', this.site.name);
-        await context.ui.showWarningMessage(message, { modal: true, stepName: 'confirmDelete' }, DialogResponses.deleteResponse);
-        await window.withProgress({ location: ProgressLocation.Notification, title: deleting }, async (): Promise<void> => {
-            ext.outputChannel.appendLog(deleting);
-            const client = await createWebSiteClient([context, proxyTree.subscription]);
-            await client.webApps.delete(nonNullValueAndProp(this.site, 'resourceGroup'), nonNullValueAndProp(this.site, 'name'));
-            void window.showInformationMessage(deleteSucceeded);
-            ext.outputChannel.appendLog(deleteSucceeded);
+        const title: string = localize('DeleteFunctionApp', 'Delete function app');
```
MicroFish91, line 151: I think it might look nice to add the app name to the activity title here here :)

src/tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResource.ts
```diff
@@ -139,18 +141,24 @@ export class ResolvedContainerizedFunctionAppResource extends ResolvedFunctionAp
...
+
+        const wizard = new AzureWizard(wizardContext, {
+            promptSteps: [new DeleteConfirmationStep(message)],
+            executeSteps: [new DeleteContainerizedFunctionappStep()],
+            title
         });
+
+        await wizard.prompt();
+        await wizard.execute();
+
```
MicroFish91, line 161: Nit: Extra new line

src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
```diff
@@ -0,0 +1,32 @@
...
+import { type DeleteFunctionappWizardContext } from "./DeleteFunctionAppWizardContext";
+
+export class DeleteContainerizedFunctionappStep extends AzureWizardExecuteStep<DeleteFunctionappWizardContext> {
+    public priority: number = 100;
+
+    public async execute(context: DeleteFunctionappWizardContext): Promise<void> {
+        const deleting: string = localize('DeletingFunctionApp', 'Deleting function app "{0}"...', context.site.name);
+        const deleteSucceeded: string = localize('DeleteFunctionAppSucceeded', 'Successfully deleted function app "{0}".', context.site.name);
+
+        await window.withProgress({ location: ProgressLocation.Notification, title: deleting }, async (): Promise<void> => {
```
MicroFish91, line 20: Question, can we skip showing this `withProgress` since we already have the live update showing in the activity log?

MicroFish91 (approved):
Mostly nits, otherwise LGTM

src/commands/deleteContainerizedFunctionApp/DeleteFunctionAppWizardContext.ts
```diff
@@ -0,0 +1,13 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type Site } from "@azure/arm-appservice";
+import { type ExecuteActivityContext, type IActionContext } from "@microsoft/vscode-azext-utils";
+import { type ContainerTreeItem } from "../../tree/containerizedFunctionApp/ContainerTreeItem";
+
+export interface DeleteFunctionappWizardContext extends IActionContext, ExecuteActivityContext {
+    site: Site;
+    proxyTree: ContainerTreeItem;
```
MicroFish91, line 12: Actually one other thing, is the only reason we have the proxyTree on the context for the subscription?  If so, maybe you can just build an `ISubscriptionActionContext` at the beginning so you don't have to store the whole tree item reference on the context.

src/commands/deleteContainerizedFunctionApp/DeleteContainerizedFunctionAppStep.ts
nturinski: Yeah, I think so. It should be our pattern going forward-- maybe a notification that says:
`Deleting function "{0}". To check progress, view Azure Activity Log.` with a button to take them there?

MicroFish91: Double check me on this, but it seems like anytime the activity log is activated through the wizard, the activity log pane automatically gets focus so it'd probably be hard for the user to miss.

MicroFish91: Oh, you mean tree item description, nevermind, yeah that makes perfect sense.

nturinski: Actually, I didn't realize that the Azure activity log was already focused whenever something happened there so maybe just disregard what I said.

The tree item description might be a bit too verbose to show it there.

src/commands/deleteContainerizedFunctionApp/DeleteFunctionAppWizardContext.ts
motm32: I don't think I can do this since the I don't have access to the `AzureSubscription` either but I made some changes and put the subscription in the constructor so I could reuse it. Let me know if that's better/worse.

src/commands/deleteContainerizedFunctionApp/DeleteFunctionAppWizardContext.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type Site } from "@azure/arm-appservice";
+import { type ExecuteActivityContext, type IActionContext, type ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+
+export interface DeleteFunctionappWizardContext extends ISubscriptionActionContext, IActionContext, ExecuteActivityContext {
```
MicroFish91, line 9: Think you forgot to remove the extra `IActionContext` here :)

MicroFish91 (approved):
Yay, ship it :D

nturinski:
> Yay, ship it :D

You need to reapprove it then.

motm32:
> > Yay, ship it :D
> 
> You need to reapprove it then.
Jokes on you

PR #4034, opened 2024-03-11
Fix containerized function app refresh on create

motm32:
I noticed when creating and new containerized function app often the app would not show in the tree right after creation. It seems like the service has a bit of delay between when it is created and when we get it. Pinging the hostname and then refreshing seems to fix this issue.


alexweininger (approved):
This works? 😮

motm32:
> This works? 😮

It does on my machine 🤷‍♀️

PR #4061, opened 2024-03-28
Fix creating a function app with no workspace open

motm32:
I was checking telemetry today and noticed the error count was a little high for the create function app telemetry event. Looking further into we were blocking users from creating a function app when no workspace was open. This happened due to the containerized function app flow checking for a dockerfile in the workspace. I added a check to see if a workspace was open before running this step which should fix that issue.

### DIFF
--- a/src/commands/createFunctionApp/containerImage/detectDockerfile.ts
+++ b/src/commands/createFunctionApp/containerImage/detectDockerfile.ts
@@ -20,7 +20,7 @@ export async function detectDockerfile(context: ICreateFunctionAppContext): Prom
         context.rootPath = workspacePath;
 
         //check for host.json location
-        if (!(await isFunctionProject(workspacePath))) {
+        if (await isFunctionProject(workspacePath)) {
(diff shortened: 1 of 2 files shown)


nturinski (approved):

src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -20,7 +20,7 @@ export async function detectDockerfile(context: ICreateFunctionAppContext): Prom
         context.rootPath = workspacePath;
 
         //check for host.json location
-        if (!(await isFunctionProject(workspacePath))) {
+        if (await isFunctionProject(workspacePath)) {
             const files = (await findFiles(context.workspaceFolder, `*/${hostFileName}`));
             if (files.length === 0) {
```
nturinski, line 25: You honestly probably don't need these checks anymore either.

motm32: Talked offline and the checks are still needed.

PR #4062, opened 2024-03-29
Switch Dockerfile and host.json check when creating function app

motm32:
Switching the two checks since it is more likely that users will have a host.json in their project than a Dockerfile. This way we can abandon the check when we don't find a Dockerfile.

### DIFF
--- a/src/commands/createFunctionApp/containerImage/detectDockerfile.ts
+++ b/src/commands/createFunctionApp/containerImage/detectDockerfile.ts
@@ -16,21 +16,21 @@ export async function detectDockerfile(context: ICreateFunctionAppContext): Prom
     if (vscode.workspace.workspaceFolders) {
         context.workspaceFolder = vscode.workspace.workspaceFolders[0];
         const workspacePath = context.workspaceFolder.uri.fsPath;
-        let hostPath: string = workspacePath
+        let dockerfilePath: string = workspacePath
         context.rootPath = workspacePath;
 
-        //check for host.json location
+        //check for dockerfile location
         if (await isFunctionProject(workspacePath)) {
-            const files = (await findFiles(context.workspaceFolder, `*/${hostFileName}`));
+            const files = (await findFiles(context.workspaceFolder, `**/${dockerfileGlobPattern}`));
             if (files.length === 0) {
-                throw new Error(localize('noHostJson', 'No host.json file found in the current workspace.'));
+                return;
             }
-            hostPath = path.dirname(files[0].fsPath);
+            dockerfilePath = path.dirname(files[0].fsPath);
         }
 
-        // check if dockerfile exists in the same folder as the host.json
-        if ((await findFiles(hostPath, dockerfileGlobPattern)).length > 0) {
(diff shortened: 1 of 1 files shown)


src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -16,21 +16,21 @@ export async function detectDockerfile(context: ICreateFunctionAppContext): Prom
...
-            hostPath = path.dirname(files[0].fsPath);
+            dockerfilePath = path.dirname(files[0].fsPath);
         }
 
-        // check if dockerfile exists in the same folder as the host.json
-        if ((await findFiles(hostPath, dockerfileGlobPattern)).length > 0) {
-            context.dockerfilePath = (await findFiles(hostPath, dockerfileGlobPattern))[0].fsPath;
+        // check if host.json is in the same directory as the Dockerfile
+        if ((await findFiles(dockerfilePath, hostFileName)).length > 0) {
+            context.dockerfilePath = (await findFiles(dockerfilePath, hostFileName))[0].fsPath;
```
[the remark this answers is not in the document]
motm32: Oh that is an issue with the switcharoo I was saving just the host.json directory so the dockerfilepath just includes the directory not the full path. I'll fix that.

src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -16,21 +16,21 @@ export async function detectDockerfile(context: ICreateFunctionAppContext): Prom
...
-            hostPath = path.dirname(files[0].fsPath);
+            dockerfilePath = path.dirname(files[0].fsPath);
         }
 
-        // check if dockerfile exists in the same folder as the host.json
-        if ((await findFiles(hostPath, dockerfileGlobPattern)).length > 0) {
-            context.dockerfilePath = (await findFiles(hostPath, dockerfileGlobPattern))[0].fsPath;
+        // check if host.json is in the same directory as the Dockerfile
+        if ((await findFiles(dockerfilePath, hostFileName)).length > 0) {
+            context.dockerfilePath = (await findFiles(dockerfilePath, dockerfileGlobPattern))[0].fsPath;
```
nturinski, line 33: Why can't you just set it as 
```suggestion
            context.dockerfilePath = dockerfilePath;
```

motm32: dockerfile path above is set to the directory not the actualy path of the dockerfile. This is because we check the host.json path using the dockerfile directory we found. When we pass this context value to ACA it won't work with just the directory path. I can make two seperate constants and set one as the dockerfile but that seemed a little over kill, but since it seems confusing I can do that if you want.

nturinski: Well, the main thing that I wanted to avoid is to have to do another `findFiles` call in here, but I guess it was always like that so whatever 🤷

nturinski (approved):

Issue #4070, opened 2024-04-01
Fail to configure deployment source for the Flex Consumption (Preview) function app


v-xinda:
**OS:** All
**Build Version:** v1.14.1-alpha.0
**Regression:** Not a Regression

**Repro Steps:**
1. Create a python function app with Flex Consumption (Preview) hosting plan and East US 2 EUAP location.
2. Right click this function app -> Configure Deployment Source... -> Select "GitHub" -> Select an organization -> Select a Repository -> Select a branch.
3. Check whether succeeds to configure deployment source.

**Expect:**
Succeed to configure deployment source.

**Actual:**
Fail to configure deployment source.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/57662131/a8ea79b0-894f-49d8-a13f-a0c885339cab)

**More Info:**
There is no error after configuring deployment source to "LocalGit".

Issue #4068, opened 2024-04-01
Pop up an error after selecting "Premium/App Service Plan" hosting plan


v-chenzha:
**OS:** All
**Build Version:** v1.14.1-alpha.0
**Regression:** Not a Regression

**Repro Steps:**

1. F1 -> Azure Functions: Create Function App in Azure...(Advanced) -> Select a subscription -> Enter a name -> Select "Premium/App Service Plan" hosting plan -> Enter.
2. Check whether the next step is "Select a runtime stack".

**Expect:**
The next step is "Select a runtime stack".

**Actual:**
Pop up an error.
![image](https://github.com/microsoft/vscode-azurefunctions/assets/107091828/6196e1c1-96c5-48f6-9940-105532d36c74)

**More Info:**
This issue doesn't reproduce after selecting "Consumption/Flex Consumption(Preview)" hosting plan.

PR #4078, opened 2024-04-04
Flex fixes and includes the adding flex to function app context values

nturinski:
Includes this https://github.com/microsoft/vscode-azurefunctions/pull/4072 PR but I ended up changing the context value portion of it slightly since a flex app is different from `productionSlot` or `slot`.

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4070 (by removing the deployment configuration command)
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4068 (by removing the sub-wizard that asks for a plan immediately after picking the host service type. We need to let users select a runtime _first_ because which OS it can run on is stack dependent)

### DIFF
--- a/src/FunctionAppResolver.ts
+++ b/src/FunctionAppResolver.ts
@@ -8,11 +8,11 @@ import { ResolvedFunctionAppResource } from "./tree/ResolvedFunctionAppResource"
 import { ResolvedContainerizedFunctionAppResource } from "./tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResource";
 import { createWebSiteClient } from "./utils/azureClients";
 
-// TODO: this is temporary until the new api-version is available in the SDK
-type Site2 = Site & { isFlex?: boolean };
+// TODO: this is temporary until the new SDK with api-version=2023-12-01 is available
+type Site20231201 = Site & { isFlex?: boolean };
 export class FunctionAppResolver implements AppResourceResolver {
     private siteCacheLastUpdated = 0;
-    private siteCache: Map<string, Site2> = new Map<string, Site2>();
+    private siteCache: Map<string, Site20231201> = new Map<string, Site20231201>();
 
     public async resolveResource(subContext: ISubscriptionContext, resource: AppResource): Promise<ResolvedFunctionAppResource | ResolvedContainerizedFunctionAppResource | undefined> {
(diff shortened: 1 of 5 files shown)


src/commands/createFunctionApp/FunctionAppHostingPlanStep.ts
```diff
@@ -42,12 +42,7 @@ export class FunctionAppHostingPlanStep extends AzureWizardPromptStep<IFunctionA
         return context.useConsumptionPlan === undefined && context.dockerfilePath === undefined;
     }
 
-    public async getSubWizard(_context: IFunctionAppWizardContext): Promise<IWizardOptions<IFunctionAppWizardContext> | undefined> {
-        if (_context.dockerfilePath) {
-            return undefined;
-        }
-        return { promptSteps: [new AppServicePlanListStep()] };
-    }
+
```
motm32, line 45: nit: Left an extra line here

motm32 (dismissed):

src/FunctionAppResolver.ts
```diff
@@ -1,14 +1,18 @@
...
+import { createGenericClient, uiUtils, type AzExtPipelineResponse, type AzExtRequestPrepareOptions } from "@microsoft/vscode-azext-azureutils";
 import { callWithTelemetryAndErrorHandling, nonNullProp, nonNullValue, nonNullValueAndProp, type IActionContext, type ISubscriptionContext } from "@microsoft/vscode-azext-utils";
 import { type AppResource, type AppResourceResolver } from "@microsoft/vscode-azext-utils/hostapi";
+import { type FunctionAppConfig } from "./commands/createFunctionApp/FunctionAppCreateStep";
 import { ResolvedFunctionAppResource } from "./tree/ResolvedFunctionAppResource";
 import { ResolvedContainerizedFunctionAppResource } from "./tree/containerizedFunctionApp/ResolvedContainerizedFunctionAppResource";
 import { createWebSiteClient } from "./utils/azureClients";
 
+// TODO: this is temporary until the new api-version is available in the SDK
+type Site2 = Site & { isFlex?: boolean };
```
JasonYeMSFT, line 12: Rename the type to Site20231201 or add that specific version to the comment so that people can easily determine if a new version of the SDK has onboarded the required api version.

motm32 (approved):

JasonYeMSFT (approved):

PR #4095, opened 2024-04-22
add sqlTrigger to verifiedtemplates

MaddyDev:
adding Sql trigger to verified templates


nturinski:
Hi @MaddyDev

Part of our process to adding templates to the verified list is to create a test for creating that specific template trigger. We do this to ensure that the template is created as we expected and we can catch anytime the input/parameter change from what we expected.

Please add a test creating the sqlTrigger. Feel free to reach out and I can give you examples of how to do that.

MaddyDev:
> Hi @MaddyDev
> 
> Part of our process to adding templates to the verified list is to create a test for creating that specific template trigger. We do this to ensure that the template is created as we expected and we can catch anytime the input/parameter change from what we expected.
> 
> Please add a test creating the sqlTrigger. Feel free to reach out and I can give you examples of how to do that.

Thanks @nturinski for calling out, I have added them and haven't pushed them. Quick qiestion, I see that some of the extensions also edited templateCount.test.ts (ref https://github.com/microsoft/vscode-azurefunctions/pull/3916/files). Can you explain if this is needed for us?

nturinski:
@MaddyDev So it seems like a lot of the tests are failing now. Is the SqlTrigger actually available in all of these runtimes? Like Python, all .NET runtimes, etc.? It doesn't appear to be showing up.

nturinski:
@MaddyDev Just want to see if there's any update on this

nturinski:
Closing in favor of https://github.com/microsoft/vscode-azurefunctions/pull/4358

PR #4104, opened 2024-04-25
Use Stacks API for Flex Consumption SKU

nturinski:
There are some different requirements for the stack API when acquiring the runtimes
1. Location is required as part of the endpoint URL. Due to this constraint, I moved the location step to be before the stacks step
2. Stack is a required query parameter which is why I have the for loop of stacks
3. If the stack has the Sku array under its LinuxRuntimeSettings, than it is flex compatible. All other runtimes have Sku as null.

### DIFF
--- a/src/commands/createFunctionApp/stacks/getStackPicks.ts
+++ b/src/commands/createFunctionApp/stacks/getStackPicks.ts
@@ -5,7 +5,7 @@
 
 import { type ServiceClient } from '@azure/core-client';
 import { createPipelineRequest } from '@azure/core-rest-pipeline';
-import { createGenericClient, type AzExtPipelineResponse } from '@microsoft/vscode-azext-azureutils';
+import { createGenericClient, LocationListStep, type AzExtPipelineResponse } from '@microsoft/vscode-azext-azureutils';
 import { openUrl, parseError, type AgentQuickPickItem, type IAzureQuickPickItem } from '@microsoft/vscode-azext-utils';
 import { funcVersionLink } from '../../../FuncVersion';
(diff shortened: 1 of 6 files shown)


src/commands/createFunctionApp/stacks/getStackPicks.ts
```diff
@@ -176,6 +178,46 @@ async function getStacks(context: IFunctionAppWizardContext & { _stacks?: Functi
...
+async function getFlexStacks(context: IFunctionAppWizardContext & { _stacks?: FunctionAppStack[] }): Promise<FunctionAppStack[]> {
+    const client: ServiceClient = await createGenericClient(context, context);
+    const location = await LocationListStep.getLocation(context);
+    const flexFunctionAppStacks: FunctionAppStack[] = [];
+    const stacks = ['dotnet', 'java', 'node', 'powershell', 'python'];
+    if (!context._stacks) {
+        const getFlexStack = async (stack: string) => {
+            const result: AzExtPipelineResponse = await client.sendRequest(createPipelineRequest({
+                method: 'GET',
+                url: requestUtils.createRequestUrl(`providers/Microsoft.Web/locations/${location.name}/functionAppStacks`, {
```
alexweininger, line 190: so flex stacks are location dependent? That's crazy

nturinski: Yeah, I can see it being helpful for this team since they have to deploy skus to regions individually (it seems like).

alexweininger (dismissed):

alexweininger (approved):

PR #4115, opened 2024-04-30
Add EG Blob Trigger templates

castrodd:
This PR is adding templates for the Event Grid-based Blob Storage trigger (as described in [this tutorial](https://learn.microsoft.com/en-us/azure/azure-functions/functions-event-grid-blob-trigger?tabs=isolated-process%2Cnodejs-v4&pivots=programming-language-csharp)).

What's included:

- Templates for JavaScript and TypeScript

cc: @ejizba

### DIFF
--- a/resources/backupTemplates/nodejs-4.x/templates/templates.json
+++ b/resources/backupTemplates/nodejs-4.x/templates/templates.json
@@ -1453,6 +1453,128 @@
         }
     },
     {
+        "id": "EventGridBlobTrigger-JavaScript",
+        "runtime": "2",
+        "files": {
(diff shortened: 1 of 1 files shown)


resources/backupTemplates/nodejs-4.x/templates/templates.json
```diff
@@ -1452,6 +1452,128 @@
             "userPrompt": []
         }
     },
+    {
+        "id": "EventGridBlobTrigger-JavaScript",
```
ejizba, line 1456: You can remove the non "-4.x" templates, they're not used. This file was originally auto-generated from a custom branch of the templates repo, which is the only reason you see other non "-4.x" templates. Since I assume you didn't auto-generate it again, there's no point in adding them manually

If you're curious, [this code here](https://github.com/microsoft/vscode-azurefunctions/blob/main/src/templates/script/NodeV4Provider.ts#L62) is what filters them out (only the v4 templates will have the triggerType property)

castrodd: Got it. Removed non-4.x templates.

ejizba (approved):

PR #4125, opened 2024-05-10
Add new command `deployByFunctionAppId`

MicroFish91:
One or more teams currently leverage our deploy command directly using `commands.executeCommand`.  We previously allowed the `functionAppId` target to be passed as a third argument into `deployProductionSlot`.  Upon investigation, I noticed that only the first arg was being passed down when wrapped with `registerCommandWithTreeNodeUnwrapping`.  

Calling with `registerCommand` instead seems to preserve all args.  As a workaround, I have created a command for these teams to utilize directly where the only thing you need to pass it is the `functionAppId`.  This is probably better also because they are not really ever passing a tree item anyway, so we don't ever need to unwrap for their use-case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -215,6 +215,12 @@
                 "enablement": "!virtualWorkspace"
             },
             {
+                "command": "azureFunctions.deployByFunctionAppId",
+                "title": "%azureFunctions.deployByFunctionAppId%",
+                "category": "Azure Functions",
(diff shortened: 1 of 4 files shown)


nturinski (approved):

nturinski:
Confirmed with partner team that needs this that it's working. We're having another issue with the defaultHostname thing, but that's unrelated to these changes.

PR #4130, opened 2024-05-14
Create from Deploy feature and some local UX changes

nturinski:
This is basically this PR https://github.com/microsoft/vscode-azurefunctions/pull/4042 but I removed some of the UX changes regarding "Create function locally...".

I really didn't like having the prompt asking for what type of project does the user want to create. It doesn't really make sense to have it in our main entry point, so until we can think of a good solution, the command list looks like this:

Old:
image
New:
image

- Removed `Create Function App...` option in the local workspace view
  - This was replaced with just `Deploy function to Azure...` which includes a `+ Create new function app` option as a quick pick when we prompt for the app
 image

_Other changes I probably should have made separate PRs for_ 😅:
- Added `Create New Project...` to workspace view if there's no project in the workspace. It actually does call `azureFunctions.createNewProject` because we know that we need to start from scratch. `createFunction` has to do some checks before it ultimately cancels and calls `createNewProject` so performance is slightly better to just call it directly.
Old:  
image
New: 
image

- Added two entry points for "Create Function..." which live on the local project node and context menu
image


package.nls.json
```diff
@@ -24,6 +24,7 @@
     "azureFunctions.deleteFunction": "Delete Function...",
     "azureFunctions.deleteFunctionApp": "Delete Function App...",
     "azureFunctions.deleteSlot": "Delete Slot...",
+    "azureFunctions.deployProject": "Deploy to Azure...",
```
alexweininger, line 27: > Removed Create Function App... option in the local workspace view
This was replaced with just Deploy function to Azure... which includes a + Create new function app option as a quick pick when we prompt for the app

image

Is it supposed to be "Deploy to Azure..." or "Deploy function to Azure..."?

nturinski: Ah, I think we originally went with "Deploy function to Azure..." but after more discussion, just went with a more generic "Deploy to Azure..."

alexweininger: We don't think we need to give more context to users on _what_ they are deploying?

nturinski: Yeah, I mean you're in the workspace part of the extension so I think it's kind of clear. That being said, I think it still prompts you for a workspace folder.

I'm thinking of adding a right-click context item on the local project node though that also has "Deploy to Azure..." on it.

alexweininger: Well I was thinking that they might need to know that the command is for deploying Functions to Azure and not App Service or anything else. But maybe they should already know?

nturinski: Well, you click the Functions symbol to open the menu for the commands. I thought that was kind of the reason we had those buttons there to give them that context.

alexweininger: Ok, that's fine

src/commands/SubscriptionListStep.ts
```diff
@@ -0,0 +1,37 @@
...
+    public async configureBeforePrompt(context: IFuncDeployContext): Promise<void> {
+        this._picks = await this.getPicks(context);
+        // auto select if only one subscription
+        if (this._picks.length === 1) {
+            this._oneSubscription = true;
+            context.subscription = this._picks[0].data;
+        }
+    }
+
+    private async getPicks(_: IFuncDeployContext): Promise<IAzureQuickPickItem<AzureSubscription>[]> {
```
alexweininger, line 32: Any reason we didn't just remove the parameter if it's not being used?

src/commands/deploy/getOrCreateFunctionApp.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export async function getOrCreateFunctionApp(context: IFuncDeployContext & Partial<IFunctionAppWizardContext>): Promise<SlotTreeItem> {
+    let node: SlotTreeItem | undefined;
+
+    const promptSteps = [new SubscriptionListStep(), new FunctionAppListStep()];
+    const title: string = l10n.t('Select Function App');
+    const wizard: AzureWizard<IAppServiceWizardContext> = new AzureWizard(context, {
+        promptSteps,
+        title
+    });
```
alexweininger, line 26: I prefer if you just put the `promptSteps` and `title` values inline here since they aren't used anywhere else.

src/commands/deploy/getOrCreateFunctionApp.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    await wizard.prompt();
+
+    node = context.site ? await ext.rgApi.tree.findTreeItem(nonNullProp(context.site, 'id'), context) : undefined;
```
alexweininger, line 30: Find tree item is sort of sketchy to use. Have you tested this when the tree hasn't been loaded yet?

src/commands/deploy/FunctionAppListStep.ts
```diff
@@ -0,0 +1,58 @@
...
+        const sites = (await uiUtils.listAllIterator(client.webApps.list()));
+        const qp: IAzureQuickPickItem<Site | undefined>[] = sites.filter(s => !!s.kind?.includes('functionapp')).map(fa => {
+            return {
+                label: nonNullProp(fa, 'name'),
+                description: parseAzureResourceGroupId(nonNullProp(fa, 'id')).resourceGroup,
+                data: fa
+            }
+        });
+
+        qp.unshift({ label: '$(plus) Create new function app', description: '', data: undefined });
```
alexweininger, line 37: Why the empty description?

src/commands/deploy/FunctionAppListStep.ts
```diff
@@ -0,0 +1,58 @@
...
+        return qp;
+    }
+
+    public async getSubWizard(context: IFuncDeployContext): Promise<IWizardOptions<IFuncDeployContext> | undefined> {
+        if (!context.site) {
```
alexweininger, line 42: Maybe add a comment here. It seems like this the pivotal place where the code decides if we're creating a FA or not for deploy.

alexweininger:
I expected there to be a way to deploy from the local project. Was that ever discussed? 
image

src/agent/agentIntegration.ts
```diff
@@ -23,8 +23,8 @@ export async function getCommands(): Promise<(WizardCommandConfig | SimpleComman
         {
             type: "simple",
             name: createFunctionProjectCommandName,
-            commandId: "azureFunctions.createNewProject",
-            displayName: "Create Function Project",
+            commandId: "azureFunctions.createFunction",
+            displayName: "Create Function Locally",
```
MicroFish91, line 27: Do you remember if there was a reason we couldn't do something like `Create Function (Local)`?  I vaguely recall a reason but can't really remember anymore

src/commands/deploy/FunctionAppListStep.ts
MicroFish91: I would also like to suggest using a guard clause here to clean this up a bit more :)

src/agent/agentIntegration.ts
nturinski: I don't know if there was a reason or not, but personally if I see (Local), it makes me think that there is another version of Create Function. Probably Create Function (Remote) or something like that.

src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
```diff
@@ -0,0 +1,164 @@
...
+
+    if (version === undefined) {
+        version = latestGAVersion;
+        context.telemetry.properties.runtimeSource = 'Backup';
+    }
+
+    return version;
+}
+
+async function createFunctionAppWizard(wizardContext: IFunctionAppWizardContext): Promise<{ promptSteps: AzureWizardPromptStep<IAppServiceWizardContext>[], executeSteps: AzureWizardExecuteStep<IAppServiceWizardContext>[] }> {
```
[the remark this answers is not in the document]
nturinski: I don't disagree with you, but this is logic that was pre-existing to this PR. I'm open to changing it, but I'd rather do it in a separate PR since I wanted to keep the core logic fundamentally the same.

https://github.com/microsoft/vscode-azurefunctions/pull/4130/files#diff-3b26e2c6749863538fbe756b9eb0e37aaa43b559e370bd6daf7aaa84a871a158L233

src/commands/deploy/FunctionAppListStep.ts
nturinski: You see the description as half empty, but I see it as half full.

src/commands/deploy/FunctionAppListStep.ts
nturinski: Could you clarify what you mean by using a guard clause here? I'm not really sure what I would be type guarding in this case.

src/commands/deploy/getOrCreateFunctionApp.ts
nturinski: Yeah, it works though it's pretty unbearably slow. I may try to refactor this to just use the data model instead because of the performance.

nturinski: It's pretty tough without some significant refactoring. We need to pass the tree ndoe to deploy because it does the whole temporary description thing on it. If the node isn't there or if the parent is undefined, it'll throw an error.

src/commands/deploy/FunctionAppListStep.ts
nturinski: Hey guy, I already did that.

MicroFish91: 🫡

alexweininger (approved):

nturinski:
Whoo boy 🥵

PR #4165, opened 2024-06-03
Show quickpicks if there are multiple runtime matches for `targetFramework`

motm32:
Related to https://github.com/microsoft/vscode-azurestaticwebapps/issues/899

Currently the runtime is chosen using .find so the first match will be chosen. We should allow users to choose which runtime they want.


src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
```diff
@@ -16,7 +16,7 @@ export class DotnetRuntimeStep extends AzureWizardPromptStep<IProjectWizardConte
     public static async createStep(context: IProjectWizardContext): Promise<DotnetRuntimeStep> {
         if (context.targetFramework) {
             context.targetFramework = typeof context.targetFramework === 'string' ? [context.targetFramework] : context.targetFramework;
-            const runtimes = await getRuntimes(context);
+            const runtimes = (await getRuntimes(context)).reverse(); // reverse so that the latest version is first
```
alexweininger, line 19: Did you verify that getRuntimes always returns them in the same order?

nturinski:
I'm a little perplexed why it would have that behavior. Shouldn't it being an exact match on the .NET framework? So looking for .NET 8.0, and .NET 6.0 shouldn't match?

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
motm32: Mine always does. I will look into it some more to see how we are exactly getting it.

alexweininger: Yes please 👍

motm32:
> I think we should probably either a.) prompt the user if there are multiple matches or b.) add a property to the elements that indicates which is default

I think originally the SWA team wanted to prompt so I could see "a" being the best option. And I do think the list is always returned the same way but either way makes more sense to me to prompt than to automatically choose.

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
motm32: I think if we are going the prompt route  as mentioned below it won't matter either way.

alexweininger (commented):
We should also add tests that cover this scenario in this PR

test/api.test.ts
```diff
@@ -7,13 +7,13 @@ import { runWithInputs } from '@microsoft/vscode-azext-dev';
...
-import { ProjectLanguage, extensionId, nonNullValue, registerOnActionStartHandler } from '../extension.bundle';
+import { FuncVersion, ProjectLanguage, extensionId, nonNullValue, registerOnActionStartHandler } from '../extension.bundle';
 // eslint-disable-next-line no-restricted-imports
 import { type AzureFunctionsExtensionApi } from '../src/vscode-azurefunctions.api';
 import { getTestWorkspaceFolder, testFolderPath } from './global.test';
-import { getJavaScriptValidateOptions, validateProject, type IValidateProjectOptions } from './project/validateProject';
+import { getCSharpValidateOptions, getJavaScriptValidateOptions, validateProject, type IValidateProjectOptions } from './project/validateProject';
 
-suite(`AzureFunctionsExtensionApi`, () => {
+suite.only(`AzureFunctionsExtensionApi`, () => {
```
alexweininger, line 16: Gotta remove dis

test/api.test.ts
```diff
@@ -73,4 +73,26 @@ suite(`AzureFunctionsExtensionApi`, () => {
...
+        const folderPath: string = path.join(workspaceFolder, projectSubpath);
+
+        await runWithInputs('api.createFunction', [language, /6/i, 'Company.Function', 'Anonymous'], registerOnActionStartHandler, async () => {
+            await api.createFunction({
+                folderPath,
+                functionName,
+                templateId: 'HttpTrigger',
+                languageFilter: /^C\#$/i,
+                functionSettings: { authLevel: 'anonymous' },
+                targetFramework: ['net8.0', 'net7.0', 'net6.0']
```
alexweininger, line 91: add a test that verifies the filters actually works

test/api.test.ts
```diff
@@ -73,4 +73,46 @@ suite(`AzureFunctionsExtensionApi`, () => {
...
+            });
+        });
+
+        const validateOptions: IValidateProjectOptions = getCSharpValidateOptions('net6.0', FuncVersion.v4);
+        await validateProject(folderPath, validateOptions);
+    });
+
+    // When passing in a language version that is not in the target framework the last two inputs should not be prompted if they are this test will fail
```
alexweininger, line 99: This test is ensuring that target framework properly filters the picks from all available framework versions.

From this comment, it's not clear to me:
1. Why `api.createFunction` doesn't throw here, since selecting version 8 shouldn't be possible
2. What last two inputs do you mean, do you mean the next steps in the createFunction wizard?
3. How do you know this test will fail if 8 is a valid pick?

I think a better comment would start like:

// Intentionally pass a version (8) that hasn't been specified in `targetFramework` (6 & 7) to verify it isn't a possible pick. In the correct case (when 8 isn't a pick)... And in the incorrect case (when 8 is available)...

test/api.test.ts
```diff
@@ -73,4 +73,47 @@ suite(`AzureFunctionsExtensionApi`, () => {
...
+                targetFramework: ['net8.0', 'net7.0', 'net6.0']
+            });
+        });
+
+        const validateOptions: IValidateProjectOptions = getCSharpValidateOptions('net6.0', FuncVersion.v4);
+        await validateProject(folderPath, validateOptions);
+    });
+
+    // Intentionally pass a version (8) that hasn't been specified in targetFramework (6 & 7) to verify it isn't a possible pick. In the correct case (when 8 isn't a pick) this will cause an error to be thrown so no more picks are needed.
+    // In the incorrect case (when 8 is a pick) this will cause the test to fail as there are no further picks being passed in.
```
alexweininger, line 100:
```suggestion
    // Intentionally pass a version (8) that hasn't been specified in targetFramework (6 & 7) to verify it isn't a possible pick. In the correct case (when 8 isn't a pick) we throw an error. api.createFunction swallows the error and returns undefined.
    // In the incorrect case (when 8 is a pick) the test fails since the 2 provided test inputs have already been used, but there are more prompts.
```

alexweininger (dismissed):

alexweininger:
Uh make sure these tests are passing in the Actions before merging. Also is the plan that this should not change any existing Functions UX right? We might want to inform CTI of this change and have them watch out for it.

motm32:
> Make sure these tests are passing in the Actions before merging. Also is the plan that this should not change any existing Functions UX right? We might want to inform CTI of this change and have them watch out for it.

I am investigating right now since one of the tests I added is failing. And you are correct this should only change when a targetFramework exists which at the moment should only be in SWA. I will inform CTI of the change.

test/MockIProjectWizardContext.ts
```diff
@@ -0,0 +1,45 @@
+/* eslint-disable no-restricted-imports */
```
nturinski, line 1: Do you need this still?

nturinski (approved):

src/commands/createNewProject/dotnetSteps/DotnetRuntimeStep.ts
```diff
@@ -16,9 +16,17 @@ export class DotnetRuntimeStep extends AzureWizardPromptStep<IProjectWizardConte
     public static async createStep(context: IProjectWizardContext): Promise<DotnetRuntimeStep> {
         if (context.targetFramework) {
             context.targetFramework = typeof context.targetFramework === 'string' ? [context.targetFramework] : context.targetFramework;
-            const runtimes = await getRuntimes(context);
+            const runtimes = (await getRuntimes(context))
```
alexweininger, line 19:
```suggestion
            const runtimes = (await getRuntimes(context));
```

alexweininger (approved):

PR #4169, opened 2024-06-04
Fix verified templates/counts

nturinski:
Updated the backup templates to include the new verified trigger, `Azure Blob Storage Trigger (using Event Grid)`. Those were throwing off the count since I added them into the verified list without adding a test.

So then I added the `Azure Blob Storage Trigger (using Event Grid)` tests for all runtimes that include them. .NET 6 (LTS) doesn't have it and it's only for V4 runtime.  The reason for this is because it's replacing storage triggers for flex consumption and these are the runtimes that it supports.

I removed the staging template tests for verified triggers because the count between latest and staging haven't been staying consistent. I don't think that they are really updating staging and there isn't a lot of benefit for us to keep testing it as far as I'm concerned.


alexweininger (approved):
Nice. Can you explain what you ended up doing to fix the tests, and why?

nturinski:
> Nice. Can you explain what you ended up doing to fix the tests, and why?

Updated original PR message with clarifications.

PR #4195, opened 2024-06-19
Enable long running tests

hossam-nasr:
Modify the nightly tests to leverage the new resource groups extension release and use federated credentials for long-running tests

### DIFF
--- a/test/nightly/global.nightly.test.ts
+++ b/test/nightly/global.nightly.test.ts
@@ -14,7 +14,7 @@ import { type AzureSubscription } from '@microsoft/vscode-azureresources-api';
 import * as vscode from 'vscode';
 
 export let testClient: WebSiteManagementClient;
-export let context: TestActionContext;
+let testContext: TestActionContext;
 export let subscriptionContext: ISubscriptionContext;
 export const resourceGroupsToDelete: string[] = [];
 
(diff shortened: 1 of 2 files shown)


MicroFish91 (approved):

test/nightly/global.nightly.test.ts
```diff
@@ -5,24 +5,30 @@
...
 import { AzureAccountTreeItemWithProjects, createAzureClient, ext } from '../../extension.bundle';
 import { longRunningTestsEnabled } from '../global.test';
 
-export let testAccount: TestAzureAccount;
+import { createSubscriptionContext, subscriptionExperience, type ISubscriptionContext } from '@microsoft/vscode-azext-utils';
+import { type AzureSubscription } from '@microsoft/vscode-azureresources-api';
+import * as vscode from 'vscode';
+
 export let testClient: WebSiteManagementClient;
+let testContext: TestActionContext;
```
MicroFish91, line 17: Nit: I prefer removing this shared test context and going with how we initiated the two separately before with `createTestActionContext()`.  It's more in line with how we usually keep set up contexts for separate processes

MicroFish91:
You can probably use this for now, however, I noticed Nathan left this note with regards to `AzureAccountTreeItemWithProjects`, so we may need to revisit this flow again soon since it's currently being used to acquire the subscription.

image

PR #4200, opened 2024-06-28
Upgrade appservice SDK for Flex Consumption support

nturinski:
Swapped out my generated requests to use the SDK for flex consumption creation. Note that some things, such as runtimes and locations still use generic requests, but that's due to the fact those are not part of the appservice sdk.

Also fixed a perf bug that caused loading to take forever, especially if we had an error retrieving a function app. 

Note that this will fix the error that Ramya was encountering with loading her function apps though the ones that were failing will still continue to error. I verified that those apps don't show up at all in the portal, so those resources must be in a weird state.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
             "license": "SEE LICENSE IN LICENSE.md",
             "dependencies": {
                 "@azure/arm-appinsights": "^5.0.0-alpha.20230530.1",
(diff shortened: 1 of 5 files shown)


alexweininger:
What was this 
image

alexweininger (approved):

nturinski:
> What was this image

Oh, it was a gif showing the Functions loading faster and also Ramya's error being resolved.
functions_loading

Issue #4206, opened 2024-07-08
Use managed identity for hooking the Function app to the storage accounts


nturinski:
We default to using the storage account's connection string to hook the FA with the SA. We should follow this tutorial loosely: https://learn.microsoft.com/en-us/azure/azure-functions/functions-identity-based-connections-tutorial

PR #4213, opened 2024-07-12
Basic implementation for managed identity for create function app

nturinski:
Partially implements https://github.com/microsoft/vscode-azurefunctions/issues/4206

This is the bare bones implementation of having managed system identities integrated with function apps.

There's still quite a lot to do in terms of checking their settings for debugging/deploying.

This code is very simple because system identity is just a flag you turn on with the `identity` object and the backend completes the rest. The rest is using the role assignment API to give the function app the `Storage Blob Data Contributor` role which is the bare minimum requirement for deploying (since it leverages storage blob containers for that).

Lastly, we need to change the `AccountWebJobsStorage` to `AccountWebJobsStorage__accountName`. This is where a lot of our customer scenarios will break since we relied on that setting for a lot, so still plenty of work to do there.


src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -56,7 +57,16 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
                 context.telemetry.properties.fileLoggingError = parseError(error).message;
             }
         }
-
+        const principalId = nonNullProp(nonNullProp(context.site, 'identity'), 'principalId');
```
alexweininger, line 60: It's a bit cleaner to do 
```suggestion
        const principalId = nonNullValueAndProp(context.site.identity, 'principalId');
```

and I think it does the same thing.

package.json
```diff
@@ -1211,6 +1211,7 @@
     "dependencies": {
         "@azure/arm-appinsights": "^5.0.0-alpha.20230530.1",
         "@azure/arm-appservice": "^15.0.0",
+        "@azure/arm-authorization-profile-2020-09-01-hybrid": "^2.1.0",
```
alexweininger, line 1214: Curious: Were you forced to use `2020-09-01-hybrid` for some reason?

src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -142,8 +153,8 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStep<IFunctionAppWi
         const storageConnectionString: string = (await getStorageConnectionString(context)).connectionString;
         let appSettings: NameValuePair[] = [
             {
-                name: ConnectionKey.Storage,
-                value: storageConnectionString
+                name: `${ConnectionKey.Storage}__accountName`,
+                value: context.newStorageAccountName ?? context.storageAccount?.name
```
alexweininger, line 157: > This is where a lot of our customer scenarios will break since we relied on that setting for a lot, so still plenty of work to do there.

Can you outline which scenarios you expect to break? Have you verified them breaking?

alexweininger (approved):

package.json
MicroFish91: https://github.com/Azure/azure-sdk-for-js/tree/main/sdk/authorization/arm-authorization
https://learn.microsoft.com/en-us/javascript/api/overview/azure/arm-authorization-readme?view=azure-node-latest

The non-hybrid version appears to be more recent to me, but I'm not really sure why both of these packages exist separately / what the difference is

MicroFish91: https://learn.microsoft.com/en-us/javascript/api/overview/azure/arm-authorization-readme?view=azure-node-latest
Yeah, not really sure what the difference is

motm32 (approved):

PR #4236, opened 2024-08-19
Update PowerShell template

andystaples:
- Set ErrorActionPreference to Stop in profile.ps1 for new PowerShell apps


nturinski:
Hi @andystaples

Thanks for your contribution. Could you please explain why you are adding this to the template? It makes sense to me that people may want an exception to stop the execution, but since that is not the default behavior, does it make sense to automatically set this for all Powershell templates?

nturinski:
Closing due to inactivity.

Issue #4278, opened 2024-09-11
Function app create in Azure should not set APPINSIGHTS_INSTRUMENTATIONKEY


BigMorty:
When using the 'Create Function App in Azure' command, the app that is created has the `APPINSIGHTS_INSTRUMENTATIONKEY` set. [APPINSIGHTS_INSTRUMENTATIONKEY is/has been deprecated](https://learn.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings), instead the `APPLICATIONINSIGHTS_CONNECTION_STRING` setting should be used.

PR #4291, opened 2024-09-20
Use app insight connection string instead of aiKey

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4278


nturinski:
See the documentation here https://learn.microsoft.com/en-us/azure/azure-monitor/app/migrate-from-instrumentation-keys-to-connection-strings

alexweininger (approved):

PR #4299, opened 2024-10-09
Expose API command to start a func task and to check if CLI tools is installed

nturinski:
Exposing these commands to be used by external teams. Basically, they just need a way to start a func task but without using our VS Code launch/task configuration.

In order to launch the func host cli, it needs to be installed, so just exposing a way for them to see if it is or not.

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -38,7 +38,7 @@ export async function startFuncProcessFromApi(
         funcHostStartCmd += ` ${args.join(' ')}`;
     }
 
-    await callWithTelemetryAndErrorHandling('startFuncProcessFromApi', async (context: IActionContext) => {
+    await callWithTelemetryAndErrorHandling('azureFunctions.api.startFuncProcess', async (context: IActionContext) => {
         try {
             await waitForPrevFuncTaskToStop(workspaceFolder);
             const funcTask = new vscode.Task({ type: 'func' },
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { sendRequestWithTimeout, type AzExtRequestPrepareOptions } from '@microsoft/vscode-azext-azureutils';
-import { UserCancelledError, type IActionContext } from '@microsoft/vscode-azext-utils';
+import { callWithTelemetryAndErrorHandling, parseError, UserCancelledError, type IActionContext } from '@microsoft/vscode-azext-utils';
 import * as unixPsTree from 'ps-tree';
 import * as vscode from 'vscode';
 import { hostStartTaskName } from '../constants';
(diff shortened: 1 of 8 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -15,12 +15,50 @@ import { localize } from '../localize';
...
+    }
+
+    await callWithTelemetryAndErrorHandling('azureFunctions.api.startFuncProcess', async (context: IActionContext) => {
+        try {
+            await waitForPrevFuncTaskToStop(workspaceFolder);
+            const funcTask = new vscode.Task({ type: 'func' },
+                workspaceFolder,
+                hostStartTaskName, 'func',
+                new vscode.ShellExecution(funcHostStartCmd, {
+                    cwd: buildPath,
```
WardenGnaw, line 47: Can we add an additional parameter to support environment variables?

src/commands/pickFuncProcess.ts
```diff
@@ -15,12 +15,50 @@ import { localize } from '../localize';
...
+    };
+
+    let funcHostStartCmd: string = 'func host start';
+    if (args) {
+        funcHostStartCmd += ` ${args.join(' ')}`;
+    }
+
+    await callWithTelemetryAndErrorHandling('azureFunctions.api.startFuncProcess', async (context: IActionContext) => {
+        try {
+            await waitForPrevFuncTaskToStop(workspaceFolder);
```
WardenGnaw, line 42: Would it be ok to track FuncTasks by `buildPath` instead of workspaceFolder? This is to support multiple AzureFunction projects in an open folder. E.g. Solution with multiple Azure Functions projects.

WardenGnaw (commented):
I'll probably fork this commit and make another PR with my requested changes.

src/commands/pickFuncProcess.ts
WardenGnaw: This is addressed in https://github.com/microsoft/vscode-azurefunctions/pull/4309

src/commands/pickFuncProcess.ts
WardenGnaw: Addressed in https://github.com/microsoft/vscode-azurefunctions/pull/4309

alexweininger (approved):
Need to try and figure out how to get the build to pass. It doesn't seem related to these changes though.

captainsafia:
@alexweininger I think I have a hunch as to what might be going on here. It appears that your build is pulling pre-release bits of the Azure Functions Core Tools.

```
[20:53:58] Starting 'getFuncLink'...
Func cli feed version: 4.101.0
Func downloadLink: https://functionscdn.azureedge.net/public/4.0.123/Azure.Functions.Cli.linux-x64.4.0.6543.zip
[20:53:58] Finished 'getFuncLink' after 197 ms
[20:53:58] Starting 'installFuncCli'...
[20:54:08] 'installFuncCli' errored after 10 s
[20:54:08] Error: ENOTDIR: not a directory, scandir '/home/runner/tools/func/in-proc6/tools'
```

Perhaps we should pin the CI machine to the non-prerelease version? I dunno if you have controller one that based on the feed things are being pulled from.

alexweininger (approved):

PR #4309, opened 2024-10-23
Use Uri.fspath as key for FuncRunningMap/Port

WardenGnaw:
This commit modifies pickFuncProcess startFuncProcessFromApi removing the workspaceFolder requirement and adding in the environment variables.

funcHostTask was originally tracked by vscode.WorkspaceFolder via Task.Scope and DebugConfiguration.WorkspaceFolder. This commit modifies it to use a new AzureFunctionTaskDefinition.functionsApp property which is either the `buildPath` from startFuncProcessFromApi or the vscode.WorkspaceFolder.uri.fsPath to maintain backwards compatibility.

runningFuncTaskMap, funcTaskStartedEmitter, and runningFuncPortMap now all using the vscode.Uri.fsPath string which can be used to map back to a WorkspaceFolder. However in the cases the WorkspaceFolder fails, APIs that used it will now see undefined passed in.

### DIFF
--- a/src/commands/pickFuncProcess.ts
+++ b/src/commands/pickFuncProcess.ts
@@ -10,7 +10,7 @@ import * as vscode from 'vscode';
 import { hostStartTaskName } from '../constants';
 import { preDebugValidate, type IPreDebugValidateResult } from '../debug/validatePreDebug';
 import { ext } from '../extensionVariables';
-import { getFuncPortFromTaskOrProject, isFuncHostTask, runningFuncTaskMap, stopFuncTaskIfRunning, type IRunningFuncTask } from '../funcCoreTools/funcHostTask';
+import { AzureFunctionTaskDefinition, getFuncPortFromTaskOrProject, isFuncHostTask, runningFuncTaskMap, stopFuncTaskIfRunning, type IRunningFuncTask } from '../funcCoreTools/funcHostTask';
 import { localize } from '../localize';
 import { delay } from '../utils/delay';
 import { requestUtils } from '../utils/requestUtils';
@@ -18,19 +18,26 @@ import { taskUtils } from '../utils/taskUtils';
 import { getWindowsProcessTree, ProcessDataFlag, type IProcessInfo, type IWindowsProcessTree } from '../utils/windowsProcessTree';
 import { getWorkspaceSetting } from '../vsCodeConfig/settings';
 
-const funcTaskReadyEmitter = new vscode.EventEmitter<vscode.WorkspaceFolder>();
+const funcTaskReadyEmitter = new vscode.EventEmitter<string>();
 export const onDotnetFuncTaskReady = funcTaskReadyEmitter.event;
 
 export async function startFuncProcessFromApi(
-    workspaceFolder: vscode.WorkspaceFolder,
     buildPath: string,
-    args?: string[]
+    args: string[],
+    env: { [key: string]: string }
 ): Promise<{ processId: string; success: boolean; error: string }> {
     const result = {
         processId: '',
         success: false,
         error: ''
     };
+
+    const uriFile: vscode.Uri = vscode.Uri.file(buildPath)
+
+    const azFuncTaskDefinition: AzureFunctionTaskDefinition = {
+        type: 'func',
+        functionsApp: uriFile.fsPath
+    }
 
     let funcHostStartCmd: string = 'func host start';
     if (args) {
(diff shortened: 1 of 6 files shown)


src/commands/pickFuncProcess.ts
```diff
@@ -10,44 +10,54 @@ import * as vscode from 'vscode';
...
     }
 
     await callWithTelemetryAndErrorHandling('azureFunctions.api.startFuncProcess', async (context: IActionContext) => {
         try {
-            await waitForPrevFuncTaskToStop(workspaceFolder);
-            const funcTask = new vscode.Task({ type: 'func' },
-                workspaceFolder,
+            await waitForPrevFuncTaskToStop(azFuncTaskDefinition.functionsApp);
+            const funcTask = new vscode.Task(azFuncTaskDefinition,
+                vscode.TaskScope.Global,
```
captainsafia, line 53: What does `vscode.TaskScope.Global` help do here?

src/funcCoreTools/funcHostTask.ts
```diff
@@ -12,15 +12,36 @@ import { getLocalSettingsJson } from '../funcConfig/local.settings';
...
 
 export interface IRunningFuncTask {
+    taskExecution: vscode.TaskExecution;
     processId: number;
 }
 
-export const runningFuncTaskMap: Map<vscode.WorkspaceFolder | vscode.TaskScope, IRunningFuncTask> = new Map<vscode.WorkspaceFolder | vscode.TaskScope, IRunningFuncTask>();
+export class AzureFunctionTaskDefinition implements vscode.TaskDefinition {
+    type: string;
+    // This is either vscode.WorkspaceFolder.uri.fsPath or a vscode.Uri.file().fsPath
```
captainsafia, line 21: Nit: Maybe clarify in which cases each scenario each type would be used?

src/funcCoreTools/funcHostTask.ts
```diff
@@ -12,15 +12,36 @@ import { getLocalSettingsJson } from '../funcConfig/local.settings';
...
+    functionsApp: string
 
-const funcTaskStartedEmitter = new vscode.EventEmitter<vscode.WorkspaceFolder | vscode.TaskScope | undefined>();
+    static is(taskDefinition: vscode.TaskDefinition): taskDefinition is AzureFunctionTaskDefinition {
+        return taskDefinition.type.startsWith('func') && "functionsApp" in taskDefinition;
```
captainsafia, line 25: Nit: Use shared `consts` here?

src/commands/pickFuncProcess.ts
```diff
@@ -10,44 +10,54 @@ import * as vscode from 'vscode';
...
         error: ''
     };
 
+    const uriFile: vscode.Uri = vscode.Uri.file(buildPath)
+
+    const azFuncTaskDefinition: AzureFunctionTaskDefinition = {
+        // VS Code will only run a single instance of a task `type`,
+        // the path will be used here to make each project be unique.
+        type: `func ${uriFile.fsPath}`,
+        functionsApp: uriFile.fsPath
```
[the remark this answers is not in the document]
WardenGnaw: Yep! VS Code seems to limit 1 Task per `type` so if there were two `type: "func"` tasks requested. The second one will wait for the first one to finish. 

Using the output path will ensure uniqueness per project.

src/commands/pickFuncProcess.ts
WardenGnaw: Technically its unused, but its a required parameter for the VS Code API. I was hoping this would let it know that this task is used globally and not just for a specific workspace. 

https://github.com/microsoft/vscode/blob/c264ebed9425f3569ab5eb2c6fbc6a6bb4abaaaf/src/vscode-dts/vscode.d.ts#L8726C1-L8738C244

src/funcCoreTools/funcHostTask.ts
```diff
@@ -32,43 +53,58 @@ export function registerFuncHostTaskEvents(): void {
...
+            // Use `process.kill` because `TaskExecution.terminate` closes the terminal pane and erases all output
+            // Also to hopefully fix https://github.com/microsoft/vscode-azurefunctions/issues/1401
+            process.kill(runningFuncTask.processId);
+        }
+        runningFuncTaskMap.delete(functionApp);
     }
 }
 
-export async function getFuncPortFromTaskOrProject(context: IActionContext, funcTask: vscode.Task | undefined, projectPathOrTaskScope: string | vscode.WorkspaceFolder | vscode.TaskScope): Promise<string> {
+export async function getFuncPortFromTaskOrProject(context: IActionContext, funcTask: vscode.Task | undefined, projectPathOrTaskScope: string | vscode.WorkspaceFolder | vscode.TaskScope | undefined): Promise<string> {
```
captainsafia, line 107: What does it mean for this value to be undefined now?

alexweininger (approved):

src/funcCoreTools/funcHostTask.ts
WardenGnaw: There no `project` to search for the function's port and will rely on the task to get the running port.

alexweininger (approved):

Issue #4311, opened 2024-11-04
Fail to create a .NET 9.0 C# project if only .NET 9.0 SDK is installed


v-meilei:
**OS:** Win 11
**Build Version:** 20241104.1
**.NET Version:** 9.0.100-rc.2.24474.11

**Repro Steps:**
1. Create a .NET 9.0 project.
2. Check whether succeeds to create the project.

**Expect:**
Succeed to create a .NET 9.0 C# project.

**Actual:**
Fail to create a .NET 9.0 C# project.
![Image](https://github.com/user-attachments/assets/d52d545e-4b98-43a7-9550-af6a518341e7)
Output log:
```
7:36:07 AM: Failed to get latest templates: Command "dotnet "c:\Users\v-meilei\.vscode-insiders\extensions\ms-azuretools.vscode-azurefunctions-1.15.5-alpha.0\resources\dotnetJsonCli\net9.0\Microsoft.TemplateEngine.JsonCli.dll" --templateDir "c:\Users\v-meilei\AppData\Roaming\Code - Insiders\User\globalStorage\ms-azuretools.vscode-azurefunctions\~4\net9.0-isolated" --operation list" failed with exit code "1":
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-c:\Users\v-meilei\.vscode-insiders\extensions\ms-azuretools.vscode-azurefunctions-1.15.5-alpha.0\resources\dotnetJsonCli\net9.0\Microsoft.TemplateEngine.JsonCli.dll does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.

7:36:07 AM: Failed to get backup templates: Command "dotnet "c:\Users\v-meilei\.vscode-insiders\extensions\ms-azuretools.vscode-azurefunctions-1.15.5-alpha.0\resources\dotnetJsonCli\net9.0\Microsoft.TemplateEngine.JsonCli.dll" --templateDir "c:\Users\v-meilei\AppData\Roaming\Code - Insiders\User\globalStorage\ms-azuretools.vscode-azurefunctions\~4\net9.0-isolated" --operation list" failed with exit code "1":
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
[...]

PR #4314, opened 2024-11-05
Use `--roll-forward Major` flag to make `.dll` tool runtime agnostic

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4311

Nathan Carlson clued us in on a flag that you can pass to make the `dotnet` tool major version agnostic. So I've used the .NET 6 .dll tools (which should be forward compatible with 7, 8, and 9) and got rid of all the other runtime specific .dlls.

This is nice because we theoretically shouldn't have to update the JsonCliTool ever again.


alexweininger (approved):
Eric would be so proud 😄

MicroFish91 (approved):

PR #4320, opened 2024-11-08
Create a RunningAzureFunctions class to handle multiple function tasks in one workspace

nturinski:
Previously, there was a limitation of only being able to run one task per workspace. This would make it impossible to run and manage multiple tasks in subfolders which was blocking a scenario for the C# Dev Kit team.

These changes make the workspace folder a key to the map, but we are storing it in an array. The array contains the cwd of the subpath for the project. This will allow us to run multiple projects in different folders simultaneously which should hopefully get us one step closer to monorepo support.


src/funcCoreTools/funcHostTask.ts
```diff
@@ -79,34 +68,57 @@ export function registerFuncHostTaskEvents(): void {
         // Used to stop the task started with pickFuncProcess.ts startFuncProcessFromApi.
         if (DotnetDebugDebugConfiguration.is(debugSession.configuration) && debugSession.configuration.launchServiceData.buildPath) {
             const buildPathUri: vscode.Uri = vscode.Uri.file(debugSession.configuration.launchServiceData.buildPath)
-            stopFuncTaskIfRunning(buildPathUri.fsPath, /* terminate */ true)
+
+            const workspaceFolder = buildPathToWorkspaceFolderMap.get(debugSession.configuration.launchServiceData.buildPath)
+            if (workspaceFolder === undefined) {
+                throw Error()
```
WardenGnaw, line 74: nit: I wasn't sure if throwing here is correct.

Maybe a `vscode.window.showWarningMessage('Unable to find workspaceFolder for debugConfiguration.buildPath");` + however its localized

src/commands/pickFuncProcess.ts
```diff
@@ -32,32 +33,37 @@ export async function startFuncProcessFromApi(
...
+                }
+            }
+
+            await waitForPrevFuncTaskToStop(workspaceFolder);
+
+            buildPathToWorkspaceFolderMap.set(buildPath, workspaceFolder);
+
+            const funcTask = new vscode.Task({ type: `func  ${buildPath}` },
+                workspaceFolder,
+                hostStartTaskName, `func`,
```
WardenGnaw, line 59: nit: place on next line since its another argument

src/commands/startFunctionApp.ts
```diff
@@ -5,16 +5,20 @@
...
+import { startFuncProcessFromApi } from './pickFuncProcess';
 
 export async function startFunctionApp(context: IActionContext, node?: SlotTreeItem): Promise<void> {
     if (!node) {
         node = await pickFunctionApp(context);
     }
 
+
     const client: SiteClient = await node.site.createClient(context);
+    await startFuncProcessFromApi(`${workspace.workspaceFolders![0].uri.path}/bin/Debug/net8.0`, [], {});
```
WardenGnaw, line 21: remove?

src/commands/pickFuncProcess.ts
```diff
@@ -32,32 +33,37 @@ export async function startFuncProcessFromApi(
...
-            const funcTask = new vscode.Task(azFuncTaskDefinition,
-                vscode.TaskScope.Global,
-                hostStartTaskName, 'func',
+            let workspaceFolder: vscode.WorkspaceFolder | undefined = buildPathToWorkspaceFolderMap.get(buildPath);
+
+            if (workspaceFolder === undefined) {
+                workspaceFolder = {
+                    uri: vscode.Uri.parse(buildPath),
+                    name: buildPath,
+                    index: -1
```
WardenGnaw, line 49: I'm also not sure if the consequences of using index: -1 if VS Code actually looks at this 😓

src/commands/startFunctionApp.ts
nturinski: Yep, will do. Was from testing 😅

MicroFish91 (approved):
Approving for now so as not to block.  Ping if you need review again later

MicroFish91 (approved):

PR #4382, opened 2025-02-21
Add Identity node to Function Apps

nturinski:
With system and user assigned enabled and expanded
{D0653932-D2AA-436D-8EDD-526EFD831934}

With system assigned disabled and no user assigned added
{B5CAEED4-C166-449F-AB2F-1F96C5CD6453}


nturinski:
I can definitely (and probably will) make these into smaller PRs. It was kind of difficult since Hossam had bits and pieces of implementation for most of it, but now that everything is done, it'll be easier to isolate.

Let me know if that will make review easier (or if this is okay as is)

src/commands/identity/enableSystemIdentity.ts
```diff
@@ -0,0 +1,34 @@
...
+import { localize } from "../../localize";
+import { type SystemIdentityTreeItemBase } from "../../tree/remoteProject/SystemIdentityTreeItemBase";
+import { type SlotTreeItem } from "../../tree/SlotTreeItem";
+import { createActivityContext } from "../../utils/activityUtils";
+import { EnableSystemIdentityAssignStep } from "./EnableSystemIdentityStep";
+import { type ManagedIdentityAssignContext } from "./ManagedIdentityAssignContext";
+
+export async function enableSystemIdentity(context: IActionContext, node: SystemIdentityTreeItemBase): Promise<undefined> {
+    const grandparentNode = node.parent.parent as SlotTreeItem;
+    const title: string = localize('enabling', 'Enabling system assigned identity for "{0}"...', grandparentNode.site.fullName);
```
MicroFish91, line 16: Should this say "Enable" instead of "Enabling" as well as no ellipses at the end?
image

src/commands/identity/enableSystemIdentity.ts
```diff
@@ -0,0 +1,34 @@
...
+import { AzureWizard, type ExecuteActivityContext, type IActionContext } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../localize";
+import { type SystemIdentityTreeItemBase } from "../../tree/remoteProject/SystemIdentityTreeItemBase";
+import { type SlotTreeItem } from "../../tree/SlotTreeItem";
+import { createActivityContext } from "../../utils/activityUtils";
+import { EnableSystemIdentityAssignStep } from "./EnableSystemIdentityStep";
+import { type ManagedIdentityAssignContext } from "./ManagedIdentityAssignContext";
+
+export async function enableSystemIdentity(context: IActionContext, node: SystemIdentityTreeItemBase): Promise<undefined> {
+    const grandparentNode = node.parent.parent as SlotTreeItem;
```
MicroFish91, line 15: With my other suggestion, I think you might be able to avoid having to typecast this value

src/tree/remoteProject/ManagedIdentityTreeItem.ts
```diff
@@ -0,0 +1,57 @@
...
+import { localize } from '../../localize';
+import { type IProjectTreeItem } from '../IProjectTreeItem';
+import { type SlotTreeItem } from '../SlotTreeItem';
+import { SystemIdentityTreeItemBase } from './SystemIdentityTreeItemBase';
+import { UserAssignedIdentitiesTreeItem } from './UserAssignedIdentitiesTreeItem';
+
+export class ManagedIdentityTreeItem extends AzExtParentTreeItem {
+    public readonly label: string = localize('Identity', 'Identity');
+    public static contextValue: string = 'azFuncManagedIdentity';
+    public readonly parent: AzExtParentTreeItem & IProjectTreeItem & SlotTreeItem;
```
MicroFish91, line 18: I think you can make parent just be `SlotTreeItem` because the base class already extends `AzExtParentTreeItem` and implements `IProjectTreeItem`

MicroFish91 (commented):
Tested `enableSystemIdentity`, going to try out the user assigned stuff tomorrow

src/commands/identity/assignManagedIdentity.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    const wizardContext: ManagedIdentityAssignContext = {
+        ...context,
+        site: node.site,
+        ...node.subscription,
+        ...(await createActivityContext())
+    }
+
+    const promptSteps: AzureWizardPromptStep<ManagedIdentityAssignContext>[] = [
+        new UserAssignedIdentityListStep()
```
MicroFish91, line 32: For the `UserAssignedIdentityListStep`, if I choose to create a new user assigned identity, I'm getting this error:
image

nturinski: Should be fixed by this when it's merged/updated: https://github.com/microsoft/vscode-azuretools/pull/1925

src/commands/identity/assignManagedIdentity.ts
```diff
@@ -0,0 +1,57 @@
...
+        promptSteps,
+        executeSteps,
+        showLoadingPrompt: true
+    });
+
+
+    await wizard.prompt();
+    await node.runWithTemporaryDescription(context, localize('enabling', 'Assigning identity...'), async () => {
+        await wizard.execute();
+        wizardContext.activityTitle = localize('assigning', 'Assign user assigned identity "{1}" for "{0}"', wizardContext.site?.fullName, wizardContext.managedIdentity?.name);
```
MicroFish91, line 52: This activity title only shows up when it's placed before the `wizard.execute()`, however, then we run into a different problem of the managed identity not existing yet (making name undefined).  

It's tricky because a related name gets generated at the time of the execute step... maybe just move the activity title up and remove the user-assigned identity name unless it's already known (i.e. an existing one was selected)? Up to you on whatever you think makes the most sense here.

src/tree/remoteProject/SystemIdentityTreeItemBase.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    public get description(): string | undefined {
+        return localize('enabled', 'Enabled');
+    }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean, context: IActionContext): Promise<AzExtTreeItem[]> {
+        const msiClient = await createManagedServiceIdentityClient([context, this.subscription]);
+        const systemIdentity = await msiClient.systemAssignedIdentities.getByScope(this.parent.parent.id);
+        const roleDefinitionsItem: RoleDefinitionsItem[] = await createRoleDefinitionsItems(context, this.parent.subscription, systemIdentity)
+        return roleDefinitionsItem.map(rd => new RoleDefinitionsTreeItem(this, rd));
```
MicroFish91, line 40: Kind of a nit, but when User-Assigned Identity Tree Items and System-Assigned Identity Tree Items are added / enabled but have no role assignments, the tree view shows an expandable node which has nothing underneath it when expanded:

image

Users could be confused as to what this means.  Perhaps if an empty list is returned, we could default to a generic tree item stating no role assignments assigned yet?  Idk, thoughts?

nturinski: I'm not against the idea, but would definitely prefer this to be a part 2 iteration on the feature. It's definitely more of a nice to have feature.

Also for context, you can open the managed identity nodes in the portal where you can add roles. I may be able to open them directly to the add roles page (and do the same for system identity as well).

MicroFish91 (dismissed):

src/tree/remoteProject/SystemIdentityTreeItemBase.ts
alexweininger, line 1: We normally split all classes into their own files. I think it's fine here to keep them together, just add newlines between the definitions.

src/commands/identity/EnableSystemIdentityStep.ts
```diff
@@ -0,0 +1,47 @@
...
+        await client.webApps.update(site.resourceGroup, site.siteName, site.rawSite);
+        ext.outputChannel.appendLog(enabled);
+    }
+
+    public shouldExecute(_context: ManagedIdentityAssignContext): boolean {
+        return true;
+    }
+}
+
+const addSystemAssignedType = (identity: Identity): ManagedServiceIdentityType => {
```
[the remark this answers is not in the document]
nturinski: Well, it kind of does.

Identity.type is an enum that looks like this:
`export declare type ResourceIdentityType = "SystemAssigned" | "UserAssigned" | "SystemAssigned, UserAssigned" | "None";`

So it takes the app's identity and adds SystemAssigned to the type if necessary.

src/commands/identity/ManagedIdentityAssignStep.ts
```diff
@@ -0,0 +1,61 @@
...
+        ext.outputChannel.appendLog(assigned);
+    }
+
+    public shouldExecute(_context: ManagedIdentityAssignContext): boolean {
+        return true;
+    }
+
+}
+
+const addUserAssignedType = (type: ManagedServiceIdentityType | undefined): ManagedServiceIdentityType => {
```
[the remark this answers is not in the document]
nturinski: Same explanation as above.

src/tree/remoteProject/ManagedIdentityTreeItem.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    public hasMoreChildrenImpl(): boolean {
+        return false;
+    }
+
+    public async loadMoreChildrenImpl(_clearCache: boolean): Promise<AzExtTreeItem[]> {
+        const systemIdentitiesTreeItem: AzExtTreeItem = createSystemIdentityTreeItem(this);
+        const userAssignedIdentitiesTreeItem = new UserAssignedIdentitiesTreeItem(this.parent);
+        const children: AzExtTreeItem[] = [systemIdentitiesTreeItem, userAssignedIdentitiesTreeItem];
+        return children;
```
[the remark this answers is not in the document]
nturinski: Yeah, I'll leave it at that.  Doing
return [createSystemIdentityTreeItem(), new UserAssignedIdentititesTreeItem] looks pretty bad, but also if we ever make the create async (which it was at one point) we'll have to revert that anyway.

src/tree/remoteProject/SystemIdentityTreeItemBase.ts
nturinski: Yeah we have a couple of examples where we would consolidate them. The tree items used to inherent from the same base, but I didn't want that child node on DisabledIdentityTreeItem anymore because I thought it looked kind of bad.

alexweininger (approved):

Issue #4376, opened 2025-02-12
There are no visibility icons for local settings


Amar-QA:
**OS:** All
**Build Version:** 20250211.1
**Regression:** Not a regression

**Repro Steps:**

1. Create a new function project.
2. Expand "Local Project" node under Workspace view -> Expand "Local Settings" node.
3. Check whether there are visibility icons for local settings.

**Expect:**
There are visibility icons for local settings.
![Image](https://github.com/user-attachments/assets/fae08867-8f62-4dd0-85e7-3b46d9e781f7)

**Actual:** 
There are no visibility icons for local settings.
![Image](https://github.com/user-attachments/assets/64b89232-061f-424c-a69f-4c506abcbf6e)

PR #4387, opened 2025-02-28
Add Local and Remote Settings conversion

motm32:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4376

For both local and remote there are multiple entry points: 
**_Local_**
On the top level Local Project node in the workspace view and on each connection under the local settings node: 
image
I have also added a warning node on the specific connections that need to be converted.

**_Remote_**
On the top level App Setting Item and on the specific connection: 
image

After converting connections and assigning roles to the chosen managed identity the activity log will look like this: 
image

Notes:

- We are not deleting local or remote settings as this can break users apps as changes to the trigger code needs to be made in order for it to work. We currently are just adding the new settings and setting up the roles for the remote
- We cannot add roles for Cosmos DB triggers as they have their own RBAC roles see [here](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/security/how-to-grant-data-plane-role-based-access?tabs=built-in-definition%2Ctypescript&pivots=azure-interface-cli#permission-model) for more information
- When locally debugging role assignments aren't necessary as your local machine cannot access the managed identity. 

Todo:
- [x] Release appsettings and azure utils package and update the package json
- [ ] Test each connection type 
- [x] When right clicking on a remote app setting the command should only show if there are connections to be converted


Cxz32:
@ #
| Header | Header | Header | Header | Header |
|--------|--------|--------|--------|--------|
| Cell | Cell | Cell | Cell | Cell |
| Cell | Cell | Cell | Cell | Cell |
| Cell | Cell | Cell | Cell | Cell |
| Cell | Cell | Cell | Cell | Cell |
| Cell | Cell | Cell | Cell | Cell |

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,210 @@
...
+                        if (remoteSettings.properties) {
+                            remoteSettings.properties[connection.name] = connection.value;
+                            delete remoteSettings.properties[nonNullProp(connection, 'originalValue')];
+                        } else {
+                            await client.updateApplicationSettings({ properties: { [connection.name]: connection.value } });
+                        }
+                        context.activityChildren?.push(
+                            new GenericTreeItem(undefined, {
+                                contextValue: createUniversallyUniqueContextValue(['useExistingResourceGroupInfoItem', activitySuccessContext]),
+                                label: localize('deletedSetting', `Deleted app setting "${connection.originalValue}"`),
```
MicroFish91, line 178: Should all the activity child labels be present tense instead of past tense?

Left a similar comment to Nathan as well:
https://github.com/microsoft/vscode-azurefunctions/pull/4382#discussion_r1978361356

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+import { CommonRoleDefinitions, createRoleId, uiUtils } from "@microsoft/vscode-azext-azureutils";
+import { activitySuccessContext, activitySuccessIcon, AzExtFsExtra, AzureWizardExecuteStep, createUniversallyUniqueContextValue, GenericTreeItem, nonNullProp, nonNullValueAndProp } from "@microsoft/vscode-azext-utils";
+import { ext } from "../../extensionVariables";
+import { type ILocalSettingsJson } from "../../funcConfig/local.settings";
+import { localize } from "../../localize";
+import { getLocalSettingsFile } from "../appSettings/localSettings/getLocalSettingsFile";
+import { type IConvertConnectionsContext } from "./IConvertConnectionsContext";
+import { type Connection } from "./SelectConnectionsStep";
+
+export class ConvertSettingsStep extends AzureWizardExecuteStep<IConvertConnectionsContext> {
```
nturinski, line 16: nit: I prefer having the step names following a \<noun\>\<verb\>Step convention. The reason being, if you have multiple steps that are for Settings, they would all be listed together in the folder, like SettingsConvertStep, SettingsListStep, etc.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+    public priority: number = 100;
+
+    public async execute(context: IConvertConnectionsContext): Promise<void> {
+        if (context.connections) {
```
nturinski, line 20: I'd prefer to take this out of the if statement and just do `nonNullProp(context, connections)`.  

The reason being that this should only execute if `!!context.connections` evaluates to true, and if there is some reason that this is running when it is not, I'd rather the step throw an internal error than just silently skipping it, making it a harder to uncover error.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+                            value: `https://${storageAccountName}.queue.core.windows.net`,
+                        },
+                        {
+                            name: 'AzureWebJobsStorage__tableServiceUri',
+                            value: `https://${storageAccountName}.table.core.windows.net`,
+                        },
+                        ...getClientIdAndCredentialProperties(context, storageAccountName)
+                    );
+                    context.roles?.push({
+                        scopeId: await getScopeHelper(context, storageAccountName, `resourceType eq 'Microsoft.Storage/storageAccounts'`),
```
nturinski, line 42: Is the filter supposed to be resourceType eq on all of these? Can you explain this a little bit?

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+                        roleDefinitionId: createRoleId(context.subscriptionId, CommonRoleDefinitions.storageBlobDataOwner),
+                        roleDefinitionName: CommonRoleDefinitions.storageBlobDataOwner.roleName
+                    });
+                } else if (connection.name.includes('STORAGE')) {
```
nturinski, line 46: I think it's safe to combine this and `connection.name.includes('AzureWebJobsStorage')` as one conditional?

`if (connection.name.includes('AzureWebJobsStorage') || connection.name.includes('STORAGE')` to save on a lot of space/copied logic.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+import { type ILocalSettingsJson } from "../../funcConfig/local.settings";
+import { localize } from "../../localize";
+import { getLocalSettingsFile } from "../appSettings/localSettings/getLocalSettingsFile";
+import { type IConvertConnectionsContext } from "./IConvertConnectionsContext";
+import { type Connection } from "./SelectConnectionsStep";
+
+export class ConvertSettingsStep extends AzureWizardExecuteStep<IConvertConnectionsContext> {
+    public priority: number = 100;
+
+    public async execute(context: IConvertConnectionsContext): Promise<void> {
```
nturinski, line 19: For readability, I'd like to refactor out a lot of the functionality into their own functions outside of execute. I think you could create helper functions to pull out common code, like creating the different settings that get pushed into the connections array, etc.

Also, a switch case might help readability over if/else if statements, but I think just pulling out a lot of code could help tremendously in terms of readability.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+        if (context.connections) {
+            context.roles = [];
+            context.convertedConnections = [];
+            for (const connection of context.connections) {
+                if (connection.name.includes('AzureWebJobsStorage')) {
+                    const storageAccountName = connection.value.split(';')[1].split('=')[1];
```
nturinski, line 25: I think you should include 2 things:
1. A comment that shows an example connection string so people can see why you're doing this madness
2. Either a try/catch or something that verifies that the connection string is correct or that the storageAccountName is actually there. Just some error handling that alerts the user that the connection string was in a form that we didn't expected would be a better error than `Can't read .split of undefined` or whatever that error would look like,

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+
+            if (context.local) {
+                const localSettings = await AzExtFsExtra.readJSON<ILocalSettingsJson>(nonNullProp(context, 'localSettingsPath'));
+                if (localSettings.Values) {
+                    for (const connection of context.convertedConnections) {
+                        localSettings.Values[connection.name] = connection.value;
+                        context.activityChildren?.push(
+                            new GenericTreeItem(undefined, {
+                                contextValue: createUniversallyUniqueContextValue(['useExistingResourceGroupInfoItem', activitySuccessContext]),
+                                label: localize('addedSetting', `Add Local setting "${connection.name}"`),
```
nturinski, line 167: This isn't localized correctly.

`localize('addedSetting', 'Add Local setting "${0}"', connection.name);`

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+                    for (const connection of context.convertedConnections) {
+                        if (remoteSettings.properties) {
+                            remoteSettings.properties[connection.name] = connection.value;
+                        } else {
+                            await client.updateApplicationSettings({ properties: { [connection.name]: connection.value } });
+                        }
+                        context.activityChildren?.push(
+                            new GenericTreeItem(undefined, {
+                                contextValue: createUniversallyUniqueContextValue(['useExistingResourceGroupInfoItem', activitySuccessContext]),
+                                label: localize('addedSetting', `Add app setting "${connection.name}" `),
```
nturinski, line 188: Same as above.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+                    );
+                }
+            }
+
+            if (!context.localSettingsPath) {
+                const message: string = localize('selectLocalSettings', 'Select the local settings to convert.');
+                context.localSettingsPath = await getLocalSettingsFile(context, message);
+            }
+
+            if (context.local) {
```
nturinski, line 159: You know when it is local versus when it remote before you get to the execute step, so I'd recommend just creating two different steps off of a base. Managing these two behaviors in one giant step is harder for us to follow.

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+    const client = await createResourceClient(context);
+    const resources = await uiUtils.listAllIterator(client.resources.list({ filter }));
+
+    for await (const resource of resources) {
+        if (resource.name === accountName) {
+            return nonNullProp(resource, 'id');
+        }
+    }
+
+    throw new Error(localize('noResourceFound', `No resource found with name "${accountName}" in subscription "${context.subscriptionDisplayName}"`));
```
nturinski, line 216: nit: localized incorrectly

src/commands/convertConnections/ConvertSettingsStep.ts
```diff
@@ -0,0 +1,236 @@
...
+}
+
+function getClientIdAndCredentialProperties(context: IConvertConnectionsContext, connectionName: string): Connection[] {
+    const clientIdAndConfigurationProperties: Connection[] = [];
+    if (!context.local) {
```
nturinski, line 221: This is something that could go into a `RemoteSettingsConvertStep` if you factor it out.

src/commands/convertConnections/IConvertConnectionsContext.ts
```diff
@@ -0,0 +1,18 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type IResourceGroupWizardContext, type Role } from "@microsoft/vscode-azext-azureutils";
+import { type ExecuteActivityContext, type ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { type SlotTreeItem } from "../../tree/SlotTreeItem";
+import { type Connection } from "./SelectConnectionsStep";
+
+export interface IConvertConnectionsContext extends ExecuteActivityContext, IResourceGroupWizardContext, ISubscriptionActionContext {
+    local: boolean;
```
nturinski, line 12: Small nit: Is there a point of having a local and function app property? I feel like it is just more things to keep track of-- if there is no functionapp on the context, can we assume that it is local?

src/commands/convertConnections/SelectConnectionsStep.ts
```diff
@@ -0,0 +1,114 @@
...
+
+export interface Connection {
+    name: string;
+    value: string;
+}
+
+export class SelectConnectionsStep extends AzureWizardPromptStep<IConvertConnectionsContext> {
+    public async prompt(context: IConvertConnectionsContext): Promise<void> {
+        let picks: IAzureQuickPickItem<Connection>[];
+        if (context.local) {
```
nturinski, line 25: I'm okay with this being one step since the differences between the two are so minimal. 

For readability, I think it'd be good to have a `this.getPicks` step, and in there, you do the if logic to return this.getLocalQuickPicks or this.getRemoteQuickPicks instead of in the prompt function itself.

src/commands/convertConnections/SelectConnectionsStep.ts
```diff
@@ -0,0 +1,114 @@
...
+                canPickMany: true,
+            })).map(item => item.data);
+        }
+    }
+
+    public shouldPrompt(context: IConvertConnectionsContext): boolean {
+        return !context.connections || context.connections.length === 0;
+    }
+
+    private async getLocalQuickPics(context: IConvertConnectionsContext, workspaceFolder?: vscode.WorkspaceFolder): Promise<IAzureQuickPickItem<Connection>[]> {
```
nturinski, line 51: nit: 
```suggestion
    private async getLocalQuickPicks(context: IConvertConnectionsContext, workspaceFolder?: vscode.WorkspaceFolder): Promise<IAzureQuickPickItem<Connection>[]> {
```

src/commands/convertConnections/SelectConnectionsStep.ts
```diff
@@ -0,0 +1,114 @@
...
+                        }
+                    });
+                }
+            }
+        }
+
+        return picks;
+    }
+
+    private async getRemoteQuickPics(context: IConvertConnectionsContext): Promise<IAzureQuickPickItem<Connection>[]> {
```
nturinski, line 89: nit:
```suggestion
    private async getRemoteQuickPicks(context: IConvertConnectionsContext): Promise<IAzureQuickPickItem<Connection>[]> {
```

src/commands/convertConnections/SelectConnectionsStep.ts
```diff
@@ -0,0 +1,114 @@
...
+        const picks: IAzureQuickPickItem<Connection>[] = [];
+
+        if (context.functionapp) {
```
nturinski, line 92: Similar thing to what I had said in another step, but I'd prefer you use nonNullValue here because otherwise, it'll just return an empty array and we won't know if that's because there is no function app or if it's because the function app isn't listing app settings correctly.

src/commands/convertConnections/convertLocalConnections.ts
```diff
@@ -0,0 +1,65 @@
...
+import { localize } from "../../localize";
+import { createActivityContext } from "../../utils/activityUtils";
+import { ConvertSettingsStep } from "./ConvertSettingsStep";
+import { type IConvertConnectionsContext } from "./IConvertConnectionsContext";
+import { SelectConnectionsStep, type Connection } from "./SelectConnectionsStep";
+
+export async function convertLocalConnections(context: IActionContext, node?: AppSettingTreeItem): Promise<void> {
+    const connections: Connection[] = [];
+    if (node instanceof AppSettingTreeItem) {
+        connections.push({ name: node.id, value: node.value })
```
nturinski, line 20:
```suggestion
        connections.push({ name: node.id, value: node.value });
```

src/commands/convertConnections/convertLocalConnections.ts
```diff
@@ -0,0 +1,65 @@
...
+        promptSteps,
+        executeSteps
+    });
+
+    await wizard.prompt();
+    await wizard.execute();
+
+    const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+    const message: string = localize('setConnectionsProperty', 'Successfully converted local connections in order to use identity based connections you may need to set the connection property within your trigger.');
+    await context.ui.showWarningMessage(message, { learnMoreLink: "https://aka.ms/AAuroke", modal: true }, continueOn);
```
nturinski, line 64: I believe that we should show this message _before_ they start converting. We can show it again after in the activity log, but if they're worried about messing up their project, it's a little bit late by this point.

src/commands/convertConnections/convertRemoteConnections.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (node) {
+        await node.runWithTemporaryDescription(context, localize('converting', 'Converting...'), async () => {
+            await convertRemoteConnectionsInternal(context, connections, node);
+        });
+    } else {
+        await convertRemoteConnectionsInternal(context);
+    }
+}
+
+export async function convertRemoteConnectionsInternal(context: IActionContext, connections?: Connection[], node?: AppSettingTreeItem): Promise<void> {
```
nturinski, line 29: Are these internal functions identical besides setting local to false?

nturinski (commented):
🥲

src/commands/convertConnections/ConvertSettingsStep.ts
motm32: Yeah I thought about doing this. Originally I had AzureWebJobsStorage set to the __accountName property and only recently found out we shouldn't be using that. I'll combine the two.

src/commands/convertConnections/convertRemoteConnections.ts
motm32: There are some slight differences beyond that but once I split up the ConvertSettingsStep into remote and local they will also use different execute steps.

src/commands/addMIConnections/SettingsAddBaseStep.ts
```diff
@@ -0,0 +1,175 @@
...
+                name: webJobsStorage ? 'AzureWebJobsStorage__queueServiceUri' : `${storageAccountName}__queueServiceUri`,
+                value: `https://${storageAccountName}.queue.core.windows.net`,
+            },
+            {
+                name: webJobsStorage ? 'AzureWebJobsStorage__tableServiceUri' : `${storageAccountName}__tableServiceUri`,
+                value: `https://${storageAccountName}.table.core.windows.net`,
+            },
+            ...getClientIdAndCredentialProperties(context, webJobsStorage ? 'AzureWebJobsStorage' : storageAccountName)
+        );
+        const scope = await getScopeHelper(context, storageAccountName, 'Microsoft.Storage/storageAccounts')
```
nturinski, line 86: Put `if (context.functionapp)` here instead of in `addRole`.

src/commands/addMIConnections/SettingsAddBaseStep.ts
```diff
@@ -0,0 +1,175 @@
...
+            },
+            {
+                name: webJobsStorage ? 'AzureWebJobsStorage__queueServiceUri' : `${storageAccountName}__queueServiceUri`,
+                value: `https://${storageAccountName}.queue.core.windows.net`,
+            },
+            {
+                name: webJobsStorage ? 'AzureWebJobsStorage__tableServiceUri' : `${storageAccountName}__tableServiceUri`,
+                value: `https://${storageAccountName}.table.core.windows.net`,
+            },
+            ...getClientIdAndCredentialProperties(context, webJobsStorage ? 'AzureWebJobsStorage' : storageAccountName)
```
nturinski, line 84: I think you can include this in the `if (context.functionApp)` too. I believe it's only called in a remote situation as well.

src/commands/addMIConnections/addLocalMIConnections.ts
```diff
@@ -0,0 +1,58 @@
...
+        title,
+        promptSteps,
+        executeSteps
+    });
+
+    await wizard.prompt();
+    await wizard.execute();
+
+    const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+    const message: string = localize('setConnectionsProperty', 'Successfully added local connections in order to use identity based connections you may need to set the connection property within your trigger.'); //Todo: edit this message
```
nturinski, line 56: I see that you still have a todo comment here, but very simple way to upgrade it to an actual English sentence.

```suggestion
    const message: string = localize('setConnectionsProperty', 'Successfully added local connections. In order to use identity based connections, you may need to set the connection properties within your trigger.'); //Todo: edit this message
```

src/commands/addMIConnections/addLocalMIConnections.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export async function addLocalMIConnections(context: IActionContext, node?: AppSettingTreeItem): Promise<void> {
+    const connections: Connection[] = [];
+    if (node instanceof AppSettingTreeItem) {
+        connections.push({ name: node.id, value: node.value });
+        await node.parent.runWithTemporaryDescription(context, localize('adding', 'Adding...'), async () => {
+            await addLocalMIConnectionsInternal(context, connections);
+        });
+    } else {
+        await addLocalMIConnectionsInternal(context);
```
nturinski, line 24: If you care enough, you can use the `ICreateChildImplContext` to show a "Creating..." item in the tree while the create is in progress. Not sure how long it takes, I imagine not that long.

src/commands/addMIConnections/addLocalMIConnections.ts
```diff
@@ -0,0 +1,58 @@
...
+    wizardContext.connections = connections;
+
+    wizardContext.activityChildren = [];
+
+    const title: string = localize('addLocalMIConnectionsInternal', 'Add Identity-Based Connections to Local Project');
+
+    const promptSteps: AzureWizardPromptStep<IAddMIConnectionsContext>[] = [];
+    const executeSteps: AzureWizardExecuteStep<IAddMIConnectionsContext>[] = [];
+
+    promptSteps.push(new ConnectionsListStep())
```
nturinski, line 43: nit: You're not pushing steps in dynamically anymore so might as well just instantiate promptSteps and executeSteps with the steps that they need instead of creating empty arrays first,

src/commands/addMIConnections/addRemoteMIConnections.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (node instanceof AppSettingTreeItem) {
+        connections.push({ name: node.id, value: node.value });
+        await node.parent.runWithTemporaryDescription(context, localize('adding', 'Adding...'), async () => {
+            await addRemoteMIConnectionsInternal(context, connections, node);
+        });
+    } else if (node instanceof AppSettingsTreeItem) {
+        await node.runWithTemporaryDescription(context, localize('adding', 'Adding...'), async () => {
+            await addRemoteMIConnectionsInternal(context, undefined, node);
+        });
+    } else {
```
nturinski, line 30: Why do you need this else? What's the third case here?

src/commands/addMIConnections/addRemoteMIConnections.ts
```diff
@@ -0,0 +1,77 @@
...
+        promptSteps,
+        executeSteps
+    });
+
+    await wizard.prompt();
+    const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+    await context.ui.showWarningMessage(localize('rolesWillBeAssignedMessage', 'This command will assign a managed identity and roles would you like to continue?'), { modal: true }, continueOn);
+    await wizard.execute();
+
+    const message: string = localize('setConnectionsProperty', 'Successfully added remote connections in order to use identity connections your application may require additional permisssions based on your code. You also may need to modify the connection property within your trigger.');
```
nturinski, line 75: nit:

src/commands/registerCommands.ts
```diff
@@ -18,8 +18,12 @@ import { getAgentBenchmarkConfigs, getCommands, runWizardCommandWithInputs, runW
 import { ext } from '../extensionVariables';
 import { installOrUpdateFuncCoreTools } from '../funcCoreTools/installOrUpdateFuncCoreTools';
 import { uninstallFuncCoreTools } from '../funcCoreTools/uninstallFuncCoreTools';
+import { type DurableTaskSchedulerClient } from '../tree/durableTaskScheduler/DurableTaskSchedulerClient';
```
nturinski, line 21: You really just went and reorganized this whole file huh?

src/commands/addMIConnections/addLocalMIConnections.ts
motm32: 💩

src/commands/registerCommands.ts
motm32: 💩

src/commands/addMIConnections/ConnectionsListStep.ts
```diff
@@ -0,0 +1,120 @@
...
+            })).map(item => item.data);
+        }
+    }
+
+    public shouldPrompt(context: AddMIConnectionsContext): boolean {
+        return !context.connections || context.connections.length === 0;
+    }
+
+    private async getPicks(context: AddMIConnectionsContext): Promise<IAzureQuickPickItem<Connection>[]> {
+        if (context.functionapp) {
```
nturinski, line 47: nit: `return context.functionapp ? this.getRemoteQuickPicks(context) : this.getLocalQuickPicks(context)`

src/commands/addMIConnections/ConnectionsListStep.ts
```diff
@@ -0,0 +1,120 @@
...
+
+        if (picks.length === 0) {
+            const noItemFoundMessage: string = localize('noConnectionsFound', 'No connections found in local settings');
+            (await context.ui.showQuickPick(picks, {
+                placeHolder: localize('selectConnections', 'Select the connections you want to add managed identity support for'),
+                suppressPersistence: true,
+                noPicksMessage: noItemFoundMessage
+            }));
+        } else {
+            context.connections = (await context.ui.showQuickPick(picks, {
```
nturinski, line 34: Do you really need to have two separate calls for this? If the picks array is empty, isn't it going to have the same behavior as it does in the if clause?

src/commands/addMIConnections/ConnectionsListStep.ts
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    private async getLocalQuickPicks(context: AddMIConnectionsContext, workspaceFolder?: vscode.WorkspaceFolder): Promise<IAzureQuickPickItem<Connection>[]> {
+        const picks: IAzureQuickPickItem<Connection>[] = [];
+        const message: string = localize('selectLocalSettings', 'Select the local settings to add identity settings for.');
+        const localSettingsPath: string = await getLocalSettingsFile(context, message, workspaceFolder);
+        context.localSettingsPath = localSettingsPath;
+
+        if (await AzExtFsExtra.pathExists(localSettingsPath)) {
```
nturinski, line 60: Now that you're using this function, I think you can get rid of the if statement above.

src/commands/addMIConnections/addLocalMIConnections.ts
```diff
@@ -0,0 +1,55 @@
...
+        title,
+        promptSteps,
+        executeSteps
+    });
+
+    await wizard.prompt();
+    await wizard.execute();
+
+    const continueOn: MessageItem = { title: localize('continueOn', 'Continue') };
+    const message: string = localize('setConnectionsProperty', 'Successfully added local connections. In order to use identity based connections, you may need to set the connection properties within your trigger.');
```
nturinski, line 53: Successfully added local connections. To use identity-based connections, you may need to configure the connection properties within your trigger.

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -60,11 +65,45 @@ export class LocalProjectTreeItem extends LocalProjectTreeItemBase implements Di
...
         });
     }
 
+    static async createLocalProjectTreeItem(parent: AzExtParentTreeItem, localProject: LocalProjectInternal): Promise<LocalProjectTreeItem> {
+        const result = await callWithTelemetryAndErrorHandling<LocalProjectTreeItem>('listApplicationSettings', async (context: IActionContext) => {
+            const localSettingsPath: string | undefined = await getLocalSettingsFileNoPrompt(context, localProject.options.folder);
+            if (localSettingsPath) {
+                if (await AzExtFsExtra.pathExists(localSettingsPath)) {
+                    const localSettingsUri: vscode.Uri = vscode.Uri.file(localSettingsPath);
+                    let localSettings: ILocalSettingsJson = <ILocalSettingsJson>await AzExtFsExtra.readJSON(localSettingsPath);
```
nturinski, line 79: Use your helper function here.

src/commands/addMIConnections/ConnectionsListStep.ts
motm32: Talked in person but we used two seperate calls since one is using `canPickMany` and if there are no picks we don't want `canPickMany` to be true.

src/commands/addMIConnections/ConnectionsListStep.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    private async getLocalQuickPicks(context: AddMIConnectionsContext, workspaceFolder?: vscode.WorkspaceFolder): Promise<IAzureQuickPickItem<Connection>[]> {
+        const picks: IAzureQuickPickItem<Connection>[] = [];
+        const message: string = localize('selectLocalSettings', 'Select the local settings to add identity settings for.');
+        const localSettingsPath: string = await getLocalSettingsFile(context, message, workspaceFolder);
+        context.localSettingsPath = localSettingsPath;
+
+        const localSettings = await getLocalSettingsJson(context, localSettingsPath);
+        if (localSettings.Values) {
+            for (const [key, value] of Object.entries(localSettings.Values)) {
```
nturinski, line 58: nit: If you want to be even cleaner, this can be a helper function since it's used by getRemoteQuickPicks too (don't hate me)

src/commands/appSettings/localSettings/LocalSettingsClient.ts
```diff
@@ -8,36 +8,35 @@ import { type AppSettingsClientProvider, type IAppSettingsClient } from "@micros
...
+        this._workspaceFolder = workspaceFolder;
     }
 
     public async listApplicationSettings(): Promise<StringDictionary> {
         const result = await callWithTelemetryAndErrorHandling<StringDictionary | undefined>('listApplicationSettings', async (context: IActionContext) => {
-            const localSettingsPath: string | undefined = await getLocalSettingsFileNoPrompt(context, this._node.workspaceFolder);
+            const localSettingsPath: string | undefined = await getLocalSettingsFileNoPrompt(context, this._workspaceFolder);
             if (localSettingsPath === undefined) {
                 return { properties: {} };
             } else {
```
nturinski, line 42: You can use your helper function again in here (a couple lines below this).

It also may make more sense to export `getLocalSettingsJson` from here since it's more related to the LocalSettingsClient than anything else. You could consider adding it to a util file too.

src/commands/addMIConnections/addLocalMIConnections.ts
```diff
@@ -0,0 +1,55 @@
...
+import { type MessageItem } from "vscode";
+import { localize } from "../../localize";
+import { createActivityContext } from "../../utils/activityUtils";
+import { type AddMIConnectionsContext } from "./AddMIConnectionsContext";
+import { ConnectionsListStep, type Connection } from "./ConnectionsListStep";
+import { LocalSettingsAddStep } from "./LocalSettingsAddStep";
+import { SettingsAddBaseStep } from "./SettingsAddBaseStep";
+
+export async function addLocalMIConnections(context: IActionContext, node?: AppSettingTreeItem): Promise<void> {
+    const connections: Connection[] = [];
```
nturinski, line 17: I think it's possible to have multiple local project nodes, so if you enter through the command palette, you'll need something like:

```
if (!node) {
  node = await pickLocalProject(); // not real code
}
```

src/tree/localProject/LocalProjectTreeItem.ts
```diff
@@ -60,11 +62,33 @@ export class LocalProjectTreeItem extends LocalProjectTreeItemBase implements Di
...
+                    for (const [key, value] of Object.entries(localSettings.Values)) {
+                        if (!isSettingConvertible(key, value)) {
+                            continue;
+                        }
+
+                        this.contextValue = 'azFuncLocalProject;convert';
+                    }
+                }
+            }
+
```
nturinski, line 85: nit: Extra line breaks.

nturinski (approved):

PR #4405, opened 2025-03-09
Integrate DTS emulator with workspace view

philliphoff:
Adds a DTS emulator node to the Workspace view of the Azure Resources explorer, which will show running DTS emulator instances.  This node offers the ability to start new emulator instances.  Emulator nodes, like in the Azure Resources view, offers the abilities:

 - Copy connections strings
 - Open the corresponding task hub dashboard
 - Shutdown the emulator

The DTS settings have been updated to allow customization (i.e. registry, image, tag) of the emulator container image.  Emulators started by the VS Code instance will be shutdown when the instance is closed.  (Emulators started outside of the VS Code instance will *not* be shutdown.)

Note: the emulator features are not hidden behind the "preview features" toggle as they are (as of now) available to all.


philliphoff:
@nturinski Thanks for taking a look.

> Phil, would it be possible to add a command to install/uninstall the emulator similar to Azure Functions Core Tools?

The emulator is just a Docker container; there's nothing to install or uninstall really (save pulling/removing an image).  I'll think about whether we can show/hide the workspace node based on the existence of the image (and then have a command that just pulls it).  I feel that's not terribly discoverable, though.

> When you activate the extension, it immediately throws an error as well:

Well that's not good; I'll try to repro and resolve that.

nturinski:
> The emulator is just a Docker container; there's nothing to install or uninstall really (save pulling/removing an image). I'll think about whether we can show/hide the workspace node based on the existence of the image (and then have a command that just pulls it). I feel that's not terribly discoverable, though.

I think if the node could instruct users on where/how to get the image, that could be helpful. Like a disabled/enabled state? That way we could still have the discoverability regardless.

philliphoff:
> I think if the node could instruct users on where/how to get the image, that could be helpful. Like a disabled/enabled state? That way we could still have the discoverability regardless.

@nturinski The image will be auto-pulled if/when the user attempts to start the container, so there's nothing they need to do (aside from Docker being running).

That said, I've added some robustness to the Docker management logic, and on errors the DTS Emulators node will show a warning node with the error (e.g. "is Docker installed and running?"). I've also updated the root node to auto-expand *only* if emulators are already running.  This *should* make the feature a little less "in your face" if you're not actually using it.

alexweininger:
Would you want to try to start the Docker runtime for the user if it's not already running? I know there are other runtimes, but I think most devs still use Docker. Mostly suggesting it because I was using the dev containers feature in VS Code today and it auto started Docker desktop for me so it's fresh in my mind.

alexweininger:
Non-blocking feedback here:

image
image

Since there aren't that many actions available on the emulator tree item and the task hub item, I think it makes sense to expose the "big" ones as buttons on the tree item itself. Similar to the buttons we have on the Functions local project tree item:
Screenshot 2025-03-19 at 5 13 23 PM

I think at least a stop/start button on the emulator item would make a lot of sense. And an open in dashboard button on the task hub. I figure we might as well make these things as discoverable as possible.

philliphoff:
> Would you want to try to start the Docker runtime for the user if it's not already running? I know there are other runtimes, but I think most devs still use Docker. Mostly suggesting it because I was using the dev containers feature in VS Code today and it auto started Docker desktop for me so it's fresh in my mind.

@alexweininger I talked to the Docker extension team and they, themselves, do not automatically start the container runtime so I'm hesitant to try to do that in this case.  (Just trying to understand *which* runtime is in use and how to start it would be a significant concern--I'm using the Docker extension client library to abstract away the runtime being used, so the extension doesn't actually know.)

philliphoff:
> I think at least a stop/start button on the emulator item would make a lot of sense. And an open in dashboard button on the task hub. I figure we might as well make these things as discoverable as possible.

@alexweininger Those are good ideas and I've added inline commands for opening the taskhub dashboard, as well as for starting and stopping emulator instances.

philliphoff:
@alexweininger Bump.  (Not in any hurry, just don't want to forget about it myself.)

src/tree/durableTaskScheduler/DurableTaskSchedulerEmulatorsWorkspaceResourceModel.ts
```diff
@@ -0,0 +1,70 @@
...
+        treeItem.iconPath = treeUtils.getIconPath('durableTaskScheduler/DurableTaskScheduler');
+
+        return treeItem;
+    }
+
+    id?: string | undefined;
+
+    private async emulatorsExist(): Promise<boolean> {
+        // Used any cached result that might exist...
+        this.getEmulatorsTask ??= this.emulatorClient.getEmulators();
```
nturinski, line 53: Is there any action that the user could take to change what `getEmulators` would output during a session? 

I ask because `getEmulatorsTask` never gets set to undefined, so this list would never get updated even if the user stopped an emulator. But I may be misunderstanding what "exists" means in this case.

src/tree/durableTaskScheduler/DurableTaskSchedulerErrorWorkspaceResourceModel.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import { type ProviderResult, ThemeIcon, TreeItem } from "vscode";
+import { type DurableTaskSchedulerWorkspaceResourceModel } from "./DurableTaskSchedulerWorkspaceResourceModel";
+
+export class DurableTaskSchedulerErrorWorkspaceResourceModel implements DurableTaskSchedulerWorkspaceResourceModel {
+    constructor(
+        private readonly error: string) {
+    }
+
+    getChildren(): ProviderResult<DurableTaskSchedulerWorkspaceResourceModel[]> {
```
nturinski, line 14: Should this item have any children? I believe it'll make it expandable, but then nothing would show below which is an odd UI experience.

nturinski (commented):
There were a couple of times get children returned an empty array or the context values were joined with a space. I didn't feel like I needed to comment on every single one, but if you'd like me to just so it's easier to identify where those all are, I can do that.

package.json
```diff
@@ -797,8 +823,33 @@
...
+                    "group": "1@1"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.openTaskHubDashboard",
+                    "when": "view =~ /(azureResourceGroups|azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.taskHubDashboard/",
+                    "group": "inline"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.startEmulator",
+                    "when": "view =~ /(azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.emulators/",
```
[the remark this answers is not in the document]
philliphoff: I think I was just following the pattern found elsewhere.  I'll make it `azureWorkspace` specific.

package.json
```diff
@@ -880,6 +935,14 @@
                     "command": "azureFunctions.durableTaskScheduler.openTaskHubDashboard",
                     "when": "never"
                 },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.startEmulator",
```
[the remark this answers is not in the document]
philliphoff: Good point. I'll keep it visible.

package.json
```diff
@@ -997,6 +1060,21 @@
...
+                    },
+                    "azureFunctions.durableTaskScheduler.emulatorImage": {
+                        "type": "string",
+                        "description": "%azureFunctions.durableTaskScheduler.emulatorImage%",
+                        "default": "dts/dts-emulator"
+                    },
+                    "azureFunctions.durableTaskScheduler.emulatorTag": {
+                        "type": "string",
+                        "description": "%azureFunctions.durableTaskScheduler.emulatorTag%",
+                        "default": "v0.0.4"
```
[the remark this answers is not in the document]
philliphoff: As of last week, we do now have a rolling `latest` tag. I'll update the default.

src/tree/durableTaskScheduler/DurableTaskSchedulerEmulatorClient.ts
```diff
@@ -0,0 +1,122 @@
...
+}
+
+interface ImageFullTag {
+    registry: string;
+    image: string;
+    tag: string;
+}
+
+function getEmulatorFullTag(): ImageFullTag {
+    const configuration = workspace.getConfiguration('azureFunctions');
```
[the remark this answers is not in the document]
philliphoff: Nice! Will use them.

src/tree/durableTaskScheduler/DurableTaskSchedulerErrorWorkspaceResourceModel.ts
philliphoff: The `TreeItemCollapsableState` controls whether the expansion UX is shown (and, by default, not).  That said, I'll update the model interface to make this optional so it can be omitted, and be clearer, for those that don't have children.

src/tree/durableTaskScheduler/DurableTaskSchedulerEmulatorsWorkspaceResourceModel.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        try {
+            await this.getEmulatorsTask;
+
+            const emulators = await this.getEmulatorsTask;
```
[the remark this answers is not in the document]
philliphoff: Ah, yes, that is redundant.

src/tree/durableTaskScheduler/DurableTaskSchedulerEmulatorsWorkspaceResourceModel.ts
philliphoff: `getEmulatorsTask` is updated when the node is asked for children.  That said, the logic takes advantage of the fact that VS Code will refresh/recreate the parent node entirely in the event of user action (e.g. manually refreshing the window, or starting or stopping an emulator).  (See the provider's subscription to emulator changes.). The observation was that VS Code tended to call getTreeItem() quite often, for the same node, so the thought was to try to eliminate redundant calls to relatively slow container commands.

package.json
```diff
@@ -765,14 +786,19 @@
                     "when": "view =~ /(azureResourceGroups|azureFocusView)/ && viewItem =~ /userAssignedIdentity/",
                     "group": "1@1"
                 },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.copyEmulatorConnectionString",
+                    "when": "view =~ /(azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.emulatorInstance/",
```
nturinski, line 791: This probably doesn't need `azureFocusView` in the view regex.

package.json
```diff
@@ -797,8 +823,33 @@
...
+                    "group": "inline"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.startEmulator",
+                    "when": "view == azureWorkspace && viewItem =~ /azFunc.dts.emulators/",
+                    "group": "1@1"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.startEmulator",
+                    "when": "view =~ /(azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.emulators/",
```
nturinski, line 841: This probably doesn't need `azureFocusView` in the view regex.

package.json
```diff
@@ -797,8 +823,33 @@
...
+                    "group": "inline"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.stopEmulator",
+                    "when": "view =~ /(azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.emulatorInstance/",
```
nturinski, line 846: This probably doesn't need `azureFocusView` in the view regex.

package.json
```diff
@@ -797,8 +823,33 @@
...
                     "group": "1@1"
+                },
+                {
+                    "command": "azureFunctions.durableTaskScheduler.stopEmulator",
+                    "when": "view =~ /(azureWorkspace|azureFocusView)/ && viewItem =~ /azFunc.dts.emulatorInstance/",
```
nturinski, line 851: This probably doesn't need `azureFocusView` in the view regex.

src/tree/durableTaskScheduler/DurableTaskSchedulerErrorWorkspaceResourceModel.ts
nturinski: I think you missed this one when removing `getChildren` definitions.

nturinski (commented):
Overall looks good to me. Let me know whenever you're done-- any approval I do now is just going to get stale.

src/tree/durableTaskScheduler/DurableTaskSchedulerErrorWorkspaceResourceModel.ts
philliphoff: Bah!  Sorry about that.

package.json
philliphoff: Updated.

package.json
philliphoff: Updated.

package.json
philliphoff: Updated.

package.json
philliphoff: Updated.

nturinski:
@philliphoff I'm going to hold off on merging this for just a couple of days. We had CTI run a release test pass this week and we were planning to release today or Monday. 

We'll release again soon including this and other DTS changes.

philliphoff:
No worries!  Thanks for taking the time to review!

PR #4423, opened 2025-03-21
Check for connection strings and managed identity settings on deployment

nturinski:
This PR is checking the local settings of the project being deployed and the remote settings of the app being deployed to. It is classifying the setting type as 'ConnectionString' | 'ManagedIdentity' | 'Emulator'.

Based on the settings, it will warn the user on deploy about the following two scenarios:

1. If the user has a managed identity in their local settings, but their function app is not associated to any managed identity
2. If the user has a connection string in their local or remote settings

The warnings are just to alert the user that they _should_ use managed identities, but it doesn't offer a way to convert their local settings yet. After Megan's PR, we may be able to offer a way to add the settings for them though.

I originally was trying to implement this matrix:

image

But I ran into a few of issues with trying to do this:

First of all, I think it's somewhat difficult for us to know if a local setting is the equivalent of a remote setting. I think technically the connection string can be compared to something that is hooked up to a managed identity, but I'm not sure how accurate it is. Not to mention, there are potentially many connection strings to one app, and it seemed like a lot to try to get it working 100% correctly.

Secondly, even if we do set up the managed identity for the user, there's no guarantee that the roles will be configured correctly. This could leave the user thinking that once they assign the MI, their app is configured correctly which is not the case. This is something we should revisit once we are able to assign roles.

I think I'm going to go back and make the default behavior for deploying to include the local settings with it. In those cases, we will not deploy anything that is an emulator string.


src/commands/deploy/getWarningsForConnectionSettings.ts
```diff
@@ -0,0 +1,98 @@
...
+            name: property.propertyName,
+            value: property.value,
+            type: 'ManagedIdentity'
+        };
+    }
+
+    return undefined;
+}
+
+function checkForEmulatorSettings(property: { [propertyName: string]: string }): ConnectionSetting | undefined {
```
MicroFish91, line 74: In the future if we get more emulator settings to add, we'll need to add them here as well correct?  Is there anything we can do here to help future engineers remember to update this logic as well if a new emulator string is added?

MicroFish91 (approved):
Otherwise looks okay to me, I didn't actually test this live but I did a once or twice over on the code changes

src/commands/deploy/getWarningsForConnectionSettings.ts
nturinski: Yeah that's a good idea.

motm32 (approved):

PR #4431, opened 2025-03-26
Logic to look for azurite, install/run it for user

nturinski:
If Azurite isn't running, we used to just give the user the option to debug anyway.  This will let users install and/or run the emulator from the modal.

installAzurite


alexweininger:
I still think making Azurite a dependency is a good idea

MicroFish91 (approved):
Pretty slick

nturinski:
> I still think making Azurite a dependency is a good idea

I'll consider it.

> Pretty slick. It'd be nice if we could give users an easy way to stop / clean in the future as well.

I do wanna be careful about cleaning it since it's basically a destructive action even if it is just an emulator.

PR #4451, opened 2025-03-31
Fix "Add Function App Identity Connections..." command when selecting "+ Create new user assigned identity"

### LINKED ISSUE #4446: Fail to execute "Add Function App Identity Connections..." command after selecting "+ Create new user assigned identity"
**OS:** Win 11
**Build Version:** 20250331.1

**Repro Steps:**
1. Create a Flex Consumption function app.
2. F1 -> Select "Azure Functions: Add Function App Identity Connections..." -> Select a function app -> Select the connections and click "OK" -> Select "+ Create new user assigned identity" to finish this operation -> Click "Continue" button in the below dialog.
Image
3. Check whether succeeds to execute "Add Function App Identity Connections..." command.

**Expect:**
Succeed to execute "Add Function App Identity Connections..." command.

**Actual:**
Fail to execute "Add Function App Identity Connections..." command with an error.
Image

**More Info:**
This issue does not reproduce if executing "Add Function App Identity Connections..." command with an existing user assigned identity selected.

motm32:
Fixes #4446.

This is happening since the managed identity wasn't being created before trying to set settings for the clientId property. By changing the priorities to make these steps happen after the managed identity is created the problem seems to be solved.

### DIFF
--- a/src/commands/addMIConnections/RemoteSettingsAddStep.ts
+++ b/src/commands/addMIConnections/RemoteSettingsAddStep.ts
@@ -9,7 +9,7 @@ import { localize } from "../../localize";
 import { type AddMIConnectionsContext } from "./AddMIConnectionsContext";
 
 export class RemoteSettingsAddStep extends AzureWizardExecuteStep<AddMIConnectionsContext> {
-    public priority: number = 110;
+    public priority: number = 160;
 
     public async execute(context: AddMIConnectionsContext): Promise<void> {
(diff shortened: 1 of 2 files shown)


nturinski (approved):
What was the MI priority? I thought I had changed it to like 101 or something?

We should probably keep a list of all the steps similar to what Matt ~~did in container apps~~ I thought it was there but wherever it was.

motm32:
@nturinski  The priority on the `UserAssignedIdentityCreateStep` is 140.

PR #4452, opened 2025-03-31
Remove warning for `Deployment_Storage_Connection_String` setting

motm32:
We don't want this to setting to have a warning or be able to be called by the `Add Function App Identity Connections...` command or `Add Local Project Identity Connections...` command.


nturinski (approved):

nturinski:
To add some further context, we don't want to have a warning for `Deployment_Storage_Connection_String` yet because this is a reserved app setting for flex consumption app deployment. There is a way to configure it for MI, but we didn't have time to do that in this iteration. So we should just be ignoring it for now.

PR #4453, opened 2025-03-31
Fix priority for `Add Local Project Identity Connections...` command

motm32:
The `LocalSettingsAddStep` needs to come after the `SettingsAddBaseStep`. The priority for `SettingsAddBaseStep` was changed here #4451


nturinski (approved):
Keeping track of step priorities be like

PR #4487, opened 2025-04-18
Convert execute activity steps to v2 children

nturinski:
Note that this is only converting the create commands since that's where the bulk of the azuretools steps were contributing items to the activity log. 

Every other command just has the top level item (if it uses the activity log at all). But getting Functions up to par with ACA is a task for another day.

PR #4490, opened 2025-04-23
Add children for create new project and create new function

nturinski:
Built off of: https://github.com/microsoft/vscode-azurefunctions/pull/4487

This is what the commands look like in the logs. Note that the V2 python model uses the ActionSchema so the activity children labels are a little different from the typical scenario (such as the JavaScript one)

{84738631-5349-46AE-B576-0D9245DAFC53}


MicroFish91:
I noticed some minor inconsistencies between the labels, for example:
1. Add httpTrigger function
2. Create new HTTP trigger
3. Create New Project with HttpTrigger function

Is there a way for us to easily standardize these labels more, or are we populating most of this through template arguments that we don't have much control over?

nturinski:
Do you remember what you created? From my PR description:

> Note that the V2 python model uses the ActionSchema so the activity children labels are a little different from the typical scenario

MicroFish91:
> Do you remember what you created? From my PR description:
> 
> > Note that the V2 python model uses the ActionSchema so the activity children labels are a little different from the typical scenario

Oh I was just referring to your screenshot in the PR description.

MicroFish91 (approved):
Looks good.  Could be worth filing an upstream issue at some point to standardize these template labels some more

MicroFish91 (approved):

PR #4506, opened 2025-05-14
Try using oauth2 token for storage client

nturinski:
Oauth2 token should work for both cases, so I'm trying that first. If it fails, we can try the connection string. If that fails too, then there's not much we can do.


meeraharidasa (approved):

MicroFish91 (approved):
If Meera approves, I approve 🫡

Issue #4516, opened 2025-05-16
There is no "Successfully created function app..." log in the output window after creating a containerized function app


Amar-QA:
**OS:** All
**Build Version:** 20250516.2
**Regression:** Previous Release

**Repro Steps:**

1. Create a containerized function app.
2. Open "Azure Functions" output window.
3. Check whether there is a "Successfully created function app..." log in the output window. 

**Expect:** 
There is a "Successfully created function app..." log in the output window.
![Image](https://github.com/user-attachments/assets/4e12355a-35ee-492e-908c-56919a56cbba)

**Actual:** 
There is no "Successfully created function app..." log in the output window.
![Image](https://github.com/user-attachments/assets/8e7d36b2-fff2-48e9-aecf-ff8f5e1cf9a8)

**More info:** 
1. There is also no "Creating new function app..." log in the output window when creating a containerized function app.
2. This issue does not reproduce on the current released Functions extension v1.17.2.

Issue #4512, opened 2025-05-15
Fail to create an advanced containerized function app if selecting "+ Create new storage account" in the creation wizard


Eva-Feng:
**OS:** All
**Build Version:** 20250515.3

**Repro Steps:**
1. Create an advanced containerized Function App and select "+ Create new storage account" in the creation wizard.
2. Check whether succeeds to create an advanced containerized function app or not.

**Expect:**
Succeed to create an advanced containerized function app.

**Actual:**
Fail to create an advanced containerized function app.
![Image](https://github.com/user-attachments/assets/3a076fc8-1df8-4e45-a258-4cf8eb29ad64)

**More Info:**
1. The above containerized Function App still lists under the Function App node.
2. This issue does not reproduce if selecting an existing storage account in the creation wizard.

v-ruizh:
This issue has been fixed on 20250520.1.

PR #4518, opened 2025-05-16
Add output to containerized steps and fix basic create

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4516
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4512#issuecomment-2886213999

### DIFF
--- a/src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
+++ b/src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
@@ -69,7 +69,8 @@ export async function createCreateFunctionAppComponents(context: ICreateFunction
 
     if (!wizardContext.advancedCreation) {
         LocationListStep.addStep(wizardContext, promptSteps);
-        wizardContext.useFlexConsumptionPlan = true;
+        // if the user is deploying to a container app, do not use a flex consumption plan
+        wizardContext.useFlexConsumptionPlan = true && !context.dockerfilePath;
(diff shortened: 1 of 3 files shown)


src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
```diff
@@ -69,7 +69,8 @@ export async function createCreateFunctionAppComponents(context: ICreateFunction
 
     if (!wizardContext.advancedCreation) {
         LocationListStep.addStep(wizardContext, promptSteps);
-        wizardContext.useFlexConsumptionPlan = true;
+        // if the user is deploying to a container app, do not use a flex consumption plan
+        wizardContext.useFlexConsumptionPlan = true && !context.dockerfilePath;
```
MicroFish91, line 73: Nit: I think the `true` is unnecessary here

MicroFish91 (approved):

src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
nturinski: It doesn't, but I wanted to keep it so it's more explicitly clear that it _should_ be set to true in all cases except where there's a dockerfilePath.

Leaving it as just `context.dockerfilePath` makes it read to me like it's something dockerfile related as in, because we don't have a dockerfile, we need to default to this versus like, this is the normal default, but having a dockerfile will screw it up, so we need to make sure that we don't have one.

Not sure if that distinction makes sense or not.

Issue #4394, opened 2025-03-05
Add DTS to Durable Functions storage provider list


lilyjma:
Add **Durable Task Scheduler (preview)** to the following list that shows up during Function project creation (when Durable Orchestration is picked)

![Image](https://github.com/user-attachments/assets/d48e0160-5aad-4d3e-adf0-922eb5a78b75)

This would require associated updates to the `local.settings.json` and `host.json` file. 

1. `local.settings.json` needs to have these two environment variables
```
 "DURABLE_TASK_SCHEDULER_CONNECTION_STRING": "Endpoint=http://localhost:<port number>;Authentication=None",
 "TASKHUB_NAME": "default"
```
2. `host.json` needs to have storage provider type `azureManaged` and task hub name:
```
 "durableTask": {
      "hubName": "%TASKHUB_NAME%",
      "storageProvider": {
        "type": "azureManaged",
        "connectionStringName": "DURABLE_TASK_SCHEDULER_CONNECTION_STRING"
      }
    }
```

3. For .NET users, the following package needs to be added to `.csproj` when **Durable Task Scheduler (preview)** is picked:

[Microsoft.Azure.Functions.Worker.Extensions.DurableTask.AzureManaged](https://www.nuget.org/packages/Microsoft.Azure.Functions.Worker.Extensions.DurableTask.AzureManaged)

For non .NET users, the following version of preview bundles is needed (specified in `host.json`)
```json
 "extensionBundle": {
   "id": "Microsoft.Azure.Functions.ExtensionBundle.Preview",
   "version": "[4.29.0, 5.0.0)"
  }
```

cc @nturinski

Issue #4393, opened 2025-03-05
Remove Netherite from Durable Functions storage provider list


lilyjma:
Remove Netherite from this list:

![Image](https://github.com/user-attachments/assets/7ac60da8-f831-411e-aca3-2986bef7760c)

PR #4527, opened 2025-05-27
Add `DTS` preview support for `Create Function` and remove `Netherite` from list of new storage options

### LINKED ISSUE #3599: Remove durable functions package manual installation for in process Dotnet
See `durableUtils`: 

image

Once templates start bringing in 2.9.1 or later, we can probably remove the above statement.  I don't know if it will end up being a big deal, I just don't want to be maintaining this installation manually if we don't need to be.  I don't want there to ever be a potential for mismatch between what the template supports and what scheme the latest df package supports because of something we're doing...

MicroFish91:
Closes #3599 #4393 #4394

### DIFF
--- a/src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
+++ b/src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
@@ -6,7 +6,7 @@
 import { AzExtFsExtra, AzureWizardExecuteStepWithActivityOutput } from '@microsoft/vscode-azext-utils';
 import * as path from "path";
 import { type Progress } from 'vscode';
-import { ConnectionKey, DurableBackend, hostFileName } from '../../../constants';
+import { ConnectionKey, DurableBackend, hostFileName, ProjectLanguage } from '../../../constants';
 import { viewOutput } from '../../../constants-nls';
 import { ext } from '../../../extensionVariables';
 import { type IHostJsonV2 } from '../../../funcConfig/host';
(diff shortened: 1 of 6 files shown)


src/commands/createFunction/durableSteps/DurableProjectConfigureStep.ts
```diff
@@ -72,6 +72,19 @@ export class DurableProjectConfigureStep<T extends IFunctionWizardContext> exten
                 hostJson.extensions.durableTask = durableUtils.getDefaultNetheriteTaskConfig();
                 await setLocalAppSetting(context, context.projectPath, ConnectionKey.EventHubs, '', MismatchBehavior.Overwrite);
                 break;
+            case DurableBackend.DTS:
+                hostJson.extensions.durableTask = durableUtils.getDefaultDTSTaskConfig();
+                // Non- .NET projects require a special preview extension bundle to work properly
+                // Todo: Remove once this functionality is out of preview
+                if (context.language !== ProjectLanguage.CSharp && context.language !== ProjectLanguage.FSharp) {
+                    hostJson.extensionBundle = {
```
nturinski, line 80: I think that the fact we are changing their host bundle version maybe should be output somewhere. Maybe as an activity child or just the output channel?

src/utils/durableUtils.ts
```diff
@@ -150,39 +154,46 @@ export namespace durableUtils {
...
         switch (context.newDurableStorageType) {
             case DurableBackend.Netherite:
                 isDotnetIsolated ?
-                    packageNames.push(dotnetIsolatedDfNetheritePackage) :
-                    packageNames.push(dotnetInProcDfNetheritePackage);
+                    packages.push({ name: dotnetIsolatedDfNetheritePackage }) :
+                    packages.push({ name: dotnetInProcDfNetheritePackage });
+                break;
+            case DurableBackend.DTS:
+                // Todo: Remove prerelease flag once this functionality is out of preview
```
nturinski, line 167: nit: why call the property prerelease when you call it preview everywhere else (including this comment)

src/utils/durableUtils.ts
```diff
@@ -150,39 +154,46 @@ export namespace durableUtils {
...
         }
 
         const failedPackages: string[] = [];
-        for (const packageName of packageNames) {
+        for (const p of packages) {
             try {
-                await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'dotnet', 'add', 'package', packageName);
+                const packageArgs: string[] = [p.name];
+                if (p.prerelease) {
+                    packageArgs.push('--prerelease');
```
nturinski, line 192: post-nit clarity: Ah, I see why now

nturinski (approved):

nturinski (approved):

PR #4532, opened 2025-06-03
Add support for domain name label scopes

MicroFish91:


src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
```diff
@@ -54,7 +54,13 @@ export async function createCreateFunctionAppComponents(context: ICreateFunction
 
     await detectDockerfile(context);
 
-    promptSteps.push(new SiteNameStep(context.dockerfilePath ? "containerizedFunctionApp" : "functionApp"));
+    // SiteNameStep pre-requisites
+    if (!context.advancedCreation) {
+        LocationListStep.addStep(wizardContext, promptSteps);
+    } else {
+        CustomLocationListStep.addStep(wizardContext, promptSteps);
```
MicroFish91, line 65: Would appreciate a double check on this location logic to make sure these are the scenarios where each type of list step applies

src/commands/createFunctionApp/FunctionAppCreateStep.ts
```diff
@@ -63,8 +62,40 @@ export class FunctionAppCreateStep extends AzureWizardExecuteStepWithActivityOut
...
+        return await client.webApps.beginCreateOrUpdateAndWait(rgName, siteName, site)
+    }
+
+    private async getNewFlexSite(context: IFlexFunctionAppWizardContext, sku: Sku): Promise<Site> {
+        const site: Site = await this.getNewBaseSite(context);
+        site.functionAppConfig = this.getNewFlexFunctionAppConfig(context, sku);
+        return site;
+    }
+
+    private async getNewSite(context: IFunctionAppWizardContext, stack?: FullFunctionAppStack): Promise<Site> {
```
MicroFish91, line 97: Some of the diff here is showing up as new code, but much of it is just me reorganizing existing code sections and grouping related logic under their own foldable regions.  

Reorganizing was especially important because we temporarily need to duplicate (double) the number of branching logic paths due to needing to make a manual call without the SDK (which requires forming different shaped site payloads than what we're used to) when tenant DNLS is chosen.

src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
```diff
@@ -58,7 +58,13 @@ export async function createCreateFunctionAppComponents(context: ICreateFunction
         wizardContext.hasDurableTaskScheduler = await durableUtils.getStorageTypeFromWorkspace(language, wizardContext.workspaceFolder.uri.fsPath) === DurableBackend.DTS;
     }
 
-    promptSteps.push(new SiteNameStep(context.dockerfilePath ? "containerizedFunctionApp" : "functionApp"));
+    // SiteNameStep pre-requisites
+    if (!context.advancedCreation) {
+        LocationListStep.addStep(wizardContext, promptSteps);
+    } else {
+        CustomLocationListStep.addStep(wizardContext, promptSteps);
+        promptSteps.push(new ResourceGroupListStep());
```
nturinski, line 66: One bad experience with resource group list step being here is that even it doesn't use my RG name for anything including the app name. I was kind of expecting an auto complete there.

PR #4564, opened 2025-06-20
Refactor getSite to initSite and use a get site to type guard

nturinski:
TLDR:
Basically, the async getSite calls weren't really working everywhere due to SiteClient referencing .site in places (I think.) It was easier to just type guard everything with a get site that throws an error if site isn't initialized, and use initSite to initialize the sites anytime it was required.

Overall, I think it's a bit clunkier of a solution, but it looks cleaner since you can reference node.site directly again.

COPILOT SUMMARY:

This pull request introduces several changes to improve the handling of site initialization and caching in the Azure Functions extension codebase. The most significant updates include replacing multiple calls to `getSite` with a new `initSite` method for initializing site data, modifying the `FunctionAppModel` type to include an `isFlex` property, and optimizing cache management within the `FunctionAppResolver` class.

### DIFF
--- a/src/tree/ResolvedFunctionAppResource.ts
+++ b/src/tree/ResolvedFunctionAppResource.ts
@@ -105,7 +105,7 @@ export class ResolvedFunctionAppResource extends ResolvedFunctionAppBase impleme
                 void this.initSite(context);
             });
             // If the site is not initialized, we should throw an error
-            throw new Error(localize('siteNotSet', 'Site has not initialized. Please try again in a moment.'));
+            throw new Error(localize('siteNotSet', 'Site is not initialized. Please try again in a moment.'));
         }
         return this._site;
     }
@@ -128,6 +128,7 @@ export class ResolvedFunctionAppResource extends ResolvedFunctionAppBase impleme
         return {
             id: site.id ?? '',
             name: site.name ?? '',
+            type: site.type ?? '',
             kind: site.kind ?? '',
             location: site.location,
             resourceGroup: site.resourceGroup ?? '',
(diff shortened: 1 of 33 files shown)


src/tree/ResolvedFunctionAppResource.ts
```diff
@@ -91,18 +90,21 @@ export class ResolvedFunctionAppResource extends ResolvedFunctionAppBase impleme
...
             this.addValuesToMask(this._site);
         }
+    }
 
-        return site;
+    public get site(): ParsedSite {
+        if (!this._site) {
+            throw new Error(localize('siteNotSet', 'Site is not set. Ensure the site is initialized before accessing it.'));
+        }
+        return this._site;
```
alexweininger: When we throw this, can we also just load the site in the background so that if a user tries again it'll work?

src/tree/ResolvedFunctionAppResource.ts
```diff
@@ -68,20 +67,20 @@ export class ResolvedFunctionAppResource extends ResolvedFunctionAppBase impleme
...
-            this.data = dataModel;
-        } else {
-            this._site = new ParsedSite(dataModel, subscription);
-            this._isFlex = !!dataModel.functionAppConfig;
-            this.data = this._site.rawSite;
-            this.addValuesToMask(this._site);
+        if (dataModel) {
+            this._isFlex = dataModel.isFlex;
+            this.dataModel = dataModel;
+            this.data = dataModel
```
MicroFish91: Q: Why is `this.data` also set to `dataModel` here?

src/tree/ResolvedFunctionAppResource.ts
nturinski: Oh yeah, that's a smart idea.

src/tree/ResolvedFunctionAppResource.ts
nturinski: Originally I just wanted a basic data model on the site if someone right-clicked view properties, but now that I'm doing the async getData call, I can get rid of both instantiations. 👍

PR #4566, opened 2025-06-23
Add Azure deployment support for DTS projects

MicroFish91:
High level flow - if DTS project detected, check for necessary DTS app settings on the function app. If missing, prompt user to select existing or create new DTS.  Then:
  - Prompt the user for a connected user assigned identity
  - Add Durable Task Contributor role if missing
  - Prepare that connection

Then resume deploy.

Of special note, I created a new `FunctionAppUserAssignedIdentitiesListStep` that can be used going forward to select existing function app identities to assign new roles to.  It is currently missing the ability to create and assign new identities, which is a todo, but it works fine for selecting identities that are already associated with the function app.  This PR is already pretty chonky as is so we can explore that implementation later.

Related: https://github.com/microsoft/azcode-internal/issues/212

Todo:
- [x] Better activity log integration coming in follow-up PR (https://github.com/microsoft/vscode-azurefunctions/pull/4574)
- [x] I need to improve the logic so that it doesn't hard code the default connection keys (upcoming PR - this ended up being a somewhat extensive refactor but it will be much cleaner)
- [x] Create and assign new UAID in the new list step? (upcoming PR)


src/commands/appSettings/connectionSettings/durableTaskScheduler/DTSConnectionSetSettingStep.ts
```diff
@@ -6,17 +6,21 @@
...
 
 export class DTSConnectionSetSettingStep<T extends IDTSConnectionWizardContext> extends SetConnectionSettingStepBase<T> {
     public priority: number = 240;
     public debugDeploySetting: ConnectionKey = ConnectionKey.DTS;
 
     public async execute(context: T): Promise<void> {
-        await this.setConnectionSetting(context, nonNullProp(context, 'newDTSConnection'));
+        let newDTSConnectionSetting = nonNullProp(context, 'newDTSConnectionSetting');
+        if ((context as unknown as IDTSAzureConnectionWizardContext).managedIdentity) {
+            newDTSConnectionSetting = newDTSConnectionSetting.replace('<ClientID>', (context as unknown as IDTSAzureConnectionWizardContext).managedIdentity?.clientId ?? '');
```
nturinski: I understand sometimes we just gotta make the typings happy, but this is pretty uggo. Why not just have the context be:
`IDTSConnectionWizardContext | IDTSAzureConnectionWizardContext` or `Partial<IDTSAzureConnectionWizardContext>`?

nturinski: Or since `IDTSAzureConnectionWizardContext` extends `IDTSConnectionWizardContext` anyway, just replace it?

src/commands/appSettings/connectionSettings/durableTaskScheduler/validateDTSConnection.ts
```diff
@@ -0,0 +1,109 @@
...
+type DTSConnection = { [ConnectionKey.DTS]?: string, [ConnectionKey.DTSHub]?: string };
+
+/**
+ * A pre-flight operation that ensures that:
+ *
+ * a) A remote DTS connection already exists for the given function app or
+ *
+ * b) That a new DTS resource and hub is created and ready for connection to the function app
+ */
+export async function validateDTSConnection(context: DTSConnectionContext, client: SiteClient, site: ParsedSite, projectPath: string): Promise<DTSConnection | undefined> {
```
nturinski: It's a little bit odd that this is called `validateDTSConnection` considering that it actually returns the connection values. I'd think it's more like a `getDTSConnection` then.

src/commands/appSettings/connectionSettings/durableTaskScheduler/validateDTSConnection.ts
```diff
@@ -0,0 +1,109 @@
...
+    const wizardContext: IDTSAzureConnectionWizardContext = {
+        ...context,
+        ...await createActivityContext(),
+        site,
+        projectPath,
+        action: CodeAction.Deploy,
+        dtsConnectionType: ConnectionType.Azure,
+        dts: remoteDTSEndpoint ? await getDTSResource(context, remoteDTSEndpoint) : undefined,
+        // If the local settings are using the emulator (i.e. localhost), it's fine because it won't match up with any remote resources and there will be no suggestion for the user
+        suggestedDTSEndpointLocalSettings: localDTSEndpoint ? tryGetDTSEndpoint(localDTSConnection) : undefined,
```
nturinski: I think it might be worth mentioning that a localDTS endpoint can be a remote resource. I get that local in this case is referencing local settings vs remote/app settings, but the variable name makes it sound like the DTS endpoint is local (ie an emulator or something)

src/commands/appSettings/connectionSettings/durableTaskScheduler/DTSConnectionSetSettingStep.ts
MicroFish91: Although type-wise using `IDTSAzureConnectionWizardContext` works fine, I think it implies that wizards that use that step are for Azure connections, which isn't always the case.  I prefer your first suggestion for the additional accuracy, although it's a little more verbose: `IDTSConnectionWizardContext | IDTSAzureConnectionWizardContext`

Unfortunately I still have to do some typecasting this way, but at least I can omit the `as unknown` which I agree did not look great!

src/commands/appSettings/connectionSettings/durableTaskScheduler/validateDTSConnection.ts
MicroFish91: Yeah that's fair, how about something like `getDTSConnectionIfNeeded` since it doesn't always return a new connection?

src/commands/deploy/deploy.ts
```diff
@@ -101,6 +103,17 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     await node.initSite(context);
     const site = node.site;
 
+    const subscriptionContext: ISubscriptionContext & { subscription: AzureSubscription } = {
+        ...node.subscription,
+        subscription: await subscriptionExperience(context, ext.rgApiV2.resources.azureResourceTreeDataProvider, {
```
nturinski, line 108: Does expanding the subscription not get you the subscription already? Why do you need to call the subscriptionExperience stuff? Does it make the quick pick wizard come up?

src/commands/identity/listUserAssignedIdentities/FunctionAppUserAssignedIdentitiesListStep.ts
```diff
@@ -0,0 +1,109 @@
...
+     * If no role is provided, or if the step has not yet been run, this will return `undefined`.
+     */
+    get hasIdentityWithTargetRole(): boolean | undefined {
+        return this._hasTargetRole;
+    }
+
+    // Verify if any of the existing user assigned identities for the function app have the required role already
+    public async configureBeforePrompt(context: T): Promise<void> {
+        if (!this.targetRole || !this.targetRole?.scopeId) {
+            this._hasTargetRole = undefined;
```
nturinski, line 42: Is there any functional difference between having this as undefined or false?

nturinski (approved):
I think everything else looks pretty good to me

src/commands/appSettings/connectionSettings/durableTaskScheduler/validateDTSConnection.ts
nturinski: That's fine with me. I usually do `tryGet` if it can return undefined.

src/commands/identity/listUserAssignedIdentities/FunctionAppUserAssignedIdentitiesListStep.ts
MicroFish91: Yeah, `undefined` after the step has been run implies no target role was provided to check for (see comment for getter above), whereas `false` implies a role was provided but not found.

That said this ended up being a somewhat temporary solution, I have a better solution coming down the pipeline.  With this change in Azure Tools (https://github.com/microsoft/vscode-azuretools/pull/2036), once I incorporate the change, I no longer need to expose this getter because I can just run an async check in the role callback instead.  I'll include that change in a near term PR.

src/commands/deploy/deploy.ts
MicroFish91: Yeah, `node.subscription` only includes the `ISubscriptionActionContext`.  However, I also need the `AzureSubscription` specifically because the `DurableTaskSchedulerClient` requires it as a parameter.  

The quick pick wizard shouldn't ever show up because `node.subscription` should always include the subscription id, and `selectSubscriptionById` should silently grab the `AzureSubscription` for the user without prompting.

src/commands/appSettings/connectionSettings/durableTaskScheduler/validateDTSConnection.ts
MicroFish91: Cool, same but in this case I think I'm still leaning towards `ifNeeded` just because it doesn't actually try at all if the remote connection already exists

PR #4569, opened 2025-06-24
Update MicroBuild.Plugins.Signing nuget package in JsonCli tool

### LINKED ISSUE #4525: SFI-ES5.2 Update ADAL.NET packages to MSAL.NET equivalents or remove the dependency
Some internal packages contain a repackaged assembly from the ADAL.NET package, which is deprecated. Please upgrade the packages to the nearest version which does not contain ADAL.NET assembly to ensure that they are either replaced with their MSAL.NET equivalents or are no longer depended on.

More details in SFI-ES5.2.  Due by 6/30/25.

alexweininger:
I had to add and connect to that feed url that I added to the nuget.config. We'll see wha the Azure Pipelines do and hopefully that resolves #4525


alexweininger:
Need to get the updated package into our nuget feed

MicroFish91 (approved):

PR #4576, opened 2025-06-28
Fix container name always lowercase

### LINKED ISSUE #4575: Function app not created properly for flex consumption with secrets as access
When the function app is created with advanced options and the plan is 'flex consumption' and access is 'secrets', the function app created isn't able to retrieve the app keys. This is because there are no keys in the storage.

Creating an app with the same plan and access from the azure portal works fine (so it is allowed to use secrets for the flex plan, I guess).

d-s-dc:
Closes #4575. The container name should be lowercase. `newSiteName` is same as function app name which can contain capital letters. This causes the deployment to fail sometimes.


nturinski (approved):
Thanks again for your contribution! It's very helpful for us to catch these types of bugs encountered by real users.

PR #4609, opened 2025-07-24
Account for the fact that properties can be null

nturinski:
Based on the error that we are seeing in telemetry is: `Cannot read properties of null (reading "toLocaleLowerCase")`, it's almost certainly this line of code causing the issue.

I assume that not everyone can access the properties of the apps which is causing this issue.


alexweininger:
We should use zod for this stuff

nturinski:
That's a good suggestion. I think in my case, it still would have come back as properties being strictly defined since all my results have them, but I think it'd still be a good idea especially since we are probably going to be use graph queries in other places now too.

PR #4617, opened 2025-07-25
Fix slot deploy context value not being used

### LINKED ISSUE #4354: Deploy to Slot action does not allow selecting deployment slot
I tried to use the Deploy to Slot action to deploy to a development slot, but it deployed to the production slot immediately without allowing me to select the additional slot I previously created.

I'm using v1.16.1.

nturinski:
## Summary
Fixed the `deploySlot` command to properly use the `expectedContextValue` parameter, ensuring that the "Deploy to Slot" command only shows slots in the picker, not production instances.

## Changes
- Modified the `deploy` function signature to properly use `expectedContextValue` instead of ignoring it
- Added conditional logic to use `pickFunctionApp` with context filtering when `expectedContextValue` is provided and no target node is specified
- Added import for `pickFunctionApp` utility function
- Ensures the Deploy to Slot command behaves correctly by only showing deployment slots

## Root Cause
The `deploySlot` function was creating an `expectedContextValue` parameter with the correct regex (`ResolvedFunctionAppResource.pickSlotContextValue`) but the `deploy` function was ignoring this parameter (prefixed with `_`). This meant that when users ran "Deploy to Slot" from the command palette, they would see all function apps (both production and slots) instead of just slots.

## Testing
- Manual testing can verify that "Deploy to Slot" command from command palette now only shows slots
- The fix follows the same pattern used in other commands like `swapSlot`

Fixes #4354

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -21,6 +21,7 @@ import { dotnetUtils } from '../../utils/dotnetUtils';
 import { durableUtils } from '../../utils/durableUtils';
 import { isPathEqual } from '../../utils/fs';
 import { treeUtils } from '../../utils/treeUtils';
+import { pickFunctionApp } from '../../utils/pickFunctionApp';
 import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
 import { verifyInitForVSCode } from '../../vsCodeConfig/verifyInitForVSCode';
 import { type ISetConnectionSettingContext } from '../appSettings/connectionSettings/ISetConnectionSettingContext';
@@ -54,7 +55,7 @@ export async function deploySlot(context: IActionContext, target?: vscode.Uri | 
     await deploy(context, target, functionAppId, new RegExp(ResolvedFunctionAppResource.pickSlotContextValue));
 }
 
-async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string | SlotTreeItem | undefined, arg2: string | {} | undefined, _expectedContextValue?: string | RegExp): Promise<void> {
+async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string | SlotTreeItem | undefined, arg2: string | {} | undefined, expectedContextValue?: string | RegExp): Promise<void> {
     const deployPaths: IDeployPaths = await getDeployFsPath(actionContext, arg1);
 
     addLocalFuncTelemetry(actionContext, deployPaths.workspaceFolder.uri.fsPath);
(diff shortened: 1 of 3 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -96,9 +97,18 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         }
     }
 
-    const node: SlotTreeItem = await getDeployNode(context, ext.rgApi.tree, arg1, arg2, async () => {
-        return await getOrCreateFunctionApp(context)
-    });
+    let node: SlotTreeItem;
```
nturinski, line 100: **Good: Fixed slot deploy context value usage**

The PR correctly addresses the deploy command logic by using `expectedContextValue` when it's provided and `arg1` is undefined. This ensures that slot deploys properly filter to show only deployment slots rather than all function apps.

The addition of `pickFunctionApp` import and the conditional logic properly implements the intended behavior:
- When deploying to slots (`expectedContextValue` is provided), use the filtered picker
- Otherwise, use the existing deployment node logic

This fixes the issue where `deploySlot` command wasn't using its context filtering capability.

src/templates/TemplateProviderBase.ts
```diff
@@ -106,7 +106,11 @@ export abstract class TemplateProviderBase implements Disposable {
     }
 
     public async getBackupTemplateVersion(): Promise<string> {
-        return (await AzExtFsExtra.readFile(await this.getBackupVersionPath())).toString().trim();
+        const versionContent = await AzExtFsExtra.readFile(await this.getBackupVersionPath());
+        if (!versionContent) {
+            throw new Error(localize('backupVersionFileEmpty', 'Backup template version file is empty or could not be read'));
+        }
```
nturinski, line 112: **Good: Fixed template version file error handling**

This change properly addresses Issue #4599 where `Cannot read properties of undefined (reading 'trim')` error occurred when the backup template version file was empty or unreadable.

The fix adds appropriate null/undefined checking before calling `.toString().trim()`, providing a clear error message when the backup version file is problematic.

The error handling follows good practices:
- Check for falsy values before proceeding
- Throw a localized error message for better user experience  
- The error message is actionable (indicates the specific file is empty/unreadable)

src/templates/TemplateProviderBase.ts
```diff
@@ -106,7 +106,11 @@ export abstract class TemplateProviderBase implements Disposable {
     }
 
     public async getBackupTemplateVersion(): Promise<string> {
-        return (await AzExtFsExtra.readFile(await this.getBackupVersionPath())).toString().trim();
+        const versionContent = await AzExtFsExtra.readFile(await this.getBackupVersionPath());
+        if (!versionContent) {
+            throw new Error(localize('backupVersionFileEmpty', 'Backup template version file is empty or could not be read'));
```
nturinski, line 111: **Bug: Missing import for localize function**

The code uses `localize('backupVersionFileEmpty', ...)` but there's no import statement for the `localize` function in this file. 

This will cause a compilation error. You need to add:

```typescript
```

at the top of the file with the other imports.

nturinski (commented):
This PR addresses two separate issues:

1. **Deploy Command Context Value**: Fixes slot deployment to properly use the `expectedContextValue` filtering
2. **Template Version Error Handling**: Fixes the `Cannot read properties of undefined (reading 'trim')` error from Issue #4599

Both fixes are well-implemented and address the root causes correctly. However, there's a missing import for the `localize` function that needs to be added to prevent compilation errors.

The deploy logic improvement ensures that slot deployments show only deployment slots rather than all function apps, which improves the user experience. The template error handling provides better error messaging and prevents runtime crashes.

Overall good fixes, just needs the missing import resolved.

PR #4622, opened 2025-07-28
Fix host.json version not showing for AzureExtensionsBundle

### LINKED ISSUE #4594: The host.json of python timer trigger project does not contain version of extension bundle
Steps followed
- Select `Create New Project` from the command palette.
- Follow steps and create timer trigger function in python

While debugging the function the following error shows up
```
The value of version property in extensionBundle section of host.json file is invalid or missing. See https://aka.ms/functions-hostjson for more information
```

The host.json file looks like this
```json
{
  "version": "2.0",
  "logging": {
    "applicationInsights": {
      "samplingSettings": {
        "isEnabled": true,
        "excludedTypes": "Request"
      }
    }
  },
  "extensionBundle": {
    "id": "Microsoft.Azure.Functions.ExtensionBundle"
  }
}
```

d-s-dc:
Fixes #4594

Added a nullish coalescing operator to take care of the case when the returned value is `null` or `undefined`. Changing the `getLatestVersionRange` to throw error didn't make sense as it is just for trying to get the latest version. It's fine if the latest version cannot be retreived and the function returns null.

### DIFF
--- a/src/utils/bundleFeedUtils.ts
+++ b/src/utils/bundleFeedUtils.ts
@@ -90,7 +90,7 @@ export namespace bundleFeedUtils {
     export async function addDefaultBundle(context: IActionContext, hostJson: IHostJsonV2): Promise<void> {
         let versionRange: string;
         try {
-            versionRange = await getLatestVersionRange(context);
+            versionRange = (await getLatestVersionRange(context)) ?? defaultVersionRange;
         } catch {
             versionRange = defaultVersionRange;
         }


src/utils/bundleFeedUtils.ts
```diff
@@ -90,7 +90,7 @@ export namespace bundleFeedUtils {
     export async function addDefaultBundle(context: IActionContext, hostJson: IHostJsonV2): Promise<void> {
         let versionRange: string;
         try {
-            versionRange = await getLatestVersionRange(context);
+            versionRange = (await getLatestVersionRange(context)) ?? defaultVersionRange;
```
nturinski, line 93: **Good fix for null/undefined handling!** This change correctly addresses the issue described in #4594 where `getLatestVersionRange()` can return `null` or `undefined`, causing the extension bundle version to be missing from host.json.

The nullish coalescing operator (`??`) is the right solution here because:
1. It handles both `null` and `undefined` return values from `getLatestVersionRange()`
2. It falls back to `defaultVersionRange` which ensures the extension bundle gets a valid version
3. It's more precise than the existing try/catch block since it only handles the null/undefined case, not all possible errors

This should fix the Python timer trigger project creation issue where host.json was missing the version property in extensionBundle.

nturinski (approved):
This is a solid fix for issue #4594! The nullish coalescing operator appropriately handles the case where `getLatestVersionRange()` returns null/undefined, ensuring that extension bundles always get a valid version in host.json.

The change is minimal, targeted, and preserves the existing error handling while adding the necessary null safety. This should resolve the Python timer trigger project creation issue.

One minor consideration is whether the function signature should be updated to reflect that it can return null/undefined, but the fix itself is correct and ready for merge.

nturinski:
Thanks for your contribution! This will be in our next release ❤️

PR #4626, opened 2025-08-05
Remove emulator preference caching behavior per session

MicroFish91:


nturinski:
Can you elaborate what issue this is solving? Was this feedback from Lily or Philip?

MicroFish91:
Actually let me revisit this, I think I actually need to make a different change here than I originally thought.  I think I interpreted the feedback wrong, but it was from Phillip.

nturinski (approved):

PR #4649, opened 2025-08-11
Create a centralized `CommandAttributes` and add attributes for DTS setup

MicroFish91:


src/commands/CommandAttributes.ts
```diff
@@ -0,0 +1,34 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type ActivityAttributes } from "@microsoft/vscode-azext-utils";
+
+export class CommandAttributes {
+
```
nturinski, line 9: Nit: Consider localizing user-facing strings. The `description` and `troubleshooting` entries appear user-visible via activity UI; wrapping in `localize()` would keep consistency with the extension’s localization pattern.

src/commands/CommandAttributes.ts
```diff
@@ -0,0 +1,34 @@
...
+    static readonly Deploy: ActivityAttributes = {
```
nturinski, line 10: Question: Would links help? `ActivityAttributes` might benefit from optional doc links per troubleshooting item (e.g., Learn docs for deployment and DTS). A `links?: string[]` or `docLinks?: Record<string, string>` could improve guidance.

src/commands/appSettings/connectionSettings/durableTaskScheduler/getDTSConnection.ts
```diff
@@ -47,6 +48,8 @@ export async function getDTSConnectionIfNeeded(context: DTSConnectionContext, ap
     const wizardContext: IDTSAzureConnectionWizardContext = {
         ...context,
         ...await createActivityContext({ withChildren: true }),
+        activityAttributes: CommandAttributes.DeployDTSSetup,
```
nturinski, line 51: Telemetry correlation: `callbackId: `${context.callbackId}.dtsSetup`` (line 51). Is this suffix convention documented/consumed elsewhere? If other components rely on exact formats, consider centralizing the pattern to avoid drift.

src/commands/deploy/deploy.ts
```diff
@@ -68,25 +69,11 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     }
 
     const context: IFuncDeployContext = Object.assign(actionContext, deployPaths, {
```
nturinski, line 71: Import + duplication: You now call `...await createActivityContext()` (line 71) and assign `activityAttributes`. Previously you spread a second `createActivityContext()` at the end. Verify there isn’t a second spread later (lines ~86–90) to prevent accidentally overwriting attributes.

src/commands/CommandAttributes.ts
nturinski: These are not user interfacing. maybe leave a comment in this file indicating that this is all for copilot?

src/commands/deploy/deploy.ts
nturinski: Don't think that this is a concern.

src/commands/appSettings/connectionSettings/durableTaskScheduler/getDTSConnection.ts
MicroFish91: We could also call with `callWithTelemetryAndErrorHandling`, but I thought that might be overkill when all we need to change is the callback id that shows up in the activity log.  I wanted it to have its own callback id so it's easier for us to log in telemetry when users call into the activity log with copilot.

Issue #4497, opened 2025-05-08
Investigate using core tools to deploy function app to flex consumption


soninaren:
## Issue: Deployment Fails Due to Missing Execute Permissions in Zipped Payload
Currently, Azure Functions Core Tools performs several behind-the-scenes operations when zipping a function app payload for deployment. These operations are primarily aimed at preserving execute permissions for binaries, which are often stripped when zipping on Windows. This logic can be found here:

[ZipHelper.cs (Lines 95–140)](https://github.com/Azure/azure-functions-core-tools/blob/main/src/Cli/func/Helpers/ZipHelper.cs#L95-L140)

This step is critical for Linux-based environments, as missing execute permissions can cause the function host to fail at runtime. As a result, deploying custom handlers to Flex Consumption plans via VS Code can lead to non-functional apps due to these permission issues.

## Proposed Solutions
1. Use Core Tools `func azure functionapp publish {AppName}` for deploying custom handlers to ensure proper file permissions.
2. Implement equivalent logic in the VS Code extension to handle file permissions correctly during the zipping process.

PR #4650, opened 2025-08-11
Deploy with CLI for custom runtimes on flex

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4497


src/commands/deploy/deploy.ts
```diff
@@ -258,8 +258,8 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
         context,
         localize('deploying', 'Deploying...'),
         async () => {
-            // prioritize func cli deployment if installed
-            if (isFuncToolsInstalled) {
+            // deply with func cli for custom runtimes on flex consumption due to additional requirements
+            if (isFuncToolsInstalled && language === ProjectLanguage.Custom && isFlexConsumption) {
```
MicroFish91: Do we need equivalent logic to prevent skipping here as well?

image

nturinski: Skipping what? I think we always want to use the old default deployment method unless these 3 conditions are true.

MicroFish91: Yeah but the way it's written right now, if func cli is installed, it will skip all the steps for setting connection settings etc which isn't the old default deployment logic right?

nturinski: Ah, I see. In the [lines above](https://github.com/microsoft/vscode-azurefunctions/pull/4650/files#diff-ed0d81595c753431c907d0d2a29e9cce17b1e4ebf0af0237c3e059416d6d7b62R230) -- good catch.

Issue #4646, opened 2025-08-10
VS Code startup error: Failed to get latest templates: Cannot read properties of undefined (reading 'resources')


hhubik:
When I open VS Code on an existing C# project, I get the following error in the `Output` window from `Azure Functions`:

`Failed to get latest templates: Cannot read properties of undefined (reading 'resources')`

This particular C# project was last worked on more than a year ago. At that time, there was no error.

Extension Version: 1.18.1

`About VS Code` output:

```
Version: 1.103.0 (user setup)
Commit: e3550cfac4b63ca4eafca7b601f0d2885817fd1f
Date: 2025-08-06T21:40:10.271Z
Electron: 37.2.3
ElectronBuildId: 12035395
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.26100
```

PR #4652, opened 2025-08-11
Use the v1 index.json for all templates not just v2

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4646
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4599


alexweininger (approved):
Can you tl;dr this? For me and future us?

nturinski:
Sure. Basically, at some point, the templates team stopped maintaining the v2 section of the templates feed for index-v2.json. Remember, templates were broken into v1 and v2 (though as far as I know, Python is the only runtime that uses v2. Typescript/Javascript programming model is still hard-coded.)

However, the decision to completely deprecate the index-v2.json was made a few weeks ago, so newer versions (I believe it's at 4.22.0 now) are not being added to the list which was causing the bug.

The templates team still maintains the old index.json where you have to build the urls yourself. I was already using it for the v2 templates, so now I'm just doing it for all the templates.

PR #4653, opened 2025-08-11
Use staticProperties for default version range

nturinski:
Similar with the index-v2.json being discontinued, the defaultVersionRange is also no longer being updated. You may have seen an issue where defaultVersionRange wasn't  being added to host.jsons anymore... Someone actually deleted the property all together, but added it back after sufficient amounts of begging.

### DIFF
--- a/src/utils/bundleFeedUtils.ts
+++ b/src/utils/bundleFeedUtils.ts
@@ -5,7 +5,6 @@
 
 import { type IActionContext } from '@microsoft/vscode-azext-utils';
 import * as semver from 'semver';
-import { ext, TemplateSource } from '../extensionVariables';
 import { type IBundleMetadata, type IHostJsonV2 } from '../funcConfig/host';
 import { localize } from '../localize';
 import { type IBindingTemplate } from '../templates/IBindingTemplate';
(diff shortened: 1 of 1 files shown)


src/utils/bundleFeedUtils.ts
```diff
@@ -108,21 +107,8 @@ export namespace bundleFeedUtils {
...
-        if (!envVarUri && bundleId === defaultBundleId && templateProvider.templateSource !== TemplateSource.Staging) {
-            url = 'https://aka.ms/bundleFeedUtilsV2';
-        } else {
-            const suffix: string = templateProvider.templateSource === TemplateSource.Staging ? '-staging' : '';
-            const baseUrl: string = envVarUri || `https://cdn${suffix}.functions.azure.com/public`;
-            url = `${baseUrl}/ExtensionBundles/${bundleId}/index-v2.json`;
-        }
-
+    async function getBundleFeed(context: IActionContext): Promise<IBundleFeed> {
+        const url: string = 'https://aka.ms/funcStaticProperties';
```
MicroFish91, line 111: How did you know to use this url?

nturinski: Because I created it haha.

Unless you meant where it leads to-- the templates engineer told me that's where they will keep static properties from now on.

PR #4656, opened 2025-08-12
Ensure durable storage type telemetry is set in all workflows

MicroFish91:


src/commands/createFunction/createFunction.ts
```diff
@@ -71,11 +71,13 @@ export async function createFunctionInternal(context: IActionContext, options: a
     }
 
     const { language, languageModel, version, templateSchemaVersion } = await verifyInitForVSCode(context, projectPath, options.language, options.languageModel, options.version);
-    const hasDurableStorage: boolean = await durableUtils.verifyHasDurableStorage(language, projectPath);
+    const durableStorageType = await durableUtils.getStorageTypeFromWorkspace(language, projectPath);
+    context.telemetry.properties.hasDurableStorageProject = String(!!durableStorageType);
+    context.telemetry.properties.durableStorageType = durableStorageType;
```
nturinski, line 76: **Bug: Potential null reference on line 76**

The `durableStorageType` variable can be `undefined` (from `getStorageTypeFromWorkspace`), but it's being assigned directly to `context.telemetry.properties.durableStorageType` without null checking. This could result in `"undefined"` string in telemetry data instead of proper handling.

Consider using: `context.telemetry.properties.durableStorageType = durableStorageType || 'none';`

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
```diff
@@ -36,7 +36,9 @@ export class DurableStorageTypePromptStep<T extends IFunctionWizardContext> exte
                 await openUrl('https://aka.ms/durable-storage-providers');
             }
         }
+
         context.newDurableStorageType = pick;
+        context.telemetry.properties.durableStorageType = pick;
```
nturinski, line 41: Same potential null reference issue here. The `pick` variable (which is `durableStorageType`) could be `undefined`, but it's being assigned directly to telemetry properties without validation.

src/commands/createFunctionApp/FunctionAppHostingPlanStep.ts
```diff
@@ -8,6 +8,7 @@ import { createHttpHeaders, createPipelineRequest } from '@azure/core-rest-pipel
 import { setLocationsTask, WebsiteOS, type IAppServiceWizardContext } from '@microsoft/vscode-azext-azureappservice';
 import { createGenericClient, LocationListStep, type AzExtPipelineResponse, type AzExtRequestPrepareOptions } from '@microsoft/vscode-azext-azureutils';
 import { AzureWizardPromptStep, type IAzureQuickPickItem } from '@microsoft/vscode-azext-utils';
+import { DurableBackend } from '../../constants';
```
nturinski, line 11: **Bug: Unused import**

The `DurableBackend` import is added but the file still only uses it in one location (`context.durableStorageType === DurableBackend.DTS`). However, looking at the previous code, the import was needed because the old logic was `=== DurableBackend.DTS` comparison. This import is correct for this usage.

Actually, reviewing this more carefully - this import is necessary and correctly used. This is not a bug.

src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
```diff
@@ -55,7 +55,8 @@ export async function createCreateFunctionAppComponents(context: ICreateFunction
 
     await detectDockerfile(context);
     if (wizardContext.workspaceFolder) {
-        wizardContext.hasDurableTaskScheduler = await durableUtils.getStorageTypeFromWorkspace(language, wizardContext.workspaceFolder.uri.fsPath) === DurableBackend.DTS;
+        wizardContext.durableStorageType = await durableUtils.getStorageTypeFromWorkspace(language, wizardContext.workspaceFolder.uri.fsPath);
+        wizardContext.telemetry.properties.durableStorageType = wizardContext.durableStorageType;
```
nturinski, line 59: Potential null reference issue: `wizardContext.durableStorageType` could be `undefined` from `getStorageTypeFromWorkspace`, but it's assigned directly to telemetry without null checking. Consider using a fallback value for telemetry consistency.

nturinski (changes requested):
The PR successfully standardizes telemetry collection for durable storage types across all workflows. However, there are several instances where `durableStorageType` could be `undefined` but is being assigned directly to telemetry properties without proper null handling. This could result in inconsistent telemetry data with `"undefined"` strings instead of meaningful values.

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
nturinski: This is a valid concern since telemetry properties need to be strings. But I think setting it to undefined doesn't cause any issues. That being said, would we want to track the label instead since it's more descriptive of what the user picked?

nturinski (approved):
I'll leave it to you if you want to address those concerns. I think they're valid but somewhat nitpicky.

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
MicroFish91: I think it's only an issue if you were to do something like String(pick).  In that case `undefined` becomes `"undefined"`.  I believe the value `undefined` won't be registered as a string otherwise.

PR #4660, opened 2025-08-12
Add warning popup when incompatible function app plan is used for durable projects

MicroFish91:


src/commands/deploy/deploy.ts
```diff
@@ -172,7 +172,9 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 
     if (durableStorageType && durableStorageType !== DurableBackend.Storage && isFlexConsumption) {
         // https://github.com/Azure/azure-functions-durable-extension/issues/2957
-        ext.outputChannel.appendLog(localize('durableStorageTypeWarning', 'Warning: Detected a flex consumption app using an incompatible Durable Functions storage provider. Only the Azure Storage backend is supported for Durable Function apps hosted in the Flex Consumption plan.'));
+        const warning: string = localize('durableStorageTypeWarning', 'Detected a flex consumption app using an incompatible Durable Functions storage provider. Only the Azure Storage backend is supported for Durable Function apps hosted in the Flex Consumption plan.');
```
nturinski: Not so sure about this wording... Maybe something like this?

> This flex consumption app is using an unsupported durable functions storage provider. For a flex consumption plan, durable functions are only supported with an Azure Storage backend."

nturinski: Also maybe consider adding a learn more link?

MicroFish91: I mostly copied the wording from the issue haha

MicroFish91: That sounds good, I copied that last sentence directly from the issue haha

nturinski (approved):

PR #4665, opened 2025-08-13
Add copilot instructions

MicroFish91:
It looks like copilot instructions got deleted at some point so I'm adding it back.

Also, I'm being more explicit with copilot about not editing `main.js` since it sometimes thinks it's still okay to edit `main.js`

https://github.com/microsoft/vscode-azurefunctions/pull/4638#issuecomment-3168980321


nturinski (approved):
Sorry, I deleted it because my giant file accidentally got included into a PR.

Issue #4668, opened 2025-08-14
Fail to create a C#/Python/PowerShell project with Event Hub trigger template selected


v-ruizh:
**OS:** Win 11
**Build Version:** 20250814.2

**Repro Steps:**
1. Create a C# .NET9 project with Event Hub trigger template selected.
2. Check succeeds to create the project or not.

**Expect:**
Succeed to create a C# .NET9 project with Event Hub trigger template selected.

**Actual:**
Fail to create a C# .NET9 project with Event Hub trigger template selected.
<img width="441" height="113" alt="Image" src="https://github.com/user-attachments/assets/f6c2574d-beb7-457c-bf06-c81c8635cce9" />

**More Info:**
This issue also reproduces when creating a Python/PowerShell project with Event Hub trigger template selected.

PR #4674, opened 2025-08-14
fix: improve EventHub wizard flow and resource group retrieval

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4668

Resource group probably was getting set somewhere before, but now it's undefined so it kept throwing the non null error. Getting the rgName from the id instead now.

Also we were getting prompted for EventHub twice. I'm assuming that there are different places these steps are being used (one for Azure related, one for local setting related) but there is a chance that you can be in a scenario where you need both (create a new local setting, and then select an Azure event hub).

Also, the new auth rule list step wasn't setting it as a namespace auth rule (it is doing an iteration on namespaces, not the event hub itself) which was causing an error when trying to get the auth policy later.

### DIFF
--- a/src/commands/appSettings/connectionSettings/netherite/azure/EventHubListStep.ts
+++ b/src/commands/appSettings/connectionSettings/netherite/azure/EventHubListStep.ts
@@ -30,6 +30,7 @@ export class EventHubListStep<T extends INetheriteAzureConnectionWizardContext> 
         if (context.eventHub?.name) {
             context.valuesToMask.push(context.eventHub.name);
             context.newEventHubConnectionSettingValue = context.eventHub.name;
+            context.eventhubname = nonNullProp(context.eventHub, 'name');
         }
     }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
                 "@azure/storage-blob": "^12.5.0",
                 "@microsoft/vscode-azext-azureappservice": "^3.6.4",
                 "@microsoft/vscode-azext-azureappsettings": "^0.2.8",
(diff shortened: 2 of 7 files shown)


src/commands/appSettings/connectionSettings/netherite/azure/EventHubListStep.ts
```diff
@@ -30,6 +30,7 @@ export class EventHubListStep<T extends INetheriteAzureConnectionWizardContext>
         if (context.eventHub?.name) {
             context.valuesToMask.push(context.eventHub.name);
             context.newEventHubConnectionSettingValue = context.eventHub.name;
+            context.eventhubname = nonNullProp(context.eventHub, 'name');
```
nturinski, line 33: **Bug**: Missing null/undefined check for `context.eventHub?.name` before calling `nonNullProp()` on line 33. If `context.eventHub` exists but has a null/undefined `name` property, `nonNullProp()` will throw an error.

Consider using `nonNullValueAndProp(context.eventHub, 'name')` consistently or add a proper null check before the assignment.

nturinski (commented):
The changes improve the EventHub wizard flow by extracting resource group information from namespace IDs rather than requiring separate context properties, which should create a more logical prompt sequence.

src/commands/appSettings/connectionSettings/netherite/azure/EventHubListStep.ts
nturinski: You should check that it's within an if clause which means that this _has_ to be defined. In fact, use of nonNullProp is probably unnecessary in this case.

PR #4680, opened 2025-08-19
Improve separation of concerns for the different event hub steps and contexts

MicroFish91:
Related: https://github.com/microsoft/vscode-azurefunctions/issues/4668

### DIFF
--- a/src/commands/appSettings/connectionSettings/netherite/EventHubsConnectionListStep.ts
+++ b/src/commands/appSettings/connectionSettings/netherite/EventHubsConnectionListStep.ts
@@ -13,6 +13,9 @@ import { localize } from '../../../../localize';
 import { EventHubSetSettingStep } from './EventHubSetSettingStep';
 import { EventHubsNamespaceSetSettingStep } from './EventHubsNamespaceSetSettingStep';
 import { type INetheriteAzureConnectionWizardContext, type INetheriteConnectionWizardContext } from './INetheriteConnectionWizardContext';
+import { EventHubGetConnectionStep } from './azure/EventHubGetConnectionStep';
+import { EventHubListStep } from './azure/EventHubListStep';
+import { EventHubsNamespaceAuthRuleListStep } from './azure/EventHubsNamespaceAuthRuleListStep';
 import { EventHubsNamespaceGetConnectionStep } from './azure/EventHubsNamespaceGetConnectionStep';
 import { EventHubsNamespaceListStep } from './azure/EventHubsNamespaceListStep';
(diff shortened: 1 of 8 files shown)


src/commands/appSettings/connectionSettings/netherite/EventHubsConnectionListStep.ts
```diff
@@ -57,20 +60,25 @@ export class EventHubsConnectionListStep<T extends INetheriteConnectionWizardCon
...
                 }
 
-                LocationListStep.addProviderForFiltering(context as unknown as ILocationWizardContext, EventHubsProvider, EventHubsNamespaceResourceType);
+                // Make sure this is added ahead of time because location might get prompted during `ResourceGroupListStep`
+                LocationListStep.addProviderForFiltering(azureContext as unknown as ILocationWizardContext, EventHubsProvider, EventHubsNamespaceResourceType);
 
                 promptSteps.push(
                     new ResourceGroupListStep() as AzureWizardPromptStep<INetheriteAzureConnectionWizardContext>,
                     new EventHubsNamespaceListStep(),
+                    new EventHubsNamespaceAuthRuleListStep(),
```
MicroFish91, line 69: Moving these steps up one level to make `EventHubsNamespaceListStep` more reuseable

src/commands/appSettings/connectionSettings/netherite/INetheriteConnectionWizardContext.ts
```diff
@@ -18,3 +19,14 @@ export interface INetheriteConnectionWizardContext extends ISetConnectionSetting
 }
 
 export type INetheriteAzureConnectionWizardContext = INetheriteConnectionWizardContext & IEventHubWizardContext;
+
+interface IEventHubWizardContext extends IResourceGroupWizardContext {
```
MicroFish91, line 25: Todo: We should create a common `IEventHubWizardContext` that can be partially reused by both Netherite and event hub function triggers. 

For now, I think we should keep them separate as the steps can't be completely swapped 1:1 without risking accidentally messing something up.  Unfortunately the namespace list step is still being shared so they're somewhat coupled, but not as much as before.

Known steps that we will need to take to allow complete sharing of step logic:
  - Add a shared/consolidated folder with the generalized event hub step logic
  - `eventhubname` binding setting should only be known in the trigger implementation and not referenced in the general implementation
  - `suggestedLocalSettings...` should only be known in the Netherite implementation (potentially we could explore injecting pick preferences into the list steps so that the generic version of the step doesn't need to be aware of any custom context suggestion values but can still display them)

PR #4688, opened 2025-08-21
Enable durable configuration path for python v2

MicroFish91:
I tested a few scenarios, but they were mostly durable, if you have some key scenarios in mind to test, please test or let me know and I will try out testing additional scenarios

### DIFF
--- a/src/commands/createFunction/FunctionListStep.ts
+++ b/src/commands/createFunction/FunctionListStep.ts
@@ -18,8 +18,7 @@ import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/
 import { FunctionSubWizard } from './FunctionSubWizard';
 import { type IFunctionWizardContext } from './IFunctionWizardContext';
 import { JobsListStep } from './JobsListStep';
-import { DurableProjectConfigureStep } from './durableSteps/DurableProjectConfigureStep';
-import { DurableStorageTypePromptStep } from './durableSteps/DurableStorageTypePromptStep';
+import { DurableStorageTypeListStep } from './durableSteps/DurableStorageTypePromptStep';
 
 export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardContext> {
     public hideStepCount: boolean = true;
@@ -71,13 +70,13 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
 
         if (context.functionTemplate?.templateSchemaVersion === TemplateSchemaVersion.v2) {
             return requiresDurableStorageSetup ?
-                { promptSteps: [new JobsListStep(this._isProjectWizard), new DurableStorageTypePromptStep()], executeSteps: [new DurableProjectConfigureStep()] } :
+                { promptSteps: [new DurableStorageTypeListStep(), new JobsListStep(this._isProjectWizard)] } :
                 { promptSteps: [new JobsListStep(this._isProjectWizard)] };
         }
 
         const { promptSteps = [], executeSteps = [] } = await FunctionSubWizard.createSubWizard(context, this._functionSettings, this._isProjectWizard) ?? { promptSteps: [], executeSteps: [] };
         if (requiresDurableStorageSetup) {
-            promptSteps.unshift(new DurableStorageTypePromptStep());
+            promptSteps.unshift(new DurableStorageTypeListStep());
         }
 
         return { promptSteps, executeSteps };
--- a/src/commands/createFunction/FunctionListStep.ts
+++ b/src/commands/createFunction/FunctionListStep.ts
@@ -18,6 +18,7 @@ import { getWorkspaceSetting, updateWorkspaceSetting } from '../../vsCodeConfig/
 import { FunctionSubWizard } from './FunctionSubWizard';
 import { type IFunctionWizardContext } from './IFunctionWizardContext';
 import { JobsListStep } from './JobsListStep';
+import { DurableProjectConfigureStep } from './durableSteps/DurableProjectConfigureStep';
 import { DurableStorageTypePromptStep } from './durableSteps/DurableStorageTypePromptStep';
 
 export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardContext> {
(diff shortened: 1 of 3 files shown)


src/commands/createFunction/FunctionListStep.ts
```diff
@@ -66,17 +66,20 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
     }
 
     public async getSubWizard(context: IFunctionWizardContext): Promise<IWizardOptions<IFunctionWizardContext> | undefined> {
+        const requiresDurableStorageSetup: boolean = durableUtils.requiresDurableStorageSetup(context);
+
         if (context.functionTemplate?.templateSchemaVersion === TemplateSchemaVersion.v2) {
-            return { promptSteps: [new JobsListStep(this._isProjectWizard)] };
+            return requiresDurableStorageSetup ?
+                { promptSteps: [new DurableStorageTypeListStep(), new JobsListStep(this._isProjectWizard)] } :
```
nturinski, line 73: **Bug:** The `requiresDurableStorageSetup` check is moved above the v2 template check, but it's only used in the v2 branch and the non-v2 branch. This variable declaration should be moved to after the v2 template check to avoid unnecessary computation when the early return happens in the v2 case.

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -66,17 +66,20 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
...
+                { promptSteps: [new JobsListStep(this._isProjectWizard)] };
         }
 
-        const requiresDurableStorageSetup: boolean = durableUtils.requiresDurableStorageSetup(context);
+        const { promptSteps = [], executeSteps = [] } = await FunctionSubWizard.createSubWizard(context, this._functionSettings, this._isProjectWizard) ?? { promptSteps: [], executeSteps: [] };
         if (requiresDurableStorageSetup) {
-            return { promptSteps: [new DurableStorageTypePromptStep()] };
```
nturinski, line 75: **Bug:** The import statement on line 21 still references `DurableStorageTypePromptStep` but the class was renamed to `DurableStorageTypeListStep`. However, line 21 shows the corrected import. The actual bug is in line 75 - it still uses `new DurableStorageTypePromptStep()` which should be `new DurableStorageTypeListStep()` to match the renamed class.

Wait, looking at the diff again, I need to check what the current state actually is...

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -66,17 +66,20 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
     }
 
     public async getSubWizard(context: IFunctionWizardContext): Promise<IWizardOptions<IFunctionWizardContext> | undefined> {
+        const requiresDurableStorageSetup: boolean = durableUtils.requiresDurableStorageSetup(context);
```
nturinski, line 69: The movement of `requiresDurableStorageSetup` calculation above the v2 template check looks correct since it's now needed in both branches (v2 and non-v2). The previous logic had this calculation duplicated or positioned incorrectly for the v2 case.

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
```diff
@@ -7,17 +7,10 @@ import { AzureWizardPromptStep, openUrl, type IAzureQuickPickItem, type IWizardO
 import { DurableBackend } from "../../../constants";
 import { defaultDescription, previewDescription } from "../../../constants-nls";
 import { localize } from "../../../localize";
-import { FunctionSubWizard } from "../FunctionSubWizard";
 import { type IFunctionWizardContext } from "../IFunctionWizardContext";
+import { DurableProjectConfigureStep } from "./DurableProjectConfigureStep";
```
nturinski, line 11: **Bug:** Missing import statement. The `DurableProjectConfigureStep` is used on line 42 but there's no import statement for it. You need to add:
```typescript
import { DurableProjectConfigureStep } from "./DurableProjectConfigureStep";
```

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
```diff
@@ -45,7 +38,7 @@ export class DurableStorageTypePromptStep<T extends IFunctionWizardContext> exte
         return !context.hasDurableStorage && !context.newDurableStorageType;
     }
 
-    public async getSubWizard(context: T): Promise<IWizardOptions<T> | undefined> {
-        return await FunctionSubWizard.createSubWizard(context, this._functionSettings);
+    public async getSubWizard(): Promise<IWizardOptions<T> | undefined> {
+        return { executeSteps: [new DurableProjectConfigureStep()] };
```
nturinski, line 42: The logic looks correct - the `getSubWizard` method now directly returns the execute step instead of calling `FunctionSubWizard.createSubWizard`, which is consistent with the architectural change to move durable configuration handling into this step's sub-wizard.

nturinski (commented):
This refactoring looks solid overall. The key architectural change is moving durable storage configuration from `FunctionSubWizard` to the `DurableStorageTypeListStep`'s sub-wizard, which enables the Python v2 template path to properly handle durable storage setup. The class rename and constructor simplification are consistent with this change.

src/commands/createFunction/FunctionListStep.ts
MicroFish91: Incorrect, it's used in all follow-up code-branches and should be checked right away.

MicroFish91: I wonder if it's looking at the wrong commit or something

nturinski: Yeah maybe. I gave it a funky PR link on accident so I wonder if something went wrong there too.

src/commands/createFunction/durableSteps/DurableStorageTypePromptStep.ts
nturinski: Dog it's literally on the line that you're commenting on.

PR #4708, opened 2025-09-07
Fix containerized functions deployment regression along w/ misc. improvements

### LINKED ISSUE #4707: Not being prompted to deploy containerized function app when project has Dockerfile
Was testing out the container app api endpoint with functions and noticed that the prompt to deploy a containerized project was not appearing even when the project had a Dockerfile.  This is likely a regression from last release related to how we check and set the project workspace folder.

MicroFish91:
Fixes #4707 

The logic was skipping entirely when there was already a workspace folder set.  Correct me if I'm wrong, but my assumption is we should always prompt if we find a Dockerfile in the current workspace project.

Other improvements I made:
- Improve workspace folder and root path finding logic
- Prompt the user to choose when more than one Dockerfile is detected
- Add more telemetry


src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -2,66 +2,98 @@
...
+        if (isContainerizedDockerfile) {
+            containerizedDockerfileCount++;
         }
+
+        const relativeDirectory: string = '.' + path.sep + path.relative(nonNullProp(context, 'rootPath'), dockerfile.fsPath);
+        const functionsImage: string = localize('functionsImage', 'Functions Image');
+
+        picks.push({
+            label: path.basename(dockerfile.fsPath),
+            description: isContainerizedDockerfile ? `${relativeDirectory} (${functionsImage})` : relativeDirectory,
```
MicroFish91, line 77: Showing pick description as `(Functions Image)` when we detect the Dockerfile references the Azure Functions base image.  Hopefully this is a useful note to add, if not I'm happy to remove it.

src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -2,66 +2,98 @@
...
+    }
+
+    // ensure container apps extension is installed before proceeding
+    await getAzureContainerAppsApi(context);
 }
 
-export async function detectFunctionsDockerfile(file: string): Promise<boolean> {
-    const content = await AzExtFsExtra.readFile(file);
-    const lines: string[] = content.split('\n');
+async function promptUseContainerImage(context: IActionContext): Promise<boolean> {
```
MicroFish91, line 51: None of this function is new logic, it's basically a 1:1 that got moved into its own function but the diff wasn't smart enough to detect that nothing else changed.

PR #4712, opened 2025-09-10
Fix custom handler runtimes response due to Stacks API regression

nturinski:
Due to a regression in the stacks API, we have to add a queryParam sku that specifies that we want FC1 results.  Without it, Custom runtimes do not show up.

However, when you add the query parameters, for some reason, the SKU property is now an array of _all_ possible SKUs instead of just FC1 (which is the old behavior).

We used to be able to assume the first result, but had to change that behavior.

Old payload:
image

New payload:
image


MicroFish91 (approved):
Weird, do we expect at some point the Stacks API team will want to fix this?  Wondering if there is more we should try to do in anticipation of if they revert back to the behavior they had before the regression

nturinski:
They are planning to fix this, but even if they do, it shouldn't impact the changes that I made.

I guess key word is shouldn't...

PR #4731, opened 2025-10-06
Refactor func CLI pretest logic

MicroFish91:
Addresses two issues:
1. For some reason the `gulp preTest` started breaking a while back, you can see this when running a long running test
https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=12514565&view=logs&j=36b3f9c6-e29e-5b82-0001-f267ab2d3d26&t=6adf503e-47f3-5605-25a5-142045d92ea9
image

2. We have a critical level security dependency flag coming from the `request` package, which I believe has been deprecated a while ago.  This removes that dependency and lets us just rely on the internal node https module instead.

I also leverage the extract package, but we already have that as a dependency, so didn't have to install anything new.

Tested locally on mac, windows, and the long running remote test pipeline and it's passing


gulpfile.ts
```diff
@@ -53,32 +50,93 @@ function matchesCliFeedOS(platform: string) {
     }
 }
 
-function installFuncCli() {
-    const funcDir = path.join(os.homedir(), 'tools', 'func');
+async function cleanReadme() {
```
MicroFish91, line 53: I literally made zero changes to `cleanReadme`.  I think it started seeing this as a change when I reordered the functions

PR #4734, opened 2025-10-07
Fix overwrite of required debugger environment settings

### LINKED ISSUE #4715: Fail to debug Azure Python function project with Local Emulator
**OS:**  Win 11
**Python version:**  3.10.11
**Build Version:** 20250919.1 

**Repro Steps:**
1. Create a new Python function project with http trigger template
2. Press F5 -> Select 'Use Local Emulator' -> Select 'Run Emulator'
3.  Check debug Azure Python function project succeeds or not

**Expect:**
Debug Azure Python function project successfully

**Actual:**
Fail to debug Azure Python function project with an error: "connect ECONNREFUSED 127.0.0.1:9091"
Image

**More Info:**
This issue doesn't reproduce on C#/JavaScript project

### LINKED ISSUE #4719: VS Code does not enter debug mode for Azure Typescript functions project with F5 (no debug toolbar shown)
**OS:** Win 11
**Node version:** 22.19.0
**Build Version:** 20250924.2

**Repro Steps:**
1. Create a new Typescript function project with http trigger template
2. Press F5 -> Select 'Use Local Emulator' -> Select 'Run Emulator'
3. Check debug Azure Python function project succeeds or not

**Expect:**
VS Code enters debug mode and shows the debug toolbar

**Actual:**
The function host starts, but no debug toolbar appears

https://github.com/user-attachments/assets/827ac651-27d0-4859-b648-8871cd7bced4

**More Info:**
This issue doesn't reproduce with debug using 'Connect Azure Storage Account'

MicroFish91:
Fixes #4715
Fixes #4719

We recently started intercepting and adding a custom env variable for `AzureWebJobsStorage`.  This was to better support the fact that users could have custom connection settings for Azurite.  

It looks like in the process of doing this, we accidentally overwrote the environment setting required to properly start the python debugger.  The simple fix here would be to preserve the existing env settings and only change the one setting.

Edit: This affects other runtimes as well

### DIFF
--- a/src/debug/FuncTaskProvider.ts
+++ b/src/debug/FuncTaskProvider.ts
@@ -127,6 +127,7 @@ export class FuncTaskProvider implements TaskProvider {
                 ...options.env,
                 // The purpose of setting this manually is so that we can honor any custom Azurite connection settings that the user may have already set up
                 // See: https://github.com/microsoft/vscode-azurefunctions/pull/4703
+                // Todo: https://github.com/microsoft/vscode-azurefunctions/issues/4735
                 "AzureWebJobsStorage": azureWebJobsStorage
             };
         }


MicroFish91:
@motm32 When you get a chance, can you double check if this fixes it for you as well?

I'm also considering adding another small change where we only set the new emulator string if it deviates from the default.  Although it's mostly harmless to replace the emulator string each time, I think it likely makes users feel uneasy when they see this message from the func CLI:

`Skipping 'AzureWebJobsStorage' from local settings as it's already defined in current environment variables.`

bwateratmsft (dismissed):

MicroFish91:
Sorry to stale, wanted to add a todo issue link as well

motm32:
> @motm32 When you get a chance, can you double check if this fixes it for you as well?
>

Can confirm it also works for me.

PR #4765, opened 2025-10-24
Implementation for debug-isolated flag and streaming func CLI output

### LINKED ISSUE #4746: Move pickProcess to file-based discovery
Sometimes, early breakpoints for .NET Functions projects (e.g., in `Program.cs`) are missed in debug sessions. This is due to the way the existing `pickProcess` function works.

VSC will poll the host process spun up by Core Tools waiting for a desired response from the status endpoint, which would happen after worker start. It then scans the process tree to get the deepest child process (which is itself somewhat fragile). As a result, some amount of execution of the worker occurs prior to attach.

The Core Tools has a `--dotnet-isolated-debug` flag which will cause the process to wait super early, only to proceed once a debugger is attached. You can add this to the Core Tools start task in `tasks.json`, but this wait occurs before the status endpoint would return, so you'll just hit timeouts.

We take an alternative approach in Visual Studio. The launch includes `--dotnet-isolated-debug` as well as `--enable-json-output --json-output-file <tempFile>`. With this configuration, before the worker starts its wait, it will write out a JSON payload to the temp file, noting its process ID. Here is an example taken from a recent run:

```json
{ "name":"dotnet-worker-startup", "workerProcessId" : 32496 }
```

The component that launches the Core Tools process can decide on the specific location and then watch for it to be updated. This ensures that it can get the information, but that the customer code isn't engaged until the debugger is present, allowing early breakpoints to be hit. We should use this model for Visual Studio Code.

nturinski:
Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4746

This pull request introduces support for streaming terminal output from Azure Functions host tasks, enabling consumers of the API to access real-time output from running processes. 

~~It adds a new async string stream utility, integrates this stream into the function task lifecycle, and exposes the stream through the API.~~

Additionally, it improves handling for .NET isolated debug scenarios by extracting process IDs from terminal output if the flag is set. 

Minor updates were made to the package configuration to support these features, including func tasks now support the `args` property since --debug-isolate is ugly to write all in the command.

~~**Configuration Updates**~~

~~* Added the `enabledApiProposals` section to `package.json` to enable the `terminalDataWriteEvent` API, which is required for capturing terminal output.~~

EDIT: To read the stream, you can run this snippet:

```
for await (const chunk of runningFuncTaskItem.stream) {
                    // Process each chunk of the stream
                    console.log(chunk);
                }
```


src/commands/pickFuncProcess.ts
```diff
@@ -182,6 +206,23 @@ async function startFuncTask(context: IActionContext, workspaceFolder: vscode.Wo
     }
 }
 
+async function setEventPidByJsonOutput(taskInfo: IRunningFuncTask): Promise<number | undefined> {
```
MicroFish91: Nit: Should the name be something more like `getWorkerPidFromJsonOutput`?

src/funcCoreTools/funcHostTask.ts
```diff
@@ -86,13 +90,32 @@ export function isFuncHostTask(task: vscode.Task): boolean {
     return /func (host )?start/i.test(commandLine || '');
 }
 
+let latestTerminalShellExecutionEvent: vscode.TerminalShellExecutionStartEvent | undefined;
 export function registerFuncHostTaskEvents(): void {
     registerEvent('azureFunctions.onDidStartTask', vscode.tasks.onDidStartTaskProcess, async (context: IActionContext, e: vscode.TaskProcessStartEvent) => {
+        const terminalEventReader = vscode.window.onDidStartTerminalShellExecution(async (terminalShellExecEvent) => {
```
MicroFish91: Had some questions about this part, reaching out offline

src/commands/pickFuncProcess.ts
nturinski: Yeah, good call. The implementation used to set it within the function and I forgot to change the name.

src/funcCoreTools/funcHostTask.ts
nturinski: Talked offline; registering the event handler outside of the `onDidStartTaskProcess`

src/commands/debug/PostFuncDebugExecuteStep.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    public constructor(readonly logs: string[]) {
+        super();
+    }
+
+    public async execute(context: T): Promise<void> {
+        const errorLogs: string[] = [];
+        // eslint-disable-next-line no-control-regex
+        const redAnsiRegex = /\x1b\[(?:[0-9;]*31m|[0-9;]*91m|38;5;(9|1)m)/;
+        const functionErrors = [
```
nturinski: Not sure if we really need to check for common errors. The redAnsi seems to catch most of it and it's expensive to `array.some` every line of the log.

src/commands/pickFuncProcess.ts
```diff
@@ -182,6 +206,23 @@ async function startFuncTask(context: IActionContext, workspaceFolder: vscode.Wo
     }
 }
 
+async function getWorkerPidFromJsonOutput(taskInfo: IRunningFuncTask): Promise<number | undefined> {
```
nturinski: Apparently the preferred way to do this is with file-based since the output can have conflicts.

nturinski:
@copilot resolve the merge conflicts in this pull request

MicroFish91 (approved):

Issue #4749, opened 2025-10-15
Self-hosted MCP: Create experience


lilyjma:
## Ask summary 
For the **Azure Functions: Create new project** command, create a new project type called "MCP Custom Handler" and remove "Ballerina":

<img width="1004" height="442" alt="Image" src="https://github.com/user-attachments/assets/3bfa03de-072c-4a7b-851d-1bbb04a590fe" />

When "MCP Custom Handler" is picked, prompt the user with the following:
1. Select server language: **Python, TypeScript, .NET**
2. Include sample server code: **Yes, No**

### Include sample server code is: No
1. _host.json_
  ```json
  {
      "version": "2.0",
      "configurationProfile": "mcp-custom-handler",
      "customHandler": {
          "description": {
              "defaultExecutablePath": "",
              "arguments": [""] 
          }
      }
  }	
  ```
2. _local.settings.json_, `FUNCTIONS_WORKER_RUNTIME` depends on the server language
  ```json
  {
      "IsEncrypted": false,
      "Values": {
          "FUNCTIONS_WORKER_RUNTIME": "<python|node|dotnet-isolated>",
          "AzureWebJobsFeatureFlags": "EnableMcpCustomHandlerPreview”
      }
  }
  ```

### Include sample server code is: Yes

1. Depending on the language, include simple MCP server snippet. For example, for Python, a _server.py_
  ```
  # TODO: provide different language's code snippet
  ```
2. Fill in the values for properties `defaultExecutablePath` and `arguments` in _host.json_: 
    | Language | `defaultExecutablePath` | `arguments` |
    |------|--------|-----|
    | Python | `python` | `["server.py"]` | 
    | TypeScript | `npm` | `["run", "start"]` | 
    | .NET | `dotnet`| `["server.dll"]`|
3. _local.settings.json_ should be same as above, with the right `FUNCTIONS_WORKER_RUNTIME` 
4. Python also requires a _pyproject.toml_ file with the following content: 
  ```toml
      [project]
      name = "you-mcp-project-name"
      version = "0.1.0"
      description = "Add your description here"
      readme = "README.md"
      requires-python = ">=3.10"
      dependencies = [
          "mcp[cli]>=1.14.1",
      ]
  ```

PR #4789, opened 2025-11-13
Support for creating self-hosted MCP servers

nturinski:
Use this repository to follow along: https://github.com/Azure-Samples/mcp-sdk-functions-hosting-node

The idea of the MCP in Functions is pretty convoluted, but I'll try my best to explain it here concisely and then explain the changes in this pull request.

All other PRs build off of this one. Don't mind the build errors, I didn't import changes I made in a file because those were more deploy related. When all 3 PRs (https://github.com/microsoft/vscode-azurefunctions/pull/4789, https://github.com/microsoft/vscode-azurefunctions/pull/4790. https://github.com/microsoft/vscode-azurefunctions/pull/4791) are merged, everything should be fine.

There are two ways to have MCP tools in Azure Functions: 
- Using the MCP extension (it's not a VS Code extension, but an optional extension that the function host runtime will download when it sees it defined in the host.json)
- Using self-hosted MCP servers

Like how it sounds, self-hosted MCP servers are function apps that behave like a remote MCP server. In order to configure the host runtime, there a few things that you have to set in the host.json, but there's a shortcut-- set the `configurationProfile` to equal `-mcp-custom-handler`. The host will see that and set all the defaults required to run the server.

Whenever we create the project, we set the workspace setting to indicate that this is a `SelfHostedMcpServer`. That comes into play in the deploy scenario.

You also need to have a feature flag enabled, `EnableMcpCustomHandlerPreview`. This flag is _not_ required for MCP extension function apps.

Don't mind the `mcpGetHostKey`. That is for the deployment piece, but it was defined in the package.json so I just left it in.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,7 +1,6 @@
 {
     "recommendations": [
         "dbaeumer.vscode-eslint",
-        "ms-vscode.azure-account",
         "ms-azuretools.vscode-azureresourcegroups"
     ]
 }
--- a/resources/backupTemplates/nodejs-4.x/templates/templates.json
+++ b/resources/backupTemplates/nodejs-4.x/templates/templates.json
@@ -619,7 +619,7 @@
         "id": "McpToolTrigger-JavaScript-4.x",
         "runtime": "2",
         "files": {
-            "%functionName%.js": "import { app, InvocationContext, arg } from \"@azure/functions\";\n\nexport async function mcpToolHello(_toolArguments: any, context: InvocationContext): Promise<string> {\n    const mcptoolargs = _toolArguments.arguments as {\n        name?: string;\n    };\n    const name = mcptoolargs?.name;\n\n    console.info(`Hello ${name}, I am MCP Tool!`);\n\n    return `Hello ${name}, I am MCP Tool!`;\n}\n\napp.mcpTool('hello', {\n    toolName: 'hello',\n    description: 'Simple hello world MCP Tool that responses with a hello message.',\n    toolProperties: {\n      name: arg.string().describe('Name to greet'),\n    },\n    handler: mcpToolHello\n});"
+            "%functionName%.js": "const { app, arg } = require('@azure/functions');\n\nfunction mcpToolHello(_toolArguments, context) {\n    const mcptoolargs = context.triggerMetadata.mcptoolargs || {};\n    const name = mcptoolargs.name;\n    console.info(`Hello ${name}, I am MCP Tool!`);\n    return `Hello ${name}, I am MCP Tool!`;\n}\n\napp.mcpTool('hello', {\n    toolName: 'hello',\n    description: 'Simple hello world MCP Tool that responses with a hello message.',\n    toolProperties:{\n        name: arg.string().describe('Name to greet'),\n    },\n    handler: mcpToolHello\n});"
         },
         "metadata": {
             "defaultFunctionName": "mcpToolTrigger",


package.json
```diff
@@ -1365,6 +1369,25 @@
                         "type": "boolean",
                         "description": "%azureFunctions.allowProgrammingModelSelection%",
                         "default": false
+                    },
+                    "azureFunctions.showBallerinaProjectCreation": {
```
MicroFish91, line 1373: Were the ballerina changes intentional?

package.nls.json
```diff
@@ -138,5 +138,10 @@
     "azureFunctions.durableTaskScheduler.enablePreviewFeatures": "Enable Durable Task Scheduler preview features",
     "azureFunctions.durableTaskScheduler.emulatorRegistry": "The registry of the Durable Task Scheduler emulator image.",
     "azureFunctions.durableTaskScheduler.emulatorImage": "The name of the Durable Task Scheduler emulator image.",
-    "azureFunctions.durableTaskScheduler.emulatorTag": "The tag of the Durable Task Scheduler emulator image."
+    "azureFunctions.durableTaskScheduler.emulatorTag": "The tag of the Durable Task Scheduler emulator image.",
+    "azureFunctions.showBallerinaProjectCreation": "Show the option to create a Ballerina Function projects when creating a new Function project.",
+    "azureFunctions.getMcpHostKey": "Retrieve the MCP host key for the Function App specified by the resource ID.",
+    "azureFunctions.mcpProjectType.NoMcpServer": "Runs the standard Azure Functions runtime with no MCP integration.",
+    "azureFunctions.mcpProjectType.McpExtensionServer": "Runs the Functions host with an embedded MCP server provided by the Azure Functions MCP extension.",
```
MicroFish91: Is it worth clarifying that this is a Core Tools embedded MCP extension?

src/commands/createFunction/FunctionListStep.ts
```diff
@@ -118,12 +118,13 @@ export class FunctionListStep extends AzureWizardPromptStep<IFunctionWizardConte
...
         }
-
         context.telemetry.properties.templateFilter = templateFilter;
     }
 
     public shouldPrompt(context: IFunctionWizardContext): boolean {
-        return !context.functionTemplate && context['buildTool'] !== JavaBuildTool.maven;
+        return !context.functionTemplate &&
+            context['buildTool'] !== JavaBuildTool.maven &&
+            context.language !== ProjectLanguage.SelfHostedMCPServer;
```
MicroFish91, line 127: It feels too generous to call an MCP server a project language instead of maybe a project type, although I see the project flow becomes difficult if you don't do it this way (also could introduce a lot of tangential changes).  

I still think it could be worth leaving an issue or todo for us to revisit this though, when we have more time to consider all the implications of lumping these together.

src/commands/createNewProject/NewProjectLanguageStep.ts
```diff
@@ -44,18 +49,22 @@ export class NewProjectLanguageStep extends AzureWizardPromptStep<IProjectWizard
...
+        }
+
         if (context.languageFilter) {
             languagePicks = languagePicks.filter(p => {
                 return context.languageFilter?.test(p.data.language);
             });
         }
 
-        const options: QuickPickOptions = { placeHolder: localize('selectLanguage', 'Select a language') };
+        const options: QuickPickOptions = { placeHolder: localize('selectProjectType', 'Select a project type') };
```
MicroFish91: Based on this, perhaps we could eventually changes variable and enum names to reflect project type or similar

src/commands/createNewProject/mcpServerSteps/MCPDownloadSnippetsExecuteStep.ts
```diff
@@ -0,0 +1,68 @@
...
+import { feedUtils } from '../../../utils/feedUtils';
+import { nonNullProp } from '../../../utils/nonNull';
+import { parseJson } from '../../../utils/parseJson';
+import { requestUtils } from '../../../utils/requestUtils';
+import { type MCPProjectWizardContext } from '../IProjectWizardContext';
+
+export class MCPDownloadSnippetsExecuteStep extends AzureWizardExecuteStepWithActivityOutput<MCPProjectWizardContext> {
+    stepName: string = 'MCPDownloadSampleCodeExecuteStep';
+    protected getTreeItemLabel(context: MCPProjectWizardContext): string {
+        return localize('downloadSampleCode', 'Downloading {0} sample server code', nonNullProp(context, 'serverLanguage'));
```
MicroFish91, line 20: Nit: Present case for tree item labels

package.json
nturinski: Yeah, if you look at this issue: https://github.com/microsoft/vscode-azurefunctions/issues/4749, Lily asked that I remove Ballerina from the project runtimes.

But I didn't have the heart to completely rip out Ballerina since it was a significant open-source contribution. I hid it behind a VS Code setting that users can enable if they need it.

But I did remove the tests for Ballerina since we don't officially support it anymore.

src/commands/createFunction/FunctionListStep.ts
nturinski: Unfortunately, and I can explain this more tomorrow offline, self-hosted MCP servers are _completely_ handled differently by the runtime.  I think the biggest callout is that underneath the hood, the runtime is actually a "custom handler" type which is used to run any server process that you want.

You typically wouldn't want a self-hosted MCP project to be included with a normal functions runtime.

src/commands/createNewProject/mcpServerSteps/MCPDownloadSnippetsExecuteStep.ts
nturinski: Will put up a different PR since changing it here will have a bunch of annoying cascading effects.

PR #4791, opened 2025-11-13
Support deploying MCP extension triggers and self-hosted MCP servers

nturinski:
The deployment scenario applies to both MCP extension triggers and self-hosted MCP servers with some small nuanced differences.

The first change is for finding the subscription from a Azure resource ID. But if we can't find it for some reason, I fall back to prompting the user for it.

We used to not have to check any app settings for flex, but since this is only on flex and we need the feature flag, I had to change that behavior.

On deployment, there's a new notification message that has embedded links. Those are next steps that help explain easy auth (using Entra ID) and something foundry related. I didn't dig too deep into these, but Lily wanted to guide users to these docs.

There's also a connect server button.
image

The remote server is automatically added to the `mpc.json` (if it doesn't already exist). The connect server button is there to alert the user and opens the `mcp.json` for them.

When checking for feature flags, I'm comparing what's in the app settings and what's in the local.settings.json. I just make a set out of them to remove duplicates and then re-join the flags. (Flags are comma-delimited lists, so it looks like `flag1, flag2, flag3, etc.`)

`getMcpHostKey` is an internal command that will get an app key from the function app. If you were paying close attention, you would have seen it got defined in the first PR and it's part of the remote server definition. The remote server definition has an input, and then that input is defined as a `command`. That `command` is `getMcpHostKey` with 2 arguments, `resourceId` and `projectType`. (Learn more [here](https://code.visualstudio.com/docs/reference/variables-reference#_input-variables))

The point of the input is so that we can programmatically grab the host key for the function app without having to hard-code it into their `mcp.json` which may go up into SCM.

The keys differ slightly depending on the scenario:
- For self-hosted MCP servers, we can rely on the `default` key that every function app has.
- For MCP extension servers, we need to use a system key called `mcp_extension` This key gets created by the runtime _if_ the host.json has the `extensions.mcp` property defined that I discuseed in #4970.

I have a work item to throw an error if we can't find the system key because falling back to the default key doesn't work. If the system key isn't there, it means that the func host never saw the `extensions.mcp` property. This happened to me because while my deployment said succeeded, the app logs said it failed to run because I had two identical IDs for my mcp tool triggers.

Lastly, the changes in the host.json definition to support all of these changes.

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -21,7 +21,6 @@ import { createActivityContext } from '../../utils/activityUtils';
 import { dotnetUtils } from '../../utils/dotnetUtils';
 import { durableUtils } from '../../utils/durableUtils';
 import { isPathEqual } from '../../utils/fs';
-import { isMcpProject } from '../../utils/mcpUtils';
 import { treeUtils } from '../../utils/treeUtils';
 import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
 import { verifyInitForVSCode } from '../../vsCodeConfig/verifyInitForVSCode';


src/commands/deploy/deploy.ts
```diff
@@ -234,16 +235,15 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
             await updateWorkerProcessTo64BitIfRequired(context, siteConfig, site, language, durableStorageType);
         }
 
-        // app settings shouldn't be checked with flex consumption plans
-        if (isZipDeploy && !isFlexConsumption) {
```
MicroFish91, line 238: Any concern that removing the `!isFlexConsumption` check could introduce breaks to other project deployment types?

src/commands/deploy/deploy.ts
```diff
@@ -62,7 +63,7 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     const deployPaths: IDeployPaths = await getDeployFsPath(actionContext, arg1);
 
     addLocalFuncTelemetry(actionContext, deployPaths.workspaceFolder.uri.fsPath);
-
+    await isMcpProject(deployPaths.workspaceFolder);
```
MicroFish91: Why is this being called without the value being stored or checked?  Are we calling this function for its side effects but not for its return value?

MicroFish91 (dismissed):

src/commands/deploy/deploy.ts
nturinski: Always, but I wrapped the checks with an if/else clause inside `verifyAppSettings` that should preserve the old behavior.

src/commands/deploy/deploy.ts
nturinski: Oh good catch. It was originally for a side effect (a prompt to configure the project if we thought it was an mcp project but wasn't marked as such) but Lily decided against the prompt.

I'll remove it.

nturinski:
```suggestion
```

PR #4797, opened 2025-11-15
Replace C# boilerplate file names with project name

nturinski:
The repo has some boiler plate names like dotnet.csproj. Lily asked me to replace that with the name of their workspace's name. Also refactored downloadFile to be reused.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
                 "@azure/core-client": "^1.7.3",
                 "@azure/core-rest-pipeline": "^1.11.0",
                 "@azure/storage-blob": "^12.5.0",
-                "@microsoft/vscode-azext-azureappservice": "^3.6.7",
+                "@microsoft/vscode-azext-azureappservice": "^3.6.9",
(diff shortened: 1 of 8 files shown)


MicroFish91 (approved):

src/commands/createNewProject/mcpServerSteps/MCPDownloadSnippetsExecuteStep.ts
```diff
@@ -65,4 +60,24 @@ export class MCPDownloadSnippetsExecuteStep extends AzureWizardExecuteStepWithAc
     public shouldExecute(context: MCPProjectWizardContext): boolean {
         return context.includeSnippets === true;
     }
+
+    public static async downloadSingleFile(context: MCPProjectWizardContext, item: GitHubFileMetadata, dirPath: string): Promise<void> {
```
MicroFish91, line 64: Nit: Consider a more stripped down context type so it's easier to quickly parse what's required to be set on the context (if any) since this is now externally exposed as a static method

PR #4861, opened 2025-12-08
Add error message when deploying to flex apps with python 3.14

gavin-aguiar:
This PR adds validation to prevent deploying Python 3.14 function apps to Flex Consumption plans, as remote build is not currently supported for this configuration.

Changes

- Added a check in the deployment flow to detect when a Python 3.14 app is being deployed to a Flex Consumption function app
- The validation reads the [functionAppConfig.runtime.version](vscode-file://vscode-app/c:/Users/gaaguiar/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) property from the remote site configuration
- When Python 3.14 on Flex is detected, the deployment is blocked with a clear error message
- The error message includes a reference link ([https://aka.ms/py314-remote-build-flex](vscode-file://vscode-app/c:/Users/gaaguiar/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)) for more information

### DIFF
--- a/src/commands/deploy/deploy.ts
+++ b/src/commands/deploy/deploy.ts
@@ -144,6 +144,16 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 
     const isFlexConsumption: boolean = await client.getIsConsumptionV2(actionContext);
     actionContext.telemetry.properties.isFlexConsumption = String(isFlexConsumption);
+
+    if (language === ProjectLanguage.Python && isFlexConsumption) {
+        const runtimeVersion = site.rawSite.functionAppConfig?.runtime?.version;
+        if (runtimeVersion === '3.14') {
+            const errorMessage = localize('python314FlexError', 'Remote build for Python 3.14 is not supported for flex. Please refer to https://aka.ms/py314-remote-build-flex.');
+            ext.outputChannel.appendLog(errorMessage);
+            throw new Error(errorMessage);
+        }
+    }
+
     // don't use remote build setting for consumption v2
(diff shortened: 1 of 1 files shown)


src/commands/deploy/deploy.ts
```diff
@@ -144,6 +144,16 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
 
     const isFlexConsumption: boolean = await client.getIsConsumptionV2(actionContext);
     actionContext.telemetry.properties.isFlexConsumption = String(isFlexConsumption);
+
+    if (language === ProjectLanguage.Python && isFlexConsumption) {
+        const runtimeVersion = site.rawSite.functionAppConfig?.runtime?.version;
+        if (runtimeVersion === '3.14') {
+            const errorMessage = localize('python314FlexError', 'Remote build for Python 3.14 is not supported for flex. Please refer to https://aka.ms/py314-remote-build-flex.');
+            ext.outputChannel.appendLog(errorMessage);
+            throw new Error(errorMessage);
```
MicroFish91, line 153: Should we just issue a warning instead?

My thoughts are if we throw an error, this will block deployments no matter what.  However, in the linked issue it says:

`We are currently targeting December - January for this scenario to be supported.`

With the warning modal it wouldn't be blocking, so we will at least have a workaround for users so that they can continue deployments before we issue a release that removes this guard once the support is added.

gavin-aguiar: We considered making this only a warning. However, warning logs often get buried among other logs. The deployment then continues, and the last message users see is, "No Functions Found," which can be confusing.
Throwing an error solves this. It stops execution and triggers a pop up error message. The message also includes actionable steps users can take.

PR #4862, opened 2025-12-08
Allow flex consumption deployments for DTS but keep a warning for SQL

MicroFish91:
Also snuck in a small tangential change when building the DTS connection string


src/commands/deploy/deploy.ts
```diff
@@ -162,6 +162,12 @@ async function deploy(actionContext: IActionContext, arg1: vscode.Uri | string |
     const durableStorageType: DurableBackend | undefined = await durableUtils.getStorageTypeFromWorkspace(language, context.projectPath);
     context.telemetry.properties.durableStorageType = durableStorageType;
 
+    if (durableStorageType === DurableBackend.SQL && isFlexConsumption) {
```
MicroFish91, line 165: Just keep the warning for SQL.  Noticed the SQL flex consumption deployments aren't working so I'm verifying with Lily to make sure this is supposed to be supported / if there's any additional work we need to do to be able to deploy for flex.

PR #4863, opened 2025-12-09
Allow injecting available hosting plans into `FunctionAppHostingPlanStep`. Update DTS plans to include `Flex Consumption`

MicroFish91:
I added flex consumption to the list of available DTS plans.  This mostly affects the Advanced Create flow.

I also changed the behavior of `FunctionAppHostingPlanStep` a bit.  I like the idea of injecting the available plans in as a dependency as it helps uncouple a lot of the more command-specific logic and helps with separation of concerns.  This gives each command a chance to customize its own availability logic without having to worry about how it's going to affect other commands since this step can be shared anywhere.

### DIFF
--- a/src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
+++ b/src/commands/createFunctionApp/createCreateFunctionAppComponents.ts
@@ -183,14 +183,21 @@ async function createContainerizedFunctionAppWizard(): Promise<{ promptSteps: Az
 function getAvailableFunctionAppHostingPlans(context: IFunctionAppWizardContext): Set<FunctionAppHostingPlans> {
     const availablePlans: Set<FunctionAppHostingPlans> = new Set();
 
-    if (context.useFlexConsumptionPlan) {
-        availablePlans.add(FunctionAppHostingPlans.Flex);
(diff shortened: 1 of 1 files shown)


src/commands/createFunctionApp/FunctionAppHostingPlanStep.ts
```diff
@@ -39,15 +57,7 @@ export class FunctionAppHostingPlanStep extends AzureWizardPromptStep<IFunctionA
...
-        if (context.durableStorageType === DurableBackend.DTS) {
-            // premium is required for DTS
-            if (context.advancedCreation) {
-                // allows users to select/create a Elastic Premium plan
-                context.planSkuFamilyFilter = premiumSkuFilter;
-            } else {
-                setPremiumPlanProperties(context);
-            }
-        } else if (context.useFlexConsumptionPlan) {
+        if (context.useFlexConsumptionPlan) {
```
MicroFish91, line 60: I think there could be a better way to handle this now but I don't want to explore that in this PR.  Basically, I think it would be better if we removed this and relied on the injected available plans to only show flex.  Then we can auto-select if the pick was just the one pick.  Prompt logic would then handle setting the related plan properties.

Issue #4911, opened 2026-02-06
There is no response when executing "Create New Project" or "Create Function App" command


v-ruizh:
**OS:** Win11
**Build Version:** [1.20.4-alpha.0](https://github.com/microsoft/vscode-azurefunctions/actions/runs/21718809277)
**Regression:** Previous Build

**Repro Steps:**
1. F1 -> Select "Azure Functions: Create New Project...".
2. Check whether the "Select Folder" dialog appears or not.

**Expect:**
The "Select Folder" dialog appears.

**Actual:**
There is no response. 

**More Info:**
This same issue also reproduces when creating a function app or function.

PR #4914, opened 2026-02-09
Fix no response when executing "Create New Project" command

motm32:
~Update: I got it working a couple times but then it stopped working again. Must be a race condition or something. Will do some more investigating.~

Fixes https://github.com/microsoft/vscode-azurefunctions/issues/4911.

Not sure exactly why this issue was only happening on prod but I tested the new vsix and it seemed to work for me with this fix.

### DIFF
--- a/src/utils/cpUtils.ts
+++ b/src/utils/cpUtils.ts
@@ -70,14 +70,6 @@ export namespace cpUtils {
         const stdoutIntermediate = new Stream.PassThrough();
         const stderrIntermediate = new Stream.PassThrough();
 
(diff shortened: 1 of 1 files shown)


MicroFish91 (approved):
Alternative suggestion: I wonder if it's simpler to remove the combined check and setting up of the on-end logic and instead just end them all in finally, then you'd only have 4 lines of code additions:

```typescript
// ...
finally {
stdoutIntermediate.end();
stderrIntermediate.end();
stdoutFinal.end();
stdoutAndErrFinal.end();
// ...
```

PR #4926, opened 2026-03-03
Add tests for domain name label scopes, fix order of any existing tests

MicroFish91:
Test changes associated with this PR:  https://github.com/microsoft/vscode-azurefunctions/pull/4909

### DIFF
--- a/test/nightly/domainNameLabelScope.test.ts
+++ b/test/nightly/domainNameLabelScope.test.ts
@@ -107,7 +107,7 @@ suite('Domain Name Label Scope', function (this: Mocha.Suite): void {
         const testInputs: (string | RegExp)[] = [
             location,                                   // Location
             rgName,                                     // Use existing RG
-            /global.*default|global/i,                  // Domain scope (Global)
+            /global.*default|legacy/i,                  // Domain scope (Legacy)
             appName,                                    // App name
             'Flex Consumption',                         // Hosting plan
             /\.net/i,                                   // Stack


src/commands/createFunctionApp/containerImage/detectDockerfile.ts
```diff
@@ -5,27 +5,14 @@
...
+import { RelativePattern, workspace, type MessageItem, type Uri } from 'vscode';
 import { browseItem, dockerfileGlobPattern } from '../../../constants';
 import { getAzureContainerAppsApi } from '../../../getExtensionApi';
 import { localize } from '../../../localize';
 import { type ICreateFunctionAppContext } from '../../../tree/SubscriptionTreeItem';
-import { getRootWorkspaceFolder } from '../../../utils/workspace';
 import { tryGetFunctionProjectRoot } from '../../createNewProject/verifyIsProject';
 
 export async function detectDockerfile(context: ICreateFunctionAppContext): Promise<string | undefined> {
-    if (!workspace.workspaceFolders?.length) {
```
MicroFish91, line 17: I think we actually already prompt for this much earlier on, so we should just skip rather than prompt again

test/nightly/functionAppOperations.test.ts
```diff
@@ -40,11 +39,28 @@ suite('Function App Operations', function (this: Mocha.Suite): void {
         saName = getRandomHexString().toLowerCase(); // storage account must have lower case name
         aiName = getRandomHexString();
         miName = getRandomHexString();
-        location = getRotatingLocation();
+        location = 'West US 2';
```
MicroFish91, line 42: Changes to this file will not fix any existing broken tests, it just more correctly orders the prompt inputs.  Basically, with the domain name label scope changes, location & resource group need to come before the site name.  Rest should stay the same.

I had Copilot do this for me and it put each input on a new line.  I kind of like the way it looks so I kept it.

PS: I don't like rotating locations :)

motm32 (approved):

PR #4959, opened 2026-04-03
Add a setting that allows users to bypass emulator validation

MicroFish91:
Users may already have their own tasks set up to launch all required emulators. In those cases, our extension can still prompt them to start emulators manually, which is kind of annoying. I'm seeing this with Copilot on Rails local-dev testing, where the task chain eventually starts the emulators, but they aren’t running yet during pre-validation and so the extension's pre-validation check triggers unwanted modals to still appear.

### DIFF
--- a/src/debug/validatePreDebug.ts
+++ b/src/debug/validatePreDebug.ts
@@ -38,6 +38,9 @@ export async function preDebugValidate(actionContext: IActionContext, debugConfi
     let shouldContinue: boolean;
     context.telemetry.properties.debugType = debugConfig.type;
 
+    const validateEmulators: boolean = !!getWorkspaceSetting<boolean>('validateEmulators', workspace.uri.fsPath);
+    context.telemetry.properties.validateEmulators = String(validateEmulators);
+
     try {
         context.telemetry.properties.lastValidateStep = 'funcInstalled';
         const message: string = localize('installFuncTools', 'You must have the Azure Functions Core Tools installed to debug your local functions.');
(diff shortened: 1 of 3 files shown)


src/debug/getPreLaunchTaskChain.ts
alexweininger, line 1: This function seems to be a great candidate for unit tests

src/debug/validatePreDebug.ts
```diff
@@ -62,28 +71,30 @@ export async function preDebugValidate(actionContext: IActionContext, debugConfi
...
+                        case DurableBackend.Storage:
+                        default:
+                    }
+
+                    context.telemetry.properties.lastValidateStep = 'azureWebJobsStorage';
+                    await validateAzureWebJobsStorage(context, context.projectPath);
+
+                    context.telemetry.properties.lastValidateStep = 'emulatorRunning';
+                    shouldContinue = hasEmulatorTask || await validateEmulatorIsRunning(context, context.projectPath);
                 }
```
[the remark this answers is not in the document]
MicroFish91: This is an intentional choice.  If we see users are choosing to automate setup of emulators on their own in the `preLaunchTask` chain, I don't think we should attempt any related actions that could affect their success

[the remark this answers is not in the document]
MicroFish91: This is intentional

src/debug/getPreLaunchTaskChain.ts
```diff
@@ -0,0 +1,50 @@
...
+        const dependsOn: unknown = task?.dependsOn;
+        if (typeof dependsOn === 'string') {
+            getDependentTasks(dependsOn);
+        } else if (Array.isArray(dependsOn)) {
+            for (const dep of dependsOn) {
+                if (typeof dep === 'string') {
+                    getDependentTasks(dep);
+                }
+            }
+        }
```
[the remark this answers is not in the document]
MicroFish91: Tests are setup and follow the currently expected `ITask` interface.

MicroFish91:
We may want to wait on merging this until after the next release

test/getPreLaunchTaskChain.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        const tasks: ITask[] = [
+            { type: 'shell', label: 'build', command: 'npm run build' }
+        ];
+        const result = getPreLaunchTaskChain(tasks, 'build');
+        assert.deepStrictEqual(result, ['build']);
+    });
+
+    test('returns empty array when preLaunchTask is not found in tasks list', () => {
+        const result = getPreLaunchTaskChain([], 'nonexistent');
+        assert.deepStrictEqual(result, []);
```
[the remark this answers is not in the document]
MicroFish91: I don't understand this, why would we return the preLaunchTask if it's not found in the list.  I vote no on this.

src/debug/getPreLaunchTaskChain.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    if (!tasksMap.has(preLaunchTask)) {
+        return [];
+    }
+
+    getDependentTasks(preLaunchTask);
+    return Array.from(dependentTasks.values());
```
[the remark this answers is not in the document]
MicroFish91: Same answer as below

src/debug/validatePreDebug.ts
```diff
@@ -32,12 +34,23 @@ export interface IPreDebugContext extends Omit<ISetConnectionSettingContext, 'pr
     projectPath?: string;
 }
 
+const emulatorTaskRegExp: RegExp = /azurite|emulator/i;
```
bwateratmsft: Worried this could be pretty fragile

MicroFish91: Yeah it's definitely not a perfect solution.  Happy to make improvements or explore some alternative approaches.  Here are some other ideas we could consider:

- Have the local setup skill use different default labels so that the functions extension doesn't attach into the F5 process (although we lose some of the other things that it might be able to help w/ like assisting with debugging func CLI errors or helping to install / verify the func CLI)
- Expose a new user/project setting that force skips emulator validation and automatically add that to new projects created through our skill (in `.vscode/settings.json`).

bwateratmsft: Are there any other tighter heuristics?

MicroFish91: Not that I can think of.  I can't think of any heuristics that don't leave us with some level of fragility, since once the user decides they want to start customizing / changing their task chain, they can basically start using any variety of labels, scripts, etc.

The other two options above aren't fragile, but come with their own tradeoffs

MicroFish91: Bullet 2 above probably is my new preferred option.

The change required in this PR would be much simpler too

bwateratmsft: I think we go that way for now, then.

MicroFish91: Done!

bwateratmsft (approved):

PR #4964, opened 2026-04-07
Create New Project design improvements

gavin-aguiar:
## Summary

Introduces an opt-in **Template Gallery** experience for "Create New Project" that consumes the new shared webview package from `microsoft/vscode-azuretools`. All UI lives in the shared package; this extension only owns:

- **Configuration** (header text, supported languages, AI generation flag)
- **Business logic** (template fetching, project creation, Copilot generation, README fetching)
- **Wiring** into existing commands

The extension's previous in-repo Template Gallery (HTML/CSS/JS in `resources/webviews/templateGallery/` from PR #4964, and the React port in `src/webview/`) is removed in favor of the shared package.

> 🔗 Companion PR in `vscode-azuretools`: <https://github.com/microsoft/vscode-azuretools/pull/2278)> — must be merged and published before this PR can switch from `file:` to a versioned dependency.

---

## How to enable

The gallery is gated behind an opt-in setting and does not affect the default Create-Project wizard:

1. Open Command Palette → **Preferences: Open Settings (JSON)**
2. Add `"azureFunctions.enableTemplateGallery": true`
3. Run **Azure Functions: Create New Project...** — the gallery webview opens

---

## Architecture

```
┌────────────────────────────────────────────────────┐
│  @microsoft/vscode-azext-webview  (shared)         │
│  ─────────────────────────────────────────────     │
│  • TemplateGalleryController   (abstract)          │
│  • React UI (browse, details, AI, creating)        │
│  • TemplateGalleryConfig + types                   │
└──────────────────────▲─────────────────────────────┘
                       │ extends
┌──────────────────────┴─────────────────────────────┐
│  FunctionsTemplateGalleryController                │
│  ─────────────────────────────────────────────     │
│  • fetchTemplates()  → ProjectTemplateProvider     │
│  • createProject()   → git clone + scaffold        │
│  • getReadme()       → GitHub raw fetch            │
│  • generateWithCopilot() → vscode.lm chat          │
│  • createAiProject()    → write files to disk      │
│  • continueInChat()     → workbench.action.chat    │
└────────────────────────────────────────────────────┘
```

The shared controller creates the `WebviewPanel`, routes typed messages between the webview and the subclass, and renders the React UI bundled in the package. The subclass implements `fetchTemplates`, `createProject`, `getReadme`, and optionally the three AI methods.

---

## Notable behaviors

| Behavior | Where |
|---|---|
| Opt-in via `azureFunctions.enableTemplateGallery` setting | `createNewProject.ts` |
| Default project location = workspace folder | controller config |
| **"Use Template" button on card → create immediately** | UI handles default location/language |
| **Card click → Template Details** (with README preview) | UI |
| Detected non-empty target folder → modal confirmation before clobbering | `createProject()` |
| Hidden / metadata files (`.vscode`, `.git`, `.DS_Store`, `Thumbs.db`) ignored when assessing emptiness | `createProject()` |
| Cloned `.git` folder removed; `git init` creates a fresh history (no template author commits) | `createProject()` |
| README.md auto-opened in markdown preview after creation | `_tryOpenReadme()` |
| Telemetry events: `templateGallery.getTemplates`, `templateGallery.createProject`, `templateGallery.generateWithCopilot`, `templateGallery.createAiProject` | wrapped via `callWithTelemetryAndErrorHandling` |

---

### Added

| File | Purpose |
|---|---|
| `src/commands/createNewProject/FunctionsTemplateGalleryController.ts` | Subclass of shared `TemplateGalleryController`; ports all gallery business logic |
| `scripts/copyWebviewAssets.mjs` | Copies `views.js`/`views.css` from the shared package into `dist/webview/` at build time so the VSIX ships the bundle |
| `.npmrc` (`install-links=true`) | Forces npm to install the local `file:` dependency as a copy (not a symlink) so `vsce package` doesn't traverse outside the extension folder |

### Modified

| File | Change |
|---|---|
| `src/extension.ts` | Calls `registerWebviewExtensionVariables({ context, webviewAssetsDir })` so the shared `WebviewBaseController` can locate the bundled assets |
| `src/commands/createNewProject/createNewProject.ts` | Routes opt-in flow to `FunctionsTemplateGalleryController.createOrShow(ext.context)` |
| `src/commands/createNewProject/NewProjectLanguageStep.ts` | "Browse Template Gallery (Preview)..." quick pick item routes to the same controller |
| `package.json` | Adds `@microsoft/vscode-azext-webview` dependency; replaces `build:webview` script (esbuild → copy assets); adds `--no-dependencies` to `vsce package` (extension is fully bundled by esbuild) |
| `.vscodeignore` | Removed obsolete codicons whitelist (font is no longer shipped from this extension) |
| `tsconfig.json` | Removed `"src/webview"` from `exclude` (the folder is gone) |


src/commands/createNewProject/StartingPointStep.ts
```diff
@@ -0,0 +1,113 @@
...
+
+        // Add informational item
+        picks.push({
+            label: `$(info) ${localize('templateInfoLine', 'Templates include working code, tests, and Bicep deployment files')}`,
+            data: 'template', // default but won't be selected since it's suppressed
+            suppressPersistence: true,
+            onPicked: () => { /* do nothing - info only */ }
+        });
+
+        const placeHolder = localize('selectStartingPoint', 'How would you like to start your project?');
```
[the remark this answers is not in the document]
gavin-aguiar: Changed the info-only item to use QuickPickItemKind.Separator so it renders as non-selectable descriptive text.

resources/webviews/templateGallery/styles.css
```diff
@@ -0,0 +1,1321 @@
...
+    padding: 0;
+    margin: 0;
+}
+
+/* Codicon font */
+@font-face {
+    font-family: 'codicon';
+    src: url('${codiconsUri}') format('truetype');
+}
+
```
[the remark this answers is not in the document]
gavin-aguiar: Removed

src/commands/createNewProject/TemplateGalleryPanel.ts
```diff
@@ -0,0 +1,1232 @@
...
+                        await AzExtFsExtra.ensureDir(projectPath);
+                        try {
+                            await this._copyDirectory(sourceDir, projectPath);
+                        } finally {
+                            try { await AzExtFsExtra.deleteResource(tempDir, { recursive: true }); } catch { /* ignore */ }
+                        }
+                    }
+
+                    await this._sendProgress('Initializing git repository...');
+                    await cpUtils.executeCommand(undefined, projectPath, 'git init');
```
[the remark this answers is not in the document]
gavin-aguiar: Done. Wrapped the git init call in the zip download path with try/catch, matching the pattern already used in the AI project path

src/commands/createNewProject/TemplateGalleryPanel.ts
```diff
@@ -0,0 +1,1232 @@
...
+            try {
+                await this._sendProgress('Creating project files...');
+                await AzExtFsExtra.ensureDir(location);
+
+                for (const file of files) {
+                    const safePath = file.path.replace(/^[/\\]/, '');
+                    const filePath = path.join(location, safePath);
+                    await AzExtFsExtra.ensureDir(path.dirname(filePath));
+                    await fs.promises.writeFile(filePath, file.content, 'utf8');
+                }
```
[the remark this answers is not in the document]
gavin-aguiar: Done. The AI project creation now checks if the target folder exists and contains files .If non-empty, it creates a uniquely named subfolder to avoid overwriting user content. All subsequent operations (file writes, git init, folder
opening) use the resolved targetDir

src/commands/deploy/SmartDeploy.ts
```diff
@@ -0,0 +1,315 @@
...
+        // Check for host: function (handles both quoted and unquoted YAML values)
+        const hasFunctionService = /host\s*:\s*["']?function["']?/i.test(yamlContent);
+        const hasInfra = detectInfraFolder(projectRoot);
+        return {
+            tier: 'azd-functions',
+            hasFunctionService,
+            hasInfra,
+            hasAzureYaml: true,
+        };
+    }
```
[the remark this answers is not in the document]
gavin-aguiar: Updated. Now detectAzdProject only returns tier azd-functions when hasFunctionService is true

package.json
```diff
@@ -1388,6 +1477,34 @@
                             "%azureFunctions.mcpProjectType.McpExtensionServer%",
                             "%azureFunctions.mcpProjectType.SelfHostedMcpServer%"
                         ]
+                    },
+                    "azureFunctions.projectTemplates.manifestUrl": {
+                        "type": "string",
+                        "default": "https://aka.ms/azfunc-project-templates-manifest",
+                        "description": "%azureFunctions.projectTemplates.manifestUrl%"
```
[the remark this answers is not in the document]
gavin-aguiar: Fixed

src/commands/validateFunctionApp/FunctionAppValidator.ts
```diff
@@ -0,0 +1,484 @@
...
+    python: 'python.md',
+    node: 'node.md',
+    dotnet: 'dotnet.md',
+    'dotnet-isolated': 'dotnet.md',
+    java: 'java.md',
+    powershell: 'powershell.md',
+};
+
+// Files collected from the project for analysis (relative to project root)
+const CONFIG_FILES = ['host.json', 'local.settings.json', 'requirements.txt', '.funcignore'];
```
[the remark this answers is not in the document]
gavin-aguiar: Removed this file.

src/templates/projectTemplates/ProjectTemplateProvider.ts
```diff
@@ -0,0 +1,295 @@
...
+import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
+import { TemplateCategory, type IProjectTemplate, type ITemplateManifest } from './IProjectTemplate';
+
+/**
+ * Provides project templates from a remote manifest with caching and offline fallback
+ */
+export class ProjectTemplateProvider {
+    private static readonly CACHE_KEY = 'projectTemplatesManifest';
+    private static readonly CACHE_TIMESTAMP_KEY = 'projectTemplatesManifestTimestamp';
+    private static readonly DEFAULT_MANIFEST_URL = 'https://cdn.functions.azure.com/public/templates-manifest/manifest.json';
```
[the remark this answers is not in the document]
gavin-aguiar: Updated

src/templates/projectTemplates/ProjectTemplateProvider.ts
```diff
@@ -0,0 +1,295 @@
...
+        const cachedManifest = await this.getCachedManifest();
+        if (cachedManifest) {
+            context.telemetry.properties.manifestSource = 'cache';
+            ext.outputChannel.appendLog(localize('usingCachedManifest', 'Using cached template manifest ({0} templates)', cachedManifest.templates.length.toString()));
+            return cachedManifest;
+        }
+
+        // Try to fetch from remote
+        try {
+            ext.outputChannel.appendLog('DEBUG: About to fetch from remote - v2');
```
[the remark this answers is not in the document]
gavin-aguiar: Removed debug logs

fiveisprime:
Alright, I have some feedback on this and I'm happy to provide more info for any
of these items if needed.

First of all, I tested TypeScript and JavaScript templates on both Windows and
Linux via WSL. I created from a template, ran the project with breakpoints,
added additional functions, and deployed the app via the AZD integration in the
workspace view. I did not test any of the Copilot functionality so far.

It's worth noting that the "create new function" command is available and works
even when a project is created via a template. 🎉

* **Template Gallery Usability:** The template gallery allows changing location
after it was already chosen, which is unnecessary.
* **Theme Compatibility:** Some themes lack sufficient contrast in inputs and
checkboxes, affecting usability.
* **Template Editability:** The Language input appears editable but is not,
which is confusing for users.
* **Workspace Redundancy:** The notification to open the workspace at the end
feels redundant since the location is chosen prior to launching the template
gallery.
* **UI Element Placement:** The notification content is positioned below where
it should be – the “Source: Azure Functions extension” is cut off in a weird
way.
* **Project Initialization Experience:** After project creation, nothing opens
(e.g., rendered readme or source functions), and consider improving this
experience.
* **Pre-Launch Task Automation:** Consider running npm install automatically for
JavaScript and TypeScript projects to prevent errors on first launch.
* **User Guidance:** Add a suggestion to guide the customer to the next step
after project creation.
* **Test File Placement and Extension Recommendation:** Evaluate whether
test.http should be nested in src or at a higher level, and update
extensions.json to recommend the appropriate HTTP testing extension.
* **Git History Initialization:** Git history is inherited from the template
repo and recommend performing a fresh git init to remove tracking.
* **Contributing and License Files:** Update contributing.md to remove
references to CLA bot and ensure license.md reflects the correct license for
customer projects.
* **Responsive UI:** Opening the integrated terminal results in multiple scroll
bars.
* **Project Overwrite Prompt:** Creating a project from a template over an
existing workspace does not prompt for overwrite and ensure the prompt matches
existing project behavior.

gavin-aguiar:
> Alright, I have some feedback on this and I'm happy to provide more info for any of these items if needed.
> 
> First of all, I tested TypeScript and JavaScript templates on both Windows and Linux via WSL. I created from a template, ran the project with breakpoints, added additional functions, and deployed the app via the AZD integration in the workspace view. I did not test any of the Copilot functionality so far.
> 
> It's worth noting that the "create new function" command is available and works even when a project is created via a template. 🎉
> 
> * **Template Gallery Usability:** The template gallery allows changing location
>   after it was already chosen, which is unnecessary.
> * **Theme Compatibility:** Some themes lack sufficient contrast in inputs and
>   checkboxes, affecting usability.
> * **Template Editability:** The Language input appears editable but is not,
>   which is confusing for users.
> * **Workspace Redundancy:** The notification to open the workspace at the end
>   feels redundant since the location is chosen prior to launching the template
>   gallery.
> * **UI Element Placement:** The notification content is positioned below where
>   it should be – the “Source: Azure Functions extension” is cut off in a weird
>   way.
> * **Project Initialization Experience:** After project creation, nothing opens
>   (e.g., rendered readme or source functions), and consider improving this
>   experience.
> * **Pre-Launch Task Automation:** Consider running npm install automatically for
>   JavaScript and TypeScript projects to prevent errors on first launch.
> * **User Guidance:** Add a suggestion to guide the customer to the next step
>   after project creation.
> * **Test File Placement and Extension Recommendation:** Evaluate whether
>   test.http should be nested in src or at a higher level, and update
>   extensions.json to recommend the appropriate HTTP testing extension.
> * **Git History Initialization:** Git history is inherited from the template
>   repo and recommend performing a fresh git init to remove tracking.
> * **Contributing and License Files:** Update contributing.md to remove
>   references to CLA bot and ensure license.md reflects the correct license for
>   customer projects.
> * **Responsive UI:** Opening the integrated terminal results in multiple scroll
>   bars.
> * **Project Overwrite Prompt:** Creating a project from a template over an
>   existing workspace does not prompt for overwrite and ensure the prompt matches
>   existing project behavior.

Hi @fiveisprime, Thanks for testing this new change.  I have addressed your comments in https://github.com/microsoft/vscode-azuretools/pull/2278 since the webview is moved to the other repo. Let me know if you have any other feedback, Thanks.

package.json
```diff
@@ -1596,6 +1700,7 @@
         "@microsoft/vscode-azext-azureappsettings": "^1.0.0",
         "@microsoft/vscode-azext-azureutils": "^4.0.1",
         "@microsoft/vscode-azext-utils": "^4.0.5",
+        "@microsoft/vscode-azext-webview": "file:../vscode-azuretools/webview",
```
gavin-aguiar: Will replace with the correct build before merging..

package.nls.json
```diff
@@ -141,10 +142,17 @@
...
     "azureFunctions.mcpProjectType.NoMcpServer": "Runs the standard Azure Functions runtime with no MCP integration.",
     "azureFunctions.mcpProjectType.McpExtensionServer": "Runs the Functions host with an embedded MCP server provided by the Azure Functions MCP extension.",
     "azureFunctions.mcpProjectType.SelfHostedMcpServer": "Runs the Functions host in custom-handler mode, forwarding requests to a self-hosted MCP server process.",
+    "azureFunctions.projectTemplates.manifestUrl": "URL of the project template manifest. Organizations can host custom manifests to provide team-specific templates.",
+    "azureFunctions.projectTemplates.additionalManifestUrls": "Additional template manifest URLs to merge with the primary manifest.",
+    "azureFunctions.projectTemplates.cacheExpirationHours": "Template manifest cache expiration time in hours.",
+    "azureFunctions.projectTemplates.preferTemplateFlow": "Pre-select 'Start from template' in the project creation wizard.",
+    "azureFunctions.projectTemplates.showBicepPrompt": "Prompt to deploy infrastructure when templates include Bicep files."
```
nturinski: Missing trailing comma — this breaks `package.nls.json` parsing and all NLS resolution.

package.json
```diff
@@ -1546,14 +1648,15 @@
...
+        "build": "npm run build:esbuild & npm run build:webview & npm run build:check",
         "build:esbuild": "node esbuild.mjs",
+        "build:webview": "node scripts/copyWebviewAssets.mjs",
         "build:check": "tsc --noEmit",
         "check": "tsc --noEmit",
         "compile": "tsc -watch",
         "version": "tsc --version",
         "cleanReadme": "node scripts/cleanReadme.mjs",
-        "package": "vsce package --githubBranch main",
+        "package": "vsce package --githubBranch main --no-dependencies",
```
nturinski: `--no-dependencies` ships the VSIX without `node_modules`, so anything not bundled by esbuild will fail at runtime. This repo's `esbuild.mjs` wasn't extended to verify every entry under `dependencies` is bundled. Please install the packaged VSIX and exercise activation before merging, or drop this flag.

package.json
```diff
@@ -1061,7 +1113,24 @@
...
             "editor/title": [
+                {
+                    "command": "azureFunctions.runFunctionApp",
+                    "when": "editorIsOpen",
+                    "group": "navigation@0"
+                },
+                {
+                    "command": "azureFunctions.smartDeploy",
+                    "when": "editorIsOpen",
+                    "group": "navigation@1"
```
nturinski: `"when": "editorIsOpen"` adds a "Run Function App" and "Deploy Function App" icon to the title bar of **every** open editor in VS Code, including unrelated workspaces. Needs to be gated on a Functions-project context (e.g. `azureFunctions.projectLanguage`), and arguably shouldn't be in `editor/title` at all.

package.json
```diff
@@ -612,6 +634,26 @@
...
+                {
+                    "command": "azureFunctions.smartDeploy",
+                    "when": "view == azureWorkspace && viewItem =~ /azFuncLocalProject/i",
+                    "group": "inline"
+                },
+                {
+                    "command": "azureFunctions.smartDeploy",
+                    "when": "view == azureWorkspace && viewItem =~ /azFuncLocalProject/i",
+                    "group": "1@2b"
+                },
```
nturinski: `smartDeploy` is registered twice for the same view item (`inline` + `1@2b`); `runFunctionApp` below is duplicated the same way. Each command will appear twice in the context menu.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,575 @@
...
+                }
+            }
+            await AzExtFsExtra.ensureDir(targetDir);
+
+            for (const file of files) {
+                const safePath = file.path.replace(/^[/\\]/, '');
+                const filePath = path.resolve(targetDir, safePath);
+                if (!filePath.startsWith(path.resolve(targetDir))) {
+                    continue;
+                }
```
nturinski: Path-traversal check is unsafe:

- `replace(/^[/\\]/, '')` strips only one leading separator, so `..\..\evil` still escapes.
- `startsWith(path.resolve(targetDir))` without a trailing `path.sep` accepts e.g. `targetDir = C:\foo` → `C:\foobar\evil.txt`.
- Silently `continue`-ing leaves no signal that the AI's output was partially dropped.

Suggest:
```ts
const rel = path.relative(targetDir, filePath);
if (rel.startsWith('..') || path.isAbsolute(rel)) {
    throw new Error(`Refusing to write outside project: ${file.path}`);
}
```

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,575 @@
...
+        await fs.promises.writeFile(zipPath, buffer);
+
+        try {
+            await cpUtils.executeCommandLine(undefined, undefined, `tar -xf "${zipPath}" -C "${path.dirname(destDir)}"`);
+        } catch {
+            if (process.platform === 'win32') {
+                await cpUtils.executeCommandLine(undefined, undefined,
+                    `powershell -Command "Expand-Archive -Path '${zipPath}' -DestinationPath '${path.dirname(destDir)}' -Force"`);
+            } else {
+                await cpUtils.executeCommandLine(undefined, undefined, `unzip -q "${zipPath}" -d "${path.dirname(destDir)}"`);
```
nturinski: String-interpolated shell commands are a quoting-injection class bug — `os.tmpdir()` contains the username on Windows, so a `'` or `"` in the path breaks PowerShell's single-quoted args. Use `cpUtils.executeCommand` with an argv array, or extract the zip with a Node library (`yauzl`/`adm-zip`) so no shell is involved.

src/commands/createNewProject/StartingPointStep.ts
nturinski, line 1: This whole file (along with `TemplateListStep`, `CloneTemplateStep`, `PostCloneStep`, `IProjectTemplate`, `ProjectTemplateProvider`, and the 846-line bundled `manifest.json`) is dead code — nothing in `createNewProjectInternal` is modified to insert `StartingPointStep` into the wizard, and no other call site references it. Either wire it in or remove it from this PR.

nturinski (changes requested):
Reviewed for blocking issues only — inline comments cover them. High level:

- The PR description only covers the webview Template Gallery, but the diff also adds (1) an entirely new quick-pick template flow (`StartingPointStep` + 5 supporting files + 846-line bundled manifest) that is **dead code**, and (2) two brand-new top-level commands (`smartDeploy`, `runFunctionApp`) that overlap existing commands. Strongly recommend splitting into separate PRs — they have very different review surfaces.
- `package.nls.json` doesn't parse.
- `file:` sibling dep + `--no-dependencies` packaging will both fail outside the author's local checkout.
- The new editor-title menu items appear on every open editor in VS Code.
- Two security issues in the AI/clone code paths (path-traversal check, shell quoting).

Happy to look at a follow-up once the dead code is removed and the gallery is split out.

src/commands/runFunctionApp/RunFunctionApp.ts
```diff
@@ -0,0 +1,511 @@
+/*---------------------------------------------------------------------------------------------
```
nturinski: Could you explain the intention behind RunFunctionApp.ts being added to this PR? This seems similar to what the extension already ships using VS Code's task and debug APIs. For example:

FuncTaskProvider auto-contributes func: host start (plus func: extensions install and Python pack / pack --build-native-deps) tasks for every Functions project.

The per-language FuncDebugProviderBase subclasses wire those tasks up as the preLaunchTask for F5.
validatePreDebug runs the necessary pre-flight checks (Core Tools installed and version-compatible, FUNCTIONS_WORKER_RUNTIME present, durable-storage connections, AzureWebJobsStorage and Azurite actually running).

PythonVenvCreateStep, PythonAliasListStep, and venvUtils (venvExists, runPipInstallCommandIfPossible, convertToVenvCommand) own venv lifecycle.

The language-specific init steps (JavaScriptInitVSCodeStep, TypeScriptInitVSCodeStep, PythonInitVSCodeStep, Script/C# variants) already declare npm install / npm run build / dotnet build / pip install as dependsOn tasks for host start.
funcHostTask tracks the running host so the FunctionHostDebugView can surface logs, errors, and the host 
PID.

However, RunFunctionApp.ts ignores every one of those systems and reimplements each one inline — project-root discovery (a worse copy of tryGetFunctionProjectRoot), runtime detection (a worse copy of tryGetFunctionsWorkerRuntimeForProject that only knows three languages and silently falls through to "generic" for Java/PowerShell/Ballerina/Custom), Python alias probing, venv creation, pip install, npm install/build, and dotnet build. Any bug fix or new language now has two places to update, with the new copy guaranteed to drift.

It is also strictly less flexible. It hardcodes .venv while the rest of the extension respects the azureFunctions.pythonVenv workspace setting. It ignores user customizations to the func: host start task in tasks.json (custom args, env vars, problem matchers, working directory), the azureFunctions.projectLanguage / azureFunctions.preDeployTask / validateEmulators settings, custom preDeployTask hooks, and the Functions Core Tools path resolved by getFuncCliPath. It bypasses Azurite validation, durable-storage connection setup, and the worker-runtime auto-repair that preDebugValidate does.

Because it spawns func and npm / dotnet directly, the host doesn't show up in the Function Host Debug view, has no problem matcher integration, no port detection, no pickFuncProcess support, and no Activity Log entries.

Its Python workaround — a custom Pseudoterminal that silently swallows all stdin to dodge the Python extension's auto-activation injection — is a workaround for a problem the existing convertToVenvCommand-based ShellExecution path doesn't have, and it solves it by breaking every legitimate stdin use case (Ctrl+C is handled, but anything else a user types is dropped on the floor).

On top of that it ships several anti-patterns the rest of the codebase has long since moved past:

Synchronous fall-through try/catch blocks that swallow pip install / npm install / npm run build / dotnet build failures and continue to func start anyway, so the user sees a broken host instead of the real error.
JSON.parse on local.settings.json instead of the existing getLocalSettingsJson helper that handles JSONC and connection-string parsing.

A hand-rolled Windows process-tree kill via taskkill instead of the shared process utilities.
An entirely new top-level command and toolbar button that overlaps in purpose with the existing F5 debug entry point and the run icon already wired into the Functions walkthrough — adding user-visible surface area, telemetry, localization strings, and a maintenance burden for a capability the extension already provides more correctly.

src/commands/deploy/SmartDeploy.ts
```diff
@@ -0,0 +1,323 @@
...
+import * as path from 'path';
+import * as vscode from 'vscode';
+import { ext } from '../../extensionVariables';
+import { localize } from '../../localize';
+
+// ---------------------------------------------------------------------------
+// Public entry point
+// ---------------------------------------------------------------------------
+
+export async function smartDeploy(_context: IActionContext, resourceUri?: vscode.Uri): Promise<void> {
```
nturinski: I’d consider folding the smartDeploy behavior into the existing azureFunctions.deployProject command rather than adding it as a parallel deploy command.

The main reason is that smartDeploy doesn’t seem like a fundamentally different deploy path. It’s more of an AZD-aware layer on top of the existing deploy flow. In the non-AZD case, it already falls back to:

`vscode.commands.executeCommand('azureFunctions.deployProject', resourceUri)`

So structurally, it feels like the AZD detection belongs inside deployProject itself.

Having both commands registered side by side also introduces a few downsides:

- User choice becomes less clear. Two similar “Deploy…” entries make the user decide which deploy path is the right one. The AZD-aware behavior feels like the thing most users would expect from the normal Deploy command.
- Existing entry points may miss the new behavior. Tree-view menus, walkthroughs, internal callers, post-deploy flows, slot deploy, and the public extension API already point at azureFunctions.deployProject. Those would bypass AZD detection unless deployProject becomes the smart default.
- Telemetry and debugging get split. Separate command IDs make it harder to understand which deploy path a user took when looking at telemetry or investigating issues.
- The two paths may drift over time. Project-root resolution, validation, sign-in handling, slot handling, and post-deploy tasks are all areas where keeping two paths aligned could become easy to miss.

A possible shape would be:

1. In deployProject, resolve the project root as usual.
1. Before launching the normal wizard, call detectAzdProject.
1. If an azure.yaml or azure.yml is found, prompt with something like:
    - Deploy with azd up
    - Use Functions deploy
    - Cancel
1. Gate that with a setting like:
  `azureFunctions.deploy.azdPromptBehavior: 'prompt' | 'azd' | 'functions'`

and include a “Don’t ask again” option.

With that structure, `azureFunctions.deployProject` remains the main deploy entry point, but becomes AZD-aware. SmartDeploy.ts can still exist as the place for detectAzdProject and runAzd* helpers; it just wouldn’t need to register a separate command/menu item.

src/commands/deploy/SmartDeploy.ts
```diff
@@ -0,0 +1,323 @@
...
+        context.telemetry.properties.deployPath = 'traditional-bicep-project';
+        await runTraditionalDeploy(resourceUri);
+    }
+}
+
+// ---------------------------------------------------------------------------
+// AZD invocation
+// ---------------------------------------------------------------------------
+
+async function runAzdDeploy(context: IActionContext, projectRoot: string): Promise<void> {
```
nturinski: I’d also simplify the AZD execution path by relying on the AZD extension rather than maintaining our own CLI fallback.

Right now, SmartDeploy has a few layers:

- Try azure-dev.commands.cli.*
- Fall back to spawning azd in a terminal
- If that fails, show docs/install guidance

I don’t think we need all three. The AZD extension already exposes the CLI surface through commands like:

```
azure-dev.commands.cli.up
azure-dev.commands.cli.init
azure-dev.commands.cli.provision
azure-dev.commands.cli.deploy
azure-dev.commands.cli.down
azure-dev.commands.cli.install
```

Since the AZD extension owns the install and CLI-missing experience, we can route through it instead of duplicating CLI detection and terminal handling.

src/commands/deploy/SmartDeploy.ts
```diff
@@ -0,0 +1,323 @@
...
+    /** True when azure.yaml contains at least one service with host: function */
+    hasFunctionService: boolean;
+    /** True when infra/ folder with *.bicep or *.tf files exists */
+    hasInfra: boolean;
+    /** True when azure.yaml is present (regardless of service type) */
+    hasAzureYaml: boolean;
+}
+
+function detectAzdProject(projectRoot: string): AzdStatus {
+    const azureYamlPath = path.join(projectRoot, 'azure.yaml');
```
nturinski: I think technically azd handles both azure.yaml and azure.yml.

package.json
gavin-aguiar: The --no-dependencies flag is a temporary workaround tied to the file: dependency on the unreleased @microsoft/vscode-azext-webview package. Once vscode-azext-webview 1.2.0 is published, I'll switch the dependency to ^1.2.0, delete .npmrc, drop --no-dependencies,

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: Fixed. Updated to using the extract-zip library,.

src/commands/createNewProject/StartingPointStep.ts
gavin-aguiar: ProjectTemplateProvider, IProjectTemplate, and the bundled manifest.json are wired into FunctionsTemplateGalleryController._templateProvider with manifest.json being the offline fallback. This manifest.json contains only a subset of templates as compared to the remote manifest. I'll delete to other files.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: Fixed

package.json
gavin-aguiar: Added a new file: src/utils/funcProjectContext.ts

 - Exports registerFunctionProjectContext(context)
 - Sets context key azureFunctions.isFunctionProject based on whether any workspace folder contains a host.json at any depth (findFiles('**/host.json', undefined, 1))
 - Refreshes on: - Activation
 - onDidChangeWorkspaceFolders
 - File watcher create/delete events for **/host.json (so a project being created or deleted mid-session updates the icons)

package.json
gavin-aguiar: This change is intentional (and identical to the existing deployProject pattern at lines 627–636). Both commands intentionally show as inline icons next to the project tree item AND as right-click menu entries. In view/item/context, the inline group is a separate rendering area, it renders as icon buttons inline next to the tree item label. It does not appear in the right-click context menu.

src/commands/runFunctionApp/RunFunctionApp.ts
gavin-aguiar: The goal of adding a RunFunctionApp option was a single, discoverable "Run" command for first-time users who don't yet know about F5, tasks.json, or the difference between "debug" and "run". The walkthrough and the new editor-title icon both surface it. The use case is: clone a template → click Run → see the host start, without needing to read docs first.
For this PR, I'll remove the RunFunctionApp and SmartDeploy feature as some more changes might be needed. I'll rewrite these features based on your comments and open a new PR for this.

src/commands/deploy/SmartDeploy.ts
gavin-aguiar: Remove Smart Deploy

src/vsCodeConfig/verifyInitForVSCode.ts
```diff
@@ -36,6 +37,11 @@ export async function verifyInitForVSCode(context: IActionContext, fsPath: strin
         language = nonNullOrEmptyValue(getWorkspaceSetting(projectLanguageSetting, fsPath), projectLanguageSetting);
         languageModel = getWorkspaceSetting(projectLanguageModelSetting, fsPath);
         version = nonNullOrEmptyValue(tryParseFuncVersion(getWorkspaceSetting(funcVersionSetting, fsPath)), funcVersionSetting);
+    } else if (languageModel === undefined) {
+        languageModel = await detectProjectLanguageModel(<ProjectLanguage>language, fsPath);
```
nturinski, line 41: Why do we need to add this check now?

src/vsCodeConfig/verifyVSCodeConfigOnActivate.ts
```diff
@@ -30,9 +31,15 @@ export async function verifyVSCodeConfigOnActivate(context: IActionContext, fold
                 context.telemetry.suppressIfSuccessful = false;
 
                 const language: ProjectLanguage | undefined = getWorkspaceSetting(projectLanguageSetting, projectPath);
-                const languageModel = getWorkspaceSetting<number>(projectLanguageModelSetting, projectPath);
+                let languageModel = getWorkspaceSetting<number>(projectLanguageModelSetting, projectPath);
                 const version: FuncVersion | undefined = tryParseFuncVersion(getWorkspaceSetting(funcVersionSetting, projectPath));
                 if (language !== undefined && version !== undefined) {
+                    if (languageModel === undefined) {
```
nturinski: Do you need this check here? This happens on activation and I'd prefer not adding an async call that would slow down performance.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+                    } catch {
+                        // Give up silently — README is best-effort.
+                    }
+                }
+                return;
+            }
+        }
+    }
+
+    private async _isGitInstalled(): Promise<boolean> {
```
nturinski: Our repository only uses underscore-prefixes for private fields, not for methods. Please keep it consistent with our current naming convention.

src/templates/projectTemplates/ProjectTemplateProvider.ts
```diff
@@ -0,0 +1,284 @@
...
+    private async cacheManifest(manifest: ITemplateManifest): Promise<void> {
+        await ext.context.globalState.update(ProjectTemplateProvider.CACHE_KEY, manifest);
+        await ext.context.globalState.update(ProjectTemplateProvider.CACHE_TIMESTAMP_KEY, Date.now());
+    }
+
+    /**
+     * Fetch manifest from remote URL(s)
+     */
+    private async fetchManifestFromRemote(context: IActionContext): Promise<ITemplateManifest> {
+        const primaryUrl = getWorkspaceSetting<string>('projectTemplates.manifestUrl') || ProjectTemplateProvider.DEFAULT_MANIFEST_URL;
```
nturinski: You have many magic string settings in this file such as:
```
'projectTemplates.manifestUrl', 'projectTemplates.additionalManifestUrls', 'projectTemplates.cacheExpirationHours', 'requestTimeout'
```

Please add them to the existing constants.ts where all of our vscode workspace settings live:

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+                } catch {
+                    existing = [];
+                }
+                const ignoredNames = new Set(['Thumbs.db', 'desktop.ini', '.DS_Store']);
+                const significant = existing.filter(name => !name.startsWith('.') && !ignoredNames.has(name));
+                if (significant.length > 0) {
+                    actionContext.telemetry.properties.nonEmptyTarget = 'true';
+                    const proceed = localize('proceedAnyway', 'Create here anyway');
+                    // Modal dialogs add an automatic "Cancel" button — only pass the affirmative action.
+                    const choice = await vscode.window.showWarningMessage(
```
nturinski: There are 4 instances where you use `vscode.window.show...`. Please use our `context.ui.show...` so the dialogs are testable, telemetry-tracked, and have stepName properties.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+    private _buildZipUrl(repositoryUrl: string, branch: string): string {
+        const base = repositoryUrl.replace(/\.git$/, '').replace(/\/$/, '');
+        return `${base}/archive/refs/heads/${branch}.zip`;
+    }
+
+    private async _downloadAndExtractZip(repositoryUrl: string, branch: string, destDir: string): Promise<void> {
+        const zipUrl = this._buildZipUrl(repositoryUrl, branch);
+        const zipPath = `${destDir}.zip`;
+
+        const response = await fetch(zipUrl);
```
nturinski: Please use our requestUtils to make fetch calls.

src/commands/createNewProject/IProjectWizardContext.ts
```diff
@@ -33,6 +34,22 @@ export interface IProjectWizardContext extends IActionContext {
     containerizedProject?: boolean;
     // Didn't want `mcpProjectType` on `IProjectWizardContext` but all projects can have mcp extension triggers
     mcpProjectType?: McpProjectType;
+
+    // Template flow properties
+    /** Whether user chose 'template' or 'scratch' starting point */
+    startingPoint?: 'template' | 'scratch';
```
nturinski: Are these being used anywhere?

package.json
```diff
@@ -1496,6 +1496,39 @@
...
+                        },
+                        "default": [],
+                        "description": "%azureFunctions.projectTemplates.additionalManifestUrls%"
+                    },
+                    "azureFunctions.projectTemplates.cacheExpirationHours": {
+                        "type": "number",
+                        "default": 24,
+                        "description": "%azureFunctions.projectTemplates.cacheExpirationHours%"
+                    },
+                    "azureFunctions.projectTemplates.preferTemplateFlow": {
```
nturinski: Unused setting along with `azureFunctions.projectTemplates.showBicepPrompt `

src/templates/projectTemplates/IProjectTemplate.ts
```diff
@@ -0,0 +1,157 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type ProjectLanguage } from '../../constants';
+
+/**
+ * Represents a complete project template that can be cloned and used as a starting point
+ */
+export interface IProjectTemplate {
```
nturinski, line 15: A lot of these settings aren't used-- is this copied and pasted from somewhere else?

src/templates/projectTemplates/ProjectTemplateProvider.ts
```diff
@@ -0,0 +1,284 @@
...
+        const additionalUrls = getWorkspaceSetting<string[]>('projectTemplates.additionalManifestUrls') || [];
```
nturinski: Mild security concern, but technically malicious extensions could inject this sitting with any malicious cdn and it will just download it. Maybe have an allow list?

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+                ext.outputChannel.appendLog(`AI parse error. Raw response length: ${fullResponse.length}. ${parseError(parseErr).message}`);
+                this.postMessageToWebview({
+                    type: 'aiError',
+                    error: localize('aiParseError', 'Could not parse Copilot response. Please try again with a more specific description.')
+                });
+            }
+        });
+    }
+
+    protected override async createAiProject(files: Array<{ path: string; content: string }>, location: string): Promise<void> {
```
nturinski: There is a user experience inconsistency here:

createProject → modal confirm on non-empty target.
createAiProject → silently writes to azure-functions-project-${Date.now()} subfolder, then shows an info message after the fact with a path the user never asked for. No confirmation, no chance to pick a different folder.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+            this.sendProgress('Initializing git repository...');
+            try {
+                await cpUtils.executeCommandLine(undefined, targetDir, 'git init');
+            } catch {
+                // git init optional
+            }
+
+            actionContext.telemetry.properties.result = 'Succeeded';
+            await this._tryOpenReadme(targetDir);
+            this.panel.dispose();
```
nturinski: nit: If the panel is closed before this, it'll throw an error. Could verify it isn't disposed first.

src/extension.ts
```diff
@@ -60,10 +61,16 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
     ext.context = context;
     ext.outputChannel = createAzExtOutputChannel('Azure Functions', ext.prefix);
     context.subscriptions.push(ext.outputChannel);
+    ext.diagnosticCollection = vscode.languages.createDiagnosticCollection('Azure Functions');
+    context.subscriptions.push(ext.diagnosticCollection);
 
     registerUIExtensionVariables(ext);
     registerAzureUtilsExtensionVariables(ext);
     registerAppServiceExtensionVariables(ext);
+    registerWebviewExtensionVariables({
```
nturinski: I don't think we should be registering this on activation especially if the preview flag is disabled. Could you defer this to `FunctionsTemplateGalleryController.createOrShow()`?

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "vscode-azurefunctions",
     "displayName": "Azure Functions",
     "description": "%azureFunctions.description%",
-    "version": "1.21.1-alpha.0",
+    "version": "1.21.1-alpha.117",
```
nturinski: What's with this version number?

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
MicroFish91: Just wanted to provide some more context about this - but we dependency inject the vscode prompting methods onto `context.ui`.  During normal operation, `context.ui` will be very siimilar to the `vscode.window` commands since we often wrap them, but during testing we pass in our own test ui so that we can inject our own test inputs

src/extension.ts
```diff
@@ -60,10 +61,16 @@ export async function activateInternal(context: vscode.ExtensionContext, perfSta
...
+        context,
+        webviewAssetsDir: vscode.Uri.joinPath(context.extensionUri, 'dist', 'webview').fsPath,
```
nturinski: By the way, we have to change this line for it to debug. I think that the path is incorrect somehow. This is what we change it to:

`webviewAssetsDir: path.join(context.extensionPath, 'dist', 'webview'),`

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -0,0 +1,572 @@
...
+                );
+            } finally {
+                try { await AzExtFsExtra.deleteResource(tempDir, { recursive: true }); } catch { /* ignore */ }
+            }
+        });
+    }
+
+    // ── Optional overrides ──
+
+    protected override async generateWithCopilot(prompt: string, language: string): Promise<void> {
```
nturinski: I think there’s a real case for consolidating the Copilot entry points here, mostly because the current split creates more complexity than value.

The empty state asks users to choose between two options before they really understand the tradeoff. For a first-run experience, that creates a little too much decision fatigue. Both buttons have equal visual weight, so the UI makes them feel equally recommended, but they are actually different interaction models.

The Generate path also overlaps pretty heavily with what Copilot agent mode can already do. Since agent mode can write files directly, we’re effectively rebuilding a scaffolding flow around our own prompt and JSON-parsing pipeline so that we can control the output shape. That gives us some control, but it also adds fragility. Strict “return only JSON” prompting is hard to rely on long-term, especially when the underlying Copilot model can change over time. The chat path is less brittle because it does not depend as much on a perfectly formatted model response.

There’s also a maintenance cost to keeping both paths. We end up with two success flows, two error flows, separate prompt-building logic, separate telemetry, and a meaningful amount of controller code that exists mainly to preserve the split. That may be worth it if the experiences are clearly distinct and both are important, but right now I’m not sure the value justifies the extra surface area.

From a trust perspective, I also think chat is the safer default for a preview feature. Having Copilot silently generate files directly into the workspace is a bigger ask than letting the user review the plan and output in chat first. The chat flow gives users more visibility and a lower-risk way to understand what is about to happen.

The one part of the chat experience that still feels very valuable is the “continue in chat after success/error” flow. That gives users a natural way to troubleshoot, refine, or ask follow-up questions after the project has been generated. I would keep that regardless. The part that feels less clearly useful is the initial empty-state Chat button, because it currently overlaps with Generate without making the distinction obvious.

There’s also some drift risk in the current setup. The shared webview package owns the labels and ordering of the buttons, while the Functions repo owns the behavior. That means a UX change has to be coordinated across multiple surfaces and repos, which increases the chance that the product meaning and implementation get out of sync over time.

My read is that the current state is probably the least ideal version: two primary-looking buttons that imply equal endorsement, even though they solve different problems and carry different tradeoffs.

I think we should fully consolidate around chat and remove generateWithCopilot and the JSON-parsing pipeline entirely. That would trade off some speed in the best-case path, but it would leave us with a smaller, more durable implementation that is less dependent on model-specific formatting behavior.

src/vsCodeConfig/verifyInitForVSCode.ts
gavin-aguiar: This guards a Python v2 / Node v4 regression that surfaces specifically with the new Template Gallery flow.  
Before verifyInitForVSCode only ran initProjectForVSCode if !language || !version (no settings written yet). If a project already had azureFunctions.projectLanguage in settings.json but not azureFunctions.projectLanguageModel, it skipped both branches and proceeded with languageModel = undefined. With this condition, it now resolves to the correct template schema on the next "Create Function" call.

src/vsCodeConfig/verifyVSCodeConfigOnActivate.ts
gavin-aguiar: Good catch. The same recovery already happens lazily in verifyInitForVSCode.ts, which only runs when a command needs the model. I'll remove this block from activation so we don't pay the file-glob + settings-write cost on every startup.

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: renamed all 6 underscore-prefixed methods (_tryOpenReadme, _isGitInstalled, _buildZipUrl, _downloadAndExtractZip, _copyDirectory, _buildGenerationPrompt

src/templates/projectTemplates/ProjectTemplateProvider.ts
gavin-aguiar: Done

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: Updated this line to actionContext.ui.showWarningMessage. The other three call sites don't have a context.ui equivalent. There are 2 non-blocking success toasts (showInformationMessage) and 1  showTextDocument (open a file in an editor as a fallback when markdown.showPreview fails). That's an editor command, not user input. IAzureUserInput only exposes interactive prompts (showWarningMessage, showInputBox, showQuickPick, showOpenDialog) but no informational toast. So the only behavior swap needed for testability was the warning prompt, which is now on context.ui

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: Removed the "Generate with Copilot" option

src/extension.ts
gavin-aguiar: Fixed

package.json
gavin-aguiar: This was my test version that got pushed accidentally. Reverted it

src/extension.ts
gavin-aguiar: Done

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
gavin-aguiar: Fixed

src/commands/deploy/SmartDeploy.ts
gavin-aguiar: Removed smartdeploy

package.json
gavin-aguiar: Updated with the correct version

nturinski (approved):

PR #4999, opened 2026-04-28
Stop filtering Java version picker by locally installed JDK

### LINKED ISSUE #4998: Java version picker hides versions not matching locally installed JDK
When creating a new Java Azure Functions project, the Java version quick pick filters out any version higher than the locally detected JDK (via JAVA_HOME or java -version). This causes problems:

- Users who manage JDKs through tools like SDKMAN, jenv, or manual installs outside JAVA_HOME never see versions the extension can't detect.
- Users who plan to install a different JDK after scaffolding the project are blocked from selecting their target version.
- The local JDK is only needed to run the Maven archetype tool, not to determine the project's target Java version.

The extension should show all Java versions supported by the selected Azure Functions runtime version, regardless of the locally installed JDK. The locally detected version can be labeled to help users pick a default.

ahmedmuhsin:
Fixes #4998

The Java version quick pick was hiding versions higher than the locally detected JDK. Users who manage JDKs outside JAVA_HOME or plan to install a different version could not select their target.

**Changes:**
- Remove the local JDK version filter from getPicks(). All versions supported by the selected Functions runtime are now shown.
- Validate that a JDK >= 8 is detected at the start of getPicks(), before the user fills out the rest of the wizard. This gives a clear, actionable error early rather than a confusing Maven failure after completing all steps.
- getJavaVersion() returns undefined instead of throwing when no JDK is found. Its role is now detection (for labeling and validation), not gatekeeping the version list.
- Add a (Detected) label on the pick matching the locally installed JDK version as a convenience.

### DIFF
--- a/src/commands/createNewProject/javaSteps/JavaVersionStep.ts
+++ b/src/commands/createNewProject/javaSteps/JavaVersionStep.ts
@@ -45,11 +45,15 @@ export class JavaVersionStep extends AzureWizardPromptStep<IJavaProjectWizardCon
     }
 
     async getPicks(context: IJavaProjectWizardContext): Promise<IAzureQuickPickItem<string>[]> {
-        const javaVersion: number = await getJavaVersion();
+        const detectedVersion: number | undefined = await getJavaVersion();
         const result: IAzureQuickPickItem<string>[] = [];
         for (const version of versionInfo) {
-            if (await hasMinFuncCliVersion(context, version.miniFunc, context.version) && javaVersion >= Number(version.data)) {
-                result.push(version);
+            if (await hasMinFuncCliVersion(context, version.miniFunc, context.version)) {
+                const pick: IAzureQuickPickItem<string> = { ...version };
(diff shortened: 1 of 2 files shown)


nturinski:
If you don't have Java installed, you have to go through the entire wizard and then you get this error immediately

image

This is pretty frustrating UX.

ahmedmuhsin:
@nturinski Agreed, that's a bad experience. The latest push addresses this - we now validate that a JDK >= 8 is detected right when the Java version picker runs, before the user fills out the rest of the wizard. If no JDK is found, we throw a clear error: "JDK 8 or higher is required to create a Java project. Please install a JDK and ensure JAVA_HOME is set or java is available on PATH."

Separately, we also stopped filtering the version list by whatever JDK happens to be locally installed. All versions supported by the selected Functions runtime are shown, so users who manage JDKs outside of JAVA_HOME can still pick their target. The detected version gets a (Detected) label for convenience.

nturinski (approved):

PR #5002, opened 2026-04-30
Adding MCP Prompt Template

swapnil-nagar:
This PR adds MCP Prompt Trigger support to the Node.js v4 backup template set used by the Azure Functions extension.

The change introduces new backup templates for both JavaScript and TypeScript so users can create MCP prompt functions when the extension is using the bundled backup templates. It also updates template classification so the new prompt templates are treated consistently with other MCP templates in the extension.

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
@@ -8,7 +8,7 @@ import * as os from 'os';
 import * as path from 'path';
 import { type Progress } from 'vscode';
 import { FuncVersion } from '../../../FuncVersion';
-import { gitignoreFileName, hostFileName, localSettingsFileName, localStorageEmulatorConnectionString, workerRuntimeKey } from '../../../constants';
+import { gitignoreFileName, hostFileName, localSettingsFileName, workerRuntimeKey } from '../../../constants';
 import { type IHostJsonV1, type IHostJsonV2 } from '../../../funcConfig/host';
 import { type ILocalSettingsJson } from '../../../funcConfig/local.settings';
 import { bundleFeedUtils } from '../../../utils/bundleFeedUtils';
@@ -25,7 +25,7 @@ export class ScriptProjectCreateStep extends ProjectCreateStepBase {
     protected localSettingsJson: ILocalSettingsJson = {
         IsEncrypted: false,
         Values: {
-            AzureWebJobsStorage: localStorageEmulatorConnectionString
+            AzureWebJobsStorage: ''
         }
     };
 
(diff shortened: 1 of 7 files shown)


swapnil-nagar:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/vscode-azurefunctions/pull/5002#pullrequestreview-4208842551)

swapnil-nagar:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/vscode-azurefunctions/pull/5002#pullrequestreview-4208842551)

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
```diff
@@ -25,7 +25,7 @@ export class ScriptProjectCreateStep extends ProjectCreateStepBase {
     protected localSettingsJson: ILocalSettingsJson = {
         IsEncrypted: false,
         Values: {
-            AzureWebJobsStorage: ''
+            AzureWebJobsStorage: localStorageEmulatorConnectionString
```
nturinski: Please revert this.

test/templateCount.test.ts
```diff
@@ -30,14 +30,14 @@ function addSuite(source: TemplateSource | undefined): void {
     suite(`Template Count - ${source === undefined ? 'defaultOnExtensionActivation' : source}`, () => {
         const cases: TestCase[] = [
             { language: ProjectLanguage.JavaScript, version: FuncVersion.v1, expectedCount: 8 },
-            { language: ProjectLanguage.JavaScript, version: FuncVersion.v4, expectedCount: 18, backupExpectedCount: 18 },
+              { language: ProjectLanguage.JavaScript, version: FuncVersion.v4, expectedCount: 18, backupExpectedCount: 18 },
```
nturinski: The count probably has to go up one.

src/commands/createNewProject/ProjectCreateStep/ScriptProjectCreateStep.ts
swapnil-nagar: Without his change, AzureWebJobsStorage is empty and template never executes? I thought that would be an improvement.

nturinski: Discussed offline, but summary:

We want to detect when a local emulator is used during deployment, but the tricky part is how to save that for later:

1. If we prompt the user to change it when they deploy, where do we overwrite the setting? In local.settings.json? It kind of doesn't make sense because if they try to debug later, now it's pointing to a live resource
1. If we prompt them to connect to a SA, do we save it as a VS Code setting and not overwrite the local.settings.json? That way the next time they deploy, it'll automatically be picked up. However, if they ever need to change it, it may be difficult to expose/find.

nturinski: Resolving though as we will deal with this later.

test/templateCount.test.ts
swapnil-nagar: I checked there are only 18 templates, no update is needed in the tests

nturinski (approved):

PR #5017, opened 2026-05-13
Add Go language support: project creation and local run

harshivcodes:
First of three PRs adding Go support to the Azure Functions extension. After this PR, users can create a Go Functions project and run it locally with F5 / Ctrl+F5. Debug and deploy follow in subsequent PRs.

Highlights:
- Add `Go` to ProjectLanguage enum, Create New Project picker (after Python), Initialize Project for VS Code picker, and the package.json projectLanguage enum (marked Preview).
- Map Go to the `native` worker runtime (what core tools and Azure expect for FUNCTIONS_WORKER_RUNTIME). Add `native` to isKnownWorkerRuntime() so deploy verification doesn't reject Go apps.
- Register `$func-golang-watch` problem matcher and use it for Go's tasks.json. Matcher keeps the user-friendly `golang` name even though the underlying runtime value is `native`.
- New GoProjectCreateStep: thin wrapper around `func init --worker-runtime go --no-source-control`. Honors workspace funcCliPath setting via context.workspaceFolder fallback chain.
- New GoInitVSCodeStep: generates .vscode/{tasks,launch,settings, extensions}.json. Launch config is a Delve remote attach on port 2345 (the dlv lifecycle itself lands in a follow-up PR). Recommends the golang.go extension.
- Skip template selection for Go in FunctionListStep (mirrors SelfHostedMCPServer) -- `func init` already scaffolds an initial function.
- Show a modal warning when Create Function is invoked on a Go project -- `func new --language Go` isn't supported by core tools yet.
- Unit tests for the three Go-specific assertions in settings.ts (getRootFunctionsWorkerRuntime, isKnownWorkerRuntime, getFuncWatchProblemMatcher).

### DIFF
--- a/src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
+++ b/src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
@@ -1,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { composeArgs, withArg } from '@microsoft/vscode-processutils';
+import { workspace, type Progress } from 'vscode';
+import { ext } from '../../../extensionVariables';
+import { getFuncCliPath } from '../../../funcCoreTools/getFuncCliPath';
+import { localize } from '../../../localize';
+import { cpUtils } from '../../../utils/cpUtils';
+import { type IProjectWizardContext } from '../IProjectWizardContext';
+import { ProjectCreateStepBase } from './ProjectCreateStepBase';
+
+export class GoProjectCreateStep extends ProjectCreateStepBase {
+    public async executeCore(context: IProjectWizardContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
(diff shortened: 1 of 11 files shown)


src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        // Prefer the currently-open workspace's azureFunctions.funcCliPath override.
+        // context.workspacePath is the freshly-picked new-project folder, which isn't part of any open
+        // workspace yet, so passing it to getFuncCliPath only resolves global/user settings.
+        const funcCliPathScope = context.workspaceFolder ?? workspace.workspaceFolders?.[0];
+        const funcCliPath = await getFuncCliPath(context, funcCliPathScope);
+        // Core tools accepts "go" as the worker-runtime CLI argument and maps it to the "native" worker runtime internally.
+        const args = composeArgs(
+            withArg('init'),
+            withArg('--worker-runtime', 'go'),
```
MicroFish91, line 27: When I create a new project with Go, I'm getting this error:

image

I think I'm upgraded to the latest func version:
image

nturinski: It's not mentioned in the PR, but Harshitha has to use a custom build of the func cli that includes golang. We will need to get that so we can test it ourselves.

PR #5019, opened 2026-05-13
Add Go language support: project creation and local run

harshivcodes:
First of three PRs adding Go support to the Azure Functions extension. After this PR, users can create a Go Functions project and run it locally with F5 / Ctrl+F5. Debug and deploy follow in subsequent PRs.

Highlights:
- Add `Go` to ProjectLanguage enum, Create New Project picker (after Python), Initialize Project for VS Code picker, and the package.json projectLanguage enum (marked Preview).
- Map Go to the `native` worker runtime (what core tools and Azure expect for FUNCTIONS_WORKER_RUNTIME). Add `native` to isKnownWorkerRuntime() so deploy verification doesn't reject Go apps.
- Register `$func-golang-watch` problem matcher and use it for Go's tasks.json. Matcher keeps the user-friendly `golang` name even though the underlying runtime value is `native`.
- New GoProjectCreateStep: thin wrapper around `func init --worker-runtime go --no-source-control`. Honors workspace funcCliPath setting via context.workspaceFolder fallback chain.
- New GoInitVSCodeStep: generates .vscode/{tasks,launch,settings, extensions}.json. Launch config is a Delve remote attach on port 2345 (the dlv lifecycle itself lands in a follow-up PR). Recommends the golang.go extension.
- Skip template selection for Go in FunctionListStep (mirrors SelfHostedMCPServer) -- `func init` already scaffolds an initial function.
- Show a modal warning when Create Function is invoked on a Go project -- `func new --language Go` isn't supported by core tools yet.
- Unit tests for the three Go-specific assertions in settings.ts (getRootFunctionsWorkerRuntime, isKnownWorkerRuntime, getFuncWatchProblemMatcher).

### DIFF
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizard, type IActionContext } from '@microsoft/vscode-azext-utils';
-import * as vscode from 'vscode';
+import { type WorkspaceFolder } from 'vscode';
 import { type FuncVersion } from '../../FuncVersion';
 import { ProjectLanguage, projectTemplateKeySetting } from '../../constants';
 import { addLocalFuncTelemetry } from '../../funcCoreTools/getLocalFuncCoreToolsVersion';
@@ -50,7 +50,7 @@ export async function createFunctionFromCommand(
 }
 
 export async function createFunctionInternal(context: IActionContext, options: api.ICreateFunctionOptions): Promise<void> {
-    let workspaceFolder: vscode.WorkspaceFolder | undefined;
+    let workspaceFolder: WorkspaceFolder | undefined;
     let workspacePath: string | undefined = options.folderPath;
 
     if (workspacePath === undefined) {
@@ -77,11 +77,11 @@ export async function createFunctionInternal(context: IActionContext, options: a
         // Core tools does not yet support `func new --language Go`, so additional functions must be added by editing
         // the source files directly. Show a modal so the user can't miss it.
         context.telemetry.properties.goCreateFunctionUnsupported = 'true';
-        await vscode.window.showWarningMessage(
-            localize('goCreateFunctionUnsupportedTitle', 'Adding functions to a Go project is not yet supported.'),
+        await context.ui.showWarningMessage(
+            localize('goCreateFunctionUnsupported', 'Adding new functions is currently not supported for Go projects'),
             {
                 modal: true,
-                detail: localize('goCreateFunctionUnsupportedDetail', 'To add another function, edit the source files directly.'),
+                detail: localize('goCreateFunctionUnsupportedDetail', 'To add another function, edit the Go source files directly.'),
             },
         );
         return;
--- a/src/commands/createFunction/createFunction.ts
+++ b/src/commands/createFunction/createFunction.ts
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizard, type IActionContext } from '@microsoft/vscode-azext-utils';
-import { type WorkspaceFolder } from 'vscode';
+import * as vscode from 'vscode';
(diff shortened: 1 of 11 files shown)


src/commands/createFunction/createFunction.ts
```diff
@@ -4,9 +4,9 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizard, type IActionContext } from '@microsoft/vscode-azext-utils';
-import { type WorkspaceFolder } from 'vscode';
+import * as vscode from 'vscode';
```
nturinski: Could you revert this? I don't see any other vscode imports and we lose the type import.

src/commands/createFunction/createFunction.ts
```diff
@@ -71,6 +71,22 @@ export async function createFunctionInternal(context: IActionContext, options: a
     }
 
     const { language, languageModel, version, templateSchemaVersion } = await verifyInitForVSCode(context, projectPath, options.language, options.languageModel, options.version);
+
+    if (language === ProjectLanguage.Go) {
+        // Go projects scaffold an initial function during "Create New Project" via `func init --worker-runtime go`.
+        // Core tools does not yet support `func new --language Go`, so additional functions must be added by editing
+        // the source files directly. Show a modal so the user can't miss it.
+        context.telemetry.properties.goCreateFunctionUnsupported = 'true';
+        await vscode.window.showWarningMessage(
```
nturinski: We have a special UI handler in the context. You can access it by using context.ui.showWarningMessage

src/vsCodeConfig/settings.ts
```diff
@@ -89,6 +89,8 @@ export function getRootFunctionsWorkerRuntime(language: string | undefined): str
             return 'python';
         case ProjectLanguage.PowerShell:
             return 'powershell';
+        case ProjectLanguage.Go:
+            return 'native';
```
nturinski, line 96: Could you leave a comment here explaining why the runtime setting for Go is 'native'?

src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ProjectCreateStepBase } from './ProjectCreateStepBase';
+
+export class GoProjectCreateStep extends ProjectCreateStepBase {
+    public async executeCore(context: IProjectWizardContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
+        progress.report({ message: localize('initializingGoProject', 'Initializing Go Functions project...') });
+
+        // Prefer the currently-open workspace's azureFunctions.funcCliPath override.
+        // context.workspacePath is the freshly-picked new-project folder, which isn't part of any open
+        // workspace yet, so passing it to getFuncCliPath only resolves global/user settings.
+        const funcCliPathScope = context.workspaceFolder ?? workspace.workspaceFolders?.[0];
```
nturinski: Is this what you were referring to when you were talking about hard coding the Function CLI?

src/commands/createFunction/createFunction.ts
```diff
@@ -71,6 +71,22 @@ export async function createFunctionInternal(context: IActionContext, options: a
...
+            localize('goCreateFunctionUnsupportedTitle', 'Adding functions to a Go project is not yet supported.'),
```
nturinski: Maybe something like: "Adding new functions is currently not supported for Go projects"

src/vsCodeConfig/settings.ts
harshivcodes: added a note.

src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
harshivcodes: yeah - but I updated it so it aligns with what other languages are doing. could you take another look?

nturinski (approved):

PR #5023, opened 2026-05-14
Add Go language support: debug

harshivcodes:
Second of three PRs adding Go support to the Azure Functions extension. Builds on #5019. After this PR, F5 attaches Delve to a Go Functions project and breakpoints work. Deploy follows in PR3.

Highlights:

 - New GoDebugProvider extends FuncDebugProviderBase. Owns the dlv lifecycle: kill stale dlv on the target port, poll up to 120s for the Go worker (app/app.exe), spawn 
dlv attach <pid> --headless --continue --listen=:2345 --api-version=2 --accept-multiclient detached, wait for it to listen. Each step logs to the Azure Functions output 
channel and emits telemetry.
 - goDebugConfig moves from GoInitVSCodeStep to GoDebugProvider as the canonical source (PR1 left a comment anticipating this).
 - New validateDelveInstalled runs dlv version before attaching; on miss, shows a modal with a Learn more link to the Delve install docs. Honors a new 
azureFunctions.validateDelve setting (default true), mirroring azureFunctions.validateFuncCoreTools.
 - extension.ts registers GoDebugProvider for the 'go' debug type. Not wired into FuncTaskProvider — Go uses external dlv attach instead of the worker-arg env-var 
pattern.
 - validatePreDebug.ts left unchanged. Shared pre-debug checks still run for Go via FuncDebugProviderBase; Go-specific validation lives only in GoDebugProvider.
 - Unit test pinning the goDebugConfig shape so silent regressions to type/mode/port surface in CI.

Known limitations:

 - Worker process name is hardcoded to app/app.exe (matches current core tools output).
 - Detached dlv is cleaned up lazily on next attach via killStaleDlv, not on session end.
 - dlv must be on PATH; no dlvPath setting in this PR.
 - Smoke tested on Windows. macOS/Linux paths use pgrep/lsof/kill — asking reviewers on those platforms to verify.

### DIFF
--- a/src/commands/initProjectForVSCode/InitVSCodeStep/GoInitVSCodeStep.ts
+++ b/src/commands/initProjectForVSCode/InitVSCodeStep/GoInitVSCodeStep.ts
@@ -5,21 +5,10 @@
 
 import { type DebugConfiguration, type TaskDefinition } from 'vscode';
 import { func, hostStartCommand, hostStartTaskName, type ProjectLanguage } from '../../../constants';
+import { goDebugConfig } from '../../../debug/GoDebugProvider';
 import { getFuncWatchProblemMatcher } from '../../../vsCodeConfig/settings';
 import { type IProjectWizardContext } from '../../createNewProject/IProjectWizardContext';
 import { InitVSCodeStepBase } from './InitVSCodeStepBase';
-
-// Default Delve DAP port used by GoDebugProvider. Inlined here so PR 1 is self-contained;
-// PR 2 will introduce GoDebugProvider and may re-export this config.
-export const goDebugConfig: DebugConfiguration = {
-    name: 'Attach to Go Functions',
-    type: 'go',
-    request: 'attach',
-    mode: 'remote',
-    port: 2345,
-    host: '127.0.0.1',
-    preLaunchTask: hostStartTaskName,
-};
 
 export class GoInitVSCodeStep extends InitVSCodeStepBase {
     stepName: string = 'GoInitVSCodeStep';
--- a/src/debug/GoDebugProvider.ts
+++ b/src/debug/GoDebugProvider.ts
@@ -207,6 +207,11 @@ async function pollAndAttachDlv(port: number, folder: WorkspaceFolder | undefine
             dlvProcess.on('error', (err) => {
                 ext.outputChannel.appendLog(localize('dlvSpawnFailed', 'Failed to spawn Delve: {0}', err.message));
             });
+            dlvProcess.on('exit', () => {
+                // remove dead entries from the tracking map so killTrackedDlv
+                // doesn't log spurious "Cleaned up" lines later.
+                activeDlvProcesses.delete(folder.uri.fsPath);
+            });
             dlvProcess.stdout?.on('data', (chunk: Buffer) => {
                 ext.outputChannel.append(`[dlv] ${chunk.toString()}`);
             });
@@ -226,6 +231,9 @@ async function pollAndAttachDlv(port: number, folder: WorkspaceFolder | undefine
             context.telemetry.properties.dlvAttachResult = 'attached';
             ext.outputChannel.appendLog(localize('dlvAttached', 'Delve attached to Go worker (PID {0}) and listening on port {1}.', pid, port));
         } catch (err) {
+            // Clean up any dlv we already spawned before failing - killTrackedDlv is idempotent.
+            killTrackedDlv(folder.uri.fsPath);
+
             if (isUserCancelledError(err)) {
                 // callWithTelemetryAndErrorHandling marks the event as cancelled and stays silent.
                 throw err;
--- a/src/debug/GoDebugProvider.ts
+++ b/src/debug/GoDebugProvider.ts
@@ -4,14 +4,18 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { callWithTelemetryAndErrorHandling, type IActionContext } from '@microsoft/vscode-azext-utils';
-import { spawn } from 'child_process';
+import { spawn, type ChildProcess } from 'child_process';
 import { createConnection } from 'net';
+import psTree, { type PS } from 'ps-tree';
+import * as vscode from 'vscode';
 import { type CancellationToken, type DebugConfiguration, type WorkspaceFolder } from 'vscode';
-import { hostStartTaskName, localhost } from '../constants';
+import { hostStartTaskName, hostStartTaskNameRegExp, localhost } from '../constants';
 import { ext } from '../extensionVariables';
+import { runningFuncTaskMap, type IRunningFuncTask } from '../funcCoreTools/funcHostTask';
 import { localize } from '../localize';
 import { cpUtils } from '../utils/cpUtils';
 import { delay } from '../utils/delay';
+import { getWindowsProcessTree, ProcessDataFlag, type IProcessInfo, type IWindowsProcessTree } from '../utils/windowsProcessTree';
 import { FuncDebugProviderBase } from './FuncDebugProviderBase';
 import { validateDelveInstalled } from './validateDelveInstalled';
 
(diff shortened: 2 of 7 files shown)


src/commands/initProjectForVSCode/InitVSCodeStep/GoInitVSCodeStep.ts
```diff
@@ -5,22 +5,11 @@
 
 import { type DebugConfiguration, type TaskDefinition } from 'vscode';
 import { func, hostStartCommand, hostStartTaskName, type ProjectLanguage } from '../../../constants';
+import { goDebugConfig } from '../../../debug/GoDebugProvider';
 import { getFuncWatchProblemMatcher } from '../../../vsCodeConfig/settings';
 import { type IProjectWizardContext } from '../../createNewProject/IProjectWizardContext';
 import { InitVSCodeStepBase } from './InitVSCodeStepBase';
 
-// Default Delve DAP port used by GoDebugProvider. Inlined here so PR 1 is self-contained;
```
harshivcodes, line 12: note: this is defined in GoDebugProvider.ts now.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,230 @@
...
+    // FuncTaskProvider has no `case ProjectLanguage.Go` branch.
+    public readonly workerArgKey: string = 'GOLANG_WORKER_DEBUG_FLAGS';
+    protected readonly defaultPortOrPipeName: number = defaultGoDebugPort;
+    protected readonly debugConfig: DebugConfiguration = goDebugConfig;
+
+    public async getWorkerArgValue(_folder: WorkspaceFolder): Promise<string> {
+        return '';
+    }
+
+    public async resolveDebugConfiguration(folder: WorkspaceFolder | undefined, debugConfiguration: DebugConfiguration, token?: CancellationToken): Promise<DebugConfiguration | undefined> {
```
nturinski, line 116: Since you're overriding `resolveDebugConfiguration` here, you're getting rid of the `if (hostStartTaskNameRegExp.test(debugConfiguration.preLaunchTask as string))` gate that is in the base class. 

As a result, this resolve code will be called for every go debug configuration, not just function projects.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,230 @@
...
+        }
+        const match = result.cmdOutput.match(new RegExp(`"${imageName.replace('.', '\\.')}","(\\d+)"`));
+        return match ? parseInt(match[1], 10) : undefined;
+    }
+
+    const result = await cpUtils.tryExecuteCommandLine(undefined, undefined, `pgrep -x ${goWorkerProcessName}`);
+    if (result.code !== 0) {
+        return undefined;
+    }
+    const pid = parseInt(result.cmdOutput.trim().split('\n')[0], 10);
```
nturinski: Is there anyway to validate that this pid is actually a Go Functions worker? My understanding is that app/app.exe is a very common process name and could interfere with retrieving the correct pid.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,230 @@
...
+        // launches the Go worker concurrently with this. The poller watches
+        // for that worker, then spawns `dlv attach` so VS Code's Go extension
+        // can connect to dlv's DAP server on `port`.
+        void pollAndAttachDlv(port);
+
+        return result;
+    }
+}
+
+async function pollAndAttachDlv(port: number): Promise<void> {
```
nturinski: The cancellation token should get passed in here so the polling can terminate early if the user cancels debugging.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,230 @@
...
+            context.telemetry.measurements.dlvWorkerPid = pid;
+
+            // Another dlv may have started in the meantime (very fast restart).
+            if (await isPortInUse(port)) {
+                context.telemetry.properties.dlvAttachResult = 'portTaken';
+                ext.outputChannel.appendLog(localize('dlvPortTaken', 'Port {0} is already in use; assuming dlv is already attached.', port));
+                return;
+            }
+
+            const dlvProcess = spawn('dlv', [
```
nturinski, line 190: Should have an error listener here just in case the spawn fails.

nturinski (commented):
Regarding this limitation:
> Detached dlv is cleaned up lazily on next attach via killStaleDlv, not on session end.

There's a `azureFunctions.onDidTerminateDebugSession` in `funcHostTask` that you may be able to leverage to be more proactive about cleaning up dlv.

src/debug/GoDebugProvider.ts
harshivcodes: Added it.

src/debug/GoDebugProvider.ts
harshivcodes: fair, I changed the implementation so that the current folder's func host PID from runningFuncTaskMap.getAll(folder), get that PID's descendants via windowsProcessTree/ps-tree and pick a child named app/app.exe

src/debug/GoDebugProvider.ts
harshivcodes: done.

src/debug/GoDebugProvider.ts
harshivcodes: Added it.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,367 @@
...
+// folder before the first ends, the previous entry is killed and replaced.
+const activeDlvProcesses = new Map<string, ChildProcess>();
+
+/**
+ * Wires the dlv-cleanup-on-session-end behavior. Call once during extension
+ * activation; the returned disposable should be added to context.subscriptions.
+ */
+export function registerGoDebugSessionCleanup(): vscode.Disposable {
+    return vscode.debug.onDidTerminateDebugSession((session) => {
+        if (session.type !== 'go' || !session.workspaceFolder) {
```
nturinski, line 62: Not a huge concern, but this will fire for any Go debug session that gets terminated.  For example, if a user has a Functions Go session running and then briefly opens a non-Functions Go debug session in the same workspace folder, the non-Functions session ending will kill the tracked dlv that the Functions session is still using.

You can fix it by checking the debug session's id explicitly, but I'm not sure how commonly a user has a debug session with both a functions and non-functions session

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,367 @@
...
+            resolve(false);
+        });
+    });
+}
+
+async function waitForPort(port: number, timeoutMs: number, token: CancellationToken | undefined): Promise<void> {
+    const deadline = Date.now() + timeoutMs;
+    while (Date.now() < deadline) {
+        if (token?.isCancellationRequested) {
+            return;
```
nturinski: We have a custom `UserCancelledError` that throws a silent error, but in telemetry tracks the telemetry event as cancelled. That way you don't need to handle the cancel in other code paths; the error will just bubble up and terminate the command.

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,367 @@
...
+    await delay(stalePortFreedDelayMs);
+}
+
+async function pollForGoWorkerPid(timeoutMs: number, folder: WorkspaceFolder | undefined, token: CancellationToken | undefined): Promise<number | undefined> {
+    const deadline = Date.now() + timeoutMs;
+    while (Date.now() < deadline) {
+        if (token?.isCancellationRequested) {
+            return undefined;
+        }
+        const pid = await findGoWorkerPid(folder);
```
nturinski, line 293: If there's no folder opened, findGoWorkerPid will always automatically return undefined, and this 120s loop will just be stuck. Probably should validate that there is a workspace folder before calling `findGoWorkerPid` and terminate immediately.

src/debug/GoDebugProvider.ts
harshivcodes: using this instead

src/debug/GoDebugProvider.ts
harshivcodes: added the fix! as for the user scenario, probably unlikely but better to add a safe guard.

src/debug/GoDebugProvider.ts
harshivcodes: good catch, made this change!

src/debug/GoDebugProvider.ts
```diff
@@ -0,0 +1,378 @@
...
+            ext.outputChannel.appendLog(localize('dlvAttached', 'Delve attached to Go worker (PID {0}) and listening on port {1}.', pid, port));
+        } catch (err) {
+            if (isUserCancelledError(err)) {
+                // callWithTelemetryAndErrorHandling marks the event as cancelled and stays silent.
+                throw err;
+            }
+            context.telemetry.properties.dlvAttachResult = 'failed';
+            const message = err instanceof Error ? err.message : String(err);
+            ext.outputChannel.appendLog(localize('dlvAttachFailed', 'Failed to auto-attach Delve: {0}', message));
+            throw err;
```
nturinski, line 244: Before you throw the error to terminate, you should probably kill dlv? Just because the port wasn't able to attach, it doesn't necessarily mean that dlv didn't spawn, right?

harshivcodes: good catch! fixed this. also pushed another fix to clean up the map pid entry if dlv exits on its own.

nturinski (approved):

PR #5030, opened 2026-05-19
Add Go language support: deploy

harshivcodes:
Third of three PRs adding Go support. Builds on #5019 and #5023. After this PR, "Deploy to Function App" works for Go projects on Linux Flex.

Approach: local func pack + standard zip-push, same path Python uses when remote build is unavailable. Reuses every existing checkpoint (verifyAppSettings, EOL warnings, deploy confirmation) — no new branch in the deploy wizard. Does not validate the Function App SKU specifically; relies on the existing per-language guards (Linux/Windows, Flex/Consumption flags) plus the new Go-on-Windows rejection.

Highlights:
 - New validateGoInstalled runs go version; modal with Learn more on miss. Honors azureFunctions.validateGo (default true), mirroring the Func Core Tools and Delve 
checks.
 - New applyGoDeployDefaults writes the four Go deploy settings (scmDoBuildDuringDeployment=false, preDeployTask=func: pack, deploySubpath=<projectName>.zip, .gitignore entry). Idempotent; logs to the output channel.
 - FuncTaskProvider registers func: pack for Go alongside Python.
 - deploy.ts: reject Go on Windows (mirrors Python guard); apply Go defaults before doRemoteBuild is read so the setting flip is reflected downstream — prevents writing Oryx app settings on a user's first Go deploy.
 - detectProjectLanguage: detect via go.mod; infer from FUNCTIONS_WORKER_RUNTIME=native.
 - Unit test pins isGoProject.

Known limitations:
 - Requires Go on PATH locally.
 - CGO not supported.
 - Windows apps rejected; users steered to Linux Flex.
 - Smoke tested on Linux Flex only; Premium / App Service Plan / Linux Consumption should also work via the same zip-push path but unverified end-to-end.


nturinski (commented):
temp

src/funcCoreTools/validateGoInstalled.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Verifies that the Go toolchain is available on PATH before packaging a Go
+ * Functions project for deployment. If missing, prompts the user with a
+ * "Learn more" button that opens the Go install docs.
+ *
+ * Honors the `azureFunctions.validateGo` workspace setting (default true).
+ *
+ * @returns true if Go is installed (or validation is disabled), false otherwise.
+ */
+export async function validateGoInstalled(_context: IActionContext, message: string, workspacePath?: string): Promise<boolean> {
```
nturinski: Are you planning to use this in other places? Just wondering why you have the message passed in as a parameter.

src/commands/deploy/applyGoDeployDefaults.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Writes the workspace settings needed for Go Functions to deploy via local
+ * `func pack` + zip-push. Idempotent — safe to run on every deploy.
+ *
+ * Go has no remote-build path on Azure (Oryx doesn't support Go Functions),
+ * so we force `scmDoBuildDuringDeployment` off, set `preDeployTask` to
+ * `func: pack`, point `deploySubpath` at the zip that pack will produce,
+ * and ensure that zip is `.gitignore`d.
+ */
+export async function applyGoDeployDefaults(context: IDeployContext, workspaceFolder: vscode.WorkspaceFolder): Promise<void> {
```
nturinski: This will mutate workspace settings before a deploy is confirmed and persists even if user cancels which can unexpectedly alter behavior for other projects sharing the same workspace folder.

src/commands/deploy/applyGoDeployDefaults.ts
```diff
@@ -0,0 +1,40 @@
...
+    const projectPath: string = await tryGetFunctionProjectRoot(context, workspaceFolder) || workspaceFolder.uri.fsPath;
+    const zipFileName: string = path.basename(projectPath) + '.zip';
+    // Always use posix separators for config checked in to source control
+    const deploySubpathValue: string = path.posix.join(path.relative(workspaceFolder.uri.fsPath, projectPath), zipFileName);
+
+    await updateWorkspaceSetting(remoteBuildSetting, false, workspaceFolder);
```
nturinski: It might be worth letting the user know which settings are going to be changed. Though should this be set when the project is created?

src/funcCoreTools/validateGoInstalled.ts
harshivcodes: that was a copy paste error - fixing it.

src/commands/deploy/applyGoDeployDefaults.ts
harshivcodes: changing it so that the settings are written during init.

src/commands/deploy/applyGoDeployDefaults.ts
harshivcodes: making the change so that the settings are written during init.

src/funcCoreTools/validateGoInstalled.ts
```diff
@@ -39,6 +39,7 @@ export async function validateGoInstalled(_context: IActionContext, message: str
             return;
         }
 
+        const message: string = 'The Go toolchain is required to package Go Functions for deployment. Install Go and try again.';
```
nturinski: nit: localize the string

src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
```diff
@@ -3,27 +3,132 @@
...
-            withArg('--no-source-control'),
-        )();
+        const goModPath: string = path.join(context.projectPath, 'go.mod');
+        if (await confirmOverwriteFile(context, goModPath)) {
+            await AzExtFsExtra.writeFile(goModPath, getGoModTemplate(moduleName));
+        }
 
-        await cpUtils.executeCommand(ext.outputChannel, context.projectPath, funcCliPath, args);
+        // Generate go.sum so `func start` / `go build` can resolve dependencies immediately.
+        await cpUtils.executeCommand(ext.outputChannel, context.projectPath, 'go', composeArgs(withArg('mod', 'tidy'))());
```
nturinski, line 48: Now that you're using go here, you should probably validate that it is installed before running this command.

src/commands/initProjectForVSCode/InitVSCodeStep/GoInitVSCodeStep.ts
```diff
@@ -4,17 +4,22 @@
...
 import { InitVSCodeStepBase } from './InitVSCodeStepBase';
+import { ensureGitIgnoreContents } from './PythonInitVSCodeStep';
 
 export class GoInitVSCodeStep extends InitVSCodeStepBase {
     stepName: string = 'GoInitVSCodeStep';
+    protected preDeployTask: string = packTaskName;
 
     protected async executeCore(context: IProjectWizardContext): Promise<void> {
-        this.setDeploySubpath(context, '.');
+        const zipFileName: string = context.projectPath.split(/[\\/]/).pop() + '.zip';
```
nturinski: Why not use path.basename(projectPath)+ '.zip' like you had been? Splitting it manually like this is a little bit fragile,.

src/commands/createNewProject/ProjectCreateStep/GoProjectCreateStep.ts
harshivcodes: ack - added validation for this.

src/funcCoreTools/validateGoInstalled.ts
harshivcodes: added!

src/commands/initProjectForVSCode/InitVSCodeStep/GoInitVSCodeStep.ts
```diff
@@ -31,6 +32,10 @@ export class GoInitVSCodeStep extends InitVSCodeStepBase {
                 problemMatcher: getFuncWatchProblemMatcher(language),
                 isBackground: true,
             },
+            {
```
nturinski, line 35: You should include a label for this new task.

nturinski (approved):

PR #5069, opened 2026-06-01
 Fixes Template Gallery bugs

### LINKED ISSUE #5065: New project created via Template Gallery is not opened automatically in VS Code
**OS:** Win 11 x64
**Build Version:** [1.21.3](https://github.com/microsoft/vscode-azurefunctions/actions/runs/26667473388)

**Repro Steps:**
1. F1-> Create new project->Browse Template Gallery(Preview)
2. In the template selection step, click a template card (for example, TypeScript HTTP Trigger) to enter Template Details
3. Select a target folder and complete project creation

**Expect:**
After project creation, VS Code should either automatically open the newly created folder or prompt the user to open the project or add it to the workspace

**Actual:**
Project is successfully created on disk, but it is not opened in VS Code. There is no prompt is shown to open the newly created project or add it to the workspace.

### LINKED ISSUE #5068: Non-empty folder warning is triggered based on workspace instead of selected target folder
**OS:** Win 11 x64
**Build Version:** [1.21.3](https://github.com/microsoft/vscode-azurefunctions/actions/runs/26667473388)

**Repro Steps:**
1. Open a workspace with an existing project (Folder A)
2. F1-> Create new project->Select an empty folder (Folder B) as the target location ->Browse Template Gallery(Preview)
3. Continue project creation

**Expect:**
Template Gallery should validate the selected target folder (Folder B). No non-empty folder warning should appear if the selected folder is empty

**Actual:**
A non-empty folder warning is displayed based on the current workspace (Folder A), even though the selected folder (Folder B) is empty.

This may mislead users into thinking the selected folder is not empty and can lead to creating the project in the wrong location.

gavin-aguiar:
Fixes two Template Gallery bugs reported by the testing team.
 
 ## Fixes #5068 — Non-empty folder warning fires against the wrong folder
 
 **Repro:** Open Folder A as a workspace → `F1 → Create New Project` → pick empty **Folder B** in the folder picker → choose "Browse Template Gallery (Preview)…" → pick a template → "Use this template".
 
 **Before:** The gallery validates against the open workspace (Folder A), so the user sees a "folder is not empty" modal warning about Folder A. The folder they picked seconds earlier (Folder B) is silently dropped.
 
 **Root cause:** `NewProjectLanguageStep.ts:89` calls `FunctionsTemplateGalleryController.createOrShow(ext.context)` and drops `context.projectPath`. The gallery then falls back to `vscode.workspace.workspaceFolders[0]` in `fetchTemplates`.
 
 **Fix:**
 - Added an optional `initialLocation?: string` to `FunctionsTemplateGalleryController` constructor and `createOrShow`.
 - `fetchTemplates` now prefers `this.initialLocation` over the workspace fallback.
 - `NewProjectLanguageStep` passes `context.projectPath` through on the hand-off.
 - `createOrShow` also disposes an existing panel if it was created with a different `initialLocation`, so the second run with a fresh folder doesn't reveal a stale panel.
 
 ## Fixes #5065 — Newly-created project from gallery is not opened in VS Code
 
 **Repro:** `F1 → Browse New Templates` → pick a template → "Use this template" → choose an empty folder outside the current workspace → wait for success toast.
 
 **Before:** Project files are written to disk and a README opens, but VS Code never prompts to open the new folder or add it to the workspace. The user is left in their previous workspace with an orphaned README tab.
 
 **Root cause:** After scaffolding, `createProject` only called `tryOpenReadme` + `disposePanelIfOpen` + `showInformationMessage`. The classic wizard handles this with `OpenBehaviorStep` (prompt) and `OpenFolderStep` (execute), but the gallery never wired them up.
 
 **Fix:** Added `promptAndOpenProject` which:
 - No-ops if the target folder is already a workspace folder (uses `isPathEqual` from `utils/fs.ts`).
 - Builds a minimal `IProjectWizardContext`, pre-seeded from the existing `azureFunctions.projectOpenBehavior` setting (same precedence as the classic wizard).
 - Runs a 2-step `AzureWizard` with `OpenBehaviorStep` → `OpenFolderStep`, reusing the shared steps verbatim so behavior and telemetry stay consistent.
 - Silently swallows `UserCancelledError` if the user dismisses the open-behavior picker — files stay on disk where they were just created.

### DIFF
--- a/src/commands/createNewProject/FunctionsTemplateGalleryController.ts
+++ b/src/commands/createNewProject/FunctionsTemplateGalleryController.ts
@@ -57,7 +57,7 @@ export class FunctionsTemplateGalleryController extends TemplateGalleryControlle
             // wizard's folder picker) and it differs from what the existing panel was
             // created with, the old defaultLocation is stale. Dispose so we recreate
             // with the new location instead of silently ignoring it.
-            if (initialLocation !== undefined && existing.initialLocation !== initialLocation) {
+            if (initialLocation !== undefined && (existing.initialLocation === undefined || !isPathEqual(existing.initialLocation, initialLocation))) {
                 existing.dispose();
             } else {
                 existing.revealToForeground();
--- a/src/commands/createNewProject/FunctionsTemplateGalleryController.ts
+++ b/src/commands/createNewProject/FunctionsTemplateGalleryController.ts
@@ -3,7 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { AzExtFsExtra, UserCancelledError, callWithTelemetryAndErrorHandling, parseError, type IActionContext } from '@microsoft/vscode-azext-utils';
+import { AzExtFsExtra, AzureWizard, UserCancelledError, callWithTelemetryAndErrorHandling, parseError, type IActionContext } from '@microsoft/vscode-azext-utils';
 import { TemplateGalleryController, registerWebviewExtensionVariables, type IProjectTemplate as ISharedProjectTemplate, type TemplateGalleryConfig } from '@microsoft/vscode-azext-webview';
 import extract from 'extract-zip';
 import * as fs from 'fs';
@@ -15,7 +15,13 @@ import { localize } from '../../localize';
 import { type IProjectTemplate } from '../../templates/projectTemplates/IProjectTemplate';
 import { ProjectTemplateProvider } from '../../templates/projectTemplates/ProjectTemplateProvider';
 import { cpUtils } from '../../utils/cpUtils';
+import { isPathEqual } from '../../utils/fs';
 import { requestUtils } from '../../utils/requestUtils';
+import { getWorkspaceSetting } from '../../vsCodeConfig/settings';
+import { projectOpenBehaviorSetting } from '../../constants';
+import { type IProjectWizardContext, type OpenBehavior } from './IProjectWizardContext';
+import { OpenBehaviorStep } from './OpenBehaviorStep';
+import { OpenFolderStep } from './OpenFolderStep';
 
 /**
  * Azure Functions implementation of the shared TemplateGalleryController.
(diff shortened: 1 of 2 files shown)


src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -42,10 +50,19 @@ export class FunctionsTemplateGalleryController extends TemplateGalleryControlle
...
+            // wizard's folder picker) and it differs from what the existing panel was
+            // created with, the old defaultLocation is stale. Dispose so we recreate
+            // with the new location instead of silently ignoring it.
+            if (initialLocation !== undefined && existing.initialLocation !== initialLocation) {
+                existing.dispose();
+            } else {
+                existing.revealToForeground();
+                return existing;
+            }
         }
```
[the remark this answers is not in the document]
gavin-aguiar: In the current call sites the comparison is safe because initialLocation always originates from context.projectPath, which FolderListStep populates via showOpenDialog -> canonical fsPath, so both ends see byte-identical strings. The undefined case is also handled correctly today (undefined !== "C:\\Bar" evaluates to true → dispose, which is the intended behavior). So no real bug in the current flow

src/commands/createNewProject/FunctionsTemplateGalleryController.ts
```diff
@@ -42,10 +50,19 @@ export class FunctionsTemplateGalleryController extends TemplateGalleryControlle
...
-            FunctionsTemplateGalleryController.currentController.revealToForeground();
-            return FunctionsTemplateGalleryController.currentController;
+    public static createOrShow(context: vscode.ExtensionContext, initialLocation?: string): FunctionsTemplateGalleryController {
+        const existing = FunctionsTemplateGalleryController.currentController;
+        if (existing) {
+            // If the caller supplied a fresh initialLocation (typically from the classic
+            // wizard's folder picker) and it differs from what the existing panel was
+            // created with, the old defaultLocation is stale. Dispose so we recreate
+            // with the new location instead of silently ignoring it.
+            if (initialLocation !== undefined && existing.initialLocation !== initialLocation) {
```
nturinski: Should use isPathEqual here like in other places, concerned about Windows paths sometimes using / or \ in directory paths

gavin-aguiar: Fixed

nturinski (approved):

PR #5080, opened 2026-06-02
Read version number in nodev4 templates to pick up new templates

nturinski:
The issue is basically without reading the version number, the extension doesn't know when it needs to reload the templates. Users can manually do this, but there's no way for them to know that.

This updates the Node v4 template provider to read the bundled backup templates version so template updates are detected automatically, and bumps the bundled Node v4 backup template version accordingly.


nturinski:
@copilot re-review it

MicroFish91 (approved):

PR #5085, opened 2026-06-03
Stop using utils that are not available in node 24

nturinski:
I think the tests suddenly started breaking because VS Code stable must have updated to node 24. Let's see this fixes it.


bwateratmsft (approved):
Look around for any `isNullOrUndefined` too

PR #5104, opened 2026-06-10
Revert change to allow deployments with flex consumption for python 3.14

gavin-aguiar:
This PR revert change added to prevent deploying Python 3.14 function apps to Flex Consumption plans
Remote build is now supported so we need to remove this error message.

Revert PR: https://github.com/microsoft/vscode-azurefunctions/pull/4861


nturinski:
> Remote build is not supported so we need to remove this error message.

Is not supported or is now supported?

gavin-aguiar:
> > Remote build is not supported so we need to remove this error message.
> 
> Is not supported or is now supported?

Sorry typo. Its "now supported"

nturinski (approved):

PR #5126, opened 2026-07-02
Add code signature validation when installing or updating `Azure Functions Core Tools`

MicroFish91:
Add code signature verification that now runs automatically whenever we install / update func core tools.

Code signature verification references used:
- Powershell: https://github.com/Azure/azure-dev/blob/main/cli/installer/install-azd.ps1#L338
- Bash: https://github.com/Azure/azure-dev/blob/main/cli/installer/install-azd.sh#L300

Verified tests on both windows and mac.


.vscode/settings.json
```diff
@@ -1,4 +1,9 @@
 {
+    // Node 24's experimental network inspection (node:internal/inspector/network_http) is buggy and
```
MicroFish91: This randomly started happening after an update.  Disabling this flag fixed it for me, will double check this again before merging

src/funcCoreTools/ensureBrewTapTrusted.ts
```diff
@@ -0,0 +1,28 @@
...
+import { composeArgs, withArg } from '@microsoft/vscode-processutils';
+import * as vscode from 'vscode';
+import { ext } from '../extensionVariables';
+import { localize } from '../localize';
+import { cpUtils } from '../utils/cpUtils';
+
+const funcTap: string = 'azure/functions';
+
+/**
+ * Prompts the user to trust the "azure/functions" Homebrew tap, then taps and trusts it.
```
MicroFish91, line 16: Newer Homebrew behavior that requires the user to trust `azure/functions` if they haven't done so before.  It's a prerequisite to being able to install core tools.

test/nightly/validateFuncCoreToolsCodeSignature/validateFuncCoreToolsCodeSignature.test.ts
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * 2. Pass each version's cliPath into `validateCodeSignature` directly
+ *    to test and ensure the code signature validation logic is working as expected.
+ *
+ * Whether a version is expected to be signed depends on the check from `isCodeSignatureExpected`.
+ */
+
+const versionsToTest: FuncVersion[] = Object.values(FuncVersion).filter(v => isCodeSignatureExpected(v));
+
+suite.only('validateFuncCoreToolsCodeSignature', function (this: Mocha.Suite): void {
```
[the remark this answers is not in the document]
MicroFish91: Todo: remove .only before merge

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
```diff
@@ -0,0 +1,179 @@
...
+
+async function getFuncCoreToolsPath(workspacePath?: string): Promise<string | undefined> {
+    let funcLookupOutput: string | undefined;
+    switch (process.platform) {
+        case 'darwin':
+        case 'linux':
+            funcLookupOutput = await cpUtils.executeCommand(ext.outputChannel, workspacePath, 'which', composeArgs(withArg('func'))());
+            break;
+        case 'win32':
+            funcLookupOutput = await cpUtils.executeCommand(ext.outputChannel, workspacePath, 'where.exe', composeArgs(withArg('func'))());
```
bwateratmsft, line 69: The `processutils` package already includes the `which` package, which is a perfect fit for this. Less parsing to do too. I think you can use it indirectly too; IIRC the package exports a `getSafeExecPath` or similar.

Though, as I think about it, `getSafeExecPath` may not do anything on Unix. If so, directly using `which` is fine (but be sure it is declared as a dependency in package.json).

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    return isValid;
+}
+
+async function validateWin32CodeSignature(cliPath: string): Promise<boolean> {
+    // Escape single quotes for the PowerShell single-quoted string literal. PowerShell single-quoted
+    // strings have no backslash escaping; a literal ' is represented by doubling it ('').
+    const escapedPath = cliPath.replace(/'/g, "''");
+    const psCommand = `$sig = Get-AuthenticodeSignature '${escapedPath}'; if ($sig.Status -ne 'Valid') { exit 1 }; $sig.SignerCertificate.Subject`;
+    const signingResult = await cpUtils.tryExecuteCommand(ext.outputChannel, undefined, 'powershell', composeArgs(withArg('-Command', psCommand))());
```
bwateratmsft, line 164: `npm` has an `audit signatures` command, not sure if you can scope it to a specific package, but that may be a better option if we know it's installed from NPM (which, I'm not sure we do know)

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
```diff
@@ -0,0 +1,179 @@
...
+        default:
+            throw new Error(localize('codeSignatureUnsupportedPlatform', 'Code signature validation is not supported on platform "{0}".', process.platform));
+    }
+}
+
+export const microsoftSubject = 'Microsoft Corporation';
+
+async function validateDarwinCodeSignature(cliPath: string): Promise<boolean> {
+    // Verify the signature is valid (i.e. the binary has not been tampered with)
+    const codeSignResult = await cpUtils.tryExecuteCommand(ext.outputChannel, undefined, 'codesign', composeArgs(withArg('-v', cliPath))());
```
bwateratmsft, line 141: CLI path probably needs quoted

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
MicroFish91: Nice, way less parsing logic and testing to worry about with this inclusion

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
MicroFish91: Yeah not super sure about this but perhaps it could be a complementary check, although I think the verification npm provides might only be for the download/tarball integrity?  If that's the case I assume we may still want to verify the func executable directly and that it was signed by Microsoft.  

Copilot also mentioned something about a restriction on running npm audit signatures on a global install.  I haven't done any manual checking on this yet to confirm though.

src/funcCoreTools/validateFuncCoreToolsCodeSignature.ts
MicroFish91: Oh would that not be handled automatically by withArg / composeArgs ?

bwateratmsft: Not necessarily. You can do `withQuotedArg`

alexweininger (approved):

bwateratmsft (approved):

PR #5169, opened 2026-08-25
Fix skipped build tasks when debugging .NET isolated Functions

patverb:
Debugging a .NET isolated project ran `func host start` without first running its `clean` and `build` dependencies, so F5 attached to stale binaries.

To attach the debugger we have to tell func where to write the isolated worker PID, via `--json-output-file`. That flag was added by cloning the task after VS Code had already resolved it. `dependsOn` only exists in tasks.json and VS Code only honors it for tasks it resolved itself, so executing the clone silently dropped the build chain.

Add the flag while the task is still being resolved, in FuncTaskProvider.createTask, so VS Code builds the task with the flag already on it and there is nothing to clone. pickFuncProcess now executes the task it fetched.

This needs no change to tasks.json, so projects scaffolded before the fix behave the same as new ones. The injection is gated on `--dotnet-isolated-debug`, leaving every other language's host start task untouched. The terminal-stream PID fallback is left armed so a task the func task provider never resolves, or a file func never writes, degrades to scraping stdout instead of hanging until pickProcessTimeout.

### DIFF
--- a/src/funcCoreTools/jsonOutputFile.ts
+++ b/src/funcCoreTools/jsonOutputFile.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzExtFsExtra } from '@microsoft/vscode-azext-utils';
+import * as crypto from 'crypto';
 import * as os from 'os';
 import * as path from 'path';
 import * as vscode from 'vscode';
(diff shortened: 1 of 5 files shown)


src/funcCoreTools/jsonOutputFile.ts
```diff
@@ -94,47 +95,60 @@ export function generateJsonOutputFilePath(): string {
...
-    wrapped.runOptions = funcTask.runOptions;
-    wrapped.detail = funcTask.detail;
-    return wrapped;
+    try {
+        if (await AzExtFsExtra.pathExists(workerPidFile)) {
+            await AzExtFsExtra.deleteResource(workerPidFile);
+        }
+    } catch {
+        // Best effort - a file we can't delete just means we wait for func to overwrite it.
+    }
```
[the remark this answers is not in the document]
patverb: done

alexweininger (approved):

src/funcCoreTools/jsonOutputFile.ts
```diff
@@ -94,47 +95,68 @@ export function generateJsonOutputFilePath(): string {
...
+ * Deterministic rather than unique-per-run so that a file left behind by a previous session lands at
+ * the same path we delete before starting, and so temp files don't accumulate. `key` should identify
+ * the func task, i.e. its workspace folder plus its configured command and args.
  */
-export function injectJsonOutputFileArgIfNeeded(funcTask: vscode.Task): vscode.Task {
-    const exec = funcTask.execution;
-    if (!(exec instanceof vscode.ShellExecution)) {
-        return funcTask;
+export function getWorkerPidFilePath(key: string): string {
+    const hash: string = crypto.createHash('sha256').update(key).digest('hex').slice(0, 16);
```
nturinski: You should be able to just use the randomUtils.getPseudononymousStringHash for this.

src/funcCoreTools/jsonOutputFile.ts
```diff
@@ -94,47 +95,68 @@ export function generateJsonOutputFilePath(): string {
...
     }
 
-    const flatArgs = getFlatShellArgs(exec);
-    const hasDebugFlag = flatArgs.includes(dotnetIsolatedDebugFlag);
-    const hasEnableJsonOutput = flatArgs.includes(enableJsonOutputFlag);
-    const alreadyHasOutputFile = flatArgs.includes(jsonOutputFileFlag) || flatArgs.some(a => a.startsWith(`${jsonOutputFileFlag}=`));
-    if (!hasDebugFlag || !hasEnableJsonOutput || alreadyHasOutputFile) {
-        return funcTask;
+    const args: string[] = [];
+    if (!existingArgs.includes(enableJsonOutputFlag)) {
```
nturinski: I'm not sure if we want to always inject this parameter though. We are trying to account for configurations that do not include this flag. I know it's recommended to use the flag, but I could still see scenarios where the user may not want it generated.

src/funcCoreTools/jsonOutputFile.ts
patverb: done

src/funcCoreTools/jsonOutputFile.ts
patverb: Fixed, Now requires `--enable-json-output` to already be present, matching the existing shouldInjectJsonOutputFile  check.

MicroFish91 (approved):

PR #5183, opened 2026-09-16
Remove project initialization popup

### LINKED ISSUE #5167: CoR: Remove Azure Functions init pop-up for CoR projects
We should potentially remove this feature which causes this pop-up to show up:
Image

MicroFish91:
Fixes #5167

## Summary
- stop prompting to initialize detected Azure Functions projects during extension activation
- retain explicit and on-demand initialization flows
- keep `azureFunctions.showProjectWarning` as a deprecated no-op setting for existing CoR workspaces


MicroFish91:
Reminder: If this gets merged, we should consider updating Copilot on Rails to no longer set this setting to false as it will be deprecated when released

